PR #13, opened 2018-05-15
DI using Dagger

saket:


app/src/main/java/org/resolvetosavelives/red/RedApp.kt
```diff
@@ -1,14 +1,25 @@
...
 class RedApp : Application() {
 
+  companion object {
+    lateinit var appComponent: AppComponent
+  }
+
   override fun onCreate() {
     super.onCreate()
 
     Timber.i("Planing timber tree")
```
pratul: unnecessary?

app/src/main/java/org/resolvetosavelives/red/widgets/Views.kt
```diff
@@ -8,6 +8,6 @@ fun EditText.showKeyboard() {
   post({
     this.requestFocus()
     val inputMethodManager = context.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
-    inputMethodManager.showSoftInput(this, InputMethodManager.SHOW_FORCED)
+    inputMethodManager.showSoftInput(this, InputMethodManager.SHOW_IMPLICIT)
```
pratul: Why was this changed, exactly?

app/src/main/java/org/resolvetosavelives/red/RedApp.kt
saket: Leftover, yes. Will remove this.

app/src/main/java/org/resolvetosavelives/red/widgets/Views.kt
saket: I was initially using `SHOW_FORCED` because the keyboard wasn't showing up for one screen with `SHOW_IMPLICIT`. But I later saw that the keyboard never gets removed when the screen is changed.

PR #14, opened 2018-05-15
MVI in new-bp and patient-personal-details entry screens

saket:
Part of #10


pratul:
I'm not going to merge this because the (only?) test fails to pass yet Bitrise is reporting passed. Need to figure out where our configuration is broken.

saket:
Oh nice. I just realized that our git pre-push hook does not run tests right now. I'll also fix this.

Issue #18, opened 2018-05-16
Sync patient data from user to server


ssrihari:
- [x] Create patient model payload (patient, address)
- [x] Create patient model payload (phone numbers)
- [x] Schedule periodic syncing of patients
- [x] Send sync-pending patients to server
- [x] Limit the number of patients to send to the server
- [x] Finish unit tests

As a nurse, I want the patient records on device to be sent in batches to the backend, so that the central database is updated, and the data is safe.

### Patient data model
ID - UUID [to discuss]
Full Name - String
Age when created - Integer
Date of Birth - Date
Gender (male / female / transgender) - Enum/String
Status (active, dead, migrated, unresponsive, other [define other]) - Enum/String

### Address data model
id – uuid
street address - string
colony - string
village - string
district - string
state - string
country - string
pin - string

### Phone number data model
number - string
type -string
active - boolean

### Relationships
Patient belongs to an address
Patient has and belongs to many phone numbers

Questions:
> - What is the "other" patient status?
> - What should the status be when a patient is "Removed from call list"?

Not relevant for MVP. We'll still keep the status field in the code. We'll default to "active" status. @saket please default to "active" status on the device as well.

>  - Should we perform any validation on the mobile number?

No, as discussed with Daniel. To allow landlines, extensions, etc.

> - Should village/colony be the same field? Should they be mandatory?

Yes, single field. Not mandatory. As discussed with Tim/Daniel at standup on 17/05.

Re: address model
1. Locality: as entered by nurses on the app. How would we model this?
2. Aadhar address: what are the fields? How would this fit into our model?

saket:
> Not relevant for MVP. We'll still keep the status field in the code. We'll default to "active" status. @saket please default to "active" status on the device as well.

Got it.

ssrihari:
Is the patient <=> phone_number relationship many-to-many? In that case, are we sure we want to update other patients' phone numbers when one patient's shared phone number is updated?

If not, it's a one-to-many relationship. 

@timcheadle @saket @govindkrjoshi

PR #36, opened 2018-05-24
Schedule periodic syncing of patients using WorkManager

saket:
Part of #18


saket:
@ssrihari @govindkrjoshi can you go through our tests for syncing to see if we're missing anything?

https://github.com/resolvetosavelives/redapp/blob/sn/24may/app/src/test/java/org/resolvetosavelives/red/sync/PatientSyncTest.kt

Issue #19, opened 2018-05-16
Integrate sentry


dvenkatr:
- [x] Integrate Android app with Sentry SDK (for fatals)
- [x] Provide Sentry configuration and secrets from Bitrise while generating builds
- [x] Enable multidex
- [x] Sanity check that everything here is done?

pratul:
@timcheadle To do this, I need the "Sentry DSN (client key)" so I can configure the SDK on Android. More information here: https://docs.sentry.io/clients/java/modules/android/

Saket and I will also need access to the sentry.io dashboard.

PR #48, opened 2018-06-01
Add Sentry libraries and configuration

pratul:
Part of #19

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,19 +1,22 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest package="org.resolvetosavelives.red"
-  xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:tools="http://schemas.android.com/tools"
+  package="org.resolvetosavelives.red">
 
   <uses-permission android:name="android.permission.CAMERA" />
   <uses-permission android:name="android.permission.VIBRATE" />
   <uses-permission android:name="android.permission.INTERNET" />
+  <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 
   <application
-    android:name=".ReleaseRedApp"
     android:allowBackup="true"
     android:icon="@mipmap/ic_launcher"
     android:label="@string/app_name"
+    android:name=".ReleaseRedApp"
     android:roundIcon="@mipmap/ic_launcher_round"
     android:supportsRtl="true"
-    android:theme="@style/AppTheme">
+    android:theme="@style/AppTheme"
+    tools:ignore="GoogleAppIndexingWarning">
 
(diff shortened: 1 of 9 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -1,18 +1,22 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest package="org.resolvetosavelives.red"
-  xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:tools="http://schemas.android.com/tools"
+  package="org.resolvetosavelives.red">
 
   <uses-permission android:name="android.permission.CAMERA" />
   <uses-permission android:name="android.permission.VIBRATE" />
+  <uses-permission android:name="android.permission.INTERNET" />
```
saket: We did not have internet permissions so far? :O how was patient sync working then?

app/src/main/java/org/resolvetosavelives/red/RedApp.kt
```diff
@@ -1,13 +1,17 @@
...
+import io.sentry.Sentry
+import io.sentry.android.AndroidSentryClientFactory
 import org.resolvetosavelives.red.di.AppComponent
 import org.resolvetosavelives.red.di.AppModule
 import org.resolvetosavelives.red.di.DaggerAppComponent
 import timber.log.Timber
 
 class RedApp : Application() {
 
+  val dsn = "https://e8606e0cd2f7470e834e446905612ff2@sentry.io/1212614"
```
saket: Can you rename this to something other than `dsn`? Sounds very vague. Maybe you can also make this a constant by moving this to a companion object:

```kotlin
companion object {
  const val DSN = "…"
}
```

I think this is the java equivalent of static fields.

app/src/main/java/org/resolvetosavelives/red/TheActivity.kt
```diff
@@ -32,6 +32,8 @@ class TheActivity : AppCompatActivity() {
 
     val contextWithRouter = screenRouter.installInContext(baseContext, android.R.id.content, initialScreenKey())
     super.attachBaseContext(contextWithRouter)
+
+    throw UnsupportedOperationException()
```
saket: What is this? 😄

sentry.properties
```diff
@@ -0,0 +1,3 @@
+defaults.project=redapp
+defaults.org=resolve-to-save-lives
+auth.token=027c5d8041f44b139e2e3af83556122ac9dc7cc768004fcba61cf0473c0606a5
```
saket: We should avoid staging this file to Github, right?

app/src/main/java/org/resolvetosavelives/red/RedApp.kt
saket: Is this `dsn` safe to include in our repo publicly?

PR #58, opened 2018-06-05
Update README with steps for downloading and building RedApp

saket:
Readme: https://github.com/resolvetosavelives/redapp/blob/sn/readme-with-build-steps/README.md


saket:
@pratul/Steve Is the `Build` section required?

PR #64, opened 2018-06-06
Split age into age and age-updated-at + move classes around

### LINKED ISSUE #40: store age, and age_updated_at, to compute the current age.
@saket @pratul

Part of #16 and #18

saket:
Closes #40


app/src/main/java/org/resolvetosavelives/red/newentry/search/OngoingPatientEntry.kt
```diff
@@ -15,6 +15,7 @@ data class OngoingPatientEntry(
    * [ageWhenCreated] is stored as a String instead of an Int because it's easy
    * to forget that [Int.toString] will return literal "null" for null Ints.
    */
+  // TODO: Rename to 'age'.
   data class PersonalDetails(val fullName: String, val dateOfBirth: String?, val ageWhenCreated: String?, val gender: Gender?)
```
saket: `ageWhenCreated` in this class is renamed to `age` in a later commit

PR #67, opened 2018-06-06
Update patient search to support names, phone numbers, and age

pratul:
(Part of #6 and #7)

- Misc
  - [ ] Update the UI to match zeplin
  - [x] Update the search query to match name or phone numbers
  - [ ] If user selects a patient, open a dummy UI patient summary screen 
  - [x] If user clicks Back button, go back to the Home screen
- Age filter 
  - [ ] Make the UI in the toolbar
  - [ ] Update the query to show results with ±3 years of age entered
- Patient Sync API
  - [ ] Make sync push API call also send phone numbers
  - [ ] Make sync pull API call also retrieve and store phone numbers

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -2,7 +2,7 @@
   <string name="app_name">RedApp</string>
 
   <!-- Patient search -->
-  <string name="patientsearch_new_patient">New patient</string>
+  <string name="patientsearch_new_patient">Create new patient</string>
   <string name="patientsearch_go_back">Go back to home screen</string>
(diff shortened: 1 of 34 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -5,6 +5,7 @@
   <string name="patientsearch_new_patient">Create new patient</string>
   <string name="patientsearch_go_back">Go back to home screen</string>
   <string name="patientsearch_name_or_phone">Patient\'s name or phone</string>
+  <string name="patient_age">Age: %s</string>
```
saket: Can you prefix this with the screen name?

Issue #52, opened 2018-06-04
Nurse can record Prescribed Drugs


ssrihari:
## As a ... I can ... So that ...

As a nurse, after recording their BP, I can record the patient's prescribed drugs according to the state protocol.

- [x] Show prescriptions under the current protocol
- [x] Save user selections
- [x] Show custom prescriptions
- [x] Add custom prescriptions
- [x] Support deletion of prescriptions

## Bugs

- [x] P0 After entering BP, then tap "add medicine", then tapping medicine/dose, then hit back arrow, BP keyboard pops back up
- [ ] P0 In some cases, 2 dosages are selected for the same drug (example: https://vitalstrat.slack.com/files/UAKE25198/FBCLDE4GZ/img_20180625_092427_963.jpg)
- [x] P1 When "add another medicine" entry is up, add an overlay to obfuscate the background
- [x] P1 If patient doesn't have any meds, the button should not say "UPDATE", but instead "ADD MEDICINES" (copy what's on [Zeplin](https://app.zeplin.io/project/5af2812b979d9e9123803ed8/screen/5b10e1c1517446700809fc5d)) 
- [x] P1 After entering drugs, the back button sometimes displays the BP entry screen instead of the Patient Summary screen
- [x] ~P1: (This may already be recorded) Add a new patient, BP pops up, enter BP, then tap "Add" on medicine, choose a medication, then hit back arrow in top left.... BP jumps up again. The BP shouldn't be triggered again.~
- [x] P1: Before medication is added the medicine button says "Add". It should say "Add medicines" as per Zeplin.
- [x] P2 There should be a SAVE/DONE button after medicines are added

vikalp02:
P2: Here is the screen which shows a Toast whenever the medicines are added/updated.
[Zeplin Web](zpl.io/aRYG7dn)

PR #76, opened 2018-06-12
Empty patient summary screen

saket:
Part of #52

### DIFF
--- a/app/src/main/java/org/resolvetosavelives/red/newentry/PatientEntryScreen.kt
+++ b/app/src/main/java/org/resolvetosavelives/red/newentry/PatientEntryScreen.kt
@@ -80,10 +80,6 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
     }
     TheActivity.component.inject(this)
 
-    // Plan:
-    // - Show 'X' icon when a field is focused.
-    // - Switch between 'date of birth' and 'date of birth (dd/mm/yyyy)` for DOB's hint.
-
     fullNameEditText.showKeyboard()
     upButton.setOnClickListener { screenRouter.pop() }
 
--- a/app/src/main/java/org/resolvetosavelives/red/search/PatientSearchScreen.kt
+++ b/app/src/main/java/org/resolvetosavelives/red/search/PatientSearchScreen.kt
@@ -19,8 +19,8 @@ import org.resolvetosavelives.red.TheActivity
 import org.resolvetosavelives.red.newentry.PatientEntryScreen
 import org.resolvetosavelives.red.patient.PatientSearchResult
 import org.resolvetosavelives.red.router.screen.ScreenRouter
+import org.resolvetosavelives.red.summary.PatientSummaryScreenKey
 import org.resolvetosavelives.red.widgets.showKeyboard
-import timber.log.Timber
 import java.util.UUID
 import javax.inject.Inject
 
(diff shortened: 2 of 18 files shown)


app/src/main/java/org/resolvetosavelives/red/search/PatientSearchScreen.kt
```diff
@@ -34,7 +36,7 @@ class PatientSearchScreen(context: Context, attrs: AttributeSet) : RelativeLayou
   @Inject
   lateinit var controller: PatientSearchScreenController
 
-  private val backButton by bindView<ImageButton>(R.id.patientsearch_back)
+  private val upButton by bindView<ImageButton>(R.id.patientsearch_up)
```
pratul: We really should call it "Back" everywhere, and not Up, because the terminology is confusing and Google has failed to make this distinction clearly. Not to mention, users don't understand this at all, and expect it to work something-almost-but-not-quite-entirely-like a "Back" button.

pratul: We really should call it "Back" everywhere, and not Up, because the terminology is confusing and Google has failed to make this distinction clearly. Not to mention, users don't understand this at all, and expect it to work something-almost-but-not-quite-entirely-like a "Back" button.

app/src/main/java/org/resolvetosavelives/red/search/PatientSearchScreen.kt
```diff
@@ -48,8 +50,7 @@ class PatientSearchScreen(context: Context, attrs: AttributeSet) : RelativeLayou
 
     TheActivity.component.inject(this)
 
-    // TODO: Can we use sealed classes to represent events?
```
pratul: Can we keep this todo around? I want to look into this.

app/src/test/java/org/resolvetosavelives/red/search/PatientSearchScreenControllerTest.kt
```diff
@@ -66,9 +67,35 @@ class PatientSearchScreenControllerTest {
   }
 
   @Test
-  fun `when back button is clicked, home screen should open`() {
-    uiEvents.onNext(BackButtonClicked())
+  fun `when up button is clicked, home screen should open`() {
```
pratul: Same back/up thing

app/src/main/res/layout/screen_patient_search.xml
```diff
@@ -14,7 +14,7 @@
     tools:ignore="UnusedAttribute">
 
     <ImageButton
-      android:id="@+id/patientsearch_back"
+      android:id="@+id/patientsearch_up"
```
pratul: Same back/up thing.

app/src/main/java/org/resolvetosavelives/red/search/PatientSearchScreen.kt
saket: On a second thought, this is not entirely true. Users get confused by the hardware back button and not the software one. The software back key always takes you a level back/up within the same app whereas hardware back button flies apps around.

app/src/main/java/org/resolvetosavelives/red/newentry/PatientEntryScreen.kt
```diff
@@ -80,10 +80,6 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
     }
     TheActivity.component.inject(this)
 
-    // Plan:
-    // - Show 'X' icon when a field is focused.
-    // - Switch between 'date of birth' and 'date of birth (dd/mm/yyyy)` for DOB's hint.
-
     fullNameEditText.showKeyboard()
     upButton.setOnClickListener { screenRouter.pop() }
```
pratul: This should also be Back?

saket: This is not a part of this PR. I've so far used `up` everywhere else. I'll change them later.

PR #77, opened 2018-06-12
Age filter in the search

pratul:
(Continues to be part of #6)


app/src/main/res/values/strings.xml
```diff
@@ -11,6 +11,8 @@
   <string name="patientsearch_go_back">Go back to home screen</string>
   <string name="patientsearch_name_or_phone">Patient\'s name or phone</string>
   <string name="patientsearch_age">Age: %s</string>
+  <string name="patientsearch_filter_by">Filter by</string>
+  <string name="patientsearch_label_age">AGE</string>
```
saket: Probably a small thing, but it is always recommended to capitalize strings in the UI using `textAllCaps` instead of normalizing them

app/src/main/java/org/resolvetosavelives/red/patient/PatientRepository.kt
```diff
@@ -24,16 +24,37 @@ class PatientRepository @Inject constructor(private val database: AppDatabase) {
...
+        .search(query!!)
+        .toObservable()
+  }
+
+  fun searchPatientsAndPhoneNumbers(query: String?, assumedAge: Int): Observable<List<PatientSearchResult>> {
+    val ageUpperBound = assumedAge + ageFuzziness
+    val ageLowerBound = assumedAge - ageFuzziness
+
+    val dateOfBirthUpperBound = LocalDate.now().minusYears(ageUpperBound.toLong()).toString()
+    val dateOfBirthLowerBound = LocalDate.now().minusYears(ageLowerBound.toLong()).toString()
```
saket: It looks weird that both `upper` and `lower` bounds use `minusYears()`

saket: `LocalDate.now()` can lead to errors because it internally uses the system's default timezone. You'll have to use `LocalDate.now(ZoneOffset.UTC)`.

app/src/main/java/org/resolvetosavelives/red/patient/PatientRepository.kt
```diff
@@ -53,7 +53,7 @@ class PatientRepository @Inject constructor(private val database: AppDatabase) {
 
     if (query.isNullOrEmpty()) {
       return database.patientSearchDao()
-          .allRecords()
+          .recentlyUpdated100Records()
```
saket: > recentlyUpdated100Records()
Nice!

app/src/main/java/org/resolvetosavelives/red/patient/PatientSearchResult.kt
```diff
@@ -75,8 +75,8 @@ data class PatientSearchResult(
       """)
     fun search(query: String, dobUpperBound: String, dobLowerBound: String): Flowable<List<PatientSearchResult>>
 
-    @Query(mainQuery)
-    fun allRecords(): Flowable<List<PatientSearchResult>>
+    @Query("$mainQuery ORDER BY P.updatedAt LIMIT 100")
+    fun recentlyUpdated100Records(): Flowable<List<PatientSearchResult>>
```
saket: Hey… shouldn't this be ordered by latest first (DESC)?

app/src/main/java/org/resolvetosavelives/red/patient/PatientWithAddressAndPhone.kt
```diff
@@ -6,7 +6,7 @@ import org.threeten.bp.Instant
 import org.threeten.bp.LocalDate
 import java.util.UUID
 
-@Deprecated(message = "You should use PatientSearchResult instead.")
+@Deprecated(message = "You should use PatientSearchResult instead.", level = DeprecationLevel.ERROR)
 data class PatientWithAddressAndPhone(
```
saket: Will compilation fail if I use this class? In that case, can we remove this class?

app/src/main/java/org/resolvetosavelives/red/patient/PatientSearchResult.kt
```diff
@@ -69,8 +69,14 @@ data class PatientSearchResult(
...
-    @Query(mainQuery)
-    fun allRecords(): Flowable<List<PatientSearchResult>>
+    @Query("""$mainQuery
+      WHERE (P.fullname LIKE '%' || :query || '%' OR PP.number LIKE '%' || :query || '%')
+      AND ((P.dateOfBirth BETWEEN :dobUpperBound AND :dobLowerBound) OR (P.age_computedDateOfBirth BETWEEN :dobUpperBound AND :dobLowerBound))
+      """)
+    fun search(query: String, dobUpperBound: String, dobLowerBound: String): Flowable<List<PatientSearchResult>>
+
+    @Query("$mainQuery ORDER BY P.updatedAt LIMIT 100")
+    fun recentlyUpdated100Records(): Flowable<List<PatientSearchResult>>
```
saket: Hey… shouldn't this be ordered by latest first (DESC)?

pratul: You're right, fixed.

app/src/main/java/org/resolvetosavelives/red/patient/PatientWithAddressAndPhone.kt
pratul: Yes, removed. This was an oversight.

PR #90, opened 2018-06-18
Ensure release build works

ssrihari:
- Disable proguard for release build
- Add a DebugClinic application, configure debug build to use that
- Only use Stetho in debug builds

Part of https://github.com/resolvetosavelives/redapp/issues/83.

### DIFF
--- a/app/src/debug/java/org/simple/clinic/DebugClinic.kt
+++ b/app/src/debug/java/org/simple/clinic/DebugClinic.kt
@@ -1,0 +1,33 @@
+package org.simple.clinic
+
+import android.annotation.SuppressLint
+import com.facebook.stetho.Stetho
+import com.tspoon.traceur.Traceur
+import org.simple.clinic.di.AppComponent
+import org.simple.clinic.di.AppModule
+import org.simple.clinic.di.DaggerAppComponent
+import org.simple.clinic.sync.SyncScheduler
+import timber.log.Timber
+import javax.inject.Inject
+
+@SuppressLint("Registered")
+class DebugClinic : Clinic() {
+
+  @Inject
+  lateinit var syncScheduler: SyncScheduler
+
+  override fun onCreate() {
+    super.onCreate()
+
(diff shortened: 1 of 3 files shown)


app/src/debug/java/org/simple/clinic/DebugClinic.kt
```diff
@@ -0,0 +1,35 @@
...
+@SuppressLint("Registered")
+class DebugClinic : Clinic() {
+
+  @Inject
+  lateinit var syncScheduler: SyncScheduler
+
+  override fun onCreate() {
+    super.onCreate()
+
+      if (BuildConfig.DEBUG) {
```
saket: Since `DebugClinic` class only exists in the debug source set, this condition will always be true :)

app/src/debug/java/org/simple/clinic/DebugClinic.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.tspoon.traceur.Traceur
+import org.simple.clinic.di.AppComponent
+import org.simple.clinic.di.AppModule
+import org.simple.clinic.di.DaggerAppComponent
+import org.simple.clinic.sync.SyncScheduler
+import timber.log.Timber
+import javax.inject.Inject
+
+@SuppressLint("Registered")
+class DebugClinic : Clinic() {
```
saket: Thanks for adding this class. I was postponing doing this for long.

app/build.gradle
```diff
@@ -34,6 +34,7 @@ android {
 
     release {
       debuggable false
+      useProguard false
```
pratul: We don't need this here -- It's already present in gradle.properties: https://github.com/resolvetosavelives/redapp/blob/master/gradle.properties

app/src/debug/java/org/simple/clinic/DebugClinic.kt
pratul: Yeah please remove this condition.

PR #157, opened 2018-07-17
17 July

vinaysshenoy:
## Highlights
Bug fix that prevents prescriptions from syncing.

### Pending Changes
 - [x] Add a test to ensure that the soft delete operation works as intended
 - [x] Update the soft delete operation in the repository and the dao

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/drugs/PrescriptionRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/drugs/PrescriptionRepositoryAndroidTest.kt
@@ -111,6 +111,29 @@ class PrescriptionRepositoryAndroidTest {
     assertThat(savedPrescriptions2).hasSize(1)
   }
 
+  @Test
+  fun soft_delete_prescription_should_update_timestamp_and_sync_status() {
+    val protocolUUID = UUID.randomUUID()
+    val patientUUID = UUID.randomUUID()
+
+    val drug = ProtocolDrug(UUID.randomUUID(), name = "Amlodipine", rxNormCode = null, dosages = listOf("5mg", "10mg"), protocolUUID = protocolUUID)
+
+    repository.savePrescription(patientUUID, drug, "5mg").blockingAwait()
+
+    val savedPrescriptions = repository.newestPrescriptionsForPatient(patientUUID).blockingFirst()
+    assertThat(savedPrescriptions).hasSize(1)
+
+    val prescription = savedPrescriptions[0]
+
+    repository.softDeletePrescription(prescription.uuid).blockingAwait()
+
+    val softDeletedPrescription = database.prescriptionDao().getOne(prescription.uuid)!!
+
+    assertThat(softDeletedPrescription.updatedAt).isGreaterThan(prescription.updatedAt)
+    assertThat(softDeletedPrescription.createdAt).isEqualTo(prescription.createdAt)
+    assertThat(softDeletedPrescription.syncStatus).isEqualTo(SyncStatus.PENDING)
+  }
+
   @After
   fun tearDown() {
     database.clearAllTables()
--- a/app/src/main/java/org/simple/clinic/drugs/PrescriptionRepository.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/PrescriptionRepository.kt
@@ -4,6 +4,7 @@ import io.reactivex.Completable
 import io.reactivex.Observable
 import io.reactivex.Single
 import io.reactivex.rxkotlin.toObservable
+import org.simple.clinic.AppDatabase
 import org.simple.clinic.di.AppScope
 import org.simple.clinic.drugs.sync.PrescribedDrugPayload
 import org.simple.clinic.facility.FacilityRepository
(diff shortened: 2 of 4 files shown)


vinaysshenoy:
@saket I'd like a quick review of some of the changes that I have made to a couple of classes.

Since we are working with time stamps, testing becomes a little tricky since we need control over the timestamps that are created. To that end, I have modified `PrescriptionRepository` in the following ways to allow overriding of the timestamps used when creating entities:

- Added a lambda parameter, `instant` to the methods `PrescriptionRepository#savePrescription` and `PrescriptionRespository#softDelete` that have a default implementation which instantiate an `Instant` at the current time
- Added an android test in `PrescriptionsRepositoryAndroidTest` to test that when the `softDelete` operation for a prescription is done, it also updates the time at which it was updated as well as updates the `SyncStatus`. This test uses `Instant` objects created in the test to compare instants and verify behaviour.

vinaysshenoy:
I did try using default parameters instead of default lambdas, but it ends up affecting other tests in some way due to how the Kotlin compiler handles these parameters. Leaving it for investigation at a later date.

app/src/test/java/org/simple/clinic/drugs/PrescriptionRepositoryTest.kt
```diff
@@ -75,4 +77,15 @@ class PrescriptionRepositoryTest {
       verify(dao).save(argThat { isEmpty() })
     }
   }
+
+  @Test
+  fun `when soft deleting a prescription, update the timestamp as well as set the sync status to pending`() {
+
```
saket: Can you remove this empty line here?

app/src/test/java/org/simple/clinic/drugs/PrescriptionRepositoryTest.kt
```diff
@@ -75,4 +77,15 @@ class PrescriptionRepositoryTest {
...
+    val prescriptionId = UUID.randomUUID()
+
+    repository.softDeletePrescription(prescriptionId).blockingAwait()
+
+    verify(dao).softDelete(eq(prescriptionId), eq(true), any())
```
saket: Holy shit I did not know you can combine `eq` and `any()` in `verify`. I was always forced to pass in actual values or manually compare the values.

app/src/main/java/org/simple/clinic/drugs/PrescriptionRepository.kt
```diff
@@ -61,7 +61,8 @@ class PrescriptionRepository @Inject constructor(
 
   fun softDeletePrescription(prescriptionUuid: UUID): Completable {
     return Completable.fromAction {
-      dao.softDelete(prescriptionUuid, deleted = true)
+      dao.softDelete(prescriptionUuid, deleted = true, updatedAt = Instant.now())
+      dao.updateSyncStatus(listOf(prescriptionUuid), SyncStatus.PENDING)
```
saket: I think we should run these two calls inside a transaction. Thoughts?

vinaysshenoy: Good point. I'll make the change.

app/src/test/java/org/simple/clinic/drugs/PrescriptionRepositoryTest.kt
```diff
@@ -29,7 +33,11 @@ class PrescriptionRepositoryTest {
 
   @Before
   fun setUp() {
-    repository = PrescriptionRepository(dao, facilityRepository, userSession)
+    // Needed if we unit test soft delete
+    whenever(database.runInTransaction(any())).thenAnswer {
```
saket: Hey this is no longer required, right?

app/src/androidTest/java/org/simple/clinic/drugs/PrescriptionRepositoryAndroidTest.kt
```diff
@@ -111,6 +111,29 @@ class PrescriptionRepositoryAndroidTest {
...
+    val savedPrescriptions = repository.newestPrescriptionsForPatient(patientUUID).blockingFirst()
+    assertThat(savedPrescriptions).hasSize(1)
+
+    val prescription = savedPrescriptions[0]
+
+    repository.softDeletePrescription(prescription.uuid).blockingAwait()
+
+    val softDeletedPrescription = database.prescriptionDao().getOne(prescription.uuid)!!
+
+    assertThat(softDeletedPrescription.updatedAt).isGreaterThan(prescription.updatedAt)
```
saket: We were using `isGreaterThan` in another android test and we found that the two timestamps were occasionally the same because the code executes so fast. Rare case, but something to keep in mind.

vinaysshenoy: Does it? `Instant` takes the `nanos` into account as well, so those must have been some really fast tests.

Here, we are testing whether the older prescriptions `updatedAt` and the soft deleted prescriptions `updatedAt` are different(soft deleted < older). Do you feel it would be likely for this case to be triggered? What do you suggest we use instead? Because we do have to test whether the timestamp actually changes or not.

vinaysshenoy: We'll keep an eye out and figure an alternative if we see it happen then.

PR #158, opened 2018-07-17
Copy changes

saket:


app/src/main/res/values/strings.xml
```diff
@@ -90,8 +90,9 @@
 
   <string name="customprescription_enter_medicine">Enter medicine</string>
   <string name="customprescription_save_drug">Save</string>
-  <string name="customprescription_drug_dosage_hint">Dosage</string>
   <string name="customprescription_drug_name_hint">Medicine name</string>
+  <string name="customprescription_dosage_placeholder">mg</string>
```
vinaysshenoy: Do we need localized versions of these strings?

saket: We will eventually, yes. But we decided yesterday that we aren't going to pay attention to translations until we've found a reliable person/service.

saket: Actually, `mg` won't have a translation. I'll mark this as `translatable=false`.

PR #159, opened 2018-07-18
Gradle build cleanup

vinaysshenoy:
## Notes
- Read endpoints from `gradle.properties` instead of system environment variables
- QA endpoint is hardcoded so that local development is a lot faster
- Staging and Production endpoints have placeholder values, that are meant to be overriden on the build server

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -15,3 +15,11 @@ org.gradle.jvmargs=-Xmx1536m
 android.enableD8=true
 # We want the ability to turn off Proguard at will.
 runProguard=false
+
+# Add endpoints in gradle properties so that we can change at will
+# Used for local development
+qaApiEndpoint=https://api-qa.simple.org/api/
+
+# Change them on CI environments or local builds, do not check any actual staging or production endpoints in
+stagingApiEndpoint=do_not_change_here
+productionApiEndpoint=do_not_change_here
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -15,3 +15,9 @@ org.gradle.jvmargs=-Xmx1536m
 android.enableD8=true
 # We want the ability to turn off Proguard at will.
 runProguard=false
+# Add endpoints in gradle properties so that we can change at will
+# Used for local development
+qaApiEndpoint=https://api-qa.simple.org/api/
+# Change them on CI environments or local builds, do not check any actual staging or producation endpoints in
+stagingApiEndpoint=do_not_change_here
+productionApiEndpoint=do_not_change_here
```
saket: Can I ask you to format this a bit? Maybe add line break before every comment?

```properties
android.enableD8=true

# We want the ability to turn off Proguard at will.
runProguard=false

# Add endpoints in gradle properties so that we can change at will
# Used for local development
qaApiEndpoint=https://api-qa.simple.org/api/

# Change them on CI environments or local builds, do not check any actual staging or producation endpoints in
stagingApiEndpoint=do_not_change_here
productionApiEndpoint=do_not_change_here 
```

saket:
LGTM 🥇

PR #160, opened 2018-07-19
Bottomsheet improvements

saket:


app/src/main/java/org/simple/clinic/widgets/BottomSheetActivity.kt
```diff
@@ -2,19 +2,55 @@ package org.simple.clinic.widgets
...
     super.finish()
     overridePendingTransition(0, 0)
   }
+
+  override fun setContentView(layoutResId: Int) {
+    LayoutInflater.from(this).inflate(layoutResId, contentContainer)
+  }
+
+  override fun setContentView(view: View?) {
+    throw AssertionError()
```
vinaysshenoy: It's not clear why the assertion error is being thrown.

app/src/main/java/org/simple/clinic/drugs/selection/entry/CustomPrescriptionEntrySheet.kt
```diff
@@ -61,7 +61,11 @@ class CustomPrescriptionEntrySheet : BottomSheetActivity() {
   }
 
   override fun onBackgroundClick() {
-    if (drugNameEditText.text.isBlank() && drugDosageEditText.text.isBlank()) {
+    val drugNameEmpty = drugNameEditText.text.isEmpty()
+    val dosageEmpty = drugDosageEditText.text.isEmpty()
+        || drugDosageEditText.text.toString() == getString(R.string.customprescription_dosage_placeholder)
```
vinaysshenoy: Do we need to ignore whitespaces when checking for the placeholder string?

saket: Done

app/src/main/java/org/simple/clinic/widgets/BottomSheetActivity.kt
saket: Fixed

PR #161, opened 2018-07-19
Screen change animations + other improvements

saket:


saket:
TODO: remove TODOs from code

PR #162, opened 2018-07-20
20 July

saket:


saket:
Investigate: when both phone number and DOB fields are empty, then DOB gets highlighted.

app/src/main/java/org/simple/clinic/bp/entry/LinearLayoutWithPreImeKeyEventListener.kt
```diff
@@ -0,0 +1,20 @@
+package org.simple.clinic.bp.entry
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.KeyEvent
+import android.widget.LinearLayout
+
+class LinearLayoutWithPreImeKeyEventListener(context: Context, attrs: AttributeSet) : LinearLayout(context, attrs) {
+
+  lateinit var backKeyPressInterceptor: () -> Unit
```
vinaysshenoy: It seems like `lateinit` is being used wrongly here since it is meant to be for something that **will** be initialized at some point in the future, but we are depending on an external class to set it **if** they need to.

app/src/main/res/values/strings.xml
```diff
@@ -43,7 +43,7 @@
   <string name="patiententry_error_empty_colony_or_village">Enter patient’s village or colony or choose ‘None’</string>
   <string name="patiententry_error_empty_district">Enter patient’s home district</string>
   <string name="patiententry_error_state_empty">Enter patient’s home state</string>
-  <string name="patiententry_error_both_dateofbirth_and_age_empty">Enter patient’s birth date</string>
+  <string name="patiententry_error_both_dateofbirth_and_age_empty">Enter patient’s age</string>
```
vinaysshenoy: Don't we have localised versions of this string?

app/src/main/java/org/simple/clinic/bp/entry/LinearLayoutWithPreImeKeyEventListener.kt
saket: TODO: throw an error instead of using `lateinit`.

PR #167, opened 2018-07-25
25 July 2018 - Fuzzy Lookup v1

vinaysshenoy:
### Changes

- Adds a new field `searchableName` to the `Patient` table that saves a stripped down version of the Patient name without whitespaces and punctuation
- Changed `PatientRepository` to strip search queries of white spaces and punctuation when searching for names
- Changed `PatientRepository` to updated the `searchableName` when merging remote patient records as well as adding new patients
- Changed `PatientSearchResult.RoomDao` to use the `searchableName` instead of the `fullName` when doing search queries

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -53,6 +53,27 @@ class PatientRepositoryAndroidTest {
   }
 
   @Test
+  fun createAnOngoingPatientEntry_thenSaveItToDatabase_withSearchableName() {
+    val ongoingAddress = OngoingPatientEntry.Address("HSR Layout", "Bangalore South", "Karnataka")
+    val names = arrayOf("Riya Puri" to "RiyaPuri", "Manabi    Mehra" to "ManabiMehra", "Amit:Sodhi" to "AmitSodhi")
+
+    names.forEach { (fullName, expectedSearchableName) ->
+      val ongoingPersonalDetails = OngoingPatientEntry.PersonalDetails(fullName, "08/04/1985", null, Gender.TRANSGENDER)
+
+      val personalDetailsOnlyEntry = OngoingPatientEntry(personalDetails = ongoingPersonalDetails)
+
+      val patient = repository.saveOngoingEntry(personalDetailsOnlyEntry)
+          .andThen(repository.ongoingEntry())
+          .map { it.copy(address = ongoingAddress) }
+          .flatMapCompletable { repository.saveOngoingEntry(it) }
+          .andThen(repository.saveOngoingEntryAsPatient())
+          .blockingGet()
+
+      assertThat(patient.searchableName).isEqualTo(expectedSearchableName)
+    }
+  }
+
+  @Test
   fun createAnOngoingPatientEntry_withoutPhoneNumber_thenSaveItToDatabase() {
     val ongoingAddress = OngoingPatientEntry.Address("HSR Layout", "Bangalore South", "Karnataka")
     val ongoingPersonalDetails = OngoingPatientEntry.PersonalDetails("Jeevan Bima", "08/04/1985", null, Gender.TRANSGENDER)
@@ -144,6 +165,35 @@ class PatientRepositoryAndroidTest {
         .andThen(repository.saveOngoingEntryAsPatient())
         .test()
         .assertError(AssertionError::class.java)
+  }
+
+  @Test
+  fun patientSearch_shouldIgnore_SpacesAndWhitespaceCharacters() {
+    val ongoingAddress = OngoingPatientEntry.Address("HSR Layout", "Bangalore South", "Karnataka")
+    val names = arrayOf("Riya Puri", "Manabi    Mehra", "Amit:Sodhi")
+    val searches = arrayOf("ya p" to true, "bime" to true, "ito" to false)
+
+    names.forEachIndexed { index, fullName ->
+      val ongoingPersonalDetails = OngoingPatientEntry.PersonalDetails(fullName, "08/04/1985", null, Gender.TRANSGENDER)
+
+      val personalDetailsOnlyEntry = OngoingPatientEntry(personalDetails = ongoingPersonalDetails)
+
+      repository.saveOngoingEntry(personalDetailsOnlyEntry)
+          .andThen(repository.ongoingEntry())
+          .map { it.copy(address = ongoingAddress) }
+          .flatMapCompletable { repository.saveOngoingEntry(it) }
+          .andThen(repository.saveOngoingEntryAsPatient())
+          .blockingGet()
+
+      val (query, shouldFindInDb) = searches[index]
+      val search = repository.searchPatientsAndPhoneNumbers(query).blockingFirst()
+      if (shouldFindInDb) {
+        assertThat(search).hasSize(1)
+        assertThat(search.first().fullName).isEqualTo(fullName)
+      } else {
+        assertThat(search).isEmpty()
+      }
+    }
   }
 
   @Test
--- a/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
@@ -22,6 +22,15 @@ import java.util.Locale
 import java.util.UUID
 import javax.inject.Inject
 
+/**
+ * [Regex] for stripping patient names and search queries of white spaces and punctuation
+ *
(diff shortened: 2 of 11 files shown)


app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -44,7 +53,7 @@ class PatientRepository @Inject constructor(
     }
 
     return database.patientSearchDao()
-        .search(query!!)
+        .search(strip(query!!))
```
govindkrjoshi: We should also lowercase the strings here.
It's an optimization, but we can avoid computing that during search.

vinaysshenoy: The SQLite query we use is already case-insensitive (`LIKE`). This is why I didn't do it, since it seemed unnecessary.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -122,7 +131,11 @@ class PatientRepository @Inject constructor(
             val newOrUpdatedAddresses = payloads.map { it.address.toDatabaseModel() }
             database.addressDao().save(newOrUpdatedAddresses)
 
-            val newOrUpdatedPatients = payloads.map { it.toDatabaseModel(newStatus = DONE) }
+            val newOrUpdatedPatients = payloads.map {
+              it.toDatabaseModel(newStatus = DONE)
+            }.map {
+              it.copy(searchableName = strip(it.fullName))
```
govindkrjoshi: We should default searchableName to it value in the model itself.

app/src/test/java/org/simple/clinic/patient/PatientRepositoryTest.kt
```diff
@@ -31,7 +34,101 @@ class PatientRepositoryTest {
...
+    whenever(database.addressDao()).thenReturn(mockPatientAddressDao)
+
+    repository.searchPatientsAndPhoneNumbers(query).blockingFirst()
+
+    verify(mockPatientSearchResultDao).search(eq(expectedSearchQuery))
+  }
+
+  @Test
+  fun `when searching for patients with age bound, strip the search query of any whitespace or punctuation`() {
+
```
saket: empty line?

app/src/test/java/org/simple/clinic/patient/PatientRepositoryTest.kt
```diff
@@ -31,7 +34,101 @@ class PatientRepositoryTest {
...
+    whenever(mockPatientSearchResultDao.search(any(), any(), any())).thenReturn(Flowable.just(emptyList()))
+    whenever(database.patientSearchDao()).thenReturn(mockPatientSearchResultDao)
+    whenever(database.addressDao()).thenReturn(mockPatientAddressDao)
+
+    repository.searchPatientsAndPhoneNumbers("Name   Surname", 40).blockingFirst()
```
saket: Tell me if I'm wrong, but can't we move these tests to `PatientRepositoryAndroidTest`? We will then be able to assert on the data directly.

app/src/test/java/org/simple/clinic/patient/PatientRepositoryTest.kt
```diff
@@ -31,7 +34,101 @@ class PatientRepositoryTest {
...
+        ageUpdatedAt = mock(),
+        status = mock(),
+        createdAt = mock(),
+        updatedAt = mock(),
+        address = serverAddress,
+        phoneNumbers = null)
+
+    repository.mergeWithLocalData(listOf(serverPatientWithoutPhone)).blockingAwait()
+
+    verify(mockPatientDao).save(argThat<List<Patient>> { first().searchableName == expectedSearchableName })
```
saket: Does type inference fail without `<List<Patient>>`?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: Do you mean make it an optional parameter?

saket: Resolved on a call

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
saket: Resolved on a call

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -52,6 +52,27 @@ class PatientRepositoryAndroidTest {
     assertThat(search2.first().phoneNumber).isEqualTo("2277")
   }
 
+  @Test
+  fun createAnOngoingPatientEntry_thenSaveItToDatabase_withSearchableName() {
+    val ongoingAddress = OngoingPatientEntry.Address("HSR Layout", "Bangalore South", "Karnataka")
+    val names = arrayOf("Riya Puri" to "RiyaPuri", "Manabi    Mehra" to "ManabiMehra", "Amit:Sodhi" to "AmitSodhi")
```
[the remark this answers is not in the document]
vinaysshenoy: This is my take on it:
- If you're going to be doing a lot of conditional checking based on some of the properties, in which case I'd rather make multiple separate tests
- If you are not able to use parameterized tests, like in Android Tests (I am not sure if they have a parameterized test runner)

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -22,6 +22,15 @@ import java.util.Locale
...
 import javax.inject.Inject
 
+/**
+ * [Regex] for stripping patient names and search queries of white spaces and punctuation
+ *
+ * Currently matches the following characters
+ * - Any whitespace
+ * - Comma, Hyphen, SemiColon, Colon, Underscore
+ * */
+private val searchReplaceRegex = Regex("[\\s;_\\-:,]")
```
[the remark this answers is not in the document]
vinaysshenoy: Maybe make it really explicit about what it's being used for and call it `whitespacePunctuationRegex`?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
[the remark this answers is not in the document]
vinaysshenoy: We could go with that.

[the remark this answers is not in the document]
vinaysshenoy: I've decided to call it `toSearchable` for now.

app/src/test/java/org/simple/clinic/patient/PatientRepositoryTest.kt
vinaysshenoy: We can move these tests, but there are a couple of other test cases related to searching for phone numbers that ensure the digits are not stripped out when we do a search query. Those are currently not handled in `PatientRepositoryAndroidTest`. If we remove them here, we should add more tests to `PatientRepositoryAndroidTest` to verify phone number search.

app/src/test/java/org/simple/clinic/patient/PatientRepositoryTest.kt
vinaysshenoy: Yes.

app/src/test/java/org/simple/clinic/patient/PatientRepositoryTest.kt
vinaysshenoy: To add to this, the `PatientRepositoryAndroidTest` test cases do not exercise all code paths in the search patients methods.

app/src/test/java/org/simple/clinic/patient/PatientRepositoryTest.kt
```diff
@@ -31,7 +34,101 @@ class PatientRepositoryTest {
   @Before
   fun setUp() {
     database = mock()
+    mockPatientSearchResultDao = mock()
```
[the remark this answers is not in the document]
vinaysshenoy: Yes. There are multiple tests that use that dao and I don't want the invocations in one test to interfere with another test.

app/src/test/java/org/simple/clinic/patient/PatientRepositoryTest.kt
vinaysshenoy: I haven't moved any of the tests so far, just waiting for your opinion on whether we should move them or not.

saket: I'd prefer testing on data. But I think it's okay for now. :)

PR #170, opened 2018-07-27
More registration screens

saket:


app/src/main/java/org/simple/clinic/registration/RegistrationApiV1.kt
```diff
@@ -0,0 +1,17 @@
+package org.simple.clinic.registration
+
+import io.reactivex.Single
+import retrofit2.http.Body
+import retrofit2.http.POST
+
+interface RegistrationApiV1 {
+
+  companion object {
```
vinaysshenoy: Companion objects are generally discouraged in Kotlin and are left for Java interop. Can we consider moving this to a top-level constant like:
 ```kotlin
private const val API_VERSION = "v1"

interface RegistrationApiV1 {}
```

File: app/src/main/java/org/simple/clinic/registration/RegistrationResult.kt [hunk 2]
```diff
@@ -0,0 +1,8 @@
+package org.simple.clinic.registration
+
+sealed class RegistrationResult {
+
+  class Success : RegistrationResult()
```
vinaysshenoy: FYI, it is no longer necessary to have sealed class types in the class body.

app/src/test/java/org/simple/clinic/registration/pin/RegistrationPinScreenControllerTest.kt
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @Test
+  fun `while pin field is empty then the next button should remain disabled`() {
+    uiEvents.onNext(RegistrationPinTextChanged(""))
+    uiEvents.onNext(RegistrationPinTextChanged("1"))
+    uiEvents.onNext(RegistrationPinTextChanged("12"))
+    uiEvents.onNext(RegistrationPinTextChanged(""))
+
+    verify(screen, times(2)).setNextButtonEnabled(false)
```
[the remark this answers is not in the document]
saket: Passing the expected values is also correct. I think `eq()` is useful when you want to use a mix of `any` and `eq` for the arguments.

app/src/main/java/org/simple/clinic/registration/RegistrationApiV1.kt
saket: Hey @vinaysshenoy, we also use companion objects for screen keys. See `PatientEntryScreen.KEY` for example. What do you think of them?

saket: Btw, the `const` keyword is only allowed in objects or on top levels.

saket: Update 2: property initialization isn't allowed in interfaces either. I think this is the reason why we tried using a companion object.

vinaysshenoy: I think that we should stop using constants as the key instances and stick to constructing them whenever we need to.

Reasons:
- The key objects are cheap
- If we need to use multiple instances of a screen, then we'll need to construct new instances of the keys instead of using a static constant, which means we'll have to start mixing constants and instances, which is not optimal.

Yes, I want to move this out into a top level const instead of a companion object. I'm also in favour of inlining that constant in the Retrofit URL itself since it's localised to the interface and not references externally.

vinaysshenoy: @saket Are we changing this in this PR?

app/src/main/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinScreenController.kt
```diff
@@ -46,6 +46,7 @@ class RegistrationConfirmPinScreenController @Inject constructor(
           userSession.ongoingRegistrationEntry()
               .map { it.copy(pinConfirmation = pinConfirmation, createdAt = Instant.now() ) }
               .flatMapCompletable { userSession.saveOngoingRegistrationEntry(it) }
+              .andThen( userSession.loginFromOngoingRegistrationEntry())
```
vinaysshenoy: Extra space here.

app/src/main/java/org/simple/clinic/registration/RegistrationApiV1.kt
saket: @vinaysshenoy sorry I forgot to respond. I think we can review all companion objects in a new PR. Does that sound okay?

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -60,10 +60,12 @@ class UserSession @Inject constructor(
     return ongoingLoginEntry()
         .map { LoginRequest(UserPayload(it.phoneNumber!!, it.pin!!, it.otp)) }
         .flatMap { loginApi.login(it) }
-        .doOnSuccess { storeUserAndAccessToken(it) }
         .flatMap {
-          facilitySync
-              .sync()
+          storeUserAndAccessToken(it)
+              .andThen(Single.just(it))
```
vinaysshenoy: Instead of using `andThen`, can we use `toSingleDefault(it)` instead?  This applies to the next operator in the chain as well.

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -108,9 +110,9 @@ class UserSession @Inject constructor(
         .firstOrError()
         .map(::RegistrationRequest)
         .flatMap { registrationApi.createUser(it) }
-        .map<RegistrationResult> { response ->
-          storeUser(response.loggedInUser)
-          RegistrationResult.Success()
+        .flatMap<RegistrationResult> {
+          storeUser(it.loggedInUser)
+              .andThen(Single.just(RegistrationResult.Success()))
```
vinaysshenoy: This one too.

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -125,13 +127,15 @@ class UserSession @Inject constructor(
...
-  private fun storeUserAndAccessToken(response: LoginResponse) {
+  private fun storeUserAndAccessToken(response: LoginResponse): Completable {
     accessTokenPreference.set(Just(response.accessToken))
-    storeUser(response.loggedInUser)
+    return storeUser(response.loggedInUser)
   }
 
-  private fun storeUser(loggedInUser: LoggedInUser) {
-    loggedInUserPreference.set(Just(loggedInUser))
+  private fun storeUser(loggedInUser: LoggedInUser): Completable {
```
vinaysshenoy: Thoughts on making this a single line function?

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -147,11 +151,20 @@ class UserSession @Inject constructor(
   }
 
   fun loggedInUser(): Observable<Optional<LoggedInUser>> {
-    return loggedInUserPreference.asObservable()
+    return appDatabase.userDao().user()
+        .toObservable()
+        .map {
+          when {
```
vinaysshenoy: Can we use an `if-else` expression here since there are only two cases.

app/src/main/java/org/simple/clinic/sync/SyncWorker.kt
```diff
@@ -33,11 +34,19 @@ class SyncWorker : Worker() {
...
 
-    if (userSession.isUserLoggedIn().not()) {
-      Timber.i("User isn't logged in yet. Skipping sync.")
-      return WorkerResult.SUCCESS
-    }
+    return userSession.loggedInUser()
+        .firstOrError()
+        .map { it.isApprovedForSyncing() }
+        .flatMap { isApproved ->
+          when {
```
vinaysshenoy: Can we use an `if-else` expression here?

app/src/main/java/org/simple/clinic/user/LoggedInUser.kt
```diff
@@ -23,9 +35,48 @@ data class LoggedInUser(
...
+    @Insert
+    fun create(user: LoggedInUser)
+  }
+
+  fun isApprovedForSyncing(): Boolean {
+    return status == LoggedInUser.Status.APPROVED_FOR_SYNCING
+  }
+}
+
+fun Optional<LoggedInUser>.isApprovedForSyncing(): Boolean {
```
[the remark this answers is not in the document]
vinaysshenoy: I don't think should be used since you can directly call `loggedInUser.get?isApprovedForSyncing() ?: false`.

app/src/main/java/org/simple/clinic/user/LoggedInUserHttpInterceptor.kt
```diff
@@ -12,16 +13,20 @@ class LoggedInUserHttpInterceptor @Inject constructor() : Interceptor {
...
 
-    if (userSession.isUserLoggedIn()) {
-      val (loggedInUser) = userSession.loggedInUser().blockingFirst()
-      val (accessToken) = userSession.accessToken()
+    val (user) = userSession.loggedInUser().blockingFirst()
+    val (accessToken) = userSession.accessToken()
 
-      updatedRequest.addHeader("Authorization", "Bearer ${accessToken!!}")
-      updatedRequest.addHeader("X-USER-ID", loggedInUser!!.uuid.toString())
+    return when (user?.isApprovedForSyncing() == true) {
```
vinaysshenoy: Can we replace this with an `if-else` expression?

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -51,4 +56,25 @@ abstract class AppDatabase : RoomDatabase() {
...
 
   abstract fun facilityDao(): Facility.RoomDao
+
+  abstract fun userDao(): LoggedInUser.RoomDao
+
+  class Migration_3_4 : Migration(3, 4) {
+
+    override fun migrate(database: SupportSQLiteDatabase) {
+      database.execSQL("""
+        CREATE TABLE IF NOT EXISTS `LoggedInUser` (
```
vinaysshenoy: Can we add an android test to run the migrations?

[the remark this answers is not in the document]
vinaysshenoy: This is not important, just something to consider.

vinaysshenoy: Let's skip this. I have to make database changes myself and I'll implement this.

app/src/main/java/org/simple/clinic/user/LoggedInUser.kt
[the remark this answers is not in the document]
saket: Done

app/src/main/java/org/simple/clinic/activity/TheActivity.kt
```diff
@@ -106,7 +112,16 @@ class TheActivity : AppCompatActivity() {
     return when {
       userSession.isUserLoggedIn() -> HomeScreen.KEY
       hasUserCompletedOnboarding.get().not() -> OnboardingScreen.KEY
-      else -> LoginPhoneScreen.KEY("")
+      else -> {
+        registrationConfig
+            .map {
+              when (it.isRegistrationEnabled) {
```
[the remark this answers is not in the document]
saket: I'd be okay if this `if else` statement was short, but this feels like there's a lot happening in a single line. Does it not feel complex to you?

app/src/main/java/org/simple/clinic/user/LoggedInUser.kt
```diff
@@ -32,8 +41,21 @@ data class LoggedInUser(
...
+    WAITING_FOR_APPROVAL;
+
+    class RoomTypeConverter : RoomEnumTypeConverter<Status>(Status::class.java)
+  }
+
+  @Dao
+  interface RoomDao {
+
+    @Query("SELECT * FROM LoggedInUser")
+    fun user(): Flowable<List<LoggedInUser>>
```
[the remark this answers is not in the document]
saket: Room does not support Maybe unfortunately.

app/src/main/java/org/simple/clinic/AppDatabase.kt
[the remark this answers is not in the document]
saket: Discussed in person: I'm not changing this because I copied this query from Room's generated code.

app/src/main/java/org/simple/clinic/user/UserSession.kt
saket: Yeah, `toSingleDefault` looks semantically more correct.

app/src/main/java/org/simple/clinic/user/UserSession.kt
saket: Sure done

app/src/main/java/org/simple/clinic/user/LoggedInUserHttpInterceptor.kt
saket: TODO: Separate discussion about if-else vs when.

app/src/main/java/org/simple/clinic/registration/pin/RegistrationPinScreen.kt
```diff
@@ -0,0 +1,74 @@
...
+        .compose(controller)
+        .observeOn(mainThread())
+        .takeUntil(RxView.detaches(this))
+        .subscribe { uiChange -> uiChange(this) }
+  }
+
+  private fun screenCreates() = Observable.just(RegistrationPinScreenCreated())
+
+  private fun pinTextChanges() =
+      RxTextView.textChanges(pinEditText)
```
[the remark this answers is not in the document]
saket: Sorry this somehow got missed. We'll use a custom widget that won't allow blank spaces, so I think this is fine for now?

app/src/main/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinScreen.kt
```diff
@@ -0,0 +1,72 @@
...
+      RxTextView.textChanges(pinEditText)
+          .map(CharSequence::toString)
+          .map(::RegistrationConfirmPinTextChanged)
+
+  private fun nextClicks() =
+      RxView.clicks(nextButton)
+          .map { RegistrationConfirmPinNextClicked() }
+
+  fun openFacilitySelectionScreen() {
+    Timber.w("TODO")
```
vinaysshenoy: @saket For unimplemented changes on the UI, can we show a toast that the feature is under construction so that we don't file it as a bug instead?

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -74,6 +84,39 @@ class UserSession @Inject constructor(
...
+    return loggedInUser()
+        .map { (user) -> user }
+        .firstOrError()
+        .map(::RegistrationRequest)
+        .flatMap { registrationApi.createUser(it) }
+        .flatMap<RegistrationResult> {
+          storeUser(it.loggedInUser)
+              .andThen(Single.just(RegistrationResult.Success()))
+        }
+        .onErrorReturn { RegistrationResult.Error() }
```
vinaysshenoy: How do we handle the case where the server fails for some reason (Unreachable/500 error) which means the server might be able to handle the request if attempted again after a while?

The current implementation will not retry the request in these cases and will never attempt to log the user back in again. We need to separate errors that we can't recover from (say missing facility) from the errors that we can recover from. @saket @govindkrjoshi Thoughts?

The updated flow for self registration converts this call to a blocking one, so I'm guessing this is acceptable for now, but we need to figure out how to handle these cases.

app/src/main/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinScreen.kt
saket: We'll open home screen instead

PR #171, opened 2018-07-30
[WIP] 30th July 2018 - Updated layout for App Lock Screen

vinaysshenoy:


saket:
1. The distance between `FORGOT PIN?` button and the card's bottom doesn't match Zeplin
2. The `NOT YOU?` button doesn't match Zeplin

app/src/main/java/org/simple/clinic/widgets/Views.kt
```diff
@@ -109,3 +111,12 @@ fun View.topRelativeTo(superParent: ViewGroup): Int {
...
 }
+
+fun View.showToast(@StringRes stringResId: Int, longToast: Boolean = false, vararg formatArgs: Any) {
+  showToast(context.getString(stringResId, formatArgs), longToast)
+}
+
+fun View.showToast(msg: CharSequence, longToast: Boolean = false) {
+  val toastLength = if (longToast) Toast.LENGTH_LONG else Toast.LENGTH_SHORT
+  Toast.makeText(context, msg, toastLength).show()
+}
```
saket: Based on the rules used by Jake and other team members of android-ktx, I think this is a wrong usage of extension function. 

https://www.youtube.com/watch?v=st1XVfkDWqk

app/src/test/java/org/simple/clinic/login/applock/AppLockScreenControllerTest.kt
```diff
@@ -88,4 +88,18 @@ class AppLockScreenControllerTest {
     uiEvents.onNext(AppLockScreenPinTextChanged("0"))
     verify(screen).setIncorrectPinErrorVisible(false)
   }
+
+  @Test
+  fun `when logout is clicked, the ui action should be triggered`() {
```
saket: Hmmm… `the ui action` sounds wrong

vinaysshenoy: Agreed. I just added it in here because I wanted to verify that the ui events are getting triggered. I plan to rename it when we actually implement the feature.

app/src/main/java/org/simple/clinic/widgets/Views.kt
vinaysshenoy: Got it. The core ktx [package](https://developer.android.com/reference/kotlin/androidx/core/widget/package-summary#extension-functions-summary) has an extension method for toasting via the `Context`. I can either add the library or implement the same signature myself.

Since we already use `work-runtime-ktx`, I guess we can use the core ktx library as well. Thoughts?

saket: Let's add android-ktx

vinaysshenoy:
@saket 
- Fixed design inconsistencies
- I added `core-ktx`, but that had deprecated the `toast` methods in the latest version. I decided to directly call `Toast.makeText().show()` in the `AppLockScreen` and did not include `core-ktx`.

saket:
Hey @vinaysshenoy I don't see android-ktx in this PR. Is it a WIP?

Btw, I was wondering if we should also add the change-facility button before you move to some other story. When we add a screen for changing facilities in the future, we won't have to update this screen again. Thoughts?

app/src/main/java/org/simple/clinic/login/applock/AppLockScreen.kt
```diff
@@ -103,7 +121,7 @@ class AppLockScreen(context: Context, attrs: AttributeSet) : RelativeLayout(cont
   fun setIncorrectPinErrorVisible(show: Boolean) {
     when (show) {
       true -> errorTextView.visibility = View.VISIBLE
-      else -> errorTextView.visibility = View.GONE
+      else -> errorTextView.visibility = View.INVISIBLE
```
saket: The large gap between the pin text field and forgot-pin button doesn't look very good now that I'm running this new design on my device.

saket:
Closing this for now because the design may change.

PR #174, opened 2018-08-01
Remove apostrophe and period in Patient Search

vinaysshenoy:
## Changes

- Changed the patient table's `searchableName` conversion to include periods and apostrophes
- Added a migration to change all stored searchable names to the newer versions stripped of periods and apostrophes as well
- Moved conversion of searchable name to a top-level helper function in a separate file


build.gradle
```diff
@@ -11,7 +11,7 @@ buildscript {
       dagger              : '2.16',
       butterKnife         : '8.8.1',
       kotterKnife         : 'e157638df1',
-      room                : '1.1.0',
+      room                : '1.1.1',
```
vinaysshenoy: There seems to be a [bug in Room 1.1.0](https://stackoverflow.com/questions/50420649/upgrading-room-db-with-migrations-causes-db-lock) where the database gets locked once migrations are running. Based on the SO comments in the link, upgrading to `1.1.1` fixes the issue.

vinaysshenoy:
@saket @govindkrjoshi PR is ready for review.

app/src/main/java/org/simple/clinic/patient/PatientSearchHelpers.kt
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * [Regex] for stripping patient names and search queries of white spaces and punctuation
+ *
+ * Currently matches the following characters
+ * - Any whitespace
+ * - Comma, Hyphen, SemiColon, Colon, Underscore, Apostrophe, Period
+ * */
+private val spacePunctuationRegex = Regex("[\\s;_\\-:,'\\\\.]")
+
+fun convertNameToSearchableForm(string: String) = string.replace(spacePunctuationRegex, "")
```
saket: @vinaysshenoy why did you decide to move these outside of `PatientRepository`?

Asking because I really dislike creating `Utils` and `Helper` classes. I try hard to avoid them 😌

saket: What do you think of renaming this function to just `nameToSearchableForm()`? Similar to avoiding function names starting with `get`, `set`, `create`, etc., I've started avoiding verb prefixes.

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -77,4 +78,43 @@ abstract class AppDatabase : RoomDatabase() {
         """)
     }
   }
+
+  class Migration_4_5 : Migration(4, 5) {
+
+    override fun migrate(database: SupportSQLiteDatabase) {
+      // Update local searchable name in the Patient table to strip out the newly added characters
+      database.inTransaction {
+        compileStatement("""update "Patient" set "searchableName"=? where "uuid"=?""")
```
saket: You don't need three quotes for single line statements right?

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -77,4 +78,43 @@ abstract class AppDatabase : RoomDatabase() {
...
+                        }
+                  }
+            }
+      }
+    }
+  }
 }
+
+/**
+ * Execute the given block in a transaction. Will call [SupportSQLiteDatabase.setTransactionSuccessful] only if no errors were thrown within the block
```
saket: Can you break this comment into multiple lines? There's no strict rule on this, but we try keeping a maximum of 13 words in a line for readability.

saket: Can you also end this comment with a period?

saket: Actually, I just read the code. I think this comment is documenting *what* the code is doing. Considering the code is very simple to understand, I don't think we need this comment here :)

saket:
Btw, @vinaysshenoy can you configure IntelliJ to add a line break at the end of every file automatically? Otherwise every app (SourceTree, Git, Github) show a warning that the linebreak is missing.

`Preferences > Editor > General > Ensure line feed at file end on save`

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -77,4 +78,43 @@ abstract class AppDatabase : RoomDatabase() {
...
+        compileStatement("""update "Patient" set "searchableName"=? where "uuid"=?""")
+            .use { statement ->
+              query("""select "uuid","fullName" from "Patient"""")
+                  .use { cursor ->
+                    val uuidIndex = cursor.getColumnIndex("uuid")
+                    val fullNameIndex = cursor.getColumnIndex("fullName")
+
+                    generateSequence { if (cursor.moveToNext()) cursor else null }
+                        .map { it.getString(uuidIndex) to it.getString(fullNameIndex) }
+                        .map { (uuid, fullName) -> uuid to convertNameToSearchableForm(fullName) }
```
saket: I don't have any strong opinion, but what do you think of merging these two maps into one?

```
.map { getString(uuidIndex) to convertNameToSearchableForm(getString(fullNameIndex)) }
```
https://blog.danlew.net/2018/02/20/when-you-dont-need-a-map/

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -77,4 +78,43 @@ abstract class AppDatabase : RoomDatabase() {
...
+                        .forEach { (uuid, searchableName) ->
+                          statement.bindString(1, searchableName)
+                          statement.bindString(2, uuid)
+                          statement.executeUpdateDelete()
+                        }
+                  }
+            }
```
saket: These `use` blocks are adding a lot of nesting + I don't think you're taking advantage of them. Will discuss with you in person.

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -77,4 +78,43 @@ abstract class AppDatabase : RoomDatabase() {
...
+  }
 }
+
+/**
+ * Execute the given block in a transaction. Will call [SupportSQLiteDatabase.setTransactionSuccessful] only if no errors were thrown within the block
+ **/
+private fun SupportSQLiteDatabase.inTransaction(block: SupportSQLiteDatabase.() -> Unit) {
+  try {
+    beginTransaction()
+    block.invoke(this)
```
saket: Isn't `invoke` optional? Can this be replaced by just `block(this)`?

saket: Nitpick: I think `beginTransaction()` can be moved out of the `try` block

app/src/main/java/org/simple/clinic/patient/PatientSearchHelpers.kt
vinaysshenoy: It's because I needed to access the function in the migration, and I didn't want to copy the same function/regex in the migration or access the repository from the migration.

vinaysshenoy: Maybe a good place to put them would be in the top-level of the `Patient.kt` file since it is the entity.

app/src/main/java/org/simple/clinic/AppDatabase.kt
vinaysshenoy: I need it because I am using `"` for SQLite identifiers and I don't want to escape them in single line strings. I guess I could use backticks as well and make them single line strings.

app/src/main/java/org/simple/clinic/AppDatabase.kt
vinaysshenoy: The `use` blocks automatically close any `Closeable` instances, like the `Cursor` and the `SQLiteStatement` for you.

PR #85, opened 2018-06-15
Patient summary: BP history

saket:
Part of #25

PR #178, opened 2018-08-03
Implemented Fuzzy Search in Patient Lookup

vinaysshenoy:


app/src/main/java/org/simple/clinic/di/SimpleSqliteOpenHelperFactory.kt
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * **Note:** There is a reason why the configuration for in-memory databases was moved to this class
+ * instead of via Room's database builder. `sqlite-android`'s default in-memory implementation ignores
+ * all dynamic extensions in the configuration and hence, testing the databases while using the extension
+ * is a lot more problematic.
+ *
+ * However, by making it a normal database in Room, and overriding the configuration provided in this
+ * class, we can make it an in-memory database with the extension loaded
+ **/
+class SimpleSqliteOpenHelperFactory(inMemory: Boolean = false) : SupportSQLiteOpenHelper.Factory {
```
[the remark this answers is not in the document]
vinaysshenoy: Yes. Will change it to `AppSqliteOpenHelperFactory`.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -36,19 +37,30 @@ class PatientRepository @Inject constructor(
...
+          .search(actualQuery)
+          .toObservable()
+    } else {
+      database.patientSearchDao()
+          .search(actualQuery)
+          .zipWith(database.fuzzyPatientSearchDao().searchForPatientsWithNameLike(actualQuery).toFlowable()) { searchResults, fuzzySearchResults ->
+            (fuzzySearchResults + searchResults).distinctBy { it.uuid }
+          }
+          .toObservable()
+    }
```
saket: Can you extract the argument to `zipWith()` as a separate value? There's a lot happening in a single line there. I tried moving things around on my machine:

```kotlin
val fuzzySearch = database.fuzzyPatientSearchDao()
    .searchForPatientsWithNameLike(actualQuery)
    .toObservable()

database.patientSearchDao()
    .search(actualQuery)
    .toObservable()
    .zipWith(fuzzySearch)
    .map { (results, fuzzyResults) -> (fuzzyResults + results).distinctBy { it.uuid } }
```

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -61,9 +73,20 @@ class PatientRepository @Inject constructor(
...
+          .search(actualQuery, dateOfBirthUpperBound, dateOfBirthLowerBound)
+          .toObservable()
+    } else {
+      database.patientSearchDao()
+          .search(actualQuery, dateOfBirthUpperBound, dateOfBirthLowerBound)
+          .zipWith(database.fuzzyPatientSearchDao().searchForPatientsWithNameLikeAndAgeWithin(actualQuery, dateOfBirthUpperBound, dateOfBirthLowerBound).toFlowable()) { searchResults, fuzzySearchResults ->
+            (fuzzySearchResults + searchResults).distinctBy { it.uuid }
+          }
+          .toObservable()
+    }
```
saket: Same question here. Not only is there too much happening at line [#85](https://github.com/simpledotorg/simple-android/blob/8b65e3b62e3e44d6ba437b4f13b20fa507da18e3/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt#L85), but you have also gone past the margin.

What do you think about this?

```kotlin
val fuzzySearch = database.fuzzyPatientSearchDao()
    .searchForPatientsWithNameLikeAndAgeWithin(actualQuery, dateOfBirthUpperBound, dateOfBirthLowerBound)
    .toObservable()

database.patientSearchDao()
    .search(actualQuery, dateOfBirthUpperBound, dateOfBirthLowerBound)
    .toObservable()
    .zipWith(fuzzySearch)
    .map { (results, fuzzyResults) -> (fuzzyResults + results).distinctBy { it.uuid } }
```

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -95,6 +118,12 @@ class PatientRepository @Inject constructor(
 
   private fun savePatient(patient: Patient): Completable {
     return Completable.fromAction { database.patientDao().save(patient) }
+        .andThen(database.fuzzyPatientSearchDao().updateFuzzySearchTableForPatients(listOf(patient.uuid)))
+  }
+
+  private fun savePatients(patients: List<Patient>): Completable {
+    return Completable.fromAction { database.patientDao().save(patients) }
+        .andThen(database.fuzzyPatientSearchDao().updateFuzzySearchTableForPatients(patients.map { it.uuid }.distinct()))
```
saket: Hrm, why do we need `distinct` here? Don't we always get distinct patients for saving them?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -124,7 +153,7 @@ class PatientRepository @Inject constructor(
 
             val newOrUpdatedPatients = payloads.map { it.toDatabaseModel(newStatus = DONE) }
 
-            database.patientDao().save(newOrUpdatedPatients)
+            savePatients(newOrUpdatedPatients).blockingAwait()
```
saket: Whoaaa. I'm *very* paranoid of using `blockingAwait()` calls.

1. It's possible to trigger a race condition. I cannot recall the situations, but I think it can be done by calling await() on the main thread or something. 

2. It feels very wrong to mix async and sync code together.

Considering that this `flatMap` is doing multiple things now, my favorite way of solving this problem is by splitting the stream into separate jobs. What do you think?

```kotlin
fun mergeWithLocalData(serverPayloads: List<PatientPayload>): Completable {
  val payloadsToSave = serverPayloads
      .toObservable()
      .filter { payload ->
        val localCopy = database.patientDao().getOne(payload.uuid)
        localCopy?.syncStatus.canBeOverriddenByServerCopy()
      }
      .toList()
      .cache()

  val saveAddresses = payloadsToSave
      .map { payloads -> payloads.map { it.address.toDatabaseModel() } }
      .flatMapCompletable { addresses ->
        Completable.fromAction {
          database.addressDao().save(addresses)
        }
      }

  val savePatients = payloadsToSave
      .map { payloads -> payloads.map { it.toDatabaseModel(newStatus = DONE) } }
      .flatMapCompletable { patients -> savePatients(patients) }

  val savePhoneNumbers = payloadsToSave
      .map { payloads ->
        payloads
            .filter { it.phoneNumbers != null }
            .flatMap { payload ->
              payload.phoneNumbers!!.map { it.toDatabaseModel(payload.uuid) }
            }
      }
      .filter { numbers -> numbers.isNotEmpty() }
      .flatMapCompletable { numbers ->
        Completable.fromAction {
          database.phoneNumberDao().save(numbers)
        }
      }

  return saveAddresses.andThen(savePatients).andThen(savePhoneNumbers)
}
```

The obvious drawback is that this function is very long now.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -36,19 +37,30 @@ class PatientRepository @Inject constructor(
...
+          .search(actualQuery)
+          .zipWith(database.fuzzyPatientSearchDao().searchForPatientsWithNameLike(actualQuery).toFlowable()) { searchResults, fuzzySearchResults ->
+            (fuzzySearchResults + searchResults).distinctBy { it.uuid }
+          }
+          .toObservable()
+    }
   }
 
-  fun searchPatientsAndPhoneNumbers(query: String?, assumedAge: Int): Observable<List<PatientSearchResult>> {
+  fun searchPatientsAndPhoneNumbers(query: String?, assumedAge: Int, includeFuzzyNameSearch: Boolean = true): Observable<List<PatientSearchResult>> {
```
saket: Should we move params into separate lines? 

Same question for the other `searchPatientsAndPhoneNumbers()` function.

app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
```diff
@@ -0,0 +1,122 @@
...
+            val wordIndex = it.getColumnIndex("word")
+            val uuidIndex = it.getColumnIndex("uuid")
+
+            generateSequence { it.takeIf { it.moveToNext() } }
+                .map { FuzzySearchResult(it.getLong(rowIdIndex), UUID.fromString(it.getString(uuidIndex)), it.getString(wordIndex)) }
+                .toList()
+          }
+    }!!
+
+    override fun updateFuzzySearchTableForPatients(uuids: List<UUID>) =
```
saket: Considering that these functions can only be called on the object of `PatientFuzzySearch`, can we shorten their names to `updateTableForPatients()` and so on?

app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
```diff
@@ -0,0 +1,122 @@
...
+    fun searchForPatientsWithNameLikeAndAgeWithin(query: String, dobUpperBound: String, dobLowerBound: String): Single<List<PatientSearchResult>>
+  }
+
+  class PatientFuzzySearchDaoImpl(
+      private val sqLiteOpenHelper: SupportSQLiteOpenHelper,
+      private val patientSearchDao: PatientSearchResult.RoomDao
+  ) : PatientFuzzySearchDao {
+
+    override fun savedEntries() = Single.fromCallable {
+      sqLiteOpenHelper.readableDatabase.query("""
```
saket: I just realized there's one small drawback of not using block function bodies. I cannot immediately tell what the return type of this function is without reading the code.

app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
```diff
@@ -0,0 +1,122 @@
...
+            val wordIndex = it.getColumnIndex("word")
+            val uuidIndex = it.getColumnIndex("uuid")
+
+            generateSequence { it.takeIf { it.moveToNext() } }
+                .map { FuzzySearchResult(it.getLong(rowIdIndex), UUID.fromString(it.getString(uuidIndex)), it.getString(wordIndex)) }
+                .toList()
+          }
+    }!!
+
+    override fun getEntriesForIds(uuids: List<UUID>) = Single.fromCallable {
```
saket: Considering that you've used `patient` in the name of other functions here, can we also rename this to `getEntriesForPatientIds()`?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: Yes, this looks a lot nicer. I'll do it this way.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: Oh! I added it in there because I wanted to see results on the UI first, and I planned to clean up the chain later. Guess I should've added a task somewhere.

Will fix this.

app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
vinaysshenoy: It's not a problem with single line functions, it's because of type inference.

We can just as easily do a
```kotlin
override fun savedEntries(): Single<List<FuzzySearchResult>> = Single.fromCallable {} 
```

We can add this in, but I feel it's not important in this case because it already implements an interface which has the function return type defined.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: Right, good point. I don't know why I added that; maybe it was because I adding `distinct` clauses when merging the patient search results. I'll remove this.

app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
saket: Hrm, that's also valid.

app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
```diff
@@ -0,0 +1,160 @@
...
+          val rowId = it.getColumnIndex("rowid")
+          val word = it.getColumnIndex("word")
+          val rank = it.getColumnIndex("rank")
+          val distance = it.getColumnIndex("distance")
+          val score = it.getColumnIndex("score")
+          val matchLength = it.getColumnIndex("matchlen")
+          val scope = it.getColumnIndex("scope")
+          generateSequence { it.takeIf { it.moveToNext() } }
+              .map {
+                FuzzySearchResult(
```
saket: Hey, this code looks new. Do you think we can move this mapper function to the model? This idea is from `Sqlbrite`, where the model implements a mapping function and you just call `FuzzySearchResult.mapFrom(cursor)` or something.

app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
```diff
@@ -0,0 +1,160 @@
...
+    fun searchForPatientsWithNameLikeAndAgeWithin(query: String, dobUpperBound: String, dobLowerBound: String): Single<List<PatientSearchResult>>
+  }
+
+  class PatientFuzzySearchDaoImpl(
+      private val sqLiteOpenHelper: SupportSQLiteOpenHelper,
+      private val patientSearchDao: PatientSearchResult.RoomDao
+  ) : PatientFuzzySearchDao {
+
+    private val fuzzyColumnsToFetch = """
+      "P"."uuid" "uuid", "PFS"."rowid" "rowid","PFS"."word" "word","PFS"."rank" "rank","PFS"."distance" "distance","PFS"."score" "score","PFS"."matchlen" "matchlen","PFS"."scope" "scope"
```
saket: This has gone beyond the line margin. Can we break this into multiple lines?

vinaysshenoy: Noted.

app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
vinaysshenoy: Noted.

PR #179, opened 2018-08-03
Export Room schemas from AppDatabase

vinaysshenoy:
# Changes
We may not do the DB migration testing right now, but since we are actively writing migrations, we should be exporting schemas so that we don't have to retro-actively generate too many schemas in the future.

- Export schemas from `AppDatabase`. 
- Add schemas to the `androidTest` source set

See https://developer.android.com/training/data-storage/room/migrating-db-versions

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -16,6 +16,12 @@ sentry {
   autoUpload false
 }
 
+kapt {
+  arguments {
+    arg("room.schemaLocation", "$projectDir/schemas".toString())
+  }
+}
+
 android {
   compileSdkVersion versions.compileSdk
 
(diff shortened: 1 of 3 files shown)


app/build.gradle
```diff
@@ -120,6 +130,7 @@ dependencies {
   androidTestImplementation "com.android.support.test:rules:$versions.supportTest"
   androidTestImplementation "com.google.truth:truth:$versions.truth"
   androidTestImplementation "com.github.blocoio:faker:$versions.faker"
+  androidTestImplementation "android.arch.persistence.room:testing:$versions.room"
```
saket: This is only required for writing migration Android tests, but sure let's keep this

PR #184, opened 2018-08-08
Improve Faker and PatientFaker implementation in tests

vinaysshenoy:
## Changes
- Inject `Faker` and `PatientFaker` via Dagger
- Change injected `Faker` instance to generate data specific to India

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/PatientFaker.kt
+++ b/app/src/androidTest/java/org/simple/clinic/PatientFaker.kt
@@ -1,6 +1,7 @@
 package org.simple.clinic
 
 import io.bloco.faker.Faker
+import org.simple.clinic.di.AppScope
 import org.simple.clinic.patient.Gender
 import org.simple.clinic.patient.PatientPhoneNumberType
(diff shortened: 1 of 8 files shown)


app/src/androidTest/java/org/simple/clinic/PatientFaker.kt
```diff
@@ -9,10 +10,10 @@ import org.simple.clinic.patient.sync.PatientPayload
 import org.simple.clinic.patient.sync.PatientPhoneNumberPayload
 import org.threeten.bp.Instant
 import java.util.UUID
+import javax.inject.Inject
 
-object PatientFaker {
-
-  private val faker = Faker()
+@AppScope
+class PatientFaker @Inject constructor(private val faker: Faker) {
```
saket: Are we making `PatientFaker` a part of our DI graph because it now depends on `Faker`?

vinaysshenoy: Yes.

saket:
LGTM 👍

PR #185, opened 2018-08-10
Fix deadlock in Room database threads when patient sync happens

vinaysshenoy:


app/src/main/java/org/simple/clinic/user/LoggedInUserHttpInterceptor.kt
```diff
@@ -9,12 +9,13 @@ import javax.inject.Inject
 class LoggedInUserHttpInterceptor @Inject constructor() : Interceptor {
 
   // Ugly hack to avoid a cyclic dependency between this class and UserSession.
-  private val userSession by lazy { ClinicApp.appComponent.userSession() }
+  private val userSession
+    get() = ClinicApp.appComponent.userSession()
```
vinaysshenoy: I changed this to a getter because lazy initialization is unnecessary. Dagger will handle it for us.

app/src/main/java/org/simple/clinic/user/LoggedInUser.kt
```diff
@@ -59,9 +59,12 @@ data class LoggedInUser(
   @Dao
   interface RoomDao {
 
-    @Query("SELECT * FROM LoggedInUser")
+    @Query("SELECT * FROM LoggedInUser LIMIT 1")
     fun user(): Flowable<List<LoggedInUser>>
```
saket: Why have a limit when we can only have one row in this table ever?

app/src/main/java/org/simple/clinic/user/LoggedInUserHttpInterceptor.kt
saket: Looks better 👍

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -182,6 +183,14 @@ class UserSession @Inject constructor(
         .map { if (it.isEmpty()) None else Just(it.first()) }
   }
 
+  // FIXME: Figure out a better way to add access tokens to network calls in a reactive fashion
+  // FIXME: Maybe add a separate RxCallAdapterFactory that lets us transform requests without interceptors?
+  // This was added because we needed to call it from the OkHttp Interceptor
+  // in a synchronous fashion because the Rx - blocking() call used earlier
+  // was causing a deadlock in the Room threads when patient sync happened
```
saket: I think we can replace `patient sync happened` with `data sync happened`

saket:
LGTM

app/src/main/java/org/simple/clinic/user/LoggedInUser.kt
vinaysshenoy: No specific reason. I'd say it tells me at a glance when I read the SQL statement that I expect only one row to be present.

PR #186, opened 2018-08-14
14 Aug - Login

vinaysshenoy:


app/src/main/java/org/simple/clinic/di/NetworkModule.kt
```diff
@@ -27,10 +28,11 @@ class NetworkModule {
   }
 
   @Provides
-  fun retrofitBuilder(moshi: Moshi, loggedInInterceptor: LoggedInUserHttpInterceptor): Retrofit.Builder {
+  fun retrofitBuilder(moshi: Moshi, loggedInInterceptor: LoggedInUserHttpInterceptor, otpRequestInterceptor: TemporaryValidateOtpRequestInterceptor): Retrofit.Builder {
     val okHttpBuilder = OkHttpClient.Builder()
 
     okHttpBuilder.addInterceptor(loggedInInterceptor)
+    okHttpBuilder.addInterceptor(otpRequestInterceptor)
```
saket: Instead of adding this to the global retrofit builder, do you want to keep this logic private to `LoginModule`?

app/src/main/java/org/simple/clinic/login/TemporaryValidateOtpRequestInterceptor.kt
```diff
@@ -0,0 +1,65 @@
...
+    return when {
+      urlString.endsWith("v1/requestOtp") -> tempSendOtpResponse(originalRequest)
+      urlString.endsWith("v1/validateLoginOtp") -> tempValidateLoginOtp(originalRequest)
+      else -> chain.proceed(originalRequest)
+    }
+  }
+
+  private fun tempValidateLoginOtp(request: Request): Response {
+    // Some sleep to simulate a real network call
+    Thread.sleep(1000L)
```
saket: 😂

app/src/main/java/org/simple/clinic/login/TemporaryValidateOtpRequestInterceptor.kt
```diff
@@ -0,0 +1,65 @@
...
+            pinDigest = validateLoginOtpRequest.passwordDigest,
+            facilityUuids = listOf(UUID.randomUUID()),
+            createdAt = Instant.now(),
+            updatedAt = Instant.now(),
+            status = UserStatus.DISAPPROVED_FOR_SYNCING
+        )
+    )
+
+    return Response.Builder().code(200).body(ResponseBody.create(MediaType.parse("application/json"), moshi.adapter(LoginResponse::class.java)
+        .toJson(loginResponse))).build()
```
saket: I want to ask you to format your code even if it's temporary because we're working on an open source project and this will be visible to people outside. There are not a lot of eyes right now, but we will get there soon.

app/src/main/java/org/simple/clinic/login/phone/LoginPhoneScreen.kt
```diff
@@ -73,6 +74,7 @@ class LoginPhoneScreen(context: Context, attrs: AttributeSet) : RelativeLayout(c
   }
 
   fun openLoginPinScreen() {
-    screenRouter.push(LoginPinScreen.KEY)
+    val key = screenRouter.key<LoginPhoneScreenKey>(this)
+    screenRouter.push(LoginPinScreenKey(key?.otp ?: ""))
```
saket: You can safely assume that `key` will never be null here.

app/src/main/java/org/simple/clinic/login/pin/LoginPinScreen.kt
```diff
@@ -80,6 +77,15 @@ class LoginPinScreen(context: Context, attrs: AttributeSet) : RelativeLayout(con
...
 
+  private fun otpReceived() =
+      Single
+          .fromCallable {
+            val key = screenRouter.key<LoginPinScreenKey>(this)
+            key?.otp ?: ""
+          }
+          .map { LoginPinOtpReceived(it) }
+          .toObservable()
+
```
saket: I'm not going to ask you to change this because this is another throwaway piece of code, but this could have been reduced to 2 lines :p 

```
val key = screenRouter.key<LoginPinScreenKey>(this)
return Observable.just(LoginPinOtpReceived(key!!.otp))
```

app/src/main/java/org/simple/clinic/login/pin/LoginPinScreenController.kt
```diff
@@ -33,18 +35,22 @@ class LoginPinScreenController @Inject constructor(
...
         .map { it.pin }
 
+    val otpReceived = events.ofType<LoginPinOtpReceived>()
+        .map { it.otp }
+
     return events.ofType<PinSubmitClicked>()
         .withLatestFrom(pinChanges) { _, pin -> pin }
-        .flatMap { enteredPin ->
-          val cachedLogin = userSession.login()
+        .withLatestFrom(otpReceived) { pin, otp -> pin to otp }
```
saket: Did you forget that RxKotlin by default uses a `Pair`?

saket: Again, throwaay code so you could skip this ¯\_(ツ)_/

app/src/main/java/org/simple/clinic/sms/SmsReader.kt
```diff
@@ -0,0 +1,15 @@
+package org.simple.clinic.sms
+
+import io.reactivex.Single
+
+sealed class SmsReadResult {
+
+  data class SmsReadSuccess(val string: String) : SmsReadResult()
```
saket: Do you want to remove the prefix from these two classes? Because `SmsReadResult.SmsReadSuccess` sounds redundant?

app/src/main/java/org/simple/clinic/sms/SmsReaderImpl.kt
```diff
@@ -0,0 +1,10 @@
+package org.simple.clinic.sms
+
+import io.reactivex.Single
+import org.simple.clinic.activity.TheActivity
+import javax.inject.Inject
+
+class SmsReaderImpl @Inject constructor(private val activity: TheActivity) : SmsReader {
+
+  override fun waitForSms() = Single.just(SmsReadResult.SmsReadSuccess("") as SmsReadResult)!!
```
saket: I *think* this function name can be improved because we never use the term `wait` for asynchronous calls. But I'm sure you will also realize this once you use this function.

app/src/main/java/org/simple/clinic/login/TemporaryValidateOtpRequestInterceptor.kt
```diff
@@ -0,0 +1,65 @@
...
+      else -> chain.proceed(originalRequest)
+    }
+  }
+
+  private fun tempValidateLoginOtp(request: Request): Response {
+    // Some sleep to simulate a real network call
+    Thread.sleep(1000L)
+    val buffer = Buffer()
+    request.body()!!.writeTo(buffer)
+    val validateLoginOtpRequest = moshi.adapter(ValidateLoginOtpRequest::class.java).fromJson(buffer.readString(Charset.forName("UTF-8")))!!
```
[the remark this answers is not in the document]
vinaysshenoy: I didn't know that! I'll make the change!

app/src/main/java/org/simple/clinic/login/phone/LoginPhoneScreen.kt
vinaysshenoy: Then the screen router api should be changed to remote the nullable type.

app/src/main/java/org/simple/clinic/login/pin/LoginPinScreenController.kt
vinaysshenoy: Ah, yes. I'll make the change. Even if it's throwaway code, we should change it.

app/src/main/java/org/simple/clinic/di/NetworkModule.kt
vinaysshenoy: Whoops, we can't. The interceptor can only be added to the `Retrofit.Builder` instance and not to the `Retrofit` instance which `LoginModule` depends on.

app/src/main/java/org/simple/clinic/sms/SmsReaderImpl.kt
vinaysshenoy: Right. I'll revisit the naming once the SMS read functionality is complete and see if there can be a more appropriate name.

app/src/main/java/org/simple/clinic/login/phone/LoginPhoneScreen.kt
vinaysshenoy: I personally try to avoid the `!!` operator unless it's happening at the java-kotlin boundaries.

vinaysshenoy:
@saket PR feedback addressed.

app/src/main/java/org/simple/clinic/login/phone/LoginPhoneScreen.kt
saket: No, in this case it's impossible to hit this flow where the key will be null. Why would you still do a null check?

app/src/main/java/org/simple/clinic/di/NetworkModule.kt
saket: Ah okay. Not very nicely designed, our modules.

app/src/main/java/org/simple/clinic/login/phone/LoginPhoneScreen.kt
saket: TODO: Change `ScreenRouter#key()`

vinaysshenoy: Doing this on the 16-Aug PR.

PR #192, opened 2018-08-20
20 Aug - Added loggedInStatus column to LoggedInUser

vinaysshenoy:


saket:
@vinayshenoy can you change the voice of the commit message? Added -> Add

app/src/main/java/org/simple/clinic/user/User.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    val updatedAt: Instant,
+
+    val loggedInStatus: LoggedInStatus
+) {
+
+  enum class LoggedInStatus {
+    // Indicates a phone number match happened on the server,
+    // and information was stored locally, but the otp request has not yet been made
+    NOT_SIGNED_IN,
```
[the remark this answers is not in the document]
saket: I also think that `Indicates a/that` and other verbs should be avoided here. They are redundant. All code indicate something :p

app/src/main/java/org/simple/clinic/user/User.kt
```diff
@@ -0,0 +1,74 @@
...
+import android.arch.persistence.room.Query
+import io.reactivex.Flowable
+import org.simple.clinic.util.RoomEnumTypeConverter
+import org.threeten.bp.Instant
+import java.util.UUID
+
+// We intentionally have to keep the table name same because
+// Room starts complaining if you try to rename a table which is referenced
+// by another table in a foreign key
+// (technically, SQLite supports renaming tables, but Room complains)
```
saket: Nitpick: Can I also ask you to use periods in comments?

saket: Nitpick: Justified text looks even better:

```
// We intentionally have to keep the table name same because Room 
// starts complaining if you try to rename a table which is referenced
// by another table in a foreign key. Technically, SQLite supports 
// renaming tables, but Room complains.
```

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -177,6 +178,19 @@ abstract class AppDatabase : RoomDatabase() {
       }
     }
   }
+
+  /**
+   * - Adds the column `loggedInStatus` to the `LoggedInUser` table
```
saket: Leftover `-`?

saket:
@vinaysshenoy do you think `AuthStatus` is more descriptive than `LoggedInStatus`?

app/src/androidTest/java/org/simple/clinic/PatientFaker.kt
```diff
@@ -99,16 +99,18 @@ class PatientFaker @Inject constructor(private val faker: Faker) {
       name: String = faker.name.name(),
       phone: String = faker.phoneNumber.phoneNumber(),
       pinDigest: String = "pin-digest",
-      status: UserStatus = UserStatus.values().toList().shuffled().first()
-  ): LoggedInUser {
-    return LoggedInUser(
+      status: UserStatus = UserStatus.values().toList().shuffled().first(),
+      loggedInStatus: User.LoggedInStatus = User.LoggedInStatus.values().toList().shuffled().first()
```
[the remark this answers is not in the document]
vinaysshenoy: Extension isn't possible since the `Enum` class doesn't define a companion. I ended up adding a `randomOfEnum(KClass<Enum<T>>)` function to the file.

PR #204, opened 2018-08-23
Fix a crash when completing registration after selecting facilities 

saket:
RegistrationPhoneScreen wasn't recreating the ongoing registration entry once it was cleared. This was happening because UserSession#clearOngoingRegistrationEntry() and isOngoingRegistrationEntryPresent() were not in sync. 

Steps for reproducing the crash:

1. Open registration screen and enter the phone number of an existing user
2. The login-pin screen should open. Go back to the registration-phone screen
3. Proceed to facility selection and press done to complete registration

### DIFF
--- a/app/src/test/java/org/simple/clinic/user/UserSessionTest.kt
+++ b/app/src/test/java/org/simple/clinic/user/UserSessionTest.kt
@@ -167,6 +167,16 @@ class UserSessionTest {
   }
 
   @Test
+  fun `when ongoing registration entry is cleared then isOngoingRegistrationEntryPresent() should emit false`() {
+    userSession.saveOngoingRegistrationEntry(OngoingRegistrationEntry())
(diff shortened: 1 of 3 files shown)


app/src/test/java/org/simple/clinic/user/UserSessionTest.kt
```diff
@@ -164,4 +163,14 @@ class UserSessionTest {
       assertThat(it.status).isEqualTo(UserStatus.APPROVED_FOR_SYNCING)
     })
   }
+
+  @Test
+  fun `when ongoing registration entry is cleared then isOngoingRegistrationEntryPresent() should emit false`() {
```
saket: Oh hey I managed to break Github's syntax highlighter

saket:
@pratul Although I've already tested once, please run through the registration flow to ensure this does not break anything.

PR #207, opened 2018-08-24
Sync patient appointments

saket:


app/src/androidTest/java/org/simple/clinic/PatientFaker.kt
```diff
@@ -147,4 +151,31 @@ class PatientFaker @Inject constructor(private val faker: Faker) {
         facilityUuid = TestClinicApp.qaUserFacilityUuid(),
         patientUuid = patientUuid)
   }
+
+  fun followUpSchedule(): FollowUpSchedule {
```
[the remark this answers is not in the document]
saket: We'll change this function whenever needed?

app/src/main/java/org/simple/clinic/sync/DataSync.kt
```diff
@@ -0,0 +1,92 @@
...
+import io.reactivex.functions.Consumer
+import org.simple.clinic.patient.SyncStatus
+import org.simple.clinic.util.Just
+import org.simple.clinic.util.Optional
+import org.threeten.bp.Instant
+import timber.log.Timber
+import java.util.UUID
+import javax.inject.Inject
+
+interface Synceable
```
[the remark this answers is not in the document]
saket: Fixing this

app/src/androidTest/java/org/simple/clinic/MigrationAndroidTest.kt
```diff
@@ -97,4 +97,18 @@ class MigrationAndroidTest {
       assertThat(it.string("loggedInStatus")).isEqualTo("LOGGED_IN")
     }
   }
+
+  @Test
+  fun migration_8_to_9() {
+    helper.createDatabase(TEST_DB_NAME, 8)
+    val db_v9 = helper.runMigrationsAndValidate(TEST_DB_NAME, 9, true, AppDatabase.Migration_8_9())
+
+    db_v9.query("""SELECT * FROM FollowUp""").use {
```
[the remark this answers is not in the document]
saket: Fixing this

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -191,6 +197,25 @@ abstract class AppDatabase : RoomDatabase() {
       }
     }
   }
+
+  class Migration_8_9 : Migration(8, 9) {
```
vinaysshenoy: @pratul, @sanchita and I had a discussion yesterday about placing each migration in a separate file. Can you do this as part of this PR?

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
```diff
@@ -0,0 +1,82 @@
...
+import java.util.UUID
+
+@Entity(tableName = "Appointment")
+data class Appointment(
+    @PrimaryKey val id: UUID,
+    val patientId: UUID,
+    val facilityId: UUID,
+    val date: LocalDate,
+    val status: Status,
+    val statusReason: StatusReason,
```
vinaysshenoy: This should be `cancelReason` and not `statusReason`. This will also be set **ONLY** when an appointment is cancelled, so we will either have to
- Make it `nullable`, or
- Have a separate enum value which indicates no reason, which will be used when the appointment is in the `Scheduled` or `Visited` state.

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
```diff
@@ -0,0 +1,82 @@
...
+    @Json(name = "cancelled")
+    CANCELLED,
+
+    @Json(name = "visited")
+    VISITED;
+
+    class RoomTypeConverter : RoomEnumTypeConverter<Status>(Status::class.java)
+  }
+
+  enum class StatusReason {
```
vinaysshenoy: This should be `CancelReason`.

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
```diff
@@ -0,0 +1,82 @@
...
+    MOVED,
+
+    @Json(name = "dead")
+    DEAD;
+
+    class RoomTypeConverter : RoomEnumTypeConverter<StatusReason>(StatusReason::class.java)
+  }
+
+  @Dao
+  interface RoomDao {
```
vinaysshenoy: Should we start writing tests for the SQL queries that we run?

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
```diff
@@ -0,0 +1,82 @@
...
+    @Query("UPDATE Appointment SET syncStatus = :to WHERE id IN (:ids)")
+    fun updateSyncStatus(ids: List<UUID>, to: SyncStatus)
+
+    @Query("SELECT * FROM Appointment WHERE id = :id LIMIT 1")
+    fun getOne(id: UUID): Appointment?
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    fun save(schedules: List<Appointment>)
+
+    @Query("SELECT COUNT(*) FROM Appointment")
```
vinaysshenoy: We should just use the `appointmentId` in the `COUNT` expression instead of all fields.

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentSyncAndroidTest.kt
```diff
@@ -0,0 +1,100 @@
...
+  lateinit var repository: AppointmentRepository
+
+  @Inject
+  lateinit var dao: Appointment.RoomDao
+
+  @Inject
+  lateinit var userSession: UserSession
+
+  @Inject
+  @field:[Named("last_appointment_pull_timestamp")]
```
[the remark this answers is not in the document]
saket: Copy pasta. Fixing this.

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentSyncAndroidTest.kt
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  private fun insertDummyRecords(count: Int): Completable {
+    return Observable.range(0, count)
+        .map { patientFaker.appointment(syncStatus = SyncStatus.PENDING) }
+        .toList()
+        .flatMapCompletable { repository.save(*it.toTypedArray()) }
+  }
+
+  private fun dummySchedulePayloads(count: Int) = (0..count).map { patientFaker.appointmentPayload() }
```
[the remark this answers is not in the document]
saket: Ah, you're right. I overlooked this.

app/src/main/java/org/simple/clinic/AppDatabase.kt
saket: Should we then also keep the test for every migration in a different class?

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
saket: > This should be cancelReason and not statusReason

Backend seems to be using `statusReason` so I copied the same name.

> Have a separate enum value which indicates no reason, which will be used when the appointment is in the Scheduled or Visited state.

Right, so `NOT_CALLED_YET` is the default enum

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
saket: Will they not get tested as part of other tests? Like `AppointmentSyncAndroidTest` in this case

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
saket: Sure

app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
```diff
@@ -0,0 +1,60 @@
...
+import io.reactivex.Single
+import org.simple.clinic.patient.SyncStatus
+import org.simple.clinic.patient.canBeOverriddenByServerCopy
+import org.simple.clinic.sync.SynceableRepository
+import java.util.UUID
+import javax.inject.Inject
+
+class AppointmentRepository @Inject constructor(val dao: Appointment.RoomDao) : SynceableRepository<Appointment, AppointmentPayload> {
+
+  fun save(vararg schedules: Appointment): Completable {
```
vinaysshenoy: Since we are using `List` everywhere, can we use lists here as well instead of `vararg`?

saket: Hrm, I chose to use vararg because I want to be able to send both one item or multiple items. I could also just default to `List`

app/src/main/java/org/simple/clinic/overdue/AppointmentSync.kt
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  fun pull(): Completable {
+    return dataSync.pull(
+        repository = repository,
+        lastPullTimestamp = lastPullTimestamp,
+        pullNetworkCall = api::pull)
+  }
+
+  private fun toRequest(schedules: List<Appointment>): AppointmentPushRequest {
```
vinaysshenoy: Rename `schedules` -> `appointments`?

app/src/main/java/org/simple/clinic/AppDatabase.kt
vinaysshenoy: Not necessary at this point, I feel. Maybe in the future.

Let's drop this change from this PR for now, and do it in a  separate one.

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
vinaysshenoy: I am going by the model that was discussed on Slack. Since the usage of this field is actually a reason for cancellation, we should name it appropriately.

app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
vinaysshenoy: Right, I just want to have the same conventions everywhere.

app/src/main/java/org/simple/clinic/overdue/AppointmentSync.kt
saket: Whoops, running a find-all and fixing it everywhere.

app/src/main/java/org/simple/clinic/AppDatabase.kt
saket: Aye

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
saket: Valid point. I wonder why did nilenso decide to change it to `status_reason`. I will ask them once they're back. I just want to keep our model namings in sync.

PR #208, opened 2018-08-24
Add FollowUpSchedule and its stub repository

saket:


app/src/main/java/org/simple/clinic/overdue/FollowUpSchedule.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import android.arch.persistence.room.PrimaryKey
+import com.squareup.moshi.Json
+import org.simple.clinic.patient.SyncStatus
+import org.threeten.bp.Instant
+import org.threeten.bp.LocalDate
+import java.util.UUID
+
+data class FollowUpSchedule(
+    @PrimaryKey val id: UUID,
```
pratul: Why does it have `@PrimaryKey` if this is a data class?

saket: Copy pasta error

PR #210, opened 2018-08-27
Report all User Interaction events

vinaysshenoy:

PR #213, opened 2018-08-28
Screen Changes and User Interaction Event reporting

vinaysshenoy:
This includes #210, along with reporting all screen change events to Analytics

### DIFF
--- a/app/src/main/java/org/simple/clinic/activity/TheActivity.kt
+++ b/app/src/main/java/org/simple/clinic/activity/TheActivity.kt
@@ -29,6 +29,7 @@ import org.simple.clinic.router.screen.RouterDirection
 import org.simple.clinic.router.screen.ScreenRouter
 import org.simple.clinic.user.UserSession
 import org.simple.clinic.widgets.TheActivityLifecycle
+import timber.log.Timber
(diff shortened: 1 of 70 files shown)


app/src/main/java/org/simple/clinic/activity/TheActivity.kt
```diff
@@ -109,10 +110,16 @@ class TheActivity : AppCompatActivity() {
...
 
-    screenRouter.registerKeyChanger(FullScreenKeyChanger(this, android.R.id.content, R.color.window_background))
+    screenRouter.registerKeyChanger(FullScreenKeyChanger(this, android.R.id.content, R.color.window_background, this::onScreenChanged))
     return screenRouter.installInContext(baseContext, initialScreenKey())
   }
 
+  private fun onScreenChanged(outgoing: FullScreenKey?, incoming: FullScreenKey) {
+    val outgoingScreenName = if(outgoing == null) "" else outgoing.javaClass.simpleName
+    val incomingScreenName = incoming.javaClass.simpleName
+    Timber.d("$outgoingScreenName -> $incomingScreenName")
```
pratul: We shouldn't need this log anymore, right?

PR #222, opened 2018-08-30
Sync appointment communications

saket:


app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @Test
+  fun when_creating_new_appointment_then_the_appointment_should_be_saved() {
+    val patientId = UUID.randomUUID()
+    val appointmentDate = LocalDate.now()
+    repository.schedule(patientId, appointmentDate).blockingAwait()
+
+    val savedAppointment = repository.pendingSyncRecords().blockingGet()[0]
+    savedAppointment.apply {
```
vinaysshenoy: Why are we asserting individual properties instead of asserting the data class instance?

app/src/androidTest/java/org/simple/clinic/overdue/communication/CommunicationRepositoryAndroidTest.kt
```diff
@@ -0,0 +1,61 @@
+package org.simple.clinic.overdue.communication
```
vinaysshenoy: Non-static usages of `Truth` in this class.

app/src/androidTest/java/org/simple/clinic/overdue/communication/CommunicationRepositoryAndroidTest.kt
```diff
@@ -0,0 +1,61 @@
...
+  fun when_creating_new_communication_then_the_communication_should_be_saved() {
+    val appointmentId = UUID.randomUUID()
+
+    repository
+        .create(appointmentId, Communication.Type.MANUAL_CALL, result = Communication.Result.AGREED_TO_VISIT)
+        .blockingAwait()
+
+    val savedCommunication = repository.pendingSyncRecords().blockingGet()[0]
+    savedCommunication.apply {
+      Truth.assertThat(this.appointmentId).isEqualTo(appointmentId)
```
vinaysshenoy: Why asserting individual properties instead of the data class?

app/src/main/java/org/simple/clinic/overdue/communication/Communication.kt
```diff
@@ -0,0 +1,72 @@
...
+    @Query("UPDATE Communication SET syncStatus = :to WHERE id IN (:ids)")
+    fun updateSyncStatus(ids: List<UUID>, to: SyncStatus)
+
+    @Query("SELECT * FROM Communication WHERE id = :id LIMIT 1")
+    fun getOne(id: UUID): Communication?
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    fun save(schedules: List<Communication>)
+
+    @Query("SELECT COUNT(*) FROM Communication")
```
vinaysshenoy: Can we use just the primary key column when getting a count of rows?

app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
```diff
@@ -2,13 +2,40 @@ package org.simple.clinic.overdue
...
+    val dao: Appointment.RoomDao,
+    val userSession: UserSession,
+    val facilityRepository: FacilityRepository
+) : SynceableRepository<Appointment, AppointmentPayload> {
+
+  fun schedule(patientId: UUID, appointmentDate: LocalDate): Completable {
+    return facilityRepository
+        .currentFacility(userSession)
+        .take(1)
+        .flatMapCompletable { facility ->
```
vinaysshenoy: Can you move the construction of the `Appointment` instance into a `map()` operator?

app/src/main/java/org/simple/clinic/overdue/communication/CommunicationRepository.kt
```diff
@@ -0,0 +1,87 @@
...
+
+class CommunicationRepository @Inject constructor(
+    val dao: Communication.RoomDao,
+    val userSession: UserSession
+) : SynceableRepository<Communication, CommunicationPayload> {
+
+  fun create(appointmentId: UUID, type: Communication.Type, result: Communication.Result): Completable {
+    return userSession.loggedInUser()
+        .take(1)
+        .doOnNext {
```
vinaysshenoy: Is the `doOnNext` operator the right one to use here, since it isn't a side-effect? 

I'd rather do something like:

```kotlin
.flatMap { if(it === None) Single.error() else Single.just(it) }
```

vinaysshenoy:
@saket `Migration_9_10` hasn't been added to the set of migrations.

app/src/androidTest/java/org/simple/clinic/overdue/communication/CommunicationRepositoryAndroidTest.kt
saket: Oh wow I'm feeling so stupid. Will fix this.

app/src/main/java/org/simple/clinic/overdue/communication/Communication.kt
saket: Aye

app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
saket: Yea okay

app/src/androidTest/java/org/simple/clinic/overdue/communication/CommunicationRepositoryAndroidTest.kt
saket: Update: not so stupid afterall

PR #223, opened 2018-08-30
Add instrumentation for patient entry input validation errors

vinaysshenoy:


app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
```diff
@@ -223,6 +229,39 @@ class PatientEntryScreenControllerTest {
...
+    uiEvents.onNext(PatientEntrySaveClicked())
+
+    whenever(dobValidator.validate("16/07/2018")).thenReturn(Result.INVALID_PATTERN)
+    uiEvents.onNext(PatientDateOfBirthTextChanged("16/07/2018"))
+    uiEvents.onNext(PatientEntrySaveClicked())
+
+    val validationErrors = reporter.receivedEvents
+        .filter { it.name == "InputValidationError" }
+
+    Truth.assertThat(validationErrors).isNotEmpty()
```
saket: Can you statically import `Truth` functions? Let's do this change in a future commit.

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
```diff
@@ -223,6 +229,39 @@ class PatientEntryScreenControllerTest {
     verify(screen, times(4)).showEmptyStateError(true)
   }
 
+  @Test
+  fun `when input validation errors are emitted, they must be sent to analytics`() {
```
saket: I like to think that the test names should be from a business POV and not include any technical implementation details like "emitted", etc. Do you think we can change this?

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
saket: Can we check that the number of reported errors is `X` instead of any positive value? Is that a useful assertion?

vinaysshenoy: I don't think it's a useful assertion, because it makes the test fragile. If we add new kinds of errors, or decide to merge some errors internally in the validator, the test will break.

What will be useful if we have a way to access just the stream of errors, so we can assert that the total number of reported events is equal to the number of errors. Does that make sense?

PR #225, opened 2018-08-30
Fetch and update the overdue list from the appointment table.

sanchita:


app/src/main/java/org/simple/clinic/home/overdue/OverdueListAdapter.kt
```diff
@@ -23,16 +25,17 @@ class OverdueListAdapter : ListAdapter<OverdueListItem, OverdueListViewHolder>(O
...
     val age: Int,
     val bpSystolic: Int,
     val bpDiastolic: Int,
-    val bpDate: String,
+    val bpDaysAgo: Int,
     val overdueDays: Int
 )
 
-class OverdueListViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
+class OverdueListViewHolder(itemView: View, val context: Context) : RecyclerView.ViewHolder(itemView) {
```
saket: This VH can get the context from `itemView` directly. Do you think we can remove this extra param?

app/src/main/java/org/simple/clinic/home/overdue/OverdueListAdapter.kt
```diff
@@ -41,19 +44,15 @@ class OverdueListViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView)
...
 
-  override fun areItemsTheSame(oldItem: OverdueListItem?, newItem: OverdueListItem?): Boolean {
-    TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
-  }
+  override fun areItemsTheSame(oldItem: OverdueListItem, newItem: OverdueListItem): Boolean = oldItem.appointmentId == newItem.appointmentId
 
-  override fun areContentsTheSame(oldItem: OverdueListItem?, newItem: OverdueListItem?): Boolean {
-    TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
-  }
+  override fun areContentsTheSame(oldItem: OverdueListItem?, newItem: OverdueListItem?): Boolean = oldItem == newItem
```
saket: Let's also remove the `?` operator from this these params?

app/src/main/java/org/simple/clinic/home/overdue/OverdueScreen.kt
```diff
@@ -45,13 +48,20 @@ class OverdueScreen(context: Context, attrs: AttributeSet) : RelativeLayout(cont
...
-
+  fun updateOverdueList(list: List<OverdueListItem>) {
+    if (list.isEmpty()) {
+      overdueRecyclerView.visibility = View.GONE
+      viewForEmptyList.visibility = View.VISIBLE
+    } else {
+      listAdapter.submitList(list)
+      overdueRecyclerView.visibility = View.VISIBLE
+      viewForEmptyList.visibility = View.GONE
+    }
```
saket: Looks like you have some logic in your screen. This is something that should be moved to the controller so that it's also tested :)

saket: `overdueRecyclerView` will be transparent when there are no items, so you can choose to not bother changing its visibility.

app/src/main/java/org/simple/clinic/home/overdue/OverdueScreenController.kt
```diff
@@ -18,7 +24,26 @@ class OverdueScreenController @Inject constructor() : ObservableTransformer<UiEv
...
+          appointments.map {
+            OverdueListItem(
+                it.id,
+                it.patientId.toString(),
+                "Female",
+                22,
+                175,
+                55,
+                30,
+                getOverdueDays(it.date))
```
saket: Can you use named parameters? They look neater and do not get affected by changes in the order of parameters.

saket: Hrm, I just realized that it's possible you have parameter hints turned on in your IDE. Considering that the code reviewer cannot see them, they're not very useful.

saket: Do you also want to use `it` instead of `appointments` and make your Rx chain slightly neater?

PR #226, opened 2018-08-31
Add user id to analytics

vinaysshenoy:


app/src/main/java/org/simple/clinic/analytics/UpdateAnalyticsUserId.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.simple.clinic.user.UserSession
+import org.simple.clinic.util.Just
+import timber.log.Timber
+import javax.inject.Inject
+
+class UpdateAnalyticsUserId @Inject constructor(private val userSession: UserSession) {
+
+  fun update(scheduler: Scheduler) {
+    userSession.loggedInUser()
+        .subscribeOn(scheduler)
```
saket: Can we let the users subscribe to update() instead of doing it internally?

saket:
@vinaysshenoy let's refactor `keepUserIdUpdatedInAnalytics()` in a separate PR.

PR #228, opened 2018-09-03
Update error state for facility selection as per the design spec

saket:
network error

unexpected error

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -150,9 +150,10 @@
   <string name="registrationfacilities_screen_title">Select your clinic</string>
   <string name="registrationfacilities_facility_address_without_street">%s, %s</string>   <!-- District, State -->
   <string name="registrationfacilities_facility_address_with_street">%s, %s, %s</string>  <!-- Street, District, State -->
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -150,9 +150,10 @@
   <string name="registrationfacilities_screen_title">Select your clinic</string>
   <string name="registrationfacilities_facility_address_without_street">%s, %s</string>   <!-- District, State -->
   <string name="registrationfacilities_facility_address_with_street">%s, %s, %s</string>  <!-- Street, District, State -->
-  <string name="registrationfacilities_error_unexpected_error">@string/api_unexpected_error</string>
-  <string name="registrationfacilities_error_check_internet_connection">@string/api_network_error</string>
-  <string name="registrationfacilities_retry">Retry</string>
+  <string name="registrationfacilities_error_unexpected_title">Something went wrong</string>
+  <string name="registrationfacilities_error_unexpected_message">Please try again after some time \nor get in touch with your supervisor</string>
```
vinaysshenoy: 1. This sentence should end with a period as per the designs
2. I am not comfortable hard-coding a new-line in this string resource. Why can't we use the TextView's standard text layout mechanism?

saket: Fixed

PR #229, opened 2018-09-03
The very-big overdue appointments object

sanchita:


app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -47,13 +61,196 @@ class AppointmentRepositoryAndroidTest {
...
+            LocalDate.parse("1977-11-15"),
+            null,
+            PatientStatus.MIGRATED,
+            Instant.now(),
+            Instant.now(),
+            SyncStatus.DONE
+        )
+    )
+    database.bloodPressureDao().save(listOf(
+        BloodPressureMeasurement(
```
saket: Will you move these dummy object creations to `TestData` in your next commit? This test looks gigantic otherwise right now.

app/src/main/java/org/simple/clinic/home/overdue/OverdueScreenController.kt
```diff
@@ -51,6 +53,11 @@ class OverdueScreenController @Inject constructor(
     return updateListStream.mergeWith(emptyStateStream)
   }
 
+  private fun getDaysAgoForInstant(instant: Instant): Int {
```
saket: I think we should avoid using `get` prefix for functions here

saket:
@pratul @sanchita go ahead and merge this branch once you've seen my comments.

app/src/main/java/org/simple/clinic/home/overdue/OverdueScreenController.kt
pratul: Yeah this should definitely be changed.

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
pratul: If not the next commit, then at some point. I'll add this to icebox.

PR #236, opened 2018-09-04
[WIP] Overdue list UI and phone call permission

pratul:


app/src/main/res/layout/item_overdue_list.xml
```diff
@@ -30,37 +30,43 @@
 
       <TextView
         android:id="@+id/overdue_patient_bp"
-        android:textStyle="normal"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_marginBottom="@dimen/spacing_4"
         android:textColor="@color/black_opacity_50"
         android:textSize="@dimen/textsize_14"
+        android:textStyle="normal"
```
saket: Did you intentionally add `textStyle=normal`? You don't need this fyi.

app/src/main/java/org/simple/clinic/home/overdue/OverdueScreenController.kt
```diff
@@ -77,4 +84,27 @@ class OverdueScreenController @Inject constructor(
   private fun calculateOverdueDays(date: LocalDate): Int {
     return DAYS.between(date, LocalDate.now(ZoneOffset.UTC)).toInt()
```
saket: Just like how you've statically imported `DAYS`, can we also import `UTC`?

app/src/main/java/org/simple/clinic/home/overdue/OverdueScreenController.kt
```diff
@@ -77,4 +84,27 @@ class OverdueScreenController @Inject constructor(
...
+
+  private fun patientCalls(events: Observable<UiEvent>): Observable<UiChange> {
+    val callPhonePermissionChanges = events
+        .ofType<CallPhonePermissionChanged>()
+        .map(CallPhonePermissionChanged::result)
+
+    val grantedStream = callPhonePermissionChanges
+        .filter { it == RuntimePermissionResult.GRANTED }
+        .withLatestFrom(events.ofType<CallPatientClicked>())
+        .flatMap { (_, callClicked) -> Observable.just { ui: Ui -> ui.callPatientWithoutUsingDialer(callClicked.phoneNumber) } }
```
saket: You don't need a `flatMap` here. 

```
.flatMap { (_, callClicked) -> { ui: Ui -> ui.callPatientWithoutUsingDialer(callClicked.phoneNumber) } }
```

saket: We can also extract `events.ofType<CallPatientClicked>()` to make `grantedStream` and `notGrantedStream` shorter.

```
val callPhonePermissionChanges = events
        .ofType<CallPhonePermissionChanged>()
        .map(CallPhonePermissionChanged::result)

val callClicks = events
        .ofType<CallPatientClicked>()
        .map { it.phoneNumber }

val grantedStream = callPhonePermissionChanges
        .filter { it == RuntimePermissionResult.GRANTED }
        .withLatestFrom(callClicks)
        .map { (_, number) -> { ui: Ui -> ui.callPatientWithoutUsingDialer(number) } }

val notGrantedStream = callPhonePermissionChanges
        .filter { it != RuntimePermissionResult.GRANTED }
        .withLatestFrom(callClicks)
        .flatMap { (_, number) -> { ui: Ui -> ui.callPatientUsingDialer(number) } }
```

saket: It looks like `grantedStream` and `notGrantedStream` names can be improved. We've so far named streams according to their jobs. 

For e.g., `updateListStream` in the `screenSetup()` is self descriptive that the stream updates the list. 

`grantedStream` in comparison does not indicate what it is doing. It only indicates that this stream will function when the permission is granted.

PR #240, opened 2018-09-06
Update overdue query to include current facility only

sanchita:


app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
```diff
@@ -59,10 +60,16 @@ class AppointmentRepository @Inject constructor(
   }
 
   fun overdueAppointments(): Observable<List<OverdueAppointment>> {
-    return overdueDao.appointments(
-        scheduledStatus = Appointment.Status.SCHEDULED,
-        dateNow = LocalDate.now()
-    ).toObservable()
+    return facilityRepository.currentFacility(userSession)
+        .take(1)
```
saket: Let's not do `take(1)` here?

PR #242, opened 2018-09-07
Add 'mg' to dosage field only when it's focused for the first time

saket:
Bug fix for https://www.pivotaltracker.com/story/show/160350133

The placeholder was previously shown everytime the dosage field is empty, making it impossible to clear the field.


app/src/main/java/org/simple/clinic/drugs/selection/entry/CustomPrescriptionEntryController.kt
```diff
@@ -78,6 +79,7 @@ class CustomPrescriptionEntryController @Inject constructor(
 
     val setPlaceholder = Observables.combineLatest(dosageFocusChanges, dosageTextChanges)
         .filter { (hasFocus, text) -> hasFocus && text.isBlank() }
+        .take(1)
```
vinaysshenoy: `Rx` to the rescue.

saket: Praise David

PR #244, opened 2018-09-10
Sync medical history of patients

saket:


app/src/main/java/org/simple/clinic/medicalhistory/MedicalHistoryRepository.kt
```diff
@@ -0,0 +1,57 @@
...
+  override fun recordsWithSyncStatus(syncStatus: SyncStatus): Single<List<MedicalHistory>> {
+    return dao.recordsWithSyncStatus(syncStatus).firstOrError()
+  }
+
+  override fun setSyncStatus(from: SyncStatus, to: SyncStatus): Completable {
+    return Completable.fromAction { dao.updateSyncStatus(from, to) }
+  }
+
+  override fun setSyncStatus(ids: List<UUID>, to: SyncStatus): Completable {
+    if (ids.isEmpty()) {
```
vinaysshenoy: Question: Is it actually important to have this assertion? Either here, or in any of the other repositories? In the event that they are absent, the dao statement should just fail silently, right?

app/src/main/java/org/simple/clinic/medicalhistory/MedicalHistory.kt
```diff
@@ -0,0 +1,47 @@
...
+) {
+
+  @Dao
+  interface RoomDao {
+
+    @Query("SELECT * FROM MedicalHistory WHERE syncStatus = :status")
+    fun recordsWithSyncStatus(status: SyncStatus): Flowable<List<MedicalHistory>>
+
+    @Query("UPDATE MedicalHistory SET syncStatus = :to WHERE syncStatus = :from")
+    fun updateSyncStatus(from: SyncStatus, to: SyncStatus)
```
vinaysshenoy: Opinion: I feel that this statement is dangerous because it will update everything.

app/src/main/java/org/simple/clinic/medicalhistory/MedicalHistoryRepository.kt
saket: Just for the record, the dao throw an error if the arguments for an `IN` clause are empty.

PR #250, opened 2018-09-12
Schedule appointment via sheet

pratul:


app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentEvents.kt
```diff
@@ -0,0 +1,23 @@
+package org.simple.clinic.scheduleappointment
+
+import org.simple.clinic.widgets.UiEvent
+import org.threeten.bp.temporal.ChronoUnit
+import java.util.UUID
+
+data class SheetCreated(val initialIndex: Int, val patientUuid: UUID) : UiEvent
```
saket: Considering that it has values used only in scheduling of appointments, can you rename this event to make it specific?

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentEvents.kt
```diff
@@ -0,0 +1,23 @@
...
+
+data class IncrementAppointmentDate(val currentIndex: Int, val size: Int) : UiEvent {
```
saket: The voice of these events is not what we've used so far. The describe "what happened" and not "what to do" :)

`IncrementAppointmentDate` -> `IncrementAppointmentDateClicked`? Or `AppointmentDateIncremented`?

saket: Same for the rest of the events

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -0,0 +1,99 @@
...
+    val enableButtonStream = notLastPositionStream
+        .distinctUntilChanged()
+        .map { { ui: Ui -> ui.enableIncrementAndDecrementButtons() } }
+
+    return Observable.merge(lastPositionStream, updateDateStream, enableButtonStream)
+  }
+
+  private fun isSecondLastItem(current: Int, size: Int) = current == size - 2
+
+  private fun isNotLastItem(current: Int, size: Int) = current != size - 1
```
saket: Considering that both `isSecondLastItem()` and `isNotLastItem()` aren't used by any other functions, they can be inlined inside `dateIncrements()` as lambdas.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -0,0 +1,99 @@
...
+        .withLatestFrom(patientUuidStream)
+        .map { (scheduleEvent, patientUuid) -> patientUuid to appointmentLocalDate(scheduleEvent.selectedDateState.second) }
+        .flatMap {
+          repository.schedule(patientUuid = it.first, appointmentDate = it.second)
+              .andThen(Observable.just { ui: Ui -> ui.closeSheet() })
+        }
+  }
+
+  private fun appointmentLocalDate(durationPair: Pair<Int, ChronoUnit>): LocalDate {
+    return LocalDate.now(UTC).plus(durationPair.first.toLong(), durationPair.second)
```
saket: I think we should create a data class to for `possibleDates` avoid using `pair.first` and `pair.second` here. They give zero information to the reader.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -0,0 +1,99 @@
...
+        .map { { ui: Ui -> ui.closeSheet() } }
+  }
+
+  private fun scheduleCreates(events: Observable<UiEvent>): Observable<UiChange> {
+    val patientUuidStream = events.ofType<SheetCreated>()
+        .map { it.patientUuid }
+
+    return events.ofType<ScheduleAppointment>()
+        .withLatestFrom(patientUuidStream)
+        .map { (scheduleEvent, patientUuid) -> patientUuid to appointmentLocalDate(scheduleEvent.selectedDateState.second) }
```
saket: Why have such long lines da.

This can be replaced with:

```kotlin
events.ofType<ScheduleAppointment>()
  .map { appointmentLocalDate(it.selectedDateState.second) }
  .withLatestFrom(patientUuidStream)
```

PR #255, opened 2018-09-17
UI changes in patient lookup

saket:


app/src/main/res/layout/screen_patient_search.xml
```diff
@@ -1,65 +1,150 @@
...
-    tools:listitem="@layout/list_patient_search" />
+    android:layout_toEndOf="@+id/patientsearch_back"
+    android:scrollbarStyle="insideOverlay"
+    android:scrollbars="vertical">
+
+    <LinearLayout
+      android:layout_width="match_parent"
+      android:layout_height="wrap_content"
+      android:orientation="vertical"
+      android:paddingBottom="64dp"
```
vinaysshenoy: Should this be extracted to a dimension resource?

app/src/main/res/layout/screen_patient_search.xml
```diff
@@ -1,65 +1,150 @@
...
+      android:paddingEnd="@dimen/spacing_8"
+      android:paddingTop="@dimen/spacing_16"
+      tools:ignore="RtlSymmetry">
+
+      <FrameLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content">
+
+        <android.support.design.widget.TextInputLayout
```
vinaysshenoy: `TextInputLayout` is a sub-class of `LinearLayout`. Can't we add the clear button as a child of the `TextInputLayout` itself?

app/src/main/res/layout/screen_patient_search.xml
saket: Could, sure

app/src/main/res/layout/screen_patient_search.xml
saket: No, because the clear button needs to appear in front of the text field

PR #258, opened 2018-09-18
Initial implementation of Forgot PIN - Create new PIN

vinaysshenoy:


app/src/main/res/layout/screen_app_lock.xml
```diff
@@ -108,7 +108,16 @@
...
+        android:text="@string/applock_error_incorrect_pin"
+        tools:visibility="visible" />
+
+      <Button
+        android:id="@+id/applock_forgotpin"
+        style="@style/Widget.AppCompat.Button.Borderless"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="@string/applock_forgotpin"
+        android:textColor="?colorAccent" />
```
saket: FYI you can also use `Widget.AppCompat.Button.Borderless.Colored` instead of setting the text color separately

app/src/main/java/org/simple/clinic/login/applock/AppLockScreen.kt
```diff
@@ -101,6 +102,9 @@ class AppLockScreen(context: Context, attrs: AttributeSet) : RelativeLayout(cont
     }
   }
 
+  private fun forgotPinClicks() =
+      RxView.clicks(forgotPinButton).map { AppLockForgotPinClicked() }
```
saket: I have realized that for testing button clicks which simply perform action without any logic, like opening a dialog, popping the backstack, etc. does not offer any value. The error window is the same when the action is taken directly or through the controller. In both the cases, commenting out this line is sufficient to break functionality:

```kotlin
// ConfirmResetPinDialog.show(activity.supportFragmentManager)
```

I hope this was understandable. What do you think? Can we handle this click in the screen directly?

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -348,4 +351,11 @@ class UserSession @Inject constructor(
   fun accessToken(): Optional<String> {
     return accessTokenPreference.get()
   }
+
+  fun startForgotPinFlow(retryCount: Int = 0): Completable {
```
saket: Hrm, should we make this function name more explicit to what it does?

app/src/main/java/org/simple/clinic/login/applock/ConfirmResetPinDialog.kt
```diff
@@ -30,6 +44,13 @@ class ConfirmResetPinDialog : AppCompatDialogFragment() {
           val view = dialog.ownerActivity.findViewById<View>(android.R.id.content)
           val snackbar = Snackbar.make(view, getString(R.string.applock_reset_pin_waiting), Snackbar.LENGTH_INDEFINITE)
           snackbar.show()
+
+          userSession.startForgotPinFlow(1)
```
saket: Named arguments please?

app/src/main/java/org/simple/clinic/login/applock/ConfirmResetPinDialog.kt
```diff
@@ -30,6 +44,13 @@ class ConfirmResetPinDialog : AppCompatDialogFragment() {
...
+              .observeOn(AndroidSchedulers.mainThread())
```
saket: Can you statically import `mainThread()`?

saket: Oh I see you've replaced `1` with a property. Why not just use a named param?

app/src/main/java/org/simple/clinic/forgotpin/createnewpin/ForgotPinCreateNewPinScreenController.kt
```diff
@@ -13,17 +14,36 @@ typealias Ui = ForgotPinCreateNewPinScreen
...
         .map { { ui: Ui -> ui.showUserName(it.fullName) } }
         .toObservable()
   }
+
+  private fun showFacilityOnScreenCreate(events: Observable<UiEvent>): Observable<UiChange> {
+    return events
+        .ofType<ScreenCreated>()
+        .firstOrError()
+        .flatMap { userSession.requireLoggedInUser().firstOrError() }
+        .flatMap { facilityRepository.currentFacility(it).firstOrError() }
```
saket: Why are you manually turning all observables to singles?

app/src/main/java/org/simple/clinic/forgotpin/createnewpin/ForgotPinCreateNewPinScreen.kt
```diff
@@ -31,21 +42,45 @@ class ForgotPinCreateNewPinScreen(context: Context, attributeSet: AttributeSet?)
...
   }
 
   private fun screenCreates(): Observable<UiEvent> = Observable.just(ScreenCreated())
 
+  private fun pinTextChanges(): Observable<UiEvent> {
+    return RxTextView.textChanges(pinEntryEditText).map { ForgotPinCreateNewPinTextChanged(it.toString()) }
+  }
+
+  private fun pinSubmitClicked(): Observable<UiEvent> {
+   return RxTextView.editorActions(pinEntryEditText).filter { it == EditorInfo.IME_ACTION_DONE }
```
saket: Can you move `filter()` to the next line?

app/src/main/java/org/simple/clinic/forgotpin/createnewpin/ForgotPinCreateNewPinScreenController.kt
```diff
@@ -47,4 +53,29 @@ class ForgotPinCreateNewPinScreenController @Inject constructor(
...
+
+    return events
+        .ofType<ForgotPinCreateNewPinSubmitClicked>()
+        .withLatestFrom(pinTextChanges) { _, pin -> pin }
+        .filter { it.length != pinLength }
+        .map { { ui: Ui -> ui.showInvalidPinError() } }
+  }
+
+  private fun openConfirmPinEntryScreenOnValidPin(events: Observable<UiEvent>): Observable<UiChange> {
+    val validPin = events.ofType<ForgotPinCreateNewPinTextChanged>().map { it.pin }
```
saket: `map()` to next line please

app/src/main/java/org/simple/clinic/login/applock/AppLockScreen.kt
vinaysshenoy: TODO: See if these can be hooked into Heap in all cases if the events don't go through the controller.

app/src/main/java/org/simple/clinic/login/applock/ConfirmResetPinDialog.kt
vinaysshenoy: That actually would be better. Will fix.

app/src/main/java/org/simple/clinic/user/UserSession.kt
vinaysshenoy: TODO: Think of a better name when the feature is finished.

app/src/main/java/org/simple/clinic/forgotpin/createnewpin/ForgotPinCreateNewPinScreenController.kt
vinaysshenoy: Had an offline discussion; decided not to bother too much for manually converting to `Single`s each time.

saket:
@vinaysshenoy looks like we'll have to create a style for dialogs, because the colors and spacing are different on zeplin.

app/src/main/java/org/simple/clinic/login/applock/ConfirmResetPinDialog.kt
```diff
@@ -0,0 +1,68 @@
...
+        .setPositiveButton(R.string.applock_reset_pin_alert_confirm) { _, _ ->
+          val view = dialog.ownerActivity.findViewById<View>(android.R.id.content)
+          val snackbar = Snackbar.make(view, getString(R.string.applock_reset_pin_waiting), Snackbar.LENGTH_INDEFINITE)
+          snackbar.show()
+
+          userSession.startForgotPinFlow(patientRepository, syncRetryCount = 1)
+              .observeOn(mainThread())
+              .doOnTerminate { snackbar.dismiss() }
+              .subscribe {
+                screenRouter.clearHistoryAndPush(ForgotPinCreateNewPinScreenKey(), RouterDirection.REPLACE)
```
saket: Why is going back not allowed? Considering that the user can just exit and reopen the app to reach app lockscreen, is this useful?

app/src/main/res/layout/screen_app_lock.xml
```diff
@@ -108,7 +108,15 @@
...
-        android:text="@string/applock_error_incorrect_pin" />
+        android:text="@string/applock_error_incorrect_pin"
+        tools:visibility="visible" />
+
+      <Button
+        android:id="@+id/applock_forgotpin"
+        style="@style/Widget.AppCompat.Button.Borderless.Colored"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="@string/applock_forgotpin" />
```
saket: The spacing below this button doesn't match zeplin :)

app/src/main/java/org/simple/clinic/login/applock/ConfirmResetPinDialog.kt
vinaysshenoy: Whoops, copy-pasta. Fixing.

app/src/main/res/layout/screen_app_lock.xml
vinaysshenoy: Will fix.

vinaysshenoy: TODO: Fix in upcoming PR.

saket: Sure

app/src/main/java/org/simple/clinic/login/applock/ConfirmResetPinDialog.kt
```diff
@@ -0,0 +1,68 @@
...
+  @Inject
+  lateinit var patientRepository: PatientRepository
+
+  companion object {
+    fun show(fragmentManager: FragmentManager) {
+      (fragmentManager.findFragmentByTag("confirm_reset_pin_alert") as ConfirmResetPinDialog?)?.dismiss()
+
+      val fragment = ConfirmResetPinDialog()
+      // Cancellable on the dialog builder is ignored. We have to use this.
+      fragment.isCancelable = false
```
saket: Another copy pasta

saket:
I'm not sure how to handle ConfirmResetPinDialog.
1. It has untested logic that should be moved to the calling screen.
2. It leaks memory by kicking off an asynchronous Rx chain when it gets dismissed. The lambdas in the Rx chain keep a reference to the dialog.

I'll ask @pratul to take a call on this.

PR #261, opened 2018-09-19
Open create new PIN screen if user has not completed raising a PIN request

vinaysshenoy:
### Changes
1. Removes the `VERIFYING_OTP` logged in status since it was not being used anywhere, and added two more, `RESETTING_PIN` and `RESET_PIN_REQUESTED`.
2. Updated initial screen key in `TheActivity` to load the Create PIN screen if the logged in status is `RESETTING_PIN`.
3. Change the user logged in status to `RESETTING_PIN` once the patient data is cleared.

### DIFF
--- a/app/src/main/java/org/simple/clinic/activity/TheActivity.kt
+++ b/app/src/main/java/org/simple/clinic/activity/TheActivity.kt
@@ -25,10 +25,9 @@ import org.simple.clinic.router.screen.FullScreenKey
 import org.simple.clinic.router.screen.FullScreenKeyChanger
 import org.simple.clinic.router.screen.NestedKeyChanger
 import org.simple.clinic.router.screen.ScreenRouter
-import org.simple.clinic.user.User.LoggedInStatus.LOGGED_IN
-import org.simple.clinic.user.User.LoggedInStatus.NOT_LOGGED_IN
-import org.simple.clinic.user.User.LoggedInStatus.OTP_REQUESTED
-import org.simple.clinic.user.User.LoggedInStatus.VERIFYING_OTP
+import org.simple.clinic.user.User
+import org.simple.clinic.user.User.*
+import org.simple.clinic.user.User.LoggedInStatus.*
 import org.simple.clinic.user.UserSession
 import org.simple.clinic.widgets.TheActivityLifecycle
 import javax.inject.Inject
@@ -98,9 +97,10 @@ class TheActivity : AppCompatActivity() {
   private fun initialScreenKey(): FullScreenKey {
     val localUser = userSession.loggedInUser().blockingFirst().toNullable()
 
+    // TODO: Figure out how to handle the new PIN reset status
     val canMoveToHomeScreen = when (localUser?.loggedInStatus) {
-      NOT_LOGGED_IN -> false
-      LOGGED_IN, OTP_REQUESTED, VERIFYING_OTP -> true
+      NOT_LOGGED_IN, RESETTING_PIN -> false
+      LOGGED_IN, OTP_REQUESTED, RESET_PIN_REQUESTED  -> true
       null -> false
     }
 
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/simple/clinic/activity/TheActivity.kt
```diff
@@ -97,17 +102,22 @@ class TheActivity : AppCompatActivity() {
...
       canMoveToHomeScreen -> HomeScreen.KEY
       hasUserCompletedOnboarding.get().not() -> OnboardingScreen.KEY
-      else -> RegistrationPhoneScreen.KEY
+      else -> {
+        return if (localUser?.loggedInStatus == RESETTING_PIN) {
+          ForgotPinCreateNewPinScreenKey()
+        } else {
+          RegistrationPhoneScreen.KEY
+        }
+      }
```
saket: This logic is slowly becoming complex. Although we cannot handle this through the controller because we need the initial screen key immediately, we could add a function to the controller for this.

```kotlin
private fun initialScreenKey() {
  return controller.initialScreenKey()
}
```

This way, the logic can also be tested. What do you think?

app/src/main/java/org/simple/clinic/activity/TheActivityController.kt
```diff
@@ -37,7 +45,14 @@ class TheActivityController @Inject constructor(
   private fun showAppLock(events: Observable<UiEvent>): Observable<UiChange> {
     val replayedCanShowAppLock = events
         .ofType<TheActivityLifecycle.Started>()
-        .filter { userSession.isUserLoggedIn() }
+        .flatMapSingle { _ ->
+          userSession.loggedInUser()
+              .firstElement()
+              .filter { it is Just<User> }
+              .map { (user) -> user!!.loggedInStatus }
+              .toSingle(NOT_LOGGED_IN)
```
saket: It took me good 2 minutes to understand that `NOT_LOGGED_IN` is the default value in `toSingle()`. Can this be improved for code readability?

app/src/main/java/org/simple/clinic/activity/TheActivityController.kt
```diff
@@ -25,6 +31,8 @@ class TheActivityController @Inject constructor(
     @Named("should_lock_after") private val lockAfterTimestamp: Preference<Instant>
 ) : ObservableTransformer<UiEvent, UiChange> {
 
+  private val showAppLockUserStates = setOf(OTP_REQUESTED, LOGGED_IN, RESET_PIN_REQUESTED)
```
saket: Can we rename this to `showAppLockForUserStates`?

app/src/main/java/org/simple/clinic/activity/TheActivity.kt
vinaysshenoy: I think it's a good idea. Will do this.

app/src/main/java/org/simple/clinic/activity/TheActivity.kt
```diff
@@ -97,17 +102,22 @@ class TheActivity : AppCompatActivity() {
...
       null -> false
     }
 
     return when {
       canMoveToHomeScreen -> HomeScreen.KEY
       hasUserCompletedOnboarding.get().not() -> OnboardingScreen.KEY
-      else -> RegistrationPhoneScreen.KEY
+      else -> {
+        return if (localUser?.loggedInStatus == RESETTING_PIN) {
+          ForgotPinCreateNewPinScreenKey()
```
saket: Can you please use `ForgotPinCreateNewPinScreen.KEY` for consistency?

app/src/test/java/org/simple/clinic/activity/TheActivityControllerTest.kt
```diff
@@ -58,15 +60,30 @@ class TheActivityControllerTest {
...
   @Test
-  fun `when activity is started, user is logged and in user was inactive then app lock should be shown`() {
+  @Parameters(value = [
+    "NOT_LOGGED_IN|false",
+    "OTP_REQUESTED|true",
+    "LOGGED_IN|true",
+    "RESETTING_PIN|false",
+    "RESET_PIN_REQUESTED|true"
+  ])
+  fun `when activity is started, user is logged and in user was inactive then app lock should be shown`(
```
saket: Can you also fix this typo? `user is logged and in`

PR #265, opened 2018-09-20
Apply style guide and update UI for overdue list

sanchita:


app/src/main/res/values/styles_v2.xml
```diff
@@ -167,4 +175,14 @@
     <item name="android:src">@drawable/ic_clear_circular_24dp</item>
     <item name="android:layout_gravity">end</item>
   </style>
+
+  <style name="Clinic.V2.OverdueListActionItems">
```
saket: This style is for an individual action item so you can just name it as `OverdueListActionItem`.

saket: Actually, there's no need to suffix it with `Item` either. Just `OverdueListAction` is sufficient.

app/src/main/res/drawable/ic_patient_transgender.xml
```diff
@@ -1,15 +1,20 @@
-<!--
-  TODO: This is just a placeholder icon.
-  https://www.pivotaltracker.com/story/show/160616504
```
saket: Can you ask akshay to mark this story as delivered?

app/src/main/java/org/simple/clinic/home/overdue/OverdueListAdapter.kt
```diff
@@ -94,10 +97,22 @@ class OverdueListViewHolder(
     actionsContainer.visibility = if (isVisible) View.GONE else View.VISIBLE
   }
 
+  fun handlePhoneNumberViewVisibility() {
```
saket: I see you've named this function on the lines of `handleBottomLayoutVisibility()`. `handle` is a very generic name and it does not indicate anything about its functionality. When someone sees this function's name, they should immediately be able to understand what will happen when it's called.

app/src/main/java/org/simple/clinic/home/overdue/OverdueListAdapter.kt
```diff
@@ -94,10 +97,22 @@ class OverdueListViewHolder(
...
+    val isVisible = phoneNumberTextView.visibility == View.VISIBLE
+    if (!isVisible && appointment.phoneNumber != null) {
+      phoneNumberTextView.visibility =
+          View.VISIBLE
+    } else {
+      phoneNumberTextView.visibility =
+          View.GONE
```
saket: Did you intentionally place the visibilities on a new line? :p

app/src/main/java/org/simple/clinic/home/overdue/OverdueListAdapter.kt
saket: Can you also rename `handleBottomLayoutVisibility()` along with this?

saket:
@sanchita one more suggestion: the first commit, `Refactor Views class` could be named better. Every commit in a way refactors something. This is not very different than saying `Code was written`. :)

PR #267, opened 2018-09-21
Overdue: "Remind to call later..." action

pratul:
This is only the UI, actions, and controller. The model changes haven't been made, and the reminder doesn't _actually_ get created.


app/src/main/java/org/simple/clinic/home/overdue/OverdueListAdapter.kt
```diff
@@ -74,7 +77,10 @@ data class OverdueListItem(
 
 class OverdueListViewHolder(
     itemView: View,
-    private val phoneCallClickStream: PublishSubject<CallPatientClicked>
+    phoneCallClickStream: PublishSubject<CallPatientClicked>,
+    agreedToVisitClickStream: PublishSubject<AgreedToVisitClicked>,
+    remindLaterClickStream: PublishSubject<RemindToCallLaterClicked>,
+    removeFromListClickStream: PublishSubject<RemoveFromListClicked>
```
saket: Tip: these 4 streams can be combined into a single `Subject<UiEvent>` stream.

app/src/main/java/org/simple/clinic/home/overdue/OverdueListAdapter.kt
```diff
@@ -20,7 +20,10 @@ import java.util.UUID
 
 
 class OverdueListAdapter(
-    private val phoneCallClickStream: PublishSubject<CallPatientClicked>
+    private val phoneCallClickStream: PublishSubject<CallPatientClicked>,
+    private val agreedToVisitClickStream: PublishSubject<AgreedToVisitClicked>,
+    private val remindLaterClickStream: PublishSubject<RemindToCallLaterClicked>,
+    private val removeFromListClickStream: PublishSubject<RemoveFromListClicked>
```
saket: Tip: We could expose these streams from the adapter instead of receiving it in the constructor. This way, this adapter can also be injected by DI.

In the calling side, you could then do:

```kotlin
Observable.mergeArray(screenCreates(), callPermissionChanges(), adapter.uiEvents)
  .compose(controller)
  ...
```

app/src/main/res/layout/list_patientsummary_add_new_bp_measurement.xml
```diff
@@ -14,7 +15,7 @@
     android:paddingStart="@dimen/spacing_24"
     android:paddingTop="@dimen/spacing_24">
 
-    <android.support.v7.widget.AppCompatButton
+    <Button
```
saket: This was intentionally done because `app:backgroundTint` is an attribute of `AppCompatButton`.

saket: Same in other places.

app/src/main/java/org/simple/clinic/home/overdue/OverdueListAdapter.kt
pratul: I like this idea. Nice.

PR #268, opened 2018-09-24
Complete Forgot PIN feature

vinaysshenoy:


app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -59,7 +60,13 @@ class UserSession @Inject constructor(
...
     private val passwordHasher: PasswordHasher,
     @Named("preference_access_token") private val accessTokenPreference: Preference<Optional<String>>,
-    private val syncScheduler: SyncScheduler
+    private val syncScheduler: SyncScheduler,
+    @Named("last_patient_pull_timestamp") private val patientSyncPullTimestamp: Preference<Optional<Instant>>,
+    @Named("last_bp_pull_timestamp") private val bpSyncPullTimestamp: Preference<Optional<Instant>>,
+    @Named("last_prescription_pull_timestamp") private val prescriptionSyncPullTimestamp: Preference<Optional<Instant>>,
+    @Named("last_appointment_pull_timestamp") private val appointmentSyncPullTimestamp: Preference<Optional<Instant>>,
+    @Named("last_communication_pull_timestamp") private val communicationSyncPullTimestamp: Preference<Optional<Instant>>,
+    @Named("last_medicalhistory_pull_timestamp") private val medicalHistorySyncPullTimestamp: Preference<Optional<Instant>>
```
saket: We should really split shared preferences into two files and clear the one containing these timestamps entirely

app/src/main/java/org/simple/clinic/user/LoggedInUserHttpInterceptor.kt
```diff
@@ -18,9 +18,10 @@ class LoggedInUserHttpInterceptor @Inject constructor() : Interceptor {
     val user = userSession.loggedInUserImmediate()
     val (accessToken) = userSession.accessToken()
 
-    return when (user?.isApprovedForSyncing() == true) {
-      true -> chain.proceed(addHeaders(originalRequest, accessToken!!, user!!))
-      false -> chain.proceed(originalRequest)
+    return if (user != null && accessToken.isNullOrBlank().not()) {
+      chain.proceed(addHeaders(originalRequest, accessToken!!, user))
```
saket: TODO: understand this

app/src/test/java/org/simple/clinic/forgotpin/confirmpin/ForgotPinConfirmPinScreenControllerTest.kt
```diff
@@ -137,6 +137,15 @@ class ForgotPinConfirmPinScreenControllerTest {
     }
   }
 
+  @Test
+  fun `when a valid PIN is submitted and reset PIN call fails, it must show the unexpected error`() {
```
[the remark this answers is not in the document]
vinaysshenoy: Whoops, yes! But the test has been removed in a future commit in this PR.

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -368,13 +371,31 @@ class UserSession @Inject constructor(
...
-          Completable
-              .fromCallable { appDatabase.userDao().updateLoggedInStatusForUser(it.uuid, User.LoggedInStatus.LOGGED_IN) }
+    val hashedPin = passwordHasher.hash(pin)
+        .cache()
+
+    val updateUserOnSuccess = hashedPin
+        .zipWith(requireLoggedInUser().firstOrError())
+        .map { (newPinDigest, user) -> user.copy(pinDigest = newPinDigest, loggedInStatus = User.LoggedInStatus.RESET_PIN_REQUESTED) }
+        .doOnSuccess { appDatabase.userDao().createOrUpdate(it) }
+        .map { ForgotPinResult.Success as ForgotPinResult }
```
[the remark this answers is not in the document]
vinaysshenoy: Agreed. This was changed in a later commit to use `flatMapCompletable` instead.

app/src/main/java/org/simple/clinic/user/UserSession.kt
vinaysshenoy: Yup. Adding it to Pivotal.

saket:
@vinaysshenoy nearly half of your commits exceed the 80 character limit. Please try keeping them within the limit :)

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -183,19 +183,24 @@ class UserSession @Inject constructor(
...
-        .flatMapCompletable { userPayload ->
-          val user = userFromPayload(userPayload, User.LoggedInStatus.LOGGED_IN)
-          val userFacilities = userPayload.facilityUuids
-          storeUser(user, userFacilities)
+        .flatMapCompletable { loggedInUser ->
+          registrationApi.findUser(loggedInUser.phoneNumber)
+              .flatMapCompletable { userPayload ->
+                // FIXME: This is a hack to handle the case where the user logged in status will
+                // not get set to LOGGED_IN when the PIN reset request is approved.
+                val finalLoggedInStatus = if (userPayload.status == UserStatus.APPROVED_FOR_SYNCING) {
```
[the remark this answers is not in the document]
vinaysshenoy: Earlier, when we used to query the user status in the home screen (to check whether the user has been approved from the backend for syncing), we used to hardcode the local `loggedInStatus` of the user to `LOGGED_IN` since that is what we used for a newly registered user since we only cared about the change in the `status` field.

However, when it comes to PIN reset, we set the local state to `PIN_RESET_REQUESTED` and this state must be maintained for necessary messaging in the UI. So this retains the previous `loggedInStatus` of the user we are refreshing (`LOGGED_IN` for a newly registered user, and `PIN_RESET_REQUESTED` for a user who's resetting the PIN) *until* the server says they have been approved, at which point we change the local `loggedInStatus` to `LOGGED_IN` again.

app/src/main/java/org/simple/clinic/user/LoggedInUserHttpInterceptor.kt
vinaysshenoy: The `isApprovedForSyncing()` method checks two things: whether the `loggedInStatus` is equal to `LOGGED_IN` and the `status` is equal to `APPROVED_FOR_SYNCING`. When we call the `reset_password` api, the access token must be sent so the server knows its a valid user who's resetting the PIN, but by this time we have already set the local logged in status to `RESETTING_PIN`, so the access token does not get sent.

This change ensures that the access token gets sent regardless of whether the user is approved for syncing or not.

saket:
@vinaysshenoy now that we are no longer resetting app data from `ConfirmResetPinDialog`, should we avoid clearing the back stack?

saket:
@vinaysshenoy the keyboard doesn't seem to show up automatically on this screen. Can you manually call the keyboard?

untitled

vinaysshenoy:
> @vinaysshenoy now that we are no longer resetting app data from `ConfirmResetPinDialog`, should we avoid clearing the back stack?

Good point. Adding to TODO to confirm with design.

saket:
@vinaysshenoy the facility names on the two PIN screens aren't aligned with each other.

screenshot_1537857272

screenshot_1537857271

saket:
The PIN field feels a bit difficult to focus because there's no visual feedback without a cursor. I think we should consider focusing it manually when the user taps anywhere on the card.

PR #271, opened 2018-09-24
Implement 'Remind me later' action in overdue list

pratul:


app/src/main/java/org/simple/clinic/home/overdue/OverdueScreenEvents.kt
```diff
@@ -18,6 +18,10 @@ data class RemindToCallLaterClicked(val appointmentUUID: UUID) : UiEvent {
   override val analyticsName = "Overdue Screen:Remind To Call Later clicked"
 }
 
+class AppointmentActionComplete : UiEvent{
+  override val analyticsName = "Overdue Screen:Action clicked on item"
```
saket: The analytics text looks wrong

PR #284, opened 2018-09-27
Fix: Schedule Appointment sheet doesn't work if you update medicines

sanchita:
…after saving BP

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
@@ -21,6 +21,7 @@ import io.reactivex.android.schedulers.AndroidSchedulers.mainThread
 import io.reactivex.rxkotlin.ofType
 import io.reactivex.schedulers.Schedulers.io
 import io.reactivex.subjects.PublishSubject
+import kotlinx.android.parcel.Parcelize
 import kotterknife.bindView
 import org.simple.clinic.R
 import org.simple.clinic.activity.TheActivity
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -79,18 +80,24 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
...
 
   private var bpEntrySheetAlreadyShownOnStart: Boolean = false
+  private var bloodPressureWasSaved: Boolean = false
 
   override fun onSaveInstanceState(): Parcelable {
     return PatientSummaryScreenSavedState(
         super.onSaveInstanceState(),
-        bpEntryShownOnStart = bpEntrySheetAlreadyShownOnStart)
+        bpEntryShownOnStart = bpEntrySheetAlreadyShownOnStart,
+        bloodPressureWasSaved = bloodPressureWasSaved)
```
saket: Can we use the same tone for both `bpEntryShownOnStart` and `bloodPressureWasSaved`?

saket: If you decide to change `bloodPressureWasSaved` then `bpWasSavedEvents` will also need to be updated

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -145,6 +153,7 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
   private fun bloodPressureSaves() = screenRouter.streamScreenResults()
       .ofType<ActivityResult>()
       .filter { it.requestCode == REQCODE_BP_SAVED && it.succeeded() }
+      .doOnNext { bloodPressureWasSaved = BloodPressureEntrySheet.wasBloodPressureSaved(it.data!!) }
       .map { PatientSummaryBloodPressureClosed(BloodPressureEntrySheet.wasBloodPressureSaved(it.data!!)) }
```
saket: Instead of extracting the result from the `Intent` twice, we can instead add another map:

```kotlin
private fun bloodPressureSaves() = screenRouter.streamScreenResults()
      .ofType<ActivityResult>()
      .filter { it.requestCode == REQCODE_BP_SAVED && it.succeeded() }
      .map {  BloodPressureEntrySheet.wasBloodPressureSaved(it.data!!) }
      .doOnNext { saved -> bloodPressureWasSaved = saved }
      .map { saved -> PatientSummaryBloodPressureClosed(saved) }
```

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
```diff
@@ -189,11 +189,23 @@ class PatientSummaryScreenControllerTest {
...
     } else {
       verify(screen).goBackToHome()
+      verify(screen, never()).showScheduleAppointmentSheet(any())
     }
   }
 
+  @Test
+  fun `when summary screen is restored, and bp was saved earlier, schedule appointment sheet should open, on clicking back or done`() {
+    uiEvents.onNext(PatientSummaryScreenCreated(patientUuid, caller = PatientSummaryCaller.SEARCH))
+    uiEvents.onNext(PatientSummaryRestoredWithBPSaved(true))
```
saket: It looks like we are testing one flow where `PatientSummaryRestoredWithBPSaved` is emitted with `true`, but not these remaining flows:

- emitted with false
- never emitted because `onRestoreInstanceState()` doesn't get called

What do you think?

PR #286, opened 2018-09-28
Heisenbug

saket:
This PR introduces `autoConnect()` instead of `refCount()` to avoid multiple subscriptions and disposals of the Ui stream, which in turn results in multiple counts of binding and unbinding of Views. While this sounds okay in theory, doing this in a multi-threaded environment causes a race condition.

This fixes the bug in summary screen where back and save buttons stop responding. If this fix works fine, we can introduce it in all screen controllers.

**Steps to potentially reproduce**
1. Force kill app from recents
2. Open app. Lockscreen should open.
3. Enter PIN to log in.
4. Search for a patient
5. Open a patient's summary
6. Software up button will not work
7. If it works, go back to 4 and repeat till it doesn’t. Try doing this very fast. Starting a data-sync sometimes helps.

I've noticed that running data sync increases the chances of hitting this bug. Restarting the app helps because it triggers a force sync.

### DIFF
--- a/app/src/main/java/org/simple/clinic/wip/PrimarySolidButtonDrawable.kt
+++ b/app/src/main/java/org/simple/clinic/wip/PrimarySolidButtonDrawable.kt
@@ -1,0 +1,101 @@
+package org.simple.clinic.wip
+
+import android.content.res.Resources
+import android.graphics.Canvas
+import android.graphics.Color
+import android.graphics.Paint
+import android.graphics.Rect
+import android.graphics.RectF
+import android.graphics.drawable.ColorDrawable
+import android.graphics.drawable.Drawable
+import android.graphics.drawable.GradientDrawable
+import android.graphics.drawable.StateListDrawable
+import android.os.Build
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/wip/PrimarySolidButtonDrawable.kt
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * WIP.
+ *
+ * This drawable supports reading drawables from Xml selectors, making it compatible
+ * with the standard way of applying different Drawable states. The current implementation of
+ * [PrimarySolidButton] involves manually setting the enabled and disabled colors as
+ * direct attributes on the button.
+ */
+class PrimarySolidButtonDrawable(delegate: StateListDrawable) : StateListDrawable() {
```
saket: FYI: this is a work in progress thing and can be skipped for review. I just want to check it in for future usage.

PR #291, opened 2018-10-03
Add 'diagnosed with hypertension' question to medical history

saket:
Gigantic PR because it includes a DB migration.


app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -39,7 +39,7 @@ import org.simple.clinic.util.UuidRoomTypeConverter
       Appointment::class,
       Communication::class,
       MedicalHistory::class],
-    version = 14,
+    version = 15,
```
vinaysshenoy: Can you please ensure the schema diff is checked in along with the version change?

app/src/main/java/org/simple/clinic/storage/Migration_14_15.kt
```diff
@@ -0,0 +1,20 @@
...
+ * Adds [MedicalHistory.diagnosedWithHypertension] column.
+ */
+@Suppress("ClassName")
+class Migration_14_15 : Migration(14, 15) {
+
+  override fun migrate(database: SupportSQLiteDatabase) {
+    database.execSQL("""
+      ALTER TABLE `MedicalHistory`
+      ADD COLUMN `diagnosedWithHypertension` INTEGER NOT NULL
+      DEFAULT '0'
```
vinaysshenoy: This should be `DEFAULT 0` because the field is an `INTEGER`.

app/src/main/java/org/simple/clinic/AppDatabase.kt
saket: done

PR #295, opened 2018-10-04
Change toolbar title styles to medium

vinaysshenoy:


app/src/main/res/values/styles_v2.xml
```diff
@@ -28,6 +28,12 @@
 
   <style name="Clinic.V2.TextAppearance" />
 
+  <style name="Clinic.V2.TextAppearance.Title">
```
saket: Do you think `Toolbar.Title` will be better than just `Title`?

app/src/main/res/layout/screen_facility_change.xml
```diff
@@ -14,7 +14,9 @@
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
     app:navigationIcon="@drawable/ic_arrow_back_24dp"
-    app:title="@string/facilitychange_screen_title" />
+    app:title="@string/facilitychange_screen_title"
+    app:titleTextAppearance="@style/Clinic.V2.TextAppearance.Title"
+    />
```
saket: wow such bad formatting :p

saket:
We already have a default textAppearance for toolbars. Can you reuse `Clinic.V2.TextAppearance.ToolbarTitle` and maybe also rename it as per zeplin?

PR #296, opened 2018-10-04
Update Indian PhoneNumberObfuscator with new specs.

vinaysshenoy:
1. Last three digits are always visible.
2. Five digits before the last three are masked.
3. All preceding digits are also visible.

### DIFF
--- a/app/src/main/java/org/simple/clinic/search/results/PhoneNumberObfuscator.kt
+++ b/app/src/main/java/org/simple/clinic/search/results/PhoneNumberObfuscator.kt
@@ -1,5 +1,7 @@
 package org.simple.clinic.search.results
 
+import org.simple.clinic.util.Unicode.bullet
+import org.simple.clinic.util.Unicode.hairSpace
 import javax.inject.Inject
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/simple/clinic/search/results/PhoneNumberObfuscator.kt
```diff
@@ -1,20 +1,30 @@
...
-    val first2Digits = number.substring(0, 2)
-    val last3Digits = number.substring(number.length - 3, number.length)
-    val obfuscatedDigits = "• ".repeat(number.length - 5)
-    return "$first2Digits $obfuscatedDigits$last3Digits"
+    return number
+        .mapIndexed { index, char ->
+          when (index) {
+            in lastThreeIndices -> char.toString()
+            in fiveBeforeLastThree -> mask
+            fiveBeforeLastThree.start - 1 -> "$char$hairSpace"
```
saket: It'll be super nice to also name this `fiveBeforeLastThree.start - 1` just like how we've named as `fiveBeforeLastThree`.

PR #301, opened 2018-10-05
Show placeholders for BP measurements

vinaysshenoy:
Refer Zeplin for designs.

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
@@ -137,7 +137,7 @@ class PatientSummaryScreenController @Inject constructor(
     return Observables.combineLatest(prescriptionItems, bloodPressureItems, medicalHistoryItems)
         .map { (prescriptions, bp, history) ->
           { ui: Ui ->
-            ui.populateList(prescriptions, bp, history)
+            ui.populateList(prescriptions, emptyList(), bp, history)
           }
         }
   }
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
@@ -116,12 +116,34 @@ class PatientSummaryScreenController @Inject constructor(
         .flatMap { prescriptionRepository.newestPrescriptionsForPatient(it) }
         .map(::SummaryPrescribedDrugsItem)
 
-    val bloodPressureItems = patientUuids
+    val bloodPressures = patientUuids
         .flatMap { bpRepository.newest100MeasurementsForPatient(it) }
+        .replay(1)
+        .refCount()
+
+    val bloodPressureItems = bloodPressures
         .map { measurements ->
           measurements.map { measurement ->
             val timestamp = timestampGenerator.generate(measurement.updatedAt)
             SummaryBloodPressureListItem(measurement, timestamp)
+          }
+        }
+
+    val bloodPressurePlaceholders = bloodPressures
+        .map { it.size }
+        .map { numberOfBloodPressures ->
+          when (numberOfBloodPressures) {
+            0 -> listOf(
+                SummaryBloodPressurePlaceholderListItem(1, showHint = true),
+                SummaryBloodPressurePlaceholderListItem(2),
+                SummaryBloodPressurePlaceholderListItem(3)
+            )
+            1 -> listOf(
+                SummaryBloodPressurePlaceholderListItem(1),
+                SummaryBloodPressurePlaceholderListItem(2)
+            )
+            2 -> listOf(SummaryBloodPressurePlaceholderListItem(1))
+            else -> emptyList()
           }
         }
 
@@ -134,12 +156,13 @@ class PatientSummaryScreenController @Inject constructor(
 
     // combineLatest() is important here so that the first data-set for the list
     // is dispatched in one go instead of them appearing one after another on the UI.
-    return Observables.combineLatest(prescriptionItems, bloodPressureItems, medicalHistoryItems)
-        .map { (prescriptions, bp, history) ->
-          { ui: Ui ->
-            ui.populateList(prescriptions, emptyList(), bp, history)
-          }
-        }
+    return Observables.combineLatest(
+        prescriptionItems,
+        bloodPressureItems,
+        medicalHistoryItems,
+        bloodPressurePlaceholders) { prescriptions, bp, history, placeHolders ->
+      { ui: Ui -> ui.populateList(prescriptions, placeHolders, bp, history) }
+    }
   }
 
   private fun updateMedicalHistory(events: Observable<UiEvent>): Observable<UiChange> {
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
@@ -34,6 +34,8 @@ import javax.inject.Inject
 typealias Ui = PatientSummaryScreen
 typealias UiChange = (Ui) -> Unit
 
+const val numBpPlaceholders = 3
+
 class PatientSummaryScreenController @Inject constructor(
     private val patientRepository: PatientRepository,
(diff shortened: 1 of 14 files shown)


vinaysshenoy:
@saket I'm not sure abut how the IDs I am generating for the BP placeholder items will work with Groupie and the reset of the views in the Summary list. Could you take a closer look at this please?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -116,15 +117,32 @@ class PatientSummaryScreenController @Inject constructor(
...
             val timestamp = timestampGenerator.generate(measurement.updatedAt)
             SummaryBloodPressureListItem(measurement, timestamp)
           }
         }
 
+    val bloodPressurePlaceholders = bloodPressures
+        .map { it.size }
+        .map { numberOfBloodPressures ->
+          val numberOfPlaceholders = config.numberOfBpPlaceholders
+          when {
```
saket: TODO: see if this can be made simpler

app/src/main/java/org/simple/clinic/summary/SummaryBloodPressurePlaceholderListItem.kt
```diff
@@ -0,0 +1,42 @@
...
+
+data class SummaryBloodPressurePlaceholderListItem(
+    private val placeholderNumber: Int,
+    private val showHint: Boolean = false
+    // We have three placeholders (might change in the future) so we can't hardcode an id. At the same time, we
+    // we have to choose IDs that will stop it from clashing with other IDs in the adapter, so we multiply the
+    // placeholder number with -20.
+
+    // FIXME: Figure out an automatic solution which will work for all IDs. Also see why we can't use Groupie's
+    // auto generated IDs.
```
saket: Groupie auto-generates incrementing IDs which will not be stable. Different items can get the same ID.

saket: Coming up with an ID strategy is something I haven't been able to solve properly. What has worked best for me so far is to keep this strategy information in the adapter or a single place so that it's visible to everyone. The controller can then use it for generating Ui models. These IDs usually lie on the negative side to keep them from clashing with actual IDs.

```kotlin
val adapterId = ADAPTER_ID_OFFSET_FOR_BP_PLACEHOLDERS + placeholderIndex
SummaryBloodPressurePlaceholderListItem(adapterId, shouldShowHint)
```

app/src/main/java/org/simple/clinic/summary/SummaryBloodPressurePlaceholderListItem.kt
```diff
@@ -0,0 +1,42 @@
...
+  override fun createViewHolder(itemView: View): BpPlaceholderViewHolder {
+    return BpPlaceholderViewHolder(itemView)
+  }
+
+  override fun bind(holder: BpPlaceholderViewHolder, position: Int) {
+    holder.placeHolderMessageTextView.visibility = if (showHint) View.VISIBLE else View.INVISIBLE
+  }
+
+  override fun isSameAs(other: Item<*>?): Boolean {
+    return this == other
```
saket: The default implementation looks sufficient for diffing and we may not need this overridden function here. Can you check?

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
```diff
@@ -54,6 +55,15 @@ class PatientSummaryScreenControllerTest {
 
   private val uiEvents = PublishSubject.create<UiEvent>()
   private val reporter = MockReporter()
+
+  // If the placeholder count is changed, the `params for placeholder bp items`
+  // method will also need to be changed to account for the change in the
+  // number of placeholder.
+
+  // TODO: Change the method to generate test inputs dynamically based on
+  // the number of placeholders declared.
```
saket: This looks easily doable. Let's do it as part of this commit?

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
```diff
@@ -126,13 +137,93 @@ class PatientSummaryScreenControllerTest {
...
+        measurementItems = check {
+          it.forEachIndexed { index, item -> assertThat(item.measurement).isEqualTo(expectedBloodPressureMeasurementItems[index].measurement) }
+        },
+        medicalHistoryItem = any()
+    )
+  }
+
+  @Suppress("Unused")
+  private fun `params for placeholder bp items`(): List<List<Any>> {
+    val bpsForTest2 = listOf(PatientMocker.bp(patientUuid))
```
saket: Is `bpsForTest1` also present somewhere?

app/src/main/java/org/simple/clinic/summary/SummaryBloodPressurePlaceholderListItem.kt
vinaysshenoy: It's sufficient. Will remove.

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
vinaysshenoy: Well, it's meant to be `bps for test #2`. `bps for test 1` is an empty list. This will anyway be removed when I change it to dynamically generated tests.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -116,15 +117,32 @@ class PatientSummaryScreenController @Inject constructor(
...
+            numberOfBloodPressures >= numberOfPlaceholders -> emptyList()
+            else -> (0 until numberOfPlaceholders - numberOfBloodPressures).map { placeholderIndex ->
+              val shouldShowHint = numberOfBloodPressures == 0 && placeholderIndex == 0
+              SummaryBloodPressurePlaceholderListItem(placeholderIndex + 1, shouldShowHint)
+            }
+          }
+        }
```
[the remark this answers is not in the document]
vinaysshenoy: Looks good. I'll check this out.

app/src/main/java/org/simple/clinic/summary/SummaryListAdapterIds.kt
```diff
@@ -0,0 +1,12 @@
+package org.simple.clinic.summary
+
+object SummaryListAdapterIds {
+  const val PRESCRIBED_DRUGS = -1L
+  const val ADD_PRESCRIPTION = -2L
+  const val ADD_NEW_BP = -3L
```
saket: Considering that these are getting statically imported, do you think we should add an `ID_` prefix? This is also fine.

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
vinaysshenoy: Discussed: Is not very easily doable. Will revisit in the future.

app/src/main/java/org/simple/clinic/summary/SummaryListAdapterIds.kt
vinaysshenoy: The reason I didn't do it was because the enclosing object already has an `Ids` in the name.

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
saket: @vinaysshenoy you should try making it a habit to use multi-line comments for code that are outside functions. This way, you can also add a link to `params for placeholder bp items()` function. I'll do this for you.

saket: @vinaysshenoy you should try using multi-line comments for code that are outside functions, as per the standard. This way, you can also add a link to `params for placeholder bp items()` function. I'll do this for you.

PR #308, opened 2018-10-09
Patient Entry UI Changes

vinaysshenoy:
1. Remove `None` checkbox from phone number and make it optional.
2. Remove `None` checkbox from colony and make it mandatory.

### DIFF
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
@@ -21,7 +21,6 @@ import org.simple.clinic.patient.OngoingPatientEntry.PhoneNumber
 import org.simple.clinic.patient.PatientEntryValidationError.BOTH_DATEOFBIRTH_AND_AGE_ABSENT
 import org.simple.clinic.patient.PatientEntryValidationError.BOTH_DATEOFBIRTH_AND_AGE_PRESENT
 import org.simple.clinic.patient.PatientEntryValidationError.COLONY_OR_VILLAGE_EMPTY
-import org.simple.clinic.patient.PatientEntryValidationError.COLONY_OR_VILLAGE_NON_NULL_BUT_BLANK
 import org.simple.clinic.patient.PatientEntryValidationError.DATE_OF_BIRTH_IN_FUTURE
 import org.simple.clinic.patient.PatientEntryValidationError.DISTRICT_EMPTY
 import org.simple.clinic.patient.PatientEntryValidationError.EMPTY_ADDRESS_DETAILS
@@ -69,23 +68,7 @@ class PatientEntryScreenController @Inject constructor(
         savePatient(transformedEvents),
         showValidationErrorsOnSaveClick(transformedEvents),
         resetValidationErrors(transformedEvents),
-        noneCheckBoxBehaviors(transformedEvents),
         scrollToBottomOnGenderSelection(transformedEvents))
-  }
-
-  private fun noneCheckBoxBehaviors(events: Observable<UiEvent>): Observable<UiChange> {
-    val noVillageOrColonyUnchecks = events
-        .ofType<PatientColonyOrVillageTextChanged>()
-        .map { it.colonyOrVillage.isNotBlank() }
-        .distinctUntilChanged()
-        .map { hasVillageOrColony -> { ui: Ui -> ui.setNoVillageOrColonyCheckboxVisible(!hasVillageOrColony) } }
-
-    val colonyOrVillageResets = events
-        .ofType<PatientNoColonyOrVillageToggled>()
-        .filter { event -> event.noneSelected }
-        .map { { ui: Ui -> ui.resetColonyOrVillageField() } }
-
-    return Observable.merge(colonyOrVillageResets, noVillageOrColonyUnchecks)
   }
 
   private fun preFillOnStart(events: Observable<UiEvent>): Observable<UiChange> {
@@ -168,20 +151,9 @@ class PatientEntryScreenController @Inject constructor(
         .ofType<PatientPhoneNumberTextChanged>()
         .map { (phoneNumber) -> if (phoneNumber.isBlank()) None else Just(PhoneNumber(phoneNumber)) }
 
-    val colonyOrVillageChanges = Observables
-        .combineLatest(
-            events
-                .ofType<PatientNoColonyOrVillageToggled>()
-                .map { it.noneSelected },
-            events
-                .ofType<PatientColonyOrVillageTextChanged>()
-                .map { it.colonyOrVillage })
-        .map { (noneSelected, colonyOrVillage) ->
-          when {
-            noneSelected -> None
-            else -> Just(colonyOrVillage!!)
-          }
-        }
+    val colonyOrVillageChanges = events
+        .ofType<PatientColonyOrVillageTextChanged>()
+        .map { it.colonyOrVillage }
 
     val districtChanges = events
         .ofType<PatientDistrictTextChanged>()
@@ -191,13 +163,10 @@ class PatientEntryScreenController @Inject constructor(
         .ofType<PatientStateTextChanged>()
         .map { it.state }
 
-    val addressChanges = Observables.combineLatest(colonyOrVillageChanges, districtChanges, stateChanges)
-    { colonyOrVillage, district, state ->
-      Address(colonyOrVillage.toNullable().nullIfBlank(), district, state)
-    }
+    val addressChanges = Observables
+        .combineLatest(colonyOrVillageChanges, districtChanges, stateChanges, ::Address)
 
-    return Observables.combineLatest(personDetailChanges, phoneNumberChanges, addressChanges)
-    { personal, phone, address ->
+    return Observables.combineLatest(personDetailChanges, phoneNumberChanges, addressChanges) { personal, phone, address ->
       OngoingPatientEntryChanged(OngoingPatientEntry(personal, address, phone.toNullable()))
     }
   }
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
```diff
@@ -175,39 +144,16 @@ class PatientEntryScreenController @Inject constructor(
...
-        .map { (noneSelected, phoneNumber) ->
-          when {
-            noneSelected -> None
-            phoneNumber.isBlank() -> None
-            else -> Just(OngoingPatientEntry.PhoneNumber(phoneNumber!!))
-          }
-        }
+    val phoneNumberChanges = events
+        .ofType<PatientPhoneNumberTextChanged>()
+        .map { (phoneNumber) -> if (phoneNumber.isBlank()) None else Just(PhoneNumber(phoneNumber)) }
```
saket: Is `Optional` needed here anymore? `colonyOrVillageChanges` doesn't use it.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
```diff
@@ -217,13 +163,12 @@ class PatientEntryScreenController @Inject constructor(
...
         .map { it.state }
 
-    val addressChanges = Observables.combineLatest(colonyOrVillageChanges, districtChanges, stateChanges)
-    { colonyOrVillage, district, state ->
-      OngoingPatientEntry.Address(colonyOrVillage.toNullable().nullIfBlank(), district, state)
-    }
+    val addressChanges = Observables
+        .combineLatest(colonyOrVillageChanges, districtChanges, stateChanges) { colonyOrVillage, district, state ->
+          Address(colonyOrVillage, district, state)
+        }
```
saket: Should we use function reference to make this even shorter?

```kotlin
val addressChanges = Observables.combineLatest(colonyOrVillageChanges, districtChanges, stateChanges, ::Address)
```

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
```diff
@@ -217,13 +163,12 @@ class PatientEntryScreenController @Inject constructor(
...
 
-    return Observables.combineLatest(personDetailChanges, phoneNumberChanges, addressChanges)
-    { personal, phone, address ->
+    return Observables.combineLatest(personDetailChanges, phoneNumberChanges, addressChanges) { personal, phone, address ->
       OngoingPatientEntryChanged(OngoingPatientEntry(personal, address, phone.toNullable()))
```
saket: `phone.toNullable()` can also go away if you decide to remove `Optional` from `phoneNumberChanges`.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
```diff
@@ -238,28 +183,14 @@ class PatientEntryScreenController @Inject constructor(
...
 
     val colonyErrors = events
         .ofType<PatientEntrySaveClicked>()
-        .withLatestFrom(colonyChanges, noColonyToggles)
-        .map { (_, colony, isNoneSelected) ->
+        .withLatestFrom(colonyChanges)
+        .map { (_, colony) ->
           when {
-            (colony.isBlank() && !isNoneSelected) -> listOf(COLONY_OR_VILLAGE_EMPTY)
+            colony.isBlank() -> listOf(COLONY_OR_VILLAGE_EMPTY)
```
saket: Colony was previously validated in two places. Once here and inside `OngoingPatientEntry#validationErrors()` because the entry model did not know the existence of the none checkbox. Now that the checkbox is removed, I think this is no longer required.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
```diff
@@ -330,20 +259,13 @@ class PatientEntryScreenController @Inject constructor(
         .ofType<PatientGenderChanged>()
         .map { { ui: Ui -> ui.showMissingGenderError(false) } }
 
-    val noPhoneErrorResets = Observable
-        .merge(events.ofType<PatientPhoneNumberTextChanged>(), events.ofType<PatientNoPhoneNumberToggled>())
-        .map { { ui: Ui -> ui.showEmptyPhoneNumberError(false) } }
-
-    val lessPhoneLengthResets = Observable
-        .merge(events.ofType<PatientPhoneNumberTextChanged>(), events.ofType<PatientNoPhoneNumberToggled>())
+    val lessPhoneLengthResets = events.ofType<PatientPhoneNumberTextChanged>()
```
saket: Whoops, hey can you please rename this variable to `phoneLengthTooShortResets`? This rename was supposed to happen in an earlier PR by Sanchita, but I missed that.

saket: Same for `morePhoneLengthResets`

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
```diff
@@ -368,20 +289,12 @@ class PatientEntryScreenController @Inject constructor(
...
-    val noPhoneToggles = events.ofType<PatientNoPhoneNumberToggled>().map { it.noneSelected }
     val colonyChanges = events.ofType<PatientColonyOrVillageTextChanged>().map { it.colonyOrVillage }
-    val noColonyToggles = events.ofType<PatientNoColonyOrVillageToggled>().map { it.noneSelected }
-
-    val isPhoneNumberValid = events
-        .ofType<PatientEntrySaveClicked>()
-        .withLatestFrom(phoneNumberChanges, noPhoneToggles)
-        .map { (_, phoneNumber, isNoneSelected) -> phoneNumber.isNotBlank() || isNoneSelected }
 
     val isColonyValid = events
```
saket: This validation can also be removed. It's handled by `OngoingPatientEntry#validationErrors()`.

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
```diff
@@ -290,29 +254,24 @@ class PatientEntryScreenControllerTest {
...
     uiEvents.onNext(PatientColonyOrVillageTextChanged(""))
 
     verify(screen).showEmptyFullNameError(false)
-    verify(screen, times(4)).showEmptyPhoneNumberError(false)
     verify(screen, times(3)).showEmptyDateOfBirthAndAgeError(false)
     verify(screen, times(2)).showInvalidDateOfBirthError(false)
     verify(screen, times(2)).showDateOfBirthIsInFutureError(false)
     verify(screen).showMissingGenderError(false)
-    verify(screen, times(4)).showEmptyColonyOrVillageError(false)
+    verify(screen, atLeastOnce()).showEmptyColonyOrVillageError(false)
```
saket: When writing this test, I intentionally used specific `times` to ensure that I do not miss any validations. Do you think that's not required?

saket:
@vinaysshenoy will you add `X` buttons for phone numbers and colony in this PR?

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
vinaysshenoy: That's the wrong way to verify validations since it creates a tight coupling between the test and the implementation. If we are verifying this, we should be sending a list of errors to the screen and validate that the errors sent to screen are matching what we expect.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
vinaysshenoy: `colonyOrVillageChanges` is no longer an `Optional`. Did you mean `phoneNumberChanges`? It needs to be optional since phone number is still optional to enter.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
vinaysshenoy: Phone number is still optional. We just don't have an explicit option to make it nullable.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
vinaysshenoy: Wow, I did not know this was possible. <3

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
saket: Hmmmmmmmmmmmm. In that case, do you want to remove `times` from other checks in this function? Otherwise we'll have a mix of both.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
saket: Ah okay

saket: Wait, you're not making use of `Optional` anywhere in this function, right? It gets immediately converted back to a nullable string using `toNullable`.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
vinaysshenoy: TODO: Add an assertion for creating a `PhoneNumber` instance with a blank phone number.

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
vinaysshenoy: I'd rather leave it for now, and change the screen api itself in a future PR to handle this change and update the test then. WDYT?

saket: That sounds like the ideal case, but let's not leave this function hanging in the middle. I've seen Sanchita use this function as a reference for one of her tests, and it'll only confuse other people more.

PR #309, opened 2018-10-09
Change window background colour to grey4

vinaysshenoy:


vinaysshenoy:
@saket This is an extremely large PR.

PR #310, opened 2018-10-10
Order facilities by name in ascending order

saket:


app/src/main/java/org/simple/clinic/facility/FacilityRepository.kt
```diff
@@ -18,7 +18,7 @@ class FacilityRepository @Inject constructor(
     val userFacilityMappingDao: LoggedInUserFacilityMapping.RoomDao
 ) {
 
-  fun facilities(): Observable<List<Facility>> {
+  fun facilitiesInAscending(): Observable<List<Facility>> {
```
saket: Accepting better names. I also considered,

- Adding a boolean param `facilities(orderByAscending=true)`
- Adding an enum `facilities(order = ASC)`

vinaysshenoy: Naming is tricky. This name doesn't indicate what to order by ascending; it could be by the name, or the date at which it was created.

I would suggest just changing the name to `facilitiesSortedByName()`, which implies a natural ordering and leave it for now. We can revisit this if the UI needs it in future (with user selected sorting order or something similar).

app/src/main/java/org/simple/clinic/registration/name/RegistrationFullNameScreenController.kt
```diff
@@ -45,7 +45,7 @@ class RegistrationFullNameScreenController @Inject constructor(
     return events
         .ofType<RegistrationFullNameScreenCreated>()
         .flatMap {
-          facilityRepository.facilities()
+          facilityRepository.facilitiesInAscending()
```
vinaysshenoy: Since the only thing we are using this for is to check if the facilities have been synced at least once, we should change it to just get the facility count instead of loading all facilities.

saket: It's a small data-set (20 rows) so the performance benefit will be negligible. I'm not very sure if I want to touch this right now.

app/src/main/java/org/simple/clinic/facility/FacilityRepository.kt
pratul: IMHO, we should not change this to anything. Let it remain `facilities()` -- we don't need to expose the sorting order at all, because we don't care about the order (or about changing it) right now.

app/src/main/java/org/simple/clinic/registration/name/RegistrationFullNameScreenController.kt
pratul: Agree with Vinay, but +1 to Saket. Seems like a micro-optimization to do right now.

PR #311, opened 2018-10-11
Change Patient Sync to the new Synceable flow

vinaysshenoy:


app/src/main/java/org/simple/clinic/patient/PatientQueryModel.kt
```diff
@@ -0,0 +1,15 @@
+package org.simple.clinic.patient
+
+import android.arch.persistence.room.Embedded
+
+data class PatientQueryModel(
```
vinaysshenoy: This class was introduced because Room doesn't have in-built support for 1:N relationships and a single patient can have multiple phone numbers.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -172,9 +173,41 @@ class PatientRepository @Inject constructor(
         }
   }
 
-  fun mergeWithLocalData(serverPayloads: List<PatientPayload>): Completable {
+  override fun recordsWithSyncStatus(syncStatus: SyncStatus): Single<List<PatientSaveModel>> {
+    return database.patientDao()
+        .recordsWithSyncStatus(syncStatus)
+        .firstOrError()
+        .map { results ->
+          results.asSequence()
```
vinaysshenoy: This is a little complicated, and I want to explain what's going on. 

`PatientSaveModel` has a non-nullable list of patient phone numbers and it expects that a Patient without phone numbers will have an empty list of phone numbers. However, `PatientQueryModel` is constructed from a join of three tables, and the phone number table is joined with a `LEFT JOIN` so missing phone numbers will be `null`.

What this does is checks if the first phone number for a Patient query model with a given patient ID is `null`, and if it is, sets the phone numbers of the `PatientSaveModel` to an empty list, otherwise it extracts all the phone numbers.

app/src/androidTest/java/org/simple/clinic/patient/PatientSyncAndroidTest.kt
```diff
@@ -2,122 +2,55 @@ package org.simple.clinic.patient
...
-import java.text.SimpleDateFormat
 import javax.inject.Inject
 import javax.inject.Named
 
 @RunWith(AndroidJUnit4::class)
-class PatientSyncAndroidTest {
-
-  @Inject
-  lateinit var configProvider: Single<SyncConfig>
+class PatientSyncAndroidTest : BaseDataSyncAndroidTest<PatientSaveModel, PatientPayload>() {
```
vinaysshenoy: It might be easier to view this as an individual file. I created a separate `PatientSync2AndroidTest` and wrote it separately, then renamed it `PatientSyncAndroidTest` and rebased it with this commit, but that makes the diff inconvenient to parse.

app/src/test/java/org/simple/clinic/patient/PatientSyncTest.kt
```diff
@@ -1,121 +0,0 @@
...
-    verify(api, never()).push(any())
-  }
-
-  @Test
-  fun `when a push succeeds then any patients that were created or updated during the push started should not get marked as synced`() {
-    // TODO: I'm out of ideas to write this test because it involves concurrency.
-  }
-
-  @Test
-  fun `errors during push should not affect pull`() {
```
vinaysshenoy: This (and the next function) seemed unnecessary to test since we are only testing whether `Completable.mergeArrayDelayError` works.

saket: No, my intention was to test that errors in the implementation get swallowed even if someone replaces `mergeArrayDelayError` with something else.

app/src/main/java/org/simple/clinic/patient/PatientSaveModel.kt
```diff
@@ -0,0 +1,3 @@
+package org.simple.clinic.patient
+
+data class PatientSaveModel(val patient: Patient, val address: PatientAddress, val phoneNumbers: List<PatientPhoneNumber>)
```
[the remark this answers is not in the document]
saket: I try to avoid using suffixes when naming data classes whenever possible. It feels like one of those cases where people are accustomed to suffixing classes with verbs. For instance, we have `PatientSearchResult` instead of `PatientSearchResultModel` and so on.

At the same time, there are classes that use a suffix. For e.g., `SummaryAddNewBpListItem`, where we couldn't think of a better name. 

Do you think we can name `PatientSaveModel` to something better?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
saket: > However, PatientQueryModel is constructed from a join of three tables, and the phone number table is joined with a LEFT JOIN so missing phone numbers will be null.

Can default values in Kotlin solve this problem?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -175,43 +146,71 @@ class PatientRepository @Inject constructor(
...
+              .groupBy { it.patient.uuid }
+              .map { (_, patientQueryModels) ->
+                PatientSaveModel(
+                    patient = patientQueryModels.first().patient,
+                    address = patientQueryModels.first().address,
+                    phoneNumbers = patientQueryModels
+                        .takeIf { it.first().phoneNumber != null }
+                        ?.asSequence()?.map { it.phoneNumber!! }?.toList()
+                        ?: emptyList()
+                )
```
saket: As you said this is indeed really complicated. There's a lot happening here. I'd like us to simplify this. I'll try doing this tomorrow.

app/src/main/java/org/simple/clinic/patient/PatientQueryModel.kt
vinaysshenoy: I took a look at it, and I don't think `@Relation` is a good fit for this use case. If we were only fetching a single patient model by `UUID`, it would be a great fit. However, we fetch all patients with a particular `SyncStatus`.

The current implementation using `PatientQueryModel` will only access the database once. However, with the `@Relation` annotation, we'll end up accessing it `numberOfPatients + 1` times.

vinaysshenoy: Also, TIL that Room has Relation support.

app/src/test/java/org/simple/clinic/patient/PatientSyncTest.kt
vinaysshenoy: Got it. The tests which were left out have now been implemented by the `ModelSyncTest` class.

app/src/main/java/org/simple/clinic/sync/ModelSync.kt
```diff
@@ -0,0 +1,9 @@
+package org.simple.clinic.sync
+
+import io.reactivex.Completable
+
+interface ModelSync {
```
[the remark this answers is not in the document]
vinaysshenoy: I agree. I feel that the naming of `DataSync` should be changed, since we already have `PatientSync`, `BloodPressureSync` etc, which do a completely different thing than what `DataSync` itself does. 

What data sync does now is more of a `SyncCoordinator`, which is what I suggest naming it.

Do you have any suggestions?

app/src/test/java/org/simple/clinic/sync/ModelSyncTest.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.nhaarman.mockito_kotlin.whenever
+import io.reactivex.Completable
+import junitparams.JUnitParamsRunner
+import junitparams.Parameters
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.simple.clinic.patient.sync.PatientSync
+
+@RunWith(JUnitParamsRunner::class)
+class ModelSyncTest {
```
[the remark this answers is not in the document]
vinaysshenoy: Good question. I don't have a clear answer at the moment.

It seemed easier to do this as a composite test right now for `ModelSyncTest`. I could also take a stab at converting `BaseDataSyncAndroidTest` to a composite form to see if that would be cleaner.

app/src/main/java/org/simple/clinic/patient/PatientSaveModel.kt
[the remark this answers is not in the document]
vinaysshenoy: Should we just call this `PatientProfile` then?

app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -96,5 +96,20 @@ data class Patient constructor(
...
+       P.updatedAt patient_updatedAt, P.syncStatus patient_syncStatus, PA.uuid addr_uuid, PA.colonyOrVillage addr_colonyOrVillage,
+       PA.district addr_district, PA.state addr_state, PA.country addr_country, PA.createdAt addr_createdAt, PA.updatedAt addr_updatedAt,
+       PPN.uuid phone_uuid, PPN.patientUuid phone_patientUuid, PPN.number phone_number, PPN.phoneType phone_phoneType, PPN.active phone_active,
+       PPN.createdAt phone_createdAt, PPN.updatedAt phone_updatedAt
+      FROM Patient P
+      INNER JOIN PatientAddress PA ON P.addressUuid == PA.uuid
+      LEFT JOIN PatientPhoneNumber PPN ON PPN.patientUuid == P.uuid
+      WHERE P.syncStatus == :syncStatus
+    """)
+    fun recordsWithSyncStatus(syncStatus: SyncStatus): Flowable<List<PatientQueryModel>>
```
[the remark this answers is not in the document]
vinaysshenoy: Yes. Making a `SynceableRepository` implementation for `PatientSearchResult` was proving to be a large endeavour and drastically increasing the scope of changes, so I opted for making a new model instead and making `PatientSearchResult` be used only for search results.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: Could you take another look at it? I have taken a stab at simplifying it.

app/src/test/java/org/simple/clinic/sync/ModelSyncTest.kt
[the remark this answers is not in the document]
vinaysshenoy: Actually, with the introduction of the `ModelSync` interface which describes a common interface all synceable models, it is possible to make it composite. Earlier, even though the individual instances of Sync classes have an identical api, we couldn't abstract them in tests since they were not conforming to a single type.

app/src/main/java/org/simple/clinic/patient/PatientQueryModel.kt
vinaysshenoy: TODO: Add comment about the existence of this class.

vinaysshenoy: TODO: Also confirm that it uses multiple database calls.

vinaysshenoy: Confirmed that `@Relation` makes only a single extra query to fetch data from the database for all parents. However, this has other issues, primarily, it does not allow loading relations into properties via the constructor and needs a settable property.

This is fixed in the newest release of `Room`, but that's an alpha release(`2.1.0-alpha01`) and  also requires migration to AndroidX. This change will be done at a later point in time when we finish this migration.

app/src/main/java/org/simple/clinic/sync/ModelSync.kt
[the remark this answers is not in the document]
vinaysshenoy: Had a discussion offline and decision was to rename `DataSync` and related classes to `SyncCoordinator`.

app/src/test/java/org/simple/clinic/sync/ModelSyncTest.kt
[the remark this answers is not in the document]
vinaysshenoy: Decision made offline was to change `BaseDataSyncAndroidTest` (now `BaseSyncCoordinatorAndroidTest`) after implementing the `ModelSync` interface on all the Sync classes to a composed form instead of inherited.

app/src/main/java/org/simple/clinic/patient/PatientSaveModel.kt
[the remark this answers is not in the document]
vinaysshenoy: Renamed to `PatientProfile`.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: Cleaned up very well by making this entire implementation internal to the Patient Dao.

PR #314, opened 2018-10-12
Report to Sentry if data syncing fails

saket:
We swallow errors if data syncing fails, but it'll be nice to see them reported on Sentry along with breadcrumbs (if any).

### DIFF
--- a/app/src/main/java/org/simple/clinic/sync/SyncWorker.kt
+++ b/app/src/main/java/org/simple/clinic/sync/SyncWorker.kt
@@ -7,6 +7,7 @@ import io.reactivex.Completable
 import io.reactivex.Single
 import org.simple.clinic.ClinicApp
 import org.simple.clinic.bp.sync.BloodPressureSync
+import org.simple.clinic.crash.CrashReporter
 import org.simple.clinic.drugs.sync.PrescriptionSync
 import org.simple.clinic.facility.FacilitySync
 import org.simple.clinic.medicalhistory.sync.MedicalHistorySync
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/analytics/DebugAnalyticsReporter.kt
```diff
@@ -1,18 +1,16 @@
 package org.simple.clinic.analytics
 
-import timber.log.Timber
-
 class DebugAnalyticsReporter : AnalyticsReporter {
 
   override fun setUserIdentity(id: String) {
-    Timber.tag("Analytics").d("User ID: $id")
+    // No-op.
```
vinaysshenoy: Instead of No-Opping them, could you instead not add the `DebugAnalyticsReporter` to the `Analytics` class? This will allow us to add it again when we need to test new analytics events.

saket: That sounds much better. I'll do this.

app/src/main/java/org/simple/clinic/sync/SyncWorker.kt
```diff
@@ -71,9 +75,10 @@ class SyncWorker : Worker() {
             medicalHistorySync.sync(),
             facilitySync.sync()
         ))
-        .doOnError {
-          if (it !is IOException) {
-            Timber.e(it)
+        .doOnError { e ->
+          if (e !is IOException) {
+            Timber.e(e)
+            crashReporter.report(e)
```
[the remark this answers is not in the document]
saket: We discussed in person that it is possible the 401 errors we're seeing might not be bugs after all. Although we checked that SharedPreferences was empty, Stetho cannot be trusted considering it has shown incorrect data in the past.

So we won't be doing this.

saket:
@vinaysshenoy PR updated

PR #318, opened 2018-10-16
Change fuzzy age calculation to a percentage of the entered age.

vinaysshenoy:


app/src/main/java/org/simple/clinic/patient/AgeFuzzer.kt
```diff
@@ -0,0 +1,56 @@
+package org.simple.clinic.patient
+
+interface AgeFuzzer {
+
+  fun fuzzy(age: Int): AgeBounds
```
sanchita: Can we rename this to `bounds` or something

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -70,8 +69,7 @@ class PatientRepository @Inject constructor(
   }
 
   fun search(name: String, assumedAge: Int, includeFuzzyNameSearch: Boolean = true): Observable<List<PatientSearchResult>> {
-    val ageUpperBound = assumedAge + ageFuzziness
-    val ageLowerBound = assumedAge - ageFuzziness
+    val (ageLowerBound, ageUpperBound) = ageFuzzer.fuzzy(assumedAge)
```
sanchita: Nice!

app/src/main/java/org/simple/clinic/patient/AgeFuzzerModule.kt
```diff
@@ -0,0 +1,11 @@
+package org.simple.clinic.patient
+
+import dagger.Module
+import dagger.Provides
+
+@Module
+open class AgeFuzzerModule {
```
sanchita: Do we need a separate module just for `AgeFuzzer`?
Shouldn't this be part of `PatientSearchModule` ?

vinaysshenoy: The problem is that `PatientRepository` is responsible for calculating the fuzziness, and it is in the `AppScope`. This comes back to why the test setup for this is not as good as I'd have preferred.

Ideally, the fuzzer should be provided to the call instead of being injected into the repository, but that would require a lot of change.

vinaysshenoy:
@sanchita PR feedback has been addressed.

I have also broken up the last commit into smaller chunks.

PR #320, opened 2018-10-16
Add option to schedule appointment for 1 month

sanchita:


app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -85,15 +85,15 @@ class ScheduleAppointmentSheetController @Inject constructor(
...
+    val toLocalDate = { amount: Int, unit: ChronoUnit ->
+      LocalDate.now(UTC).plus(amount.toLong(), unit)
     }
 
     val patientUuidStream = events.ofType<ScheduleAppointmentSheetCreated>()
         .map { it.patientUuid }
 
     return events.ofType<AppointmentScheduled>()
-        .map { toLocalDate(it.selectedDateState.second) }
+        .map { toLocalDate(it.selectedDateState.timeAmount,it.selectedDateState.chronoUnit) }
```
vinaysshenoy: There should be a space between the parameters.

Could you please set up auto formatting in your IDE?

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -85,15 +85,15 @@ class ScheduleAppointmentSheetController @Inject constructor(
   }
 
   private fun scheduleCreates(events: Observable<UiEvent>): Observable<UiChange> {
-    val toLocalDate = { pair: Pair<Int, ChronoUnit> ->
-      LocalDate.now(UTC).plus(pair.first.toLong(), pair.second)
+    val toLocalDate = { amount: Int, unit: ChronoUnit ->
```
vinaysshenoy: This can take in the `ScheduleAppointment` directly.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheet.kt
```diff
@@ -121,4 +117,6 @@ class ScheduleAppointmentSheet : BottomSheetActivity() {
   fun enableDecrementButton(state: Boolean) {
     decrementDateButton.isEnabled = state
   }
+
+  data class ScheduleAppointment(val displayText: String, val timeAmount: Int, val chronoUnit: ChronoUnit)
```
vinaysshenoy: Can you move this out of the activity and into the package as a separate file?

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheet.kt
```diff
@@ -31,31 +31,27 @@ class ScheduleAppointmentSheet : BottomSheetActivity() {
...
-      ("4 months" to (4 to ChronoUnit.MONTHS)),
-      ("5 months" to (5 to ChronoUnit.MONTHS)),
-      ("6 months" to (6 to ChronoUnit.MONTHS)),
-      ("7 months" to (7 to ChronoUnit.MONTHS)),
-      ("8 months" to (8 to ChronoUnit.MONTHS)),
-      ("9 months" to (9 to ChronoUnit.MONTHS)),
-      ("10 months" to (10 to ChronoUnit.MONTHS)),
-      ("11 months" to (11 to ChronoUnit.MONTHS)),
-      ("12 months" to (12 to ChronoUnit.MONTHS))
+      ScheduleAppointment("1 day", 1, ChronoUnit.DAYS),
```
vinaysshenoy: We should be moving these hardcoded strings to string resources. Maybe even look at building these dynamically using the `ChronoUnit` value. Thoughts?

sanchita: That is what we want to do. We have just deferred it for later.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheet.kt
sanchita: Since the scope of this class is limited to this activity, I'd prefer keeping it here.
We are doing this in other places as well. Refer `AppointmentReminderSheet`

PR #324, opened 2018-10-19
Update sync interval to 16 minutes

pratul:
Also bump up version of WorkManager, Play Services Auth, and Play Services Vision.


saket:
Welp, I just saw this. We changed the definition of “immediate sync” to "enqueued for future" and I’m amazed nothing has broken yet (except for this crash). 😄

PR #328, opened 2018-10-23
Update BP recording.

vinaysshenoy:


app/src/main/java/org/simple/clinic/summary/PatientSummaryModule.kt
```diff
@@ -2,10 +2,11 @@ package org.simple.clinic.summary
...
 import dagger.Module
 import dagger.Provides
+import org.threeten.bp.Duration
 
 @Module
 class PatientSummaryModule {
 
   @Provides
-  fun providesSummaryConfig() = PatientSummaryConfig(numberOfBpPlaceholders = 3)
+  fun providesSummaryConfig() = PatientSummaryConfig(numberOfBpPlaceholders = 3, bpEditableFor = Duration.ofMinutes(1L))
```
sanchita: The duration set for `bpEditableFor` should be configurable from server.
If we want to test on staging, we can change Date/Time settings on the phone.

Can we change the duration to the production value?

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
```diff
@@ -16,6 +16,11 @@ import org.threeten.bp.LocalDate
 import org.threeten.bp.ZoneOffset
```
sanchita: Please move this change to the test commit.

vinaysshenoy:
> 1. The length of the edit text lines in Systolic and Diastolic are different
- Not going to fix this now because there's some custom layout there and it seems that making this work right now will be more effort than expected. I'll take a look at it when working on a PR with UI changes.

> 2. The cursor on the edit text in BP entry sheet touches the edit text lines
- This is again related to the previous comment. Fixing this would require creating a custom cursor drawable which seems unnecessary at the moment. Will fix this when fixing the previous problem.

> 3. The cursor should move to the end of Diastolic text when the BP entry sheet is opened in edit mode
- Done.

> 4. The BP entry sheet, when opened in edit mode, shouldn't it get dismissed when clicked outside?
- Discussed offline and realised it's not a bug and intended behaviour.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -68,20 +69,22 @@ class BloodPressureEntrySheetController @Inject constructor(
         .ofType<BloodPressureEntrySheetCreated>()
         .filter { it.openAs == UPDATE_BP }
         .flatMapSingle { bloodPressureRepository.findOne(it.uuid) }
-        .map { bloodPressure -> { ui: Ui -> ui.updateBpMeasurements(bloodPressure.systolic, bloodPressure.diastolic)} }
+        .map { bloodPressure -> { ui: Ui -> ui.updateBpMeasurements(bloodPressure.systolic, bloodPressure.diastolic) } }
```
sanchita: Please revert the formatting.

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
vinaysshenoy: Done.

saket:
@vinaysshenoy I'll review this first thing today.

saket:
@vinaysshenoy did you miss a word at the end in the commit, `Inject PatientSummaryConfig as a Single instead of directly`?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -77,8 +89,14 @@ class BloodPressureEntrySheet : BottomSheetActivity() {
   }
 
   private fun sheetCreates(): Observable<UiEvent> {
-    val patientUuid = intent.extras.getSerializable(KEY_PATIENT_UUID) as UUID
-    return Observable.just(BloodPressureEntrySheetCreated(patientUuid))
+    val patientUuid = intent.extras!!.getSerializable(KEY_UUID) as UUID
+    val openAs = intent.extras!!.getSerializable(KEY_OPEN_AS) as OpenAs
```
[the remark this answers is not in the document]
vinaysshenoy: This is a good idea. I'll make this change in a separate PR (probably patient update).

app/src/main/java/org/simple/clinic/summary/PatientSummaryModule.kt
```diff
@@ -2,11 +2,15 @@ package org.simple.clinic.summary
...
 import org.threeten.bp.Duration
 
 @Module
 class PatientSummaryModule {
 
   @Provides
-  fun providesSummaryConfig() = PatientSummaryConfig(numberOfBpPlaceholders = 3, bpEditableFor = Duration.ofMinutes(1L))
+  fun providesSummaryConfig(): Single<PatientSummaryConfig> = Single.just(PatientSummaryConfig(
+      numberOfBpPlaceholders = 3,
+      bpEditableFor = Duration.ofMinutes(1L)
```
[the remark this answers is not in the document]
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -59,12 +62,31 @@ class BloodPressureEntrySheetController @Inject constructor(
...
     val imeDoneClicks = events.ofType<BloodPressureSaveClicked>()
 
-    val patientUuids = events
+    val patientUuidFromUpdateBp = events
         .ofType<BloodPressureEntrySheetCreated>()
+        .filter { it.openAs == UPDATE_BP }
+        .flatMapSingle { bloodPressureRepository.findOne(it.uuid) }
         .map { it.patientUuid }
+        .take(1)
+        .cache()
```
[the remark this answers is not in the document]
vinaysshenoy: This method has now been broken up into multiple methods and is no longer valid.

However what was happening was that the database was being queried for the patient ID in two separate places in the same chain, so the `cache()` was added to prevent an unnecessary database call.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -35,7 +35,8 @@ class BloodPressureEntrySheetController @Inject constructor(
         automaticDiastolicFocusChanges(replayedEvents),
         validationErrorResets(replayedEvents),
         prefillWhenUpdatingABloodPressure(replayedEvents),
-        bpValidationsAndSaves(replayedEvents))
+        bpValidationsAndSaves(replayedEvents)
+    )
```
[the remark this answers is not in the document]
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/summary/PatientSummaryModule.kt
```diff
@@ -11,6 +11,6 @@ class PatientSummaryModule {
   @Provides
   fun providesSummaryConfig(): Single<PatientSummaryConfig> = Single.just(PatientSummaryConfig(
       numberOfBpPlaceholders = 3,
-      bpEditableFor = Duration.ofMinutes(1L)
+      bpEditableFor = Duration.ofHours(1L)
```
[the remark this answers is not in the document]
vinaysshenoy: I personally find adding the type of a field to the name unnecessary in obvious cases. 

When we speak naturally, we will still say "The BP is editable for <X minutes>". The "X minutes" part here is encoded in the type `Duration` which the IDE and type system will take care of for you.

vinaysshenoy:
> @vinaysshenoy did you miss a word at the end in the commit, `Inject PatientSummaryConfig as a Single instead of directly`?

Right. I wanted to reword this, but couldn't think of something better. Will fix it.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -308,4 +310,19 @@ class PatientSummaryScreenController @Inject constructor(
...
+
+  private fun openBloodPressureUpdateSheet(events: Observable<UiEvent>): Observable<UiChange> {
+    return events.ofType<PatientSummaryBpClicked>()
+        .map { it.bloodPressureMeasurement }
+        .withLatestFrom(configProvider.toObservable())
+        .filter { (bloodPressureMeasurement, config) ->
+          val now = Instant.now(clock)
+          val cannotEditAfter = bloodPressureMeasurement.createdAt.plus(config.bpEditableFor)
+
+          now.isBefore(cannotEditAfter) || now == cannotEditAfter
```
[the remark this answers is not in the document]
saket: Using relative terms such as `before`, `after` in variable names can make it hard to understand a calculation. And date calculations are already hard. 

Can I suggest improving this?

```kotlin
.filter { (measurement, config) ->
  val now = Instant.now(clock)
  val editExpiresAt = measurement.createdAt.plus(config.bpEditableFor)
  now <= editExpiresAt
}
```

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
```diff
@@ -381,6 +381,61 @@ class PatientSummaryScreenControllerTest {
     verify(medicalHistoryRepository).save(eq(updatedMedicalHistory), any())
   }
 
+  @Test
+  @Parameters(method = "params for editing blood pressures")
+  fun `when blood pressure is clicked for editing, only those recorded in a given period must be edited`(
```
saket: `edited` -> `editable` ?

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
```diff
@@ -381,6 +381,61 @@ class PatientSummaryScreenControllerTest {
...
+
+    if(shouldBeEditable) {
+      verify(screen).showBloodPressureUpdateSheet(bloodPressureMeasurement.uuid)
+    } else {
+      verify(screen, never()).showBloodPressureUpdateSheet(any())
+    }
+  }
+
+  @Suppress("Unused")
+  private fun `params for editing blood pressures`(): List<List<Any>> {
```
saket: @vinaysshenoy this function doesn't look it was written in readable manner for people :p There are lots of magic values like `duration1`, `duration2` which I'm having trouble understanding. Can this be improved?

saket: @vinaysshenoy this function doesn't look it was written in readable manner for people :p There are lots of magic values like `duration1`, `duration2` which I'm having trouble understanding. Can this be improved?

saket:
Can the `Handle BP click events in PatientSummary` commit be made better? `Open BP sheet when an editable BP is clicked`? `Invoke the open BP update sheet method` can also be removed from the description :p

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -35,12 +35,25 @@ class BloodPressureEntrySheet : BottomSheetActivity() {
...
+    fun intentForNewBp(context: Context, patientUuid: UUID): Intent {
+      return Intent(context, BloodPressureEntrySheet::class.java)
+          .putExtra(KEY_OPEN_AS, OpenAs.NEW_BP)
+          .putExtra(KEY_UUID, patientUuid)
+    }
+
+    fun intentForUpdateBp(context: Context, bloodPressureMeasurementUuid: UUID): Intent {
+      return Intent(context, BloodPressureEntrySheet::class.java)
+          .putExtra(KEY_OPEN_AS, OpenAs.UPDATE_BP)
+          .putExtra(KEY_UUID, bloodPressureMeasurementUuid)
```
saket: OMG the same key is being used for two different types of values. Let's not do this? 😱

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -250,11 +250,11 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
   }
 
   fun showBloodPressureEntrySheet(patientUuid: UUID) {
-    activity.startActivityForResult(BloodPressureEntrySheet.intent(context, patientUuid), REQCODE_BP_SAVED)
+    activity.startActivityForResult(BloodPressureEntrySheet.intentForNewBp(context, patientUuid), REQCODE_BP_SAVED)
   }
 
   fun showBloodPressureUpdateSheet(bloodPressureMeasurementUuid: UUID) {
-
+    activity.startActivityForResult(BloodPressureEntrySheet.intentForUpdateBp(context, bloodPressureMeasurementUuid), REQCODE_BP_SAVED)
```
saket: Can I ask you to break this statement down? A lot is happening here? Same for `showBloodPressureEntrySheet()`

Something along the lines of,

```kotlin
val intent = ...
activity.startActivityForResult(...)
```

app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
```diff
@@ -96,4 +96,6 @@ class BloodPressureRepository @Inject constructor(
         .newest100ForPatient(patientUuid)
         .toObservable()
   }
+
+  fun findOne(uuid: UUID): Single<BloodPressureMeasurement> = Single.fromCallable { dao.getOne(uuid) }
```
saket: We generally avoid using prefixes like `get` or `find` for these functions. For instance, `measurementsWithSyncStatus()` instead of `findMeasurementsWithSyncStatus()`.

Let's use `measurement(UUID)`?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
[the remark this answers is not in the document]
saket: Using a single enum would also been alright IMO, just like `PatientSummaryScreenCreated`. Using separate events will save you from one `filter()` operator, but require you to create a new event.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
saket: Is `cache()` required here anymore?

saket: Oh I see it's changed in a separate commit

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -89,7 +91,12 @@ class BloodPressureEntrySheet : BottomSheetActivity() {
   private fun sheetCreates(): Observable<UiEvent> {
     val patientUuid = intent.extras!!.getSerializable(KEY_UUID) as UUID
     val openAs = intent.extras!!.getSerializable(KEY_OPEN_AS) as OpenAs
-    return Observable.just(BloodPressureEntrySheetCreated(openAs, patientUuid))
+    return Observable
+        .just(BloodPressureEntrySheetCreated(openAs, patientUuid))
+        // This delay stops the race condition (?) that happens frequently with replay().refCount()
+        // in the controller. Temporary workaround until we figure out what exactly is going on.
+        .delay(100L, TimeUnit.MILLISECONDS)
+        .cast()
```
saket: The commit name looks wrong…? `Prefill BP when opening BP entry sheet for updates`

saket: TIL about `cast()`. Useful!

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -89,7 +91,12 @@ class BloodPressureEntrySheet : BottomSheetActivity() {
   private fun sheetCreates(): Observable<UiEvent> {
     val patientUuid = intent.extras!!.getSerializable(KEY_UUID) as UUID
     val openAs = intent.extras!!.getSerializable(KEY_OPEN_AS) as OpenAs
-    return Observable.just(BloodPressureEntrySheetCreated(openAs, patientUuid))
+    return Observable
+        .just(BloodPressureEntrySheetCreated(openAs, patientUuid))
+        // This delay stops the race condition (?) that happens frequently with replay().refCount()
+        // in the controller. Temporary workaround until we figure out what exactly is going on.
+        .delay(100L, TimeUnit.MILLISECONDS)
```
saket: Time based workarounds are not very good workarounds :p Did you not consider manually adding an `observeOn(io())` before your database call?

app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
```diff
@@ -54,6 +56,19 @@ class BloodPressureRepository @Inject constructor(
...
+        .fromCallable {
+          bloodPressureMeasurement.copy(
+              updatedAt = Instant.now(clock),
+              syncStatus = SyncStatus.PENDING
+          )
+        }
+        .flatMapCompletable { updatedBloodPressureMeasurement ->
+          Completable.fromAction { dao.save(listOf(updatedBloodPressureMeasurement)) }
+        }
+  }
```
saket: I think `Single.fromCallable` isn't required here and this can be made simpler:

```kotlin
  val updatedMeasurement = bloodPressureMeasurement.copy(
      updatedAt = Instant.now(clock),
      syncStatus = SyncStatus.PENDING)
  dao.save(listOf(updatedMeasurement))
}
```

What do you think?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
saket: Vinay, Sanchita and I were talking that `OpenAs` can be converted into a sealed class to offer type safety in the controller.

```kotlin
sealed class OpenAs {
  data class New(val patientUuid: UUID) : OpenAs()
  data class Update(val bpUuid: UUID) : OpenAs()
}
```

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
saket: and maybe also `showScheduleAppointmentSheet()`?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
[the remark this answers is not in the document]
vinaysshenoy: Wow, I completely forgot they were `Comparable`. Fixing now.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
vinaysshenoy: Yeah, I didn't like it either.

Can I use `@Parcelize` on a sealed class? That would solve a lot of problems.

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
vinaysshenoy: I'm going to take a page out of your book and use local functions to clean this up! 🥇

vinaysshenoy: Seems to be a duplicate comment. Closing this.

vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -68,20 +69,22 @@ class BloodPressureEntrySheetController @Inject constructor(
...
         .flatMapSingle { bloodPressureRepository.findOne(it.uuid) }
-        .map { bloodPressure -> { ui: Ui -> ui.updateBpMeasurements(bloodPressure.systolic, bloodPressure.diastolic)} }
+        .map { bloodPressure -> { ui: Ui -> ui.updateBpMeasurements(bloodPressure.systolic, bloodPressure.diastolic) } }
   }
 
   private fun bpValidationsAndSaves(events: Observable<UiEvent>): Observable<UiChange> {
     val imeDoneClicks = events.ofType<BloodPressureSaveClicked>()
 
-    val patientUuidFromUpdateBp = events
+    val bloodPressureFromUpdateBp = events
```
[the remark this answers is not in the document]
vinaysshenoy: The method was broken up into separate methods for handling validations, saves and updates.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
vinaysshenoy: Whoops. Will fix this.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
vinaysshenoy: Right! I'll do that instead.

vinaysshenoy: Tried this out and it doesn't seem to work. What I did notice though, was that the db query method was being invoked twice, once on the IO thread and once more on the main thread.

Leaving this workaround in for now until we figure out why the other problem actually happens.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
vinaysshenoy: Squashed this commit with the actual prefill BP fields commit since this was a workaround to fix a bug in that commit.

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
```diff
@@ -80,7 +72,7 @@ class PatientSummaryScreenControllerTest {
         medicalHistoryRepository,
         timestampGenerator,
         clock,
-        config)
+        configEmitter.firstOrError())
```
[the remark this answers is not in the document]
vinaysshenoy: Good point. I'll rename this to `configSubject`.

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
```diff
@@ -114,6 +106,9 @@ class PatientSummaryScreenControllerTest {
 
   @Test
   fun `patient's prescription summary should be populated`() {
+    val config = PatientSummaryConfig(numberOfBpPlaceholders = 0, bpEditableFor = Duration.ofSeconds(30L))
+    configEmitter.onNext(config)
```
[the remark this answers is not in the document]
vinaysshenoy: Ah. I personally find the subject method easier to reason about.

app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
vinaysshenoy: Oh yeah, it isn't necessary. It's just a habit where I want to make sure each operator in an Rx chain does only one thing. I don't mind changing it to this format because the chain is small.

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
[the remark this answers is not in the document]
saket: Hmmmm I see

app/src/main/java/org/simple/clinic/summary/PatientSummaryModule.kt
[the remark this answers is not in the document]
saket: @sanchita I don't think that suffixing a variable with its type adds more meaning to the variable. It's useful in some cases to describe what type of values does a variable holds. that way, we would start naming variables like `systolicInt`, `patientNameString`, and so on. :)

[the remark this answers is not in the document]
saket: @sanchita I don't think that suffixing a variable with its type adds more meaning to the variable. It's useful in some cases to describe what type of values does a variable holds. Otherwise, we will start naming variables like `systolicInt`, `patientNameString`, and so on. :)

app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
saket: I understand and I think that works very nicely for single line lambdas.

PR #331, opened 2018-10-25
Mask nurse's phone number offline with a dummy Twilio number

saket:
This feature is behind a flag and is currently disabled because we don't have toll-free numbers yet. In the future, the server will send the nurse's assigned toll-free number during authentication.


saket:
@vinaysshenoy this feature is behind a feature flag FYI. You can enable it in `PhoneModule`.

app/src/main/java/org/simple/clinic/phone/Caller.kt
```diff
@@ -0,0 +1,24 @@
...
+import android.content.Intent
+import android.net.Uri
+
+interface Caller {
+
+  fun call(context: Context, phoneNumber: String)
+
+  object UsingDialer : Caller {
+    override fun call(context: Context, phoneNumber: String) {
+      val intent = Intent(Intent.ACTION_CALL, Uri.fromParts("tel", phoneNumber, null))
```
vinaysshenoy: Isn't this supposed to be `ACTION_DIAL`?

app/src/main/java/org/simple/clinic/phone/TwilioPhoneNumberMasker.kt
```diff
@@ -0,0 +1,40 @@
...
+    return configProvider
+        .map { config -> maskNumber(config, numberToMask) }
+        .flatMapCompletable { maskedNumber -> callNumber(maskedNumber, caller) }
+  }
+
+  private fun maskNumber(config: PhoneNumberMaskerConfig, numberToMask: String): String {
+    return if (config.maskingEnabled) {
+      // TODO: This number is temporary. Every nurse will
+      // be assigned a number during login/registration.
+      val twilioNumber = "+1 704 452 4471"
```
vinaysshenoy: Replace with a fake number please.

saket: Whoops, keep forgetting that our source is open. Done.

app/src/main/java/org/simple/clinic/phone/Caller.kt
saket: mCopyPasta

app/src/main/java/org/simple/clinic/phone/PhoneModule.kt
```diff
@@ -0,0 +1,15 @@
...
+import dagger.Provides
+import io.reactivex.Single
+
+@Module
+class PhoneModule {
+
+  @Provides
+  fun config() = Single.just(PhoneNumberMaskerConfig(maskingEnabled = false))!!
+
+  @Provides
```
vinaysshenoy: `@Binds`, but not `@Binds`. 😆

app/src/main/java/org/simple/clinic/phone/PhoneNumberMasker.kt
```diff
@@ -0,0 +1,7 @@
+package org.simple.clinic.phone
+
+import io.reactivex.Completable
+
+interface PhoneNumberMasker {
```
vinaysshenoy: The naming of this interface indicates that it only masks a number, but it also accepts a `Caller` and it is not clear from the interface name why it would do this.

I think we can improve this by changing the interface to just return a phone number as a string, and have two implementations: a `NonOperationalPhoneNumberMasker` and a `TwilioPhoneNumberMasker`. The `NonOperationalPhoneNumberMasker` can then be injected into the controller and we avoid the need for creating a feature flag at all.

What do you think?

app/src/main/java/org/simple/clinic/phone/PhoneModule.kt
saket: 😄

app/src/main/java/org/simple/clinic/phone/PhoneNumberMasker.kt
saket: Hrm, in that case we won't be able to have an abstract function in the interface. What do you think about `MaskedPhoneCaller`?

vinaysshenoy: Sounds good to me.

app/src/test/java/org/simple/clinic/phone/TwilioMaskedPhoneCallerTest.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    val callerCaptor = CallerCaptor()
+    val plainNumber = "123"
+
+    maskedPhoneCaller.maskAndCall(plainNumber, caller = callerCaptor).blockingAwait()
+
+    assertThat(callerCaptor.calledNumber).isEqualTo(plainNumber)
+  }
+
+  class CallerCaptor : Caller {
```
vinaysshenoy: I'm going to rename this to `MockCaller`.

PR #333, opened 2018-10-29
Remove facility ID as foreign keys in BloodPressureMeasurement and PrescribedDrug

saket:


app/src/androidTest/java/org/simple/clinic/MigrationTestHelperWithForeignConstraints.kt
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  @Deprecated(message = "Use migrateTo() instead", replaceWith = ReplaceWith("helper.migrateTo(version, *migrations)"))
+  override fun runMigrationsAndValidate(
+      name: String?,
+      version: Int,
+      validateDroppedTables: Boolean,
+      vararg migrations: Migration?
+  ): SupportSQLiteDatabase {
+    return super.runMigrationsAndValidate(name, version, validateDroppedTables, *migrations)
```
vinaysshenoy: Instead of deprecating this, we should just throw an exception with a message to use `migrateTo` instead.

app/src/androidTest/java/org/simple/clinic/MigrationAndroidTest.kt
```diff
@@ -1,29 +1,25 @@
 package org.simple.clinic
```
vinaysshenoy: Can you change usages of backticks in SQL statements to double quotes please?

app/src/androidTest/java/org/simple/clinic/MigrationAndroidTest.kt
```diff
@@ -391,5 +407,111 @@ class MigrationAndroidTest {
...
+    }
+
+    db_v18.query("SELECT * FROM `PrescribedDrug`").use {
+      it.moveToNext()
+      assertThat(it.count).isEqualTo(1)
+      assertThat(it.columnCount).isEqualTo(11)
+      assertThat(it.string("uuid")).isEqualTo(prescribedDrugUuid1)
+    }
+
+    // Inserting records for a non-existent facility shouldn't fail.
```
vinaysshenoy: Should we verify that attempting this before running the migration fails?

app/src/main/java/org/simple/clinic/bp/BloodPressureMeasurement.kt
```diff
@@ -2,32 +2,18 @@ package org.simple.clinic.bp
...
-          parentColumns = ["uuid"],
-          childColumns = ["facilityUuid"],
-          onDelete = ForeignKey.CASCADE,
-          onUpdate = ForeignKey.CASCADE)
-    ],
-    indices = [
-      Index("patientUuid", unique = false),
-      Index("facilityUuid", unique = false)
-    ])
+@Entity(indices = [Index("patientUuid", unique = false)])
```
vinaysshenoy: We might want to consider keeping the index on `facilityUuid` while dropping the foreign key, depending on how frequently we query the table based on this field.

app/src/main/java/org/simple/clinic/drugs/PrescribedDrug.kt
```diff
@@ -19,19 +17,7 @@ import java.util.UUID
...
-          parentColumns = ["uuid"],
-          childColumns = ["facilityUuid"],
-          onDelete = ForeignKey.CASCADE,
-          onUpdate = ForeignKey.CASCADE)
-    ],
-    indices = [
-      Index("patientUuid", unique = false),
-      Index("facilityUuid", unique = false)
-    ])
+@Entity(indices = [Index("patientUuid", unique = false)])
```
vinaysshenoy: We might want to consider keeping the index on `facilityUuid` while dropping the foreign key, depending on how frequently we query the table based on this field.

app/src/main/java/org/simple/clinic/storage/Migration_17_18.kt
```diff
@@ -0,0 +1,72 @@
...
+class Migration_17_18 : Migration(17, 18) {
+
+  override fun migrate(database: SupportSQLiteDatabase) {
+    migrateBloodPressureMeasurement(database)
+    migratePrescribedDrug(database)
+  }
+
+  private fun migrateBloodPressureMeasurement(database: SupportSQLiteDatabase) {
+    // Sqlite doesn't support dropping constraints, so we recreate the table.
+    database.execSQL("ALTER TABLE `BloodPressureMeasurement` RENAME TO `BloodPressureMeasurement_v17`")
```
vinaysshenoy: Can you please use double quotes instead of backticks in SQL statements?

app/src/androidTest/java/org/simple/clinic/MigrationTestHelperWithForeignConstraints.kt
saket: Using this function isn't an error, why would you throw an exception?

app/src/androidTest/java/org/simple/clinic/MigrationAndroidTest.kt
```diff
@@ -391,5 +407,111 @@ class MigrationAndroidTest {
...
+    db_v17.execSQL("""
```
vinaysshenoy: - This should be `db_v18`.
- Also, why isn't this breaking since it's `db_v17` which has foreign key constraints enabled 😮

app/src/androidTest/java/org/simple/clinic/MigrationAndroidTest.kt
saket: Since this class has usages of both backticks and double quotes, with backticks being the majority, let's keep this for now until we decide on one standard. I'll add this to my discuss-once-pratul-is-back TODOs.

app/src/main/java/org/simple/clinic/bp/BloodPressureMeasurement.kt
saket: We don't query BPs based on facilities right now, so should be fine.

app/src/main/java/org/simple/clinic/drugs/PrescribedDrug.kt
saket: Same: We don't query prescriptions based on facilities right now, so should be fine.

app/src/main/java/org/simple/clinic/storage/Migration_17_18.kt
saket: Same

> Since this class has usages of both backticks and double quotes, with backticks being the majority, let's keep this for now until we decide on one standard. I'll add this to my discuss-once-pratul-is-back TODOs.

app/src/androidTest/java/org/simple/clinic/MigrationAndroidTest.kt
saket: @vinaysshenoy changing them to double quotes

app/src/main/java/org/simple/clinic/storage/Migration_17_18.kt
saket: doing this

app/src/androidTest/java/org/simple/clinic/MigrationAndroidTest.kt
saket: Done

app/src/androidTest/java/org/simple/clinic/MigrationTestHelperWithForeignConstraints.kt
saket: I don't think we should throw an exception here. Using this function is not an error. The only reason I deprecated this was because when you start typing `helper.migrate`, the person shouldn't see two migrate functions and get confused between them.

app/src/androidTest/java/org/simple/clinic/MigrationAndroidTest.kt
```diff
@@ -391,5 +413,99 @@ class MigrationAndroidTest {
...
+    val measurementUuid1 = "ee367a66-f47e-42d8-965b-7a2b5c54f4bd"
+    val prescribedDrugUuid1 = "814132f2-f440-40cb-991c-43861c295815"
+
+    val measurementUuid2 = "fe367a66-f47e-42d8-965b-7a2b5c54f4bd"
+    val prescribedDrugUuid2 = "124132f2-f440-40cb-991c-43861c295815"
+
+    val userUuid = "a1d33096-cea6-4beb-8441-82cab2befe2d"
+    val patientUuid = "464bcda8-b26a-484d-bb70-59b3675f4a38"
+
+    fun insertFacilityStatement(uuid: String, name: String) = """
```
vinaysshenoy: For the first time in my life, I see why local functions can be useful sometimes.

app/src/androidTest/java/org/simple/clinic/MigrationAndroidTest.kt
```diff
@@ -391,5 +413,99 @@ class MigrationAndroidTest {
...
+    db_v17.execSQL(insertFacilityStatement(uuid = existentFacilityUuid, name = "Facility name"))
+    db_v17.execSQL(insertBloodPressureStatement(uuid = measurementUuid1, facilityUuid = existentFacilityUuid))
+    db_v17.execSQL(insertPrescribedDrugStatement(uuid = prescribedDrugUuid1, facilityUuid = existentFacilityUuid))
+
+    try {
+      db_v17.execSQL(insertBloodPressureStatement(uuid = measurementUuid2, facilityUuid = nonExistentFacilityUuid))
+      db_v17.execSQL(insertPrescribedDrugStatement(uuid = prescribedDrugUuid2, facilityUuid = nonExistentFacilityUuid))
+
+    } catch (e: Throwable) {
+      assertThat(e).isInstanceOf(SQLiteConstraintException::class.java)
```
vinaysshenoy: There is still a problem here where the test will still pass if the error does not get thrown in the first place. I'll fix this.

PR #334, opened 2018-10-29
Add five new protocol drugs in the Prescription screen

sanchita:


vinaysshenoy:
Noticed a bug where the "Update Medicines" screen automatically scrolls to the bottom as soon as the screen is opened. I will fix this.

PR #336, opened 2018-10-31
 Extract PinEntryCardView from lockscreen as a custom View

saket:
Because brute-force protection is needed in both lockscreen and login screen, I'm slowly extracting the PIN text field (and related UI elements) from both the screens into a custom View.

### DIFF
--- a/app/src/main/java/org/simple/clinic/login/applock/AppLockScreen.kt
+++ b/app/src/main/java/org/simple/clinic/login/applock/AppLockScreen.kt
@@ -1,14 +1,8 @@
 package org.simple.clinic.login.applock
 
 import android.content.Context
-import android.support.transition.Fade
-import android.support.transition.TransitionManager
-import android.support.v4.view.animation.FastOutSlowInInterpolator
 import android.util.AttributeSet
-import android.view.View
 import android.widget.Button
-import android.widget.LinearLayout
-import android.widget.ProgressBar
 import android.widget.RelativeLayout
 import android.widget.TextView
 import android.widget.Toast
@@ -20,11 +14,10 @@ import kotterknife.bindView
 import org.simple.clinic.R
 import org.simple.clinic.activity.TheActivity
 import org.simple.clinic.facility.change.FacilityChangeScreen
+import org.simple.clinic.pin.PinEntryCardView
 import org.simple.clinic.router.screen.BackPressInterceptCallback
 import org.simple.clinic.router.screen.BackPressInterceptor
 import org.simple.clinic.router.screen.ScreenRouter
-import org.simple.clinic.widgets.StaggeredEditText
-import org.simple.clinic.widgets.hideKeyboard
 import org.simple.clinic.widgets.showKeyboard
 import javax.inject.Inject
 
@@ -46,11 +39,7 @@ class AppLockScreen(context: Context, attrs: AttributeSet) : RelativeLayout(cont
   private val facilityButton by bindView<Button>(R.id.applock_facility_name)
   private val fullNameTextView by bindView<TextView>(R.id.applock_user_fullname)
   private val logoutButton by bindView<Button>(R.id.applock_logout)
-  private val pinEditText by bindView<StaggeredEditText>(R.id.applock_pin)
-  private val pinFormLayout by bindView<LinearLayout>(R.id.applock_pin_container)
-  private val progressView by bindView<ProgressBar>(R.id.applock_progress)
-  private val errorTextView by bindView<TextView>(R.id.applock_error)
-  private val forgotPinButton by bindView<Button>(R.id.applock_forgotpin)
+  private val pinEntryCardView by bindView<PinEntryCardView>(R.id.applock_pin_entry_card)
 
   override fun onFinishInflate() {
     super.onFinishInflate()
@@ -59,7 +48,7 @@ class AppLockScreen(context: Context, attrs: AttributeSet) : RelativeLayout(cont
     }
     TheActivity.component.inject(this)
 
-    Observable.mergeArray(screenCreates(), pinTextChanges(), submitClicks(), backClicks(), facilityClicks(), forgotPinClicks())
+    Observable.mergeArray(screenCreates(), backClicks(), facilityClicks(), forgotPinClicks(), pinAuthentications())
         .observeOn(Schedulers.io())
         .compose(controller)
         .observeOn(AndroidSchedulers.mainThread())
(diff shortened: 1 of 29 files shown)


app/src/main/java/org/simple/clinic/pin/PinEntryCardView.kt
```diff
@@ -0,0 +1,117 @@
...
+
+  fun showIncorrectPinError() {
+    showError(R.string.pinentry_error_incorrect_pin)
+  }
+
+  fun clearPin() {
+    pinEditText.text = null
+  }
+
+  fun dispatchAuthenticatedCallback() {
```
sanchita: This name doesn't sound right since there's no actual callback of any sort. What it seems to be doing is informing the view that the authentication is successful. Do you feel a name like `onAuthenticated` would be better?

app/src/main/java/org/simple/clinic/pin/PinEntryCardView.kt
```diff
@@ -0,0 +1,117 @@
...
+
+  @SuppressLint("CheckResult")
+  override fun onFinishInflate() {
+    super.onFinishInflate()
+    if (isInEditMode) {
+      return
+    }
+    TheActivity.component.inject(this)
+
+    Observable.mergeArray(pinTextChanges())
```
sanchita: We are not handling *IME done* clicks here. If the user submits *IME action done* when they have entered less than the PIN length, the view won't show any feedback.

app/src/main/java/org/simple/clinic/pin/PinEntryCardController.kt
```diff
@@ -0,0 +1,81 @@
...
+typealias Ui = PinEntryCardView
+typealias UiChange = (Ui) -> Unit
+
+class PinEntryCardController @Inject constructor(
+    private val userSession: UserSession,
+    private val passwordHasher: PasswordHasher
+) : ObservableTransformer<UiEvent, UiChange> {
+
+  override fun apply(events: Observable<UiEvent>): ObservableSource<UiChange> {
+    val replayedEvents = events.replay().refCount()
```
sanchita: We need to report analytics events too. Earlier, screens would report their events individually but this controller won't do that.

app/src/main/res/layout/pin_entry_card.xml
```diff
@@ -0,0 +1,41 @@
...
+  <ProgressBar
+    android:id="@+id/pinentry_progress"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:layout_gravity="center"
+    android:visibility="invisible" />
+
+  <LinearLayout
+    android:id="@+id/pinentry_pin_container"
+    style="@style/Clinic.CardContentContainer.AppLock">
```
sanchita: There are styles in this layout that are inconsistent with the usage: `LoginPinLabel` and `Clinic.CardContentContainer.AppLock` for example.

Should we rename them to be more appropriate?

app/src/main/res/layout/pin_entry_card.xml
```diff
@@ -0,0 +1,41 @@
...
+
+    <TextView
+      style="@style/Clinic.LoginPinLabel"
+      android:labelFor="@+id/pinentry_pin"
+      android:text="@string/loginpin_pin_number_hint" />
```
sanchita: Can we also rename the string resource `loginpin_pin_number_hint` to be more specific to PinCardView?

app/src/main/java/org/simple/clinic/pin/PinEntryCardView.kt
saket: This is intentional. We're not showing any error if you press IME done without entering the full PIN.

app/src/main/java/org/simple/clinic/login/applock/ComparisonResult.kt
```diff
@@ -0,0 +1,6 @@
+package org.simple.clinic.login.applock
```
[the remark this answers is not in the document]
saket: Good idea, will do this.

app/src/main/java/org/simple/clinic/login/applock/AppLockScreen.kt
```diff
@@ -72,28 +61,21 @@ class AppLockScreen(context: Context, attrs: AttributeSet) : RelativeLayout(cont
 
     // The keyboard shows up on PIN field automatically when the app is
     // starting, but not when the user comes back from FacilityChangeScreen.
-    pinEditText.showKeyboard()
+    pinEntryCardView.pinEditText.showKeyboard()
```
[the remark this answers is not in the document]
saket: Why is it unnecessarily? I think it's okay to abstract properties when their underlying implementation can change. It looks fine here. 🤔

For instance, `forgotPinButton` is also exposed as a public field so that its visibility can be set to hidden by the screen. This will be used in `LoginPinScreen`.

app/src/main/java/org/simple/clinic/pin/PinEntryCardView.kt
saket: There's an emission and not a callback, true, but `onAuthenticated` is even less descriptive than `dispatchAuthenticatedCallback()`. It's like saying authentication happened, now do something. I could… rename this to `emitPinAuthenticated()`. What do you think?

app/src/main/java/org/simple/clinic/pin/PinEntryCardController.kt
saket: Whoops, totally missed that. Will do this.

app/src/main/res/layout/pin_entry_card.xml
saket: Yes, thanks for catching this.

app/src/main/res/layout/pin_entry_card.xml
saket: Yes, thanks for catching this x2.

app/src/main/java/org/simple/clinic/login/applock/ComparisonResult.kt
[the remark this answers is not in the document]
saket: Brute-force protection related classes can also go there.

app/src/main/res/layout/pin_entry_card.xml
saket: Don't think I can change `LoginPinLabel` or `LoginPinErrorText`. They're being used in other places too.

saket: `Clinic.CardContentContainer.AppLock` isn't shared with any other screen so I'm removing it.

saket: Update: False alarm. `Clinic.CardContentContainer` is still being used so I'm renaming `Clinic.CardContentContainer.AppLock` to `Clinic.CardContentContainer.PinEntry`

app/src/main/res/layout/pin_entry_card.xml
saket: Adding `@string/pinentry_pin_number_hint`.

app/src/main/java/org/simple/clinic/pin/PinEntryCardView.kt
vinaysshenoy: This will break when you add this widget to the login pin screen then because screens have behaviour on submit clicked.

saket: Oh yea, I'll also change that because the new design is to auto-submit PINs everywhere.

app/src/test/java/org/simple/clinic/login/applock/BCryptPasswordHasherTest.kt
```diff
@@ -1,6 +1,8 @@
 package org.simple.clinic.login.applock
 
 import org.junit.Test
+import org.simple.clinic.security.BCryptPasswordHasher
+import org.simple.clinic.security.ComparisonResult
 
 class BCryptPasswordHasherTest {
```
vinaysshenoy: Shouldn't this test also get moved to the `security` package?

app/src/main/java/org/simple/clinic/pin/PinEntryCardView.kt
vinaysshenoy: Let's leave it for now and revisit this when we add `PinCardView` to another screen?

app/src/main/java/org/simple/clinic/login/applock/AppLockScreen.kt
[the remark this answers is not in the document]
vinaysshenoy: Hmm. Let's wait until this is used in multiple places before we decide.

app/src/main/java/org/simple/clinic/pin/PinEntryCardView.kt
saket: Sounds good 👍

app/src/test/java/org/simple/clinic/login/applock/BCryptPasswordHasherTest.kt
saket: Whoops, missed this. Fixing this.

PR #344, opened 2018-11-07
Edit Patient Screen initial implementation

vinaysshenoy:
## Changes
1. Screen for edit patient added, with the fields that we are editing for now.
2. Pre-fill the fields when opening the screen.
3. Put the edit patient feature behind a feature flag that's enabled only in the debug build


vinaysshenoy:
@saket This seems like a large PR but a lot of it is a new XML layout and the glue code for the screen and controller.

app/src/main/res/values/styles_v2.xml
```diff
@@ -154,6 +154,11 @@
     <item name="android:paddingBottom">18dp</item>
   </style>
 
+  <style name="Clinic.V2.RoundedClearButton">
```
saket: `RoundedClearButton` -> `RoundedClearEditTextButton`? `RoundedClearFieldButton`? `RoundedClearInputButton`?

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreen.kt
```diff
@@ -0,0 +1,20 @@
+package org.simple.clinic.editpatient
+
+import android.content.Context
+import android.util.AttributeSet
+import android.widget.RelativeLayout
+
+class PatientEditScreen(context: Context, attributeSet: AttributeSet) : RelativeLayout(context, attributeSet) {
+
+  companion object {
+    @JvmField
```
saket: `@JvmField`? We aren't using Java anymore? :p

app/src/main/res/layout/screen_patient_edit.xml
```diff
@@ -0,0 +1,229 @@
+<?xml version="1.0" encoding="utf-8"?>
+<org.simple.clinic.editpatient.PatientEditScreen android:id="@+id/patientedit_root"
+  xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:app="http://schemas.android.com/apk/res-auto"
+  xmlns:tools="http://schemas.android.com/tools"
+  android:layout_width="match_parent"
+  android:layout_height="match_parent"
+  android:orientation="horizontal">
+
```
saket: Are we going to reuse the layout for creating and editing a patient? How do you suggest we reuse things? Do we use styles? Or do we use a single layout?

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
```diff
@@ -0,0 +1,70 @@
...
+) : ObservableTransformer<UiEvent, UiChange> {
+
+  override fun apply(events: Observable<UiEvent>): ObservableSource<UiChange> {
+    val replayedEvents = events.compose(ReportAnalyticsEvents()).replay().refCount()
+
+    return prefillOnStart(replayedEvents)
+  }
+
+  private fun prefillOnStart(events: Observable<UiEvent>): Observable<UiChange> {
+
```
saket: Empty line here

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
```diff
@@ -0,0 +1,70 @@
...
+    val patientUuidStream = events.ofType<PatientEditScreenCreated>()
+        .map { it.patientUuid }
+
+    val savedPatient = patientUuidStream
+        .flatMap(patientRepository::patient)
+        .take(1)
+        .map { it as Just }
+        .map { it.value }
```
saket: Unwrapping Optionals in Rx chains looks so bad. One simpler way is to deconstruct:

```kotlin
.map { (patient) -> patient!! }
```

Can we create an extension function to simplify this somehow?

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
```diff
@@ -0,0 +1,70 @@
...
+        .cache()
+
+    val savedAddress = savedPatient
+        .flatMap { patient ->
+          patientRepository
+              .address(patient.addressUuid)
+              .take(1)
+              .map { it as Just }
+              .map { it.value }
```
saket: Deconstruction can also be applied here: 

```kotlin
.map { (address) -> address!! }
```

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
saket: I propose:

```kotlin
private fun <T : Any> Observable<Optional<T>>.filterJust(): Observable<T> {
  return map { it as Just<T> }.map { (value) -> value }
}

private fun <T : Any> Observable<Optional<T>>.mapAsJust(): Observable<T> {
  return filter { it is Just }.map { (value) -> value!! }
}
```

Not very satisfied with the naming though.

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    return Observables.combineLatest(savedPatient, savedAddress, savedPhoneNumber)
+        .map { (patient, address, phoneNumber) ->
+          { ui: Ui ->
+            ui.setPatientName(patient.fullName)
+            ui.setGender(patient.gender)
+            ui.setState(address.state)
+            ui.setDistrict(address.district)
+
+            if(address.colonyOrVillage.isNullOrBlank().not()) {
```
saket: Missing space after `if` here

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
```diff
@@ -0,0 +1,70 @@
...
+              ui.setColonyOrVillage(address.colonyOrVillage!!)
+            }
+
+            if(phoneNumber is Just) {
+              ui.setPatientPhoneNumber(phoneNumber.value.number)
```
saket: Small suggestion: I think you can avoid this `if` check and using `phoneNumber.value` by keeping the three streams for pre-filling profile, address and phone separate instead of combining them. That way, you can leverage filter().

```kotlin
val preFillPhone = savedPatient
    .flatMap { patientRepository.phoneNumbers(it.uuid) }
    .take(1)
    .filterJust()
    .map { { ui: Ui -> ui.setPatientPhoneNumber(it.number) } }

return Observable.merge(preFillProfile, preFillAddress, preFillPhone)
```

app/src/test/java/org/simple/clinic/editpatient/PatientEditScreenControllerTest.kt
```diff
@@ -0,0 +1,100 @@
...
+    if (shouldSetPhoneNumber) {
+      verify(screen).setPatientPhoneNumber(phoneNumber!!.number)
+    } else {
+      verify(screen, never()).setPatientPhoneNumber(any())
+    }
+  }
+
+  @Suppress("Unused")
+  private fun `params for prefilling fields on screen created`(): List<List<Any?>> {
+
```
saket: leftover linebreak here

app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
```diff
@@ -89,6 +91,12 @@ class DebugClinicApp : ClinicApp() {
         .crashReporterModule(object : CrashReporterModule() {
           override fun crashReporter() = NoOpCrashReporter()
         })
+        .patientSummaryModule(object : PatientSummaryModule() {
+          override fun providesSummaryConfig(): Single<PatientSummaryConfig> {
+            return super.providesSummaryConfig()
+                .map { it.copy(isPatientEditFeatureEnabled = true) }
```
saket: Oh nice! Copying super values is much better than creating the values again that I've been doing.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -136,6 +142,13 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
         .subscribe { uiChange -> uiChange(this) }
   }
 
+  private fun enableEditPatientFeature() {
```
saket: 🙀 Let's do this from the controller and avoid blocking Rx calls?

vinaysshenoy: The reason I didn't put this in the controller was because this is a throwaway flag that will be removed as soon as the edit patient feature is done. I don't have a problem moving it, however.

app/src/main/res/values/styles_v2.xml
vinaysshenoy: Ah, the name is unclear. by clear, I mean the background colour is translucent. The naming is wrong, I'll improve it.

FYI, this is the background for the buttons like "Not You?" on the app lock screen and the "Edit" button on the patient summary screen.

saket: Clearly 😄

saket: CLEARly 😄

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
vinaysshenoy: I made a class `RxOptional.kt` since this is a pattern we seem to using in a lot of places. I've added one method for now which is basically

```kotlin
fun <T : Any> Observable<Optional<T>>.unwrapIfJust(): Observable<T> {
  return filter { it is Just }.map { (value) -> value!! }
}
```

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreen.kt
saket: > getters and setters.

Ohh. TIL.

app/src/main/res/layout/screen_patient_edit.xml
saket: Sure, sounds good

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
saket: This looks much better. I noticed you're using `cache()` for sharing `savedPatient` stream. I don't have enough experience, but I think `Observable#cache()` is dangerous. From what I understand from the docs, there's some caveats:

```
<em>Note:</em> You sacrifice the ability to dispose the origin when you use the {@code cache} Observer so be careful not to use this Observer on ObservableSources that emit an infinite or very large number of items that will use up memory. A possible workaround is to apply `takeUntil` with a predicate or another source before (and perhaps after) the application of cache().
```

Let's use `replay().refCount()` instead?

As a rule of thumb, I use `cache()` with `Single` and `replay().refCount()` with `Observable`.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
saket: I see you've fixed this.

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
vinaysshenoy: As decided offline, I'll make this change as part of my next PR.

PR #353, opened 2018-11-10
Fix: Crash on 1000+ results in patient search

vinaysshenoy:
SQLite has a hard limit of 999 for the number of parameters in a parameterized query. Since it is unlikely the nurse will scroll through 999 patients, this fixes the problem by limiting the number of results from fuzzy search to 100.

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -22,6 +22,7 @@ import org.simple.clinic.user.UserSession
 import org.simple.clinic.util.TestClock
 import org.threeten.bp.Clock
 import org.threeten.bp.LocalDate
+import java.util.UUID
 import javax.inject.Inject
 
 class PatientRepositoryAndroidTest {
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -455,6 +456,39 @@ class PatientRepositoryAndroidTest {
...
 
+  /*
+  * Added to test the case where SQLite's max query param length (999) can be
+  * exceeded during fuzzy name search.
+  **/
+  @Test
+  fun when_searching_with_fuzzy_search_the_results_must_be_limited() {
+    val template = testData.patientProfile(syncStatus = SyncStatus.DONE)
+
+    val patientsToSave = (1 .. 1000).map {
```
saket: `1000` could potentially be replaced with `sqliteParamLimit + 1`, where `sqliteParamLimit = 999` just for readability instead of depending on the function documentation.

vinaysshenoy: Good point. I'll fix this in my next PR.

saket: 🙂👍

PR #354, opened 2018-11-12
Validate input when editing patient data

vinaysshenoy:


vinaysshenoy:
This is a large PR, but most of it is boilerplate for setting up events and test parameters.

app/src/main/java/org/simple/clinic/editpatient/PatientData.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.simple.clinic.registration.phone.PhoneNumberValidator.Type
+
+data class PatientData(
+    val name: String,
+    val gender: Gender,
+    val phoneNumber: String,
+    val colonyOrVillage: String,
+    val district: String,
+    val state: String
+) {
```
saket: This name is not very convincing with this class name. We're going to end up with `Patient`, `PatientProfile`, `OngoingPatientEntry` and `PatientData`. 

What do you think about splitting `OngoingPatientEntry` into `OngoingNewPatientEntry` and `OngoingEditPatientEntry` or something?

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenEvents.kt
```diff
@@ -1,6 +1,37 @@
...
 
+import org.simple.clinic.patient.Gender
 import org.simple.clinic.widgets.UiEvent
 import java.util.UUID
 
-data class PatientEditScreenCreated(val patientUuid: UUID): UiEvent
+data class PatientEditScreenCreated(val patientUuid: UUID) : UiEvent
+
+data class PatientEditPatientNameTextChanged(val name: String) : UiEvent {
+  override val analyticsName = "Patient Edit:Patient Name Text Changed"
```
saket: Can we use analytics keys similar to those in `PatientEntryScreen`? We're using `Create Patient Entry` there.

app/src/main/java/org/simple/clinic/editpatient/PatientEditValidationError.kt
```diff
@@ -0,0 +1,11 @@
+package org.simple.clinic.editpatient
+
+enum class PatientEditValidationError(val analyticsName: String) {
+  FULL_NAME_EMPTY("Patient Edit:Name is empty"),
```
saket: Same question about making these analytics name more consistent with `PatientEntryValidationError`

app/src/main/java/org/simple/clinic/editpatient/PatientData.kt
vinaysshenoy: I like this. I'll make this change.

app/src/main/java/org/simple/clinic/editpatient/PatientEditValidationError.kt
vinaysshenoy: `PatientEntryValidationError` currently uses `Patient Entry` as the prefix and `PatientEditValidationError` uses `Patient Edit`. Seems like it's more consistent right now.

saket: Hrm right. I think the problem is that `PatientEntryValidationError` uses a different analytics name than `PatientEntryScreenEvents`.

saket:
@vinaysshenoy thought for later: I realize we should rename `PatientEntryScreen` and related classes (like `PatientEntryValidationError`) to `NewPatientScreen`. But that's a large refactor.

PR #358, opened 2018-11-14
Hide validation errors when editing patient data

vinaysshenoy:


app/src/main/java/org/simple/clinic/editpatient/PatientEditScreen.kt
```diff
@@ -193,6 +196,34 @@ class PatientEditScreen(context: Context, attributeSet: AttributeSet) : Relative
...
+    errors.forEach {
+      when (it) {
+        FULL_NAME_EMPTY -> {
+          showEmptyFullNameError(false)
+        }
+
+        PHONE_NUMBER_EMPTY,
+        PHONE_NUMBER_LENGTH_TOO_SHORT,
+        PHONE_NUMBER_LENGTH_TOO_LONG -> {
+          showLengthTooShortPhoneNumberError(false)
```
vinaysshenoy: All these errors map to the same action, i.e, clearing the error on the same field.

vinaysshenoy: Do you feel it's better to add another method to just clear the error on the phone number field?

vinaysshenoy: I don't actually want to do this because then testing the validation errors as functions is hard. With the current setup, validating errors is extremely easy.

vinaysshenoy: I want to avoid behaviour testing in favour of value testing, that's why I set up the current methods the way they are right now.

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreen.kt
```diff
@@ -265,4 +268,22 @@ class PatientEditScreen(context: Context, attributeSet: AttributeSet) : Relative
...
+        stateInputLayout)
+
+
+    views
+        .firstOrNull {
+          it.error.isNullOrBlank().not()
+        }
+        ?.let { firstFieldWithError ->
+          formScrollView.scrollToChild(firstFieldWithError, onScrollComplete = { firstFieldWithError.requestFocus() })
+        }
```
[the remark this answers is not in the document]
vinaysshenoy: Look's good. I'll do this.

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreen.kt
saket: Sure okay

PR #361, opened 2018-11-15
Override IO scheduler to Trampoline in PatientsScreenControllerTest

vinaysshenoy:
We manually change the thread to IO for refreshing user status which was causing the test to be flaky.


app/src/test/java/org/simple/clinic/home/patients/PatientsScreenControllerTest.kt
```diff
@@ -43,6 +45,8 @@ class PatientsScreenControllerTest {
 
   @Before
   fun setUp() {
+    RxJavaPlugins.setIoSchedulerHandler { Schedulers.trampoline() }
```
saket: @vinaysshenoy can you add a comment here why this is being used?

vinaysshenoy: Done.

PR #363, opened 2018-11-15
 Order overdue appointments by keeping high risk patients at the top

saket:
FYI: this is behind a feature flag and is currently disabled.

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
@@ -58,7 +58,7 @@ class AppointmentRepositoryAndroidTest {
   @Test
   fun when_creating_new_appointment_then_the_appointment_should_be_saved() {
     val patientId = UUID.randomUUID()
-    val appointmentDate = LocalDate.now()
+    val appointmentDate = LocalDate.now(clock)
     repository.schedule(patientId, appointmentDate).blockingAwait()
 
     val savedAppointment = repository.recordsWithSyncStatus(SyncStatus.PENDING).blockingGet().first()
@@ -75,10 +75,10 @@ class AppointmentRepositoryAndroidTest {
   fun when_creating_new_appointment_then_all_old_appointments_for_that_patient_should_be_canceled() {
     val patientId = UUID.randomUUID()
 
-    val date1 = LocalDate.now()
+    val date1 = LocalDate.now(clock)
     repository.schedule(patientId, date1).blockingAwait()
 
-    val date2 = LocalDate.now().plusDays(10)
+    val date2 = LocalDate.now(clock).plusDays(10)
     repository.schedule(patientId, date2).blockingAwait()
 
     val savedAppointment = repository.recordsWithSyncStatus(SyncStatus.PENDING).blockingGet()
@@ -111,12 +111,12 @@ class AppointmentRepositoryAndroidTest {
             faker.address.city(),
             faker.address.state(),
             "India",
-            Instant.now(),
-            Instant.now()
+            Instant.now(clock),
+            Instant.now(clock)
         )
     )
     val patient1 = UUID.randomUUID()
-    val date1 = LocalDate.now().minusDays(100)
+    val date1 = LocalDate.now(clock).minusDays(100)
     val bp1 = UUID.randomUUID()
     database.patientDao().save(
         Patient(
@@ -128,8 +128,8 @@ class AppointmentRepositoryAndroidTest {
             LocalDate.parse("1947-08-15"),
             null,
             PatientStatus.ACTIVE,
-            Instant.now(),
-            Instant.now(),
+            Instant.now(clock),
+            Instant.now(clock),
             SyncStatus.DONE
         )
     )
@@ -142,8 +142,8 @@ class AppointmentRepositoryAndroidTest {
             userUuid = testData.qaUserUuid(),
             facilityUuid = testData.qaUserFacilityUuid(),
             patientUuid = patient1,
-            createdAt = Instant.now(),
-            updatedAt = Instant.now()
+            createdAt = Instant.now(clock),
+            updatedAt = Instant.now(clock)
         )
     ))
 
@@ -157,8 +157,8 @@ class AppointmentRepositoryAndroidTest {
             faker.address.city(),
             faker.address.state(),
             "India",
-            Instant.now(),
-            Instant.now()
+            Instant.now(clock),
+            Instant.now(clock)
         )
     )
     database.patientDao().save(
@@ -171,8 +171,8 @@ class AppointmentRepositoryAndroidTest {
             dateOfBirth = LocalDate.parse("1997-08-15"),
             age = null,
             status = PatientStatus.ACTIVE,
-            createdAt = Instant.now(),
-            updatedAt = Instant.now(),
+            createdAt = Instant.now(clock),
+            updatedAt = Instant.now(clock),
             syncStatus = SyncStatus.DONE
         )
     )
@@ -183,15 +183,15 @@ class AppointmentRepositoryAndroidTest {
             number = "983374583",
             phoneType = PatientPhoneNumberType.MOBILE,
             active = false,
-            createdAt = Instant.now(),
-            updatedAt = Instant.now()
+            createdAt = Instant.now(clock),
+            updatedAt = Instant.now(clock)
         ))
     )
 
     val patient3 = UUID.randomUUID()
     val address3 = UUID.randomUUID()
     val phoneNumber3 = UUID.randomUUID()
-    val date3 = LocalDate.now().minusDays(10)
+    val date3 = LocalDate.now(clock).minusDays(10)
     val bp30 = UUID.randomUUID()
     val bp31 = UUID.randomUUID()
     database.addressDao().save(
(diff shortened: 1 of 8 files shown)


app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -348,23 +372,126 @@ class AppointmentRepositoryAndroidTest {
...
+
+    fun savePatientAndAppointment(
+        fullName: String,
+        hasHadStroke: Boolean = false,
+        hasDiabetes: Boolean = false,
+        hasHadKidneyDisease: Boolean = false,
+        age: String = "30",
+        vararg bpMeasurements: BP
+    ) {
+      val patientUuid1 = patientRepository.saveOngoingEntry(testData.ongoingPatientEntry(fullName = fullName, age = age))
```
vinaysshenoy: This can be just called `patientUuid` right? I'm assuming this was something leftover from a refactoring.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
```diff
@@ -62,5 +66,53 @@ data class OverdueAppointment(
...
+          AND (A.remindOn < :dateNow OR A.remindOn IS NULL)
+
+          GROUP BY P.uuid HAVING max(BP.updatedAt)
+          ORDER BY isAtHighRisk DESC, A.scheduledDate, A.updatedAt ASC
+          """)
+    fun appointmentsForFacility(
+        facilityUuid: UUID,
+        scheduledStatus: Appointment.Status,
+        dateNow: LocalDate,
+        sixtyYearsDateOfBirth: LocalDate
```
vinaysshenoy: `sixtyYearsDateOfBirth` is something specific to a protocol for calculating whether someone is at high risk or not.

I feel that this age must be configurable and can be named to a more generic name like `bornBeforeDate` or something like that.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
```diff
@@ -62,5 +66,53 @@ data class OverdueAppointment(
           ORDER BY A.scheduledDate, A.updatedAt ASC
           """)
     fun appointmentsForFacility(facilityUuid: UUID, scheduledStatus: Appointment.Status, dateNow: LocalDate): Flowable<List<OverdueAppointment>>
+
+    /**
+     * FYI: IntelliJ's SQL parser highlights `isAtHighRisk` as an error, but it's not. This is probably
```
vinaysshenoy: Ugh, I hate this. Even though we know that whatever we are doing is fine, the query parser doesn't know this since it doesn't know we are using SQLite under the hood.

app/src/main/java/org/simple/clinic/overdue/AppointmentConfig.kt
```diff
@@ -0,0 +1,3 @@
+package org.simple.clinic.overdue
+
+data class AppointmentConfig(val highlightHighRiskPatients: Boolean)
```
vinaysshenoy: Do you feel we should move the age limit for calculation of high risk patients here as well?

app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
```diff
@@ -116,12 +117,24 @@ class AppointmentRepository @Inject constructor(
...
-              dateNow = LocalDate.now(UTC)
-          ).toObservable()
+        .withLatestFrom(configProvider.toObservable())
+        .flatMap { (facilityUuid, config) ->
+          val today = LocalDate.now(clock)
+          if (config.highlightHighRiskPatients) {
+            overdueDao.appointmentsForFacility(
+                facilityUuid = facilityUuid,
+                scheduledStatus = Appointment.Status.SCHEDULED,
+                dateNow = today,
```
vinaysshenoy: Instead of `dateNow`, do you feel a better name would be `appointmentScheduledBefore`?

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -348,23 +372,126 @@ class AppointmentRepositoryAndroidTest {
...
       assertThat(this.uuid).isEqualTo(uuid)
       assertThat(this.cancelReason).isNull()
       assertThat(this.status).isEqualTo(Appointment.Status.VISITED)
     }
   }
+
+  @Test
+  @Suppress("ASSIGNING_SINGLE_ELEMENT_TO_VARARG_IN_NAMED_FORM_FUNCTION")
+  fun high_risk_patients_should_be_present_at_the_top() {
+    data class BP(val systolic: Int, val diastolic: Int)
```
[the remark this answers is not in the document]
saket: Why do you want that? It's also only local to this function.

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
saket: Whoops, copy pasta! Fixing this.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
saket: Makes sense. I'll also add this to the config when I have more details about how we're categorizing high risk patients.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
saket: I don't think it has anything to do with Android. Room uses IntelliJ's parser so technically it shouldn't target any specific version of Sqlite.

app/src/main/java/org/simple/clinic/overdue/AppointmentConfig.kt
saket: Yes, from a comment I posted to another conversation:

> Makes sense. I'll also add this to the config when I have more details about how we're categorizing high risk patients.

app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
saket: Totally. Doing this.

saket: Actually, I'll name it just `scheduledBefore` because the function anyway returns appointments and there's another param called `scheduledStatus`.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
saket: Actually, doing `bornBeforeDate` -> `bornBefore`

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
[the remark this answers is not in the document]
vinaysshenoy: No specific reason, except that we have been avoiding using abbreviations for type names everywhere.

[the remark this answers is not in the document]
saket: Hrm we haven't intentionally avoided them 🤔. Since this is a test and a fairly large one, can we keep it to reduce verbosity of this test?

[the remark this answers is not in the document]
vinaysshenoy: Sure.

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -348,23 +372,127 @@ class AppointmentRepositoryAndroidTest {
...
     assertThat(updatedList).hasSize(1)
     updatedList[0].apply {
       assertThat(this.uuid).isEqualTo(uuid)
       assertThat(this.cancelReason).isNull()
       assertThat(this.status).isEqualTo(Appointment.Status.VISITED)
     }
   }
+
+  @Test
+  @Suppress("ASSIGNING_SINGLE_ELEMENT_TO_VARARG_IN_NAMED_FORM_FUNCTION")
```
vinaysshenoy: I just realized that this is a deprecation warning, and the ability to provide single arguments to named vararg parameters has been removed in Kotlin 1.3 (which is stable and released now). Upgrading our Kotlin version to 1.3 will break these tests.

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
vinaysshenoy: Has it been fixed? I still see `patientUuid1`.

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
saket: Oh no. The snytax for providing vararg parameters is not nice :( 

I'll update this.

saket: Fixed

vinaysshenoy: > Oh no. The snytax for providing vararg parameters is not nice :(
> 
> I'll update this.

Yeah. Maybe we should avoid varargs altogether and just use lists everywhere.

PR #365, opened 2018-11-16
Ignore reporting network errors to crash reporter

saket:


app/src/main/java/org/simple/clinic/sync/SyncWorker.kt
```diff
@@ -78,10 +78,11 @@ class SyncWorker(context: Context, workerParams: WorkerParameters) : Worker(cont
...
         ))
         .doOnError { e ->
-          if (e !is IOException) {
-            Timber.e(e)
-            crashReporter.report(e)
+          val resolvedError = ErrorResolver.resolve(e)
+          if (resolvedError.shouldBeReported) {
+            crashReporter.report(resolvedError.actualCause)
           }
+          Timber.e(e)
```
vinaysshenoy: Since we report Timber logs to Sentry, won't this continue to get reported even if the error was a network error?

app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
```diff
@@ -0,0 +1,43 @@
...
+import java.net.SocketTimeoutException
+import java.net.UnknownHostException
+
+object ErrorResolver {
+
+  fun resolve(error: Throwable): ResolvedError {
+    val actualCause = findActualCause(error)
+
+    return when (actualCause) {
+      is SocketException,
```
vinaysshenoy: I feel we can make this a set of Error types that can be checked against.

app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
```diff
@@ -0,0 +1,43 @@
...
+    }
+    // This place may identify more wrapped errors
+    // like UndeliverableException, etc. in the future.
+    return actualError
+  }
+}
+
+sealed class ResolvedError(val actualCause: Throwable, val shouldBeReported: Boolean) {
+
+  class NetworkRelated(actualCause: Throwable) : ResolvedError(actualCause, shouldBeReported = false)
```
vinaysshenoy: I don't think `shouldBeReported` should be a property of the `ResolvedError` type. The reason is that the responsibility of this class is to classify errors into network related errors or not. Whether they should be reported or not must be the responsibility of the reporter.

app/src/test/java/org/simple/clinic/util/ErrorResolverTest.kt
```diff
@@ -0,0 +1,62 @@
...
+    assertThat(resolvedError.shouldBeReported).isFalse()
+    assertThat(resolvedError).isInstanceOf(ResolvedError.NetworkRelated::class.java)
+  }
+
+  @Suppress("unused")
+  fun `network related errors`() = arrayOf<Any>(
+      SocketException(),
+      SocketTimeoutException(),
+      UnknownHostException(),
+      StreamResetException(ErrorCode.ENHANCE_YOUR_CALM),
```
vinaysshenoy: TIL this is an actual error in HTTP/2.

app/src/main/java/org/simple/clinic/sync/SyncWorker.kt
saket: Timber logs are only forwarded to Sentry as breadcrumbs. They won't show up as individual crash reports.

app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
saket: That's a good idea, I'll do this.

app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
saket: Makes sense, I agree with this. Will change this.

app/src/test/java/org/simple/clinic/util/ErrorResolverTest.kt
saket: It's a name for an unofficial 420 http code.

saket:

PR #376, opened 2018-11-22
Fuzzy name search method that works on the basis of individual parts in a name and search term

vinaysshenoy:
@saket @sanchita I'd like both of your eyes on the PRs that make up the new fuzzy search implementation since there a lot of changes from what was being done earlier.

FYI: This doesn't introduce any actual changes to the behaviour, just introduces some classes that will be used in a future PR.

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/filter/SearchContext.kt
+++ b/app/src/main/java/org/simple/clinic/patient/filter/SearchContext.kt
@@ -1,0 +1,15 @@
+package org.simple.clinic.patient.filter
+
+import org.simple.clinic.patient.PatientSearchResult
+
+data class EditDistanceRecord(val namePart: String, val searchPart: String, val editDistance: Double)
+
+data class SearchContext(
+    val patient: PatientSearchResult.PatientNameAndId,
+    val nameParts: List<String>,
+    val searchParts: List<String>,
+    val editDistanceRecords: List<EditDistanceRecord> = emptyList()
+) {
+  val totalEditDistance: Double
+    get() = editDistanceRecords.map { it.editDistance }.sum()
+}
--- a/app/src/main/java/org/simple/clinic/patient/filter/SearchContext.kt
+++ b/app/src/main/java/org/simple/clinic/patient/filter/SearchContext.kt
@@ -2,14 +2,14 @@ package org.simple.clinic.patient.filter
 
 import org.simple.clinic.patient.PatientSearchResult
 
-data class EditDistanceRecord(val namePart: String, val searchPart: String, val editDistance: Double)
-
 data class SearchContext(
     val patient: PatientSearchResult.PatientNameAndId,
     val nameParts: List<String>,
     val searchParts: List<String>,
-    val editDistanceRecords: List<EditDistanceRecord> = emptyList()
+    val editDistances: List<EditDistance> = emptyList()
 ) {
   val totalEditDistance: Double
-    get() = editDistanceRecords.map { it.editDistance }.sum()
+    get() = editDistances.map { it.editDistance }.sum()
+
+  data class EditDistance(val namePart: String, val searchPart: String, val editDistance: Double)
 }
--- a/app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
+++ b/app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
@@ -51,13 +51,13 @@ class WeightedLevenshteinSearch(
           .filter { it.nameParts.isNotEmpty() }
           .map {
             val distances = searchTermParts.zip(it.nameParts) { searchPart, namePart ->
-              EditDistanceRecord(
+              SearchContext.EditDistance(
                   namePart = namePart,
                   searchPart = searchPart,
                   editDistance = levenshtein.distance(searchPart, namePart))
             }
 
-            it.copy(editDistanceRecords = distances)
+            it.copy(editDistances = distances)
           }
           .filter { it.totalEditDistance <= fuzzyStringDistanceCutoff }
           .sortedWith(resultsComparator)
--- a/app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
+++ b/app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
@@ -59,11 +59,15 @@ class WeightedLevenshteinSearch(
 
             it.copy(editDistances = distances)
           }
-          .filter { it.totalEditDistance <= maximumAllowedEditDistance }
+          .filter(this::discardLowQualityResults)
           .sortedWith(resultsComparator)
 
       Single.just(searchContext.map { it.patient.uuid })
     }
+  }
+
+  private fun discardLowQualityResults(patientSearchContext: PatientSearchContext): Boolean {
+    return patientSearchContext.totalEditDistance <= maximumAllowedEditDistance
   }
 
   private fun stringToSearchableParts(string: String) = string.split(whiteSpaceRegex)
--- a/app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
+++ b/app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
@@ -34,7 +34,7 @@ class WeightedLevenshteinSearch(
   )
 
   override fun search(searchTerm: String, names: List<PatientSearchResult.PatientNameAndId>): Single<List<UUID>> {
-    val searchTermParts = nameToSearchableParts(searchTerm)
+    val searchTermParts = stringToSearchableParts(searchTerm)
         .filter { it.length >= minimumSearchTermLength }
 
     return if (searchTermParts.isEmpty()) {
@@ -45,7 +45,7 @@ class WeightedLevenshteinSearch(
           .map {
             PatientSearchContext(
                 patient = it,
-                nameParts = nameToSearchableParts(it.fullName),
+                nameParts = stringToSearchableParts(it.fullName),
                 searchParts = searchTermParts)
           }
           .filter { it.nameParts.isNotEmpty() }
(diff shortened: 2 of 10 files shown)


app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
```diff
@@ -0,0 +1,68 @@
...
+          .map {
+            SearchInfo(
+                patient = it,
+                nameParts = nameToSearchableParts(it.fullName),
+                searchParts = searchTermParts)
+          }
+          .filter { it.nameParts.isNotEmpty() }
+          .map {
+            val distances = searchTermParts
+                .zip(it.nameParts) { searchPart, namePart -> levenshtein.distance(searchPart, namePart) }
```
saket: Do you want to use `.zip(it.nameParts, levenshtein::distance)`?

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.simple.clinic.patient.PatientSearchResult
+import java.util.UUID
+
+class WeightedLevenshteinSearch(
+    private val minimumSearchTermLength: Int,
+    private val fuzzyStringDistanceCutoff: Float,
+    characterSubstitutionCost: Float,
+    characterInsertionCost: Float,
+    characterDeletionCost: Float,
+    private val resultsComparator: Comparator<SearchInfo>
```
saket: nit: can we group values and constructor params separately?

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
```diff
@@ -0,0 +1,68 @@
...
+          .filter { it.distances.sum() <= fuzzyStringDistanceCutoff }
+          .sortedWith(resultsComparator)
+
+      Single.just(searchInfo.map { it.patient.uuid })
+    }
+  }
+
+  private fun nameToSearchableParts(string: String) = string.split(whiteSpaceRegex)
+      .filter { it.length >= minimumSearchTermLength }
+      .map { it.toLowerCase() }
```
saket: Why is this function called `nameToSearchableParts()`? Is it only supposed to split them into lowercase parts?

app/src/test/java/org/simple/clinic/patient/filter/SortByTotalSumOfDistancesTest.kt
```diff
@@ -0,0 +1,69 @@
...
+      }
+      val expected = results.map { originalList[it] }
+
+      return listOf(originalList, expected)
+    }
+
+    return listOf(
+        generateTestData(
+            distances = listOf(
+                listOf(1.0),
```
saket: Will using local data classes here instead of using numbers directly make this function more understandable? I see lots of numbers here.

saket:
@vinaysshenoy this PR involves a lot of maths. Can we pair review this sometime?

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
vinaysshenoy: Much nicer. I'll do this.

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
vinaysshenoy: Not necessarily. It also filters them based on a minimum length required for a search term to be considered for fuzzy matching.

app/src/test/java/org/simple/clinic/patient/filter/SortByTotalSumOfDistancesTest.kt
vinaysshenoy: It might. I'll take a stab at it.

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
```diff
@@ -0,0 +1,72 @@
...
+    return if (searchTermParts.isEmpty()) {
+      Single.just(emptyList())
+    } else {
+      val searchContext = names
+          .filter { it.fullName.isNotBlank() }
+          .map {
+            SearchContext(
+                patient = it,
+                nameParts = nameToSearchableParts(it.fullName),
+                searchParts = searchTermParts)
```
saket: I'm getting slightly confused with the naming here. The value of `nameToSearchableParts()` is stored in `nameParts`, but there's another property called `searchParts`.

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
```diff
@@ -0,0 +1,72 @@
...
+          }
+          .filter { it.nameParts.isNotEmpty() }
```
saket: Why will name parts be empty if the name isn't empty?

saket: Oh I see, the name parts received from `nameToSearchableParts()` can be empty. Hrm, this is slightly difficult to follow. I wonder if we can use a type different from String to make this more readable? 

Do you think this is necessary?

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
```diff
@@ -0,0 +1,72 @@
...
+            val distances = searchTermParts.zip(it.nameParts) { searchPart, namePart ->
+              EditDistanceRecord(
+                  namePart = namePart,
+                  searchPart = searchPart,
+                  editDistance = levenshtein.distance(searchPart, namePart))
+            }
+
+            it.copy(editDistanceRecords = distances)
+          }
+          .filter { it.totalEditDistance <= fuzzyStringDistanceCutoff }
```
saket: This fluent chain is a complex and I'm finding it a bit uncomfortable to understand it because everything is so terse.  For example, this line,

```kotlin
filter { it.totalEditDistance <= fuzzyStringDistanceCutoff }
```

doesn't give me enough information on why are they being discarded. Does using a lambda add any more clarity?

```kotlin
filter { discardLowQualityResults() }
```

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
saket: Okay

app/src/main/java/org/simple/clinic/patient/filter/SearchContext.kt
```diff
@@ -0,0 +1,15 @@
+package org.simple.clinic.patient.filter
+
+import org.simple.clinic.patient.PatientSearchResult
+
+data class EditDistanceRecord(val namePart: String, val searchPart: String, val editDistance: Double)
```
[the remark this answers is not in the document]
vinaysshenoy: I'll make it a nested class of `SearchContext`. 

I did call it `EditDistance` at first, but then I realized it also holds the two terms it holds the edit distance for. I can still change it, though.

app/src/main/java/org/simple/clinic/patient/filter/SearchContext.kt
```diff
@@ -0,0 +1,15 @@
...
+
+data class SearchContext(
```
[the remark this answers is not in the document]
vinaysshenoy: Sounds good, I'll do this.

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
```diff
@@ -0,0 +1,72 @@
...
+
+  override fun search(searchTerm: String, names: List<PatientSearchResult.PatientNameAndId>): Single<List<UUID>> {
+    val searchTermParts = nameToSearchableParts(searchTerm)
+        .filter { it.length >= minimumSearchTermLength }
+
+    return if (searchTermParts.isEmpty()) {
+      Single.just(emptyList())
+    } else {
+      val searchContext = names
+          .filter { it.fullName.isNotBlank() }
```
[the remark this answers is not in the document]
vinaysshenoy: We won't. But I didn't want to write this class assuming that it will always receive clean input.

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
vinaysshenoy: Ah, I should probably name this function differently, probably call it `stringToSearchableParts` or something.

`nameParts` is the split parts of the patient name, and `searchParts` is the split parts of the search term.

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
vinaysshenoy: Hmm.. I'm not sure. Shall we revisit this in the upcoming PRs that complete this feature?

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
vinaysshenoy: Looks good. I'll do this.

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
vinaysshenoy: Renamed to `stringToSearchableParts `.

saket: Will `searchParts` -> `searchQueryParts` be more understandable? Feel free to do this in your next PR.

vinaysshenoy: Actually, yes. Will make the change in my coming PR.

PR #388, opened 2018-11-29
UI improvements

saket:


app/src/main/java/org/simple/clinic/registration/pin/RegistrationPinScreenController.kt
```diff
@@ -55,11 +66,12 @@ class RegistrationPinScreenController @Inject constructor(
...
   private fun updateOngoingEntryAndProceed(events: Observable<UiEvent>): Observable<UiChange> {
-    val pinTextChanges = events.ofType<RegistrationPinTextChanged>()
     val doneClicks = events.ofType<RegistrationPinDoneClicked>()
+    val pinTextChanges = events
+        .ofType<RegistrationPinTextChanged>()
+        .map { it.pin }
 
-    return doneClicks
-        .withLatestFrom(pinTextChanges.map { it.pin })
+    return Observables.combineLatest(doneClicks, pinTextChanges)
```
vinaysshenoy: Why was this done here? Won't this let us skip the need to filter on a valid PIN since `RegistrationPinDoneClicked` will emit whenever a valid PIN is entered?

```kotlin
return doneClicks
.wlf(pinTextChanges) { _, pin -> pin }
// Do whatever
```

vinaysshenoy: Nvm, I just realised that the event can be triggered from the IME Done action too.

vinaysshenoy:
@saket During registration, if there's an error in validation when confirming security, the error is no longer shown.

app/src/main/java/org/simple/clinic/registration/pin/RegistrationPinScreenController.kt
saket: Because `Done` is now a derived event that is automatically emitted when `Text` length reaches 4. As a result, `doneClicks.withLatestFrom(textChanges)` no longer works. `textChanges.withLatestFrom(doneClicks)` might have worked.

saket:
@vinaysshenoy I've fixed the bug.

app/src/main/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinScreen.kt
```diff
@@ -81,23 +87,18 @@ class RegistrationConfirmPinScreen(context: Context, attrs: AttributeSet) : Rela
           .map { RegistrationConfirmPinDoneClicked() }
 
   fun showPinMismatchError() {
+    Timber.i("Showing error")
```
vinaysshenoy: Leftover log?

PR #395, opened 2018-12-05
Complete edit age/DOB feature

vinaysshenoy:


app/src/test/java/org/simple/clinic/editpatient/PatientEditScreenControllerTest.kt
```diff
@@ -763,26 +792,66 @@ class PatientEditScreenControllerTest {
       )
     }
 
+    val oneYear = Duration.ofDays(365L)
+    val twoYears = oneYear.plus(oneYear)
+
     return listOf(
         generateTestData(
-            patientProfile = generatePatientProfile(false),
+            patientProfile = generatePatientProfile(shouldAddNumber = false, shouldHaveAge = false),
```
saket: This list of test data is so difficult to parse and understand what kind of profiles are you creating any why. I wish we had a way to declare possible values and generate all combinations of them. 😌

app/src/test/java/org/simple/clinic/editpatient/PatientEditScreenControllerTest.kt
```diff
@@ -725,6 +747,8 @@ class PatientEditScreenControllerTest {
     fun generateTestData(
         patientProfile: PatientProfile,
         numberValidationResult: PhoneNumberValidator.Result,
+        dateOfBirthValidationResult: DateOfBirthFormatValidator.Result,
+        advanceClockBy: Duration,
```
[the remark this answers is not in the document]
vinaysshenoy: Makes sense since in most of the tests, we use the valid parameters anyway. I'll do this.

vinaysshenoy:
@saket Feedback addressed.

app/src/test/java/org/simple/clinic/editpatient/PatientEditScreenControllerTest.kt
saket: > but I do not see how we'll generated expected values to test against.

By declaring collections of values and generating permutations of it? Though I'm not sure how will we handle cases where some specific combination of values have to be discarded, etc.

> Do you feel we should break Patient, Address, PatientPhoneNumber exclusive parameters into their own test methods and then leaving this only for the combination of all models?

Hrm, I'm finding it difficult to imagine if that'll help. I'm unable to think of anything else right now. Let's continue this later and merge this PR for now.

PR #396, opened 2018-12-05
Store appointment cancel reason enums as raw strings

saket:
Enums are great for representing constants as they offer a type-safe API for usage, but blow-up when unknown strings are received from the server. Our JSON library, Moshi, provides a way to use a default enum as a fallback for unknown strings, but that results in the original string getting lost. When the enum has to be sent back to the server during our two way sync, there's no way to retrieve the actual value.

This commit replaces enums with sealed classes for storing Appointment cancellation reasons. The conversion of cancellation reason strings <-> sealed classes happen only when they're being read, allowing us to store the strings received from the server as-is. The usages also get to keep their type-safe API. Kotlin FTW!

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/removepatient/RemoveAppointmentEvents.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/removepatient/RemoveAppointmentEvents.kt
@@ -1,6 +1,6 @@
 package org.simple.clinic.home.overdue.removepatient
 
-import org.simple.clinic.overdue.Appointment
+import org.simple.clinic.overdue.AppointmentCancelReason
 import org.simple.clinic.widgets.UiEvent
 import java.util.UUID
 
@@ -18,6 +18,6 @@ data class PatientDeadClicked(val patientUuid: UUID) : UiEvent {
   override val analyticsName = "Remove Appointment with Reason:Reason changed to Patient Dead"
 }
 
-data class CancelReasonClicked(val selectedReason: Appointment.CancelReason) : UiEvent {
+data class CancelReasonClicked(val selectedReason: AppointmentCancelReason) : UiEvent {
   override val analyticsName = "Remove Appointment with Reason:Reason changed to $selectedReason"
 }
--- a/app/src/main/java/org/simple/clinic/overdue/AppointmentCancelReason.kt
+++ b/app/src/main/java/org/simple/clinic/overdue/AppointmentCancelReason.kt
@@ -1,0 +1,94 @@
+package org.simple.clinic.overdue
+
+import android.arch.persistence.room.TypeConverter
+import android.support.annotation.VisibleForTesting
+import com.squareup.moshi.FromJson
+import com.squareup.moshi.ToJson
+
+sealed class AppointmentCancelReason {
+
+  /**
+   * A human-readable string that can be printed when, say a reason is logged to analytics.
+   * Unlike enums, calling toString() on sealed classes will print their fully qualified name
+   * (e.g., org.simple.clinic.overdue.AppointmentCancelReason$Moved@6aceb1a5), which is not
+   * very readable.
+   */
+  abstract override fun toString(): String
+
+  object PatientNotResponding : AppointmentCancelReason() {
+    override fun toString() = "PatientNotResponding"
+  }
+
+  object Moved : AppointmentCancelReason() {
+    override fun toString() = "Moved"
+  }
+
+  object Dead : AppointmentCancelReason() {
+    override fun toString() = "Dead"
+  }
+
+  object Other : AppointmentCancelReason() {
+    override fun toString() = "Other"
+  }
+
+  data class Unknown(val actualValue: String) : AppointmentCancelReason() {
+    override fun toString() = "Unknown ($actualValue)"
+  }
+
+  /**
+   * It'll be nice to write a code generator instead of hand-typing
+   * and manually maintaining this adapter in the future.
+   */
+  object TypeAdapter {
+
+    val KNOWN_MAPPINGS = mapOf(
+        PatientNotResponding to "not_responding",
+        Moved to "moved",
+        Dead to "dead",
+        Other to "other")
+
+    fun toEnum(reason: String?): AppointmentCancelReason? {
+      if (reason == null) {
+        return null
+      }
+
+      val foundEnum = KNOWN_MAPPINGS.entries
+          .find { (_, jsonKey) -> jsonKey == reason }
+          ?.key
+      return foundEnum ?: Unknown(actualValue = reason)
+    }
+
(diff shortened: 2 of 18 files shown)


app/src/main/java/org/simple/clinic/home/overdue/removepatient/RemoveAppointmentEvents.kt
```diff
@@ -18,6 +18,6 @@ data class PatientDeadClicked(val patientUuid: UUID) : UiEvent {
   override val analyticsName = "Remove Appointment with Reason:Reason changed to Patient Dead"
 }
 
-data class CancelReasonClicked(val selectedReason: Appointment.CancelReason) : UiEvent {
-  override val analyticsName = "Remove Appointment with Reason:Reason changed to $selectedReason"
+data class CancelReasonClicked(val selectedReason: AppointmentCancelReason) : UiEvent {
+  override val analyticsName = "Remove Appointment with Reason:Reason changed to ${AppointmentCancelReason.TypeAdapter.fromEnum(selectedReason)}"
```
saket: Because cancelation reasons are now Kotlin objects, their string representation will not be user readable anymore. E.g., `org.simple.clinic.overdue.AppointmentCancelReason$Moved@6aceb1a5`

Using the type adapter is necessary here, but it makes this string very long. I hope that's okay.

app/src/main/java/org/simple/clinic/overdue/AppointmentCancelReason.kt
```diff
@@ -0,0 +1,74 @@
...
+  class RoomTypeConverter {
+    @TypeConverter
+    fun toEnum(reason: String?): AppointmentCancelReason? = TypeAdapter.toEnum(reason)
+
+    @TypeConverter
+    fun fromEnum(reason: AppointmentCancelReason?): String? = TypeAdapter.fromEnum(reason)
+  }
+
+  companion object {
+    fun random() = TypeAdapter.KNOWN_MAPPINGS.entries.shuffled().first().key
```
saket: What do you think of keeping this function here? It's only used in the test and shouldn't belong here, but the mappings are private to this class.

app/src/main/java/org/simple/clinic/home/overdue/removepatient/RemoveAppointmentEvents.kt
vinaysshenoy: What do you feel about adding a property `analyticsName` in the sealed class itself?

```kotlin
sealed class AppointmentCancelReason {
  
  val analticsName: String
    get() = TypeAdapter.fromEnum(this)!!
}
```

app/src/main/java/org/simple/clinic/overdue/AppointmentCancelReason.kt
```diff
@@ -0,0 +1,74 @@
...
+  object Moved : AppointmentCancelReason()
+
+  object Dead : AppointmentCancelReason()
+
+  object Other : AppointmentCancelReason()
+
+  data class Unknown(val actualValue: String) : AppointmentCancelReason()
+
+  /**
+   * It'll be nice to write a code generator instead of hand-typing
```
vinaysshenoy: We have many enums in the api, and will have many in the future. We might want to use this for some, or all of them. Makes sense to invest some time, either in this or a generic wrapper that will manage this, at some point. Let's revisit this when we have to convert our next enum to a string.

app/src/main/java/org/simple/clinic/overdue/AppointmentCancelReason.kt
vinaysshenoy: I think it's fine for now. Perhaps, in the future, we can investigate using `kotlin-reflect` in tests to do something similar like we have done for enums.

vinaysshenoy: Do you want to add a `@VisibleForTesting` annotation to the method?

app/src/test/java/org/simple/clinic/overdue/AppointmentCancelReasonTypeAdapterTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        listOf("not_responding", AppointmentCancelReason.PatientNotResponding),
+        listOf("moved", AppointmentCancelReason.Moved),
+        listOf("dead", AppointmentCancelReason.Dead),
+        listOf("other", AppointmentCancelReason.Other)
+    )
+  }
+
+  @Test
+  fun `unknown reason enum should be serialized to its actual value`() {
+    val unknownReason = AppointmentCancelReason.Unknown(actualValue = "abducted_by_joker")
```
vinaysshenoy: I can see why that would make it hard for a patient to show up.

vinaysshenoy: Do you want to add one more unknown reason as a parameterized test?

app/src/test/java/org/simple/clinic/overdue/AppointmentCancelReasonTypeAdapterTest.kt
```diff
@@ -0,0 +1,67 @@
...
+    val serialized = AppointmentCancelReason.TypeAdapter.fromEnum(unknownReason)
+
+    assertThat(unknownReason.actualValue).isEqualTo(serialized)
```
vinaysshenoy: Nitpick: It's usually `assertThat(valueToBeTested).isEqualTo(expectedValue)`. Here, it's the other way round.

app/src/main/java/org/simple/clinic/overdue/AppointmentCancelReason.kt
saket: Yes, good idea

app/src/test/java/org/simple/clinic/overdue/AppointmentCancelReasonTypeAdapterTest.kt
saket: Yea sure

app/src/test/java/org/simple/clinic/overdue/AppointmentCancelReasonTypeAdapterTest.kt
saket: True, this is a copy pasta mistake. I'll fix this.

app/src/main/java/org/simple/clinic/overdue/AppointmentCancelReason.kt
saket: Agreed.

saket:
@vinaysshenoy I do not want to store `analyticsName` in the sealed classes. I don't think they belong there. Don't you think they should remain inside UI events?

PR #398, opened 2018-12-05
Fix UserSession Test by sending only one facility during registration

sanchita:


app/src/androidTest/java/org/simple/clinic/AuthenticationRule.kt
```diff
@@ -67,7 +67,7 @@ class AuthenticationRule : TestRule {
       val registrationEntry = testData.ongoingRegistrationEntry(
           phoneNumber = faker.number.number(10),
           pin = testData.qaUserPin(),
-          facilities = facilities)
+          facilities = facilities.take(1))
```
saket: Has the server stopped accepting multiple facilities?

vinaysshenoy: Yes, this change was merged in yesterday, it seems.

vinaysshenoy: @saket @sanchita I am going to merge this in.

PR #400, opened 2018-12-07
Sort fuzzy matches by likeliness of individual parts of a name rather than the complete name

vinaysshenoy:
This PR also enables the new fuzzy search feature.

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/filter/SortByWeightedNameParts.kt
+++ b/app/src/main/java/org/simple/clinic/patient/filter/SortByWeightedNameParts.kt
@@ -1,0 +1,48 @@
+package org.simple.clinic.patient.filter
+
+import kotlin.math.max
+
+/**
+ * This is a comparator that is used to sort [PatientSearchContext] objects by prioritizing matches
+ * which have the same number of words as the search term while still providing good
+ * results when the search term and the name do not have the same number of words.
+ *
+ * It works by generating an overall score for a match based on a weighted function that
+ * multiples the edit distance for a particular part with a multiplier which is decided
+ * based on the position of the term in the word.
+ *
+ * Variations in words are penalized in a progressively decreasing manner from left to
+ * right. Names with more matching words get a better score.
+ **/
+class SortByWeightedNameParts : Comparator<PatientSearchContext> {
+
+  override fun compare(first: PatientSearchContext, second: PatientSearchContext): Int {
+    val scoreOfFirst = computeScore(first)
+    val scoreOfSecond = computeScore(second)
+
+    return scoreOfFirst.compareTo(scoreOfSecond)
+  }
+
+  private fun computeScore(searchInfo: PatientSearchContext): Double {
+    val maximumWordParts = max(searchInfo.nameParts.size, searchInfo.searchParts.size)
+
+    // Generate decreasing multipliers based on the position of the
+    // term. i.e, if the maximum number of parts is 3, the multipliers are
+    // 3, 2, 1 for positions 0, 1, 2 respectively.
+    val penaltiesForPositions: List<Int> = (1..maximumWordParts).map { it }.reversed()
+
+    // Generate weighted scores for each position by assigning a default value,
+    // one divided by the max parts in this case, and multiplying that value with
+    // the position multipliers generated previously.
+    val triedAndTestedArbitraryValue = 1 / maximumWordParts.toDouble()
+    val positionBasedDefaultWeights = penaltiesForPositions.map { penalty -> penalty * triedAndTestedArbitraryValue }
+
+    val editDistances = searchInfo.editDistances.map { it.editDistance }
+
+    return positionBasedDefaultWeights
+        .mapIndexed { index, weightedScore ->
+          weightedScore * editDistances.getOrElse(index) { 1.0 }
+        }
+        .sum()
+  }
+}
--- a/app/src/test/java/org/simple/clinic/patient/filter/SortByWeightedNamePartsTest.kt
+++ b/app/src/test/java/org/simple/clinic/patient/filter/SortByWeightedNamePartsTest.kt
@@ -1,0 +1,93 @@
+package org.simple.clinic.patient.filter
+
+import com.google.common.truth.Truth.assertThat
+import junitparams.JUnitParamsRunner
+import junitparams.Parameters
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.simple.clinic.patient.PatientSearchResult.PatientNameAndId
+import java.util.UUID
+
+@RunWith(JUnitParamsRunner::class)
+class SortByWeightedNamePartsTest {
+
+  @Test
+  @Parameters(method = "params for sorting values")
+  fun `it should sort the results based on the overall match of individual parts`(
+      originalList: List<PatientSearchContext>,
+      expected: List<PatientSearchContext>
+  ) {
+    val sorted = originalList.sortedWith(SortByWeightedNameParts())
+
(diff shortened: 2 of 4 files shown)


vinaysshenoy:
@saket @sanchita The algorithm used to sort the results here is a little involved, so I'm assigning the both of you to this. Maybe we can have a detailed discussion on Monday about what's actually happening here.

saket:
@vinaysshenoy I'm on leave till Wednesday. Pull pratul if he's available. :)

app/src/main/java/org/simple/clinic/patient/filter/SortByWeightedNameParts.kt
```diff
@@ -0,0 +1,54 @@
...
+ * which have the same number of words as the search term while still providing good
+ * results when the search term and the name do not have the same number of words.
+ *
+ * It works by generating an overall score for a match based on a weighted function that
+ * multiples the edit distance for a particular part with a multiplier which is decided
+ * based on the position of the term in the word.
+ *
+ * Variations in the earlier parts of the name are penalised more than the variations in the
+ * later parts of the name and names with more matching words get a better score.
+ **/
```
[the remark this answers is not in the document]
saket: Sanchita and I came up with this:

```
Variations in words are penalized in a progressively decreasing manner from left to
right. Names with more matching words get a better score.
```

app/src/main/java/org/simple/clinic/patient/filter/SortByWeightedNameParts.kt
```diff
@@ -0,0 +1,54 @@
...
+    val scores = defaultScores.toMutableList()
+
+    searchInfo.editDistances.forEachIndexed { index, (_, _, distance) ->
+      scores[index] *= when (sumOfDistances) {
+        0.0 -> 0.0
+        else -> distance / sumOfDistances
+      }
+    }
+
+    return scores.sum()
```
saket: Sanchita and I tried rewording things to make this block of code more readable on the lines of how you helped us understand the maths:

```kotlin
// Generate decreasing multipliers based on the position of the
// term. i.e, if the maximum number of parts is 3, the multipliers are
// 3, 2, 1 for positions 0, 1, 2 respectively.
val penaltiesForPositions: List<Int> = (1..maximumWordParts).map { it }.reversed()

// Generate weighted scores for each position by assigning a default value,
// one divided by the max parts in this case, and multiplying that value with
// the position multipliers generated previously.
val triedAndTestedArbitraryValue = 1 / maximumWordParts.toDouble()
val positionBasedDefaultWeights = penaltiesForPositions.map { penalty -> penalty * triedAndTestedArbitraryValue }

val editDistances = searchInfo.editDistances.map { it.editDistance }

return positionBasedDefaultWeights
    .zip(editDistances)
    .map { (weightedScore, editDistance) -> weightedScore * editDistance }
    .sum()
```

app/src/test/java/org/simple/clinic/patient/filter/SortByWeightedNamePartsTest.kt
```diff
@@ -0,0 +1,92 @@
...
+
+  @Test
+  @Parameters(method = "params for sorting values")
+  fun `it should sort the results based on the overall match of individual parts`(
+      originalList: List<PatientSearchContext>,
+      expected: List<PatientSearchContext>
+  ) {
+    val sorted = originalList.sortedWith(SortByWeightedNameParts())
+
+    Truth.assertThat(sorted).isEqualTo(expected)
```
saket: Static import `Truth`?

app/src/test/java/org/simple/clinic/patient/filter/SortByWeightedNamePartsTest.kt
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @Suppress("Unused")
+  private fun `params for sorting values`(): List<List<Any>> {
+    fun generateTestData(
+        nameParts: List<List<String>>,
+        searchParts: List<String>,
+        distances: List<List<PatientSearchContext.EditDistance>>, results: List<Int>
```
saket: Took me a while to find where is `results` param coming from until I realized it's on the wrong line 😄

saket: `results` -> `resultIndices`?

app/src/main/java/org/simple/clinic/patient/filter/SortByWeightedNameParts.kt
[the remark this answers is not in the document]
vinaysshenoy: Sounds good. Will change.

app/src/main/java/org/simple/clinic/patient/filter/SortByWeightedNameParts.kt
```diff
@@ -0,0 +1,54 @@
...
+    val scoreOfFirst = computeScore(first)
+    val scoreOfSecond = computeScore(second)
+
+    return scoreOfFirst.compareTo(scoreOfSecond)
+  }
+
+  private fun computeScore(searchInfo: PatientSearchContext): Double {
+    val maximumWordParts = max(searchInfo.nameParts.size, searchInfo.searchParts.size)
+
+    // Generates decreasing multipliers based on the position of the
```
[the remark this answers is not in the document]
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/patient/filter/SortByWeightedNameParts.kt
vinaysshenoy: Looks good. I'll switch to this.

vinaysshenoy: Done. I had to make a small change since I can't use `zip` for the final score calculation because it will ignore scores for words that don't have a corresponding edit distance calculation.

app/src/test/java/org/simple/clinic/patient/filter/SortByWeightedNamePartsTest.kt
vinaysshenoy: Since I only had a single assertion in this class, I didn't do it. I'll add it in.

app/src/main/java/org/simple/clinic/patient/filter/SortByWeightedNameParts.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    // Generate weighted scores for each position by assigning a default value,
+    // one divided by the max parts in this case, and multiplying that value with
+    // the position multipliers generated previously.
+    val triedAndTestedArbitraryValue = 1 / maximumWordParts.toDouble()
+    val positionBasedDefaultWeights = penaltiesForPositions.map { penalty -> penalty * triedAndTestedArbitraryValue }
+
+    val editDistances = searchInfo.editDistances.map { it.editDistance }
+
+    val scores = positionBasedDefaultWeights.toMutableList()
```
saket: I'd not call this just `scores`. I thought `weightedScores` would be slightly better, but ended up skipping the problem entirely and inlining it with the return statement.

app/src/main/java/org/simple/clinic/patient/filter/SortByWeightedNameParts.kt
saket: > it will ignore scores for words that don't have a corresponding edit distance calculation.

Ah, right. That's what I was worried about with zip.

PR #409, opened 2018-12-14
Remove Age and Date of Birth filter from patient lookup

vinaysshenoy:


app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
```diff
@@ -14,20 +14,36 @@ class PatientFuzzySearch {
...
-          patientSearchDao
-              .searchByIds(uuids, dobUpperBound, dobLowerBound, PatientStatus.ACTIVE)
-              .compose(sortPatientSearchResultsByScore(uuidsSortedByScore))
-        }!!
+    override fun searchForPatientsWithNameLikeAndAgeWithin(
+        query: String,
+        dobUpperBound: String,
+        dobLowerBound: String
+    ): Single<List<PatientSearchResult>> {
+      return patientUuidsMatching(query).flatMap { uuidsSortedByScore ->
```
saket: Shouldn't we be using a `switchMap` here?

app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
```diff
@@ -14,20 +14,36 @@ class PatientFuzzySearch {
...
+        val uuids = uuidsSortedByScore.map { it.uuid }
+        patientSearchDao
+            .searchByIds(uuids, dobUpperBound, dobLowerBound, PatientStatus.ACTIVE)
+            .compose(sortPatientSearchResultsByScore(uuidsSortedByScore))
+      }
+    }
+
+    override fun searchForPatientsWithNameLike(query: String): Single<List<PatientSearchResult>> {
+      return patientUuidsMatching(query).flatMap { uuidsSortedByScore ->
```
saket: Same `switchMap` question here

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -49,6 +49,62 @@ class PatientRepository @Inject constructor(
 
   private var ongoingNewPatientEntry: OngoingNewPatientEntry = OngoingNewPatientEntry()
 
+  fun search(name: String): Observable<List<PatientSearchResult>> {
+    return configProvider.flatMapObservable { (isFuzzySearchV2Enabled) ->
```
saket: I'm slowly growing afraid of using destructuring because they're index based

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -49,6 +49,62 @@ class PatientRepository @Inject constructor(
...
+    val fuzzySearch = database.fuzzyPatientSearchDao()
+        .searchForPatientsWithNameLike(searchableName)
+        .toObservable()
+
+    return Observables.zip(substringSearch, fuzzySearch, configProvider.toObservable())
+        .map { (results, fuzzyResults, config) ->
+          // Fuzzy search has an internal limit, but the substring search also can return more than
+          // a thousand results which can fail when we sort on the current facility since it
+          // queries the blood pressure table with an IN clause which can cause an exception.
+          (fuzzyResults + results).distinctBy { it.uuid }.take(config.limitOfSearchResults)
```
saket: Operators on different lines please?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -49,6 +49,62 @@ class PatientRepository @Inject constructor(
...
+        }
+        .compose(sortByCurrentFacility())
+  }
+
+  private fun searchWithoutAgeV2(name: String): Observable<List<PatientSearchResult>> {
+    return database.patientSearchDao()
+        .nameAndId(PatientStatus.ACTIVE)
+        .toObservable()
+        .switchMapSingle { searchPatientByName.search(name, it) }
+        .zipWith(configProvider.toObservable())
```
saket: Why does this have to be a zip? We won't get updates if the database is updated.

saket: Why does this have to be a zip? We won't get updates if the database is updated.

app/src/main/java/org/simple/clinic/search/PatientSearchScreenController.kt
```diff
@@ -52,16 +42,8 @@ class PatientSearchScreenController @Inject constructor(
...
-        .ofType<SearchQueryAgeChanged>()
-        .map { it.ageString }
-
-    val dateOfBirthChanges = events
-        .ofType<SearchQueryDateOfBirthChanged>()
-        .map { it.dateOfBirth }
-
-    return Observables.combineLatest(nameChanges, ageChanges, dateOfBirthChanges)
-        .map { (name, age, dob) -> name.isNotBlank() && (age.isNotBlank() || dob.isNotBlank()) }
+    return nameChanges
```
saket: Let's inline `nameChanges` with the return statement?

```kotlin
return events
    .ofType<SearchQueryNameChanged>()
    .map { it.name.isNotBlank() }
    .map { isQueryComplete ->
      { ui: Ui ->
        if (isQueryComplete) {
          ui.showSearchButtonAsEnabled()
        } else {
          ui.showSearchButtonAsDisabled()
        }
      }
    }
```

app/src/main/java/org/simple/clinic/search/PatientSearchScreenController.kt
```diff
@@ -199,8 +110,9 @@ class PatientSearchScreenController @Inject constructor(
...
     return Observables.combineLatest(searchClicks, validationErrors)
         .filter { (_, errors) -> errors.isEmpty() }
-        .withLatestFrom(nameChanges, ageChanges, dateOfBirthChanges) { _, name, age, dob -> Triple(name, age, dob) }
-        .map { (name, age, dateOfBirth) -> { ui: Ui -> ui.openPatientSearchResultsScreen(name, age, dateOfBirth) } }
+        .withLatestFrom(nameChanges) { _, name ->
+          { ui: Ui -> ui.openPatientSearchResultsScreen(name) }
+        }
   }
 
   private fun saveAndProceeds(events: Observable<UiEvent>): Observable<UiChange> {
```
saket: Looks like this function is incorrectly named. Can you please rename this to something like `openPatientRegistrationScreen()` or something as part of this PR?

saket: We could also skip this if you want.

app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsController.kt
```diff
@@ -38,29 +36,20 @@ class PatientSearchResultsController @Inject constructor(
...
+        .flatMap { (name) ->
           val facilities = userSession
               .requireLoggedInUser()
               .switchMap { facilityRepository.currentFacility(it) }
 
-          val searchResults = patientRepository.search(name, computedAge)
+          val searchResults = patientRepository.search(name)
 
           Observables.combineLatest(searchResults, facilities)
               // We can't understand why, but search is occasionally
```
saket: Temporary solution that's clearly going to be permanent :p

app/src/main/res/values/styles_v2.xml
```diff
@@ -25,7 +25,7 @@
   </style>
 
   <style name="Clinic.V2.DialogNeutralButton" parent="Widget.AppCompat.Button.Borderless.Colored">
-     <item name="android:textAppearance">@style/Clinic.V2.DialogNeutralButton.TextAppearance</item>
+    <item name="android:textAppearance">@style/Clinic.V2.DialogNeutralButton.TextAppearance</item>
```
saket: I feel offended by this

app/src/test/java/org/simple/clinic/search/PatientSearchScreenControllerTest.kt
```diff
@@ -5,118 +5,44 @@ import com.nhaarman.mockito_kotlin.mock
...
   fun setUp() {
-    controller = PatientSearchScreenController(repository, dateValidator)
+    controller = PatientSearchScreenController(repository)
 
     uiEvents.compose(controller).subscribe { uiChange -> uiChange(screen) }
   }
 
   @Test
-  fun `search button should remain enabled only when both name and age or date of birth are present`() {
+  fun `search button should remain enabled only when both name is present`() {
```
saket: `both` can also be removed from this function's name

app/src/test/java/org/simple/clinic/search/results/PatientSearchResultsControllerTest.kt
```diff
@@ -39,65 +39,39 @@ class PatientSearchResultsControllerTest {
   private val uiEvents = PublishSubject.create<UiEvent>()
```
saket: There are some unused imports in this file. Can you clear them?

app/src/main/java/org/simple/clinic/ageanddateofbirth/DateFormatterModule.kt
```diff
@@ -13,10 +13,10 @@ class DateFormatterModule {
   @Provides
   @AppScope
   @Named("date_for_search_results")
-  fun provideShortDigitDateFormatter(): DateTimeFormatter = DateTimeFormatter.ofPattern("d/MM/yyyy", Locale.ENGLISH)
+  fun provideDateFormatterForSearchResults(locale: Locale): DateTimeFormatter = DateTimeFormatter.ofPattern("d-MMM-yyyy", locale)
```
saket: TODO: Ask why are we changing from `MM` to `MMM`

app/src/main/java/org/simple/clinic/di/AppModule.kt
```diff
@@ -113,4 +114,8 @@ open class AppModule(
   @Provides
   @AppScope
   open fun clock(): Clock = Clock.systemUTC()
+
+  @Provides
+  @AppScope
+  fun currentLocale(): Locale = Locale.ENGLISH
```
saket: Nice 👍

app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
vinaysshenoy: We could, but that would mean that we would have to convert the `patientUuidsMatching` method to a room Dao method instead of querying the data ourselves.

Given that is an unused data flow (since searchV2 flag has been set), I don't think we should spend time on doing this now. Maybe we can revisit this if we ever decide to implement the new fuzzy search in SQLite again.

app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
vinaysshenoy: Same answer as above.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: Good point. I'll remove the destructuring here. We should probably limit destructuring to things like `Pair`, `Triple` and `Map.Entry` going forward.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: Ah, good catch! I'll make this a `combineLatest`.

vinaysshenoy: Duplicate comment?

app/src/main/java/org/simple/clinic/search/PatientSearchScreenController.kt
vinaysshenoy: The name seems to be correct, as it opens the `PatientSearchResultsScreen`.

app/src/test/java/org/simple/clinic/search/PatientSearchScreenControllerTest.kt
vinaysshenoy: Done.

app/src/test/java/org/simple/clinic/search/results/PatientSearchResultsControllerTest.kt
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/ageanddateofbirth/DateFormatterModule.kt
vinaysshenoy: This was already discussed in person, but adding it here for the record.

The earlier "date_for_search_results" was not being used anywhere. The actual date formatted in the `PatientSearchResultsAdapter` was being instantiated in the class, using this format.

app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
saket: Got it

app/src/main/java/org/simple/clinic/patient/PatientFuzzySearch.kt
saket: Same answer's answer as above

PR #410, opened 2018-12-14
Cleaner disposal of UI Events with auto connect

vinaysshenoy:


vinaysshenoy:
@saket @sanchita I'm tagging both of you in this because this is what I propose we should be using for *ALL* of our screens, and I'd like everyone's feedback.

saket:
@vinaysshenoy I'm yet to run this on a device. Will do this tomorrow.

saket:
@vinaysshenoy looks like there's something missing. It's not working exactly as expected. For e.g., the done button in patient summary screen gets subscribed twice when the screen opens. Similarly, it gets disposed twice. Please check?

Having said that, the back and forth of binding and unbinding of Views isn't happening so that part is working fine.

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreen.kt
```diff
@@ -110,11 +113,13 @@ class PatientEditScreen(context: Context, attributeSet: AttributeSet) : Relative
 
     TheActivity.component.inject(this)
 
-    controller.disposeOnDetach(this)
+    val screenDestroys = RxView.detaches(this)
+        .map { ScreenDestroyed() }
```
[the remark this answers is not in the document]
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -119,11 +121,13 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
     setupSummaryList()
     setupEditButtonClicks()
 
-    controller.disposeOnDetach(this)
+    val screenDestroys = RxView.detaches(this)
+        .map { ScreenDestroyed() }
```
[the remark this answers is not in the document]
vinaysshenoy: Done.

vinaysshenoy:
> @vinaysshenoy looks like there's something missing. It's not working exactly as expected. For e.g., the done button in patient summary screen gets subscribed twice when the screen opens. Similarly, it gets disposed twice. Please check?
> 
> ```kotlin
> private fun doneClicks() = RxView.clicks(doneButtonFrame.button)
>     .doOnSubscribe { Timber.i("Subscribe done") }
>     .doOnDispose { Timber.i("Dispose done") }
>     .map { PatientSummaryDoneClicked() }
> ```
> 
> Having said that, the back and forth of binding and unbinding of Views isn't happening so that part is working fine.

Figured out the problem. `ReplayUntilScreenIsDestroyed` was subscribing to the original stream once when it set up the disposal of the stream, and then the controller itself was subscribing to the stream again.

I've changed `ReplayUntilScreenIsDestroyed` to take in a list of transformations that it applies inter ally and shares the final stream. After this, it sets up the disposal on the final shared stream instead of the original, which fixes this problem.

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
```diff
@@ -178,58 +158,62 @@ class PatientEditScreenController @Inject constructor(
...
-        .map { it.state }
-
-    val phoneNumberChanges = events
-        .ofType<PatientEditPhoneNumberTextChanged>()
-        .map { it.phoneNumber }
-
-    val ageChanges = events
-        .ofType<PatientEditAgeTextChanged>()
-        .map { EntryWithAge(it.age) as OngoingEditPatientEntry.EitherAgeOrDateOfBirth }
+  private fun mergeWithOngoingEntryPatientEntryChanges(): ObservableTransformer<UiEvent, UiEvent> {
```
saket: I'm happy that this will force us to use Transformers for every transformation. I was slowly realizing that this is better.

app/src/main/java/org/simple/clinic/ReplayUntilScreenIsDestroyed.kt
```diff
@@ -0,0 +1,45 @@
...
+              .autoConnect(1) { disposables += it }
+        }
+
+    disposeWhenScreenIsDestroyed(transformedEvents)
+
+    return transformedEvents
+  }
+
+  private fun disposeWhenScreenIsDestroyed(events: Observable<UiEvent>) {
+    disposables.add(events.ofType<ScreenDestroyed>().subscribe { disposables.clear() })
```
saket: This is slightly compact because both `disposables.add()` and `disposables.clear()` are happening on the same line. Can we break this line?

```kotlin
disposables.add(
    events
        .ofType<ScreenDestroyed>()
        .subscribe { disposables.clear() })
```

app/src/main/java/org/simple/clinic/ReplayUntilScreenIsDestroyed.kt
```diff
@@ -0,0 +1,45 @@
...
+import io.reactivex.rxkotlin.ofType
+import io.reactivex.rxkotlin.plusAssign
+import org.simple.clinic.widgets.ScreenDestroyed
+import org.simple.clinic.widgets.UiEvent
+
+class ReplayUntilScreenIsDestroyed(
+    private val transforms: List<ObservableTransformer<UiEvent, UiEvent>> = emptyList()
+) : ObservableTransformer<UiEvent, UiEvent> {
+
+  constructor(transform: ObservableTransformer<UiEvent, UiEvent>) : this(transforms = listOf(transform))
```
saket: Let's do `@Suppress("unused")`?

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
```diff
@@ -57,30 +57,10 @@ class PatientEditScreenController @Inject constructor(
...
-        .subscribe {
-          disposables.clear()
-        }
-  }
-
   override fun apply(events: Observable<UiEvent>): ObservableSource<UiChange> {
-    val replayedEvents = events.compose(ReportAnalyticsEvents())
-        .replay()
-        .autoConnect(1) { d -> disposables += d }
+    val replayUntilScreenIsDestroyed = ReplayUntilScreenIsDestroyed(ReportAnalyticsEvents(), mergeWithOngoingEntryPatientEntryChanges())
```
saket: It feels that replay-until-screen-destroyed is only being applied to the two arguments to `ReplayUntilScreenIsDestroyed` and not to `events` parameter. Do you think this will be more understandable?

```kotlin
ReplayUntilScreenIsDestroyed(events, ReportAnalyticsEvents(), mergeWithPatientSummaryChanges())
```

Not too convinced with this though.

app/src/main/java/org/simple/clinic/ReplayUntilScreenIsDestroyed.kt
vinaysshenoy: I could do it, but we'll end up removing it almost immediately because we can migrate our other screens to it (and `ReportAnalyticsEvents()` will always be present).

Do you still feel we should do it?

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
vinaysshenoy: Yes, but passing `events` to it is tricky since it itself is composed on the stream outside and passing events to the constructor is pointless since it will be ignored.

app/src/main/java/org/simple/clinic/ReplayUntilScreenIsDestroyed.kt
vinaysshenoy: Yup, I'll do this.

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
vinaysshenoy: Discussed offline: Try to get a cleaner API like

```kotlin
        .compose(mergeWithPatientSummaryChanges())
        .replay()
```

app/src/main/java/org/simple/clinic/ReplayUntilScreenIsDestroyed.kt
vinaysshenoy: Discussed offline: This will no longer be relevant when the new API is done.

saket:
@vinaysshenoy this looks fab. Merging it in.

PR #415, opened 2018-12-20
Fix broken tests

vinaysshenoy:


app/src/androidTest/java/org/simple/clinic/AuthenticationRule.kt
```diff
@@ -46,28 +73,61 @@ class AuthenticationRule : TestRule {
...
         } finally {
           logout()
+          registeredFacilityUuid = null
         }
       }
     }
   }
 
-  private fun register() {
+  private fun registerPatient(patientUuid: UUID) {
```
saket: Discussed offline: We'll move registration of patients and appointments to somewhere outside this class because:

- they're not the responsibility of `AuthenticationRule`
- the user's current facility can be fetched easily using `FacilityRepository.currentFacility(User)`

PR #420, opened 2018-12-21
Add "deletedAt" property to all sync models

vinaysshenoy:


app/src/main/java/org/simple/clinic/overdue/communication/CommunicationPullResponse.kt
```diff
@@ -10,7 +10,7 @@ data class CommunicationPullResponse(
     @Json(name = "communications")
     override val payloads: List<CommunicationPayload>,
 
-    @Json(name = "processed_since")
+    @Json(name = "process_token")
     override val processedSinceTimestamp: String
```
sanchita: We could rename the variable `processedSinceTimestamp` in all our Response models.
Also in `DataPullResponse`.

app/src/main/java/org/simple/clinic/facility/FacilityPullResponse.kt
```diff
@@ -10,6 +10,6 @@ data class FacilityPullResponse(
     @Json(name = "facilities")
     override val payloads: List<FacilityPayload>,
 
-    @Json(name = "processed_since")
+    @Json(name = "process_token")
     override val processedSinceTimestamp: String
```
sanchita: Rename this  `processedSinceTimestamp`-> `processToken`

app/src/main/java/org/simple/clinic/overdue/communication/CommunicationPullResponse.kt
vinaysshenoy: Makes sense, I'll do this.

vinaysshenoy: Done.

PR #426, opened 2018-12-26
Mark blood pressure as deleted

vinaysshenoy:
The feature to filter deleted BPs from the UI will be done in an upcoming PR.

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
@@ -37,6 +37,9 @@ class BloodPressureRepositoryAndroidTest {
   @get:Rule
   val authenticationRule = AuthenticationRule()
 
+  val testClock: TestClock
+    get() = clock as TestClock
+
   @Before
   fun setUp() {
     TestClinicApp.appComponent().inject(this)
@@ -55,13 +58,12 @@ class BloodPressureRepositoryAndroidTest {
         diastolic = 80,
         createdAt = Instant.now(clock),
         updatedAt = Instant.now(clock),
-        syncStatus = SyncStatus.DONE
-    )
+        syncStatus = SyncStatus.DONE)
 
     appDatabase.bloodPressureDao().save(listOf(bloodPressure))
 
     val durationToAdvanceBy = Duration.ofMinutes(15L)
-    (clock as TestClock).advanceBy(durationToAdvanceBy)
+    testClock.advanceBy(durationToAdvanceBy)
 
     repository.updateMeasurement(bloodPressure.copy(systolic = 130, diastolic = 90)).blockingAwait()
 
@@ -103,5 +105,32 @@ class BloodPressureRepositoryAndroidTest {
     val bpMeasurements = repository.newestMeasurementsForPatient(patientUuid, 4).blockingFirst()
 
     assertThat(bpMeasurements).isEqualTo(listOf(bloodPressure2, bloodPressure4, bloodPressure1, bloodPressure3))
+  }
+
+  @Test
+  fun marking_a_blood_pressure_as_deleted_should_work_correctly() {
+    val now = Instant.now(clock)
+    val bloodPressure = testData.bloodPressureMeasurement(
+        createdAt = now,
+        updatedAt = now,
+        deletedAt = null,
+        syncStatus = SyncStatus.DONE)
+
+    appDatabase.bloodPressureDao().save(listOf(bloodPressure))
+
+    val durationToAdvanceBy = Duration.ofMinutes(15L)
+    testClock.advanceBy(durationToAdvanceBy)
+
+    repository.markBloodPressureAsDeleted(bloodPressure).blockingAwait()
+
+    val timeAtWhichBpWasDeleted = now.plus(durationToAdvanceBy)
+    val expected = bloodPressure.copy(
+        updatedAt = timeAtWhichBpWasDeleted,
+        deletedAt = timeAtWhichBpWasDeleted,
+        syncStatus = SyncStatus.PENDING)
+
+    val savedBloodPressure = appDatabase.bloodPressureDao().getOne(expected.uuid)!!
+
+    assertThat(savedBloodPressure).isEqualTo(expected)
   }
 }
--- a/app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
@@ -17,6 +17,7 @@ import org.threeten.bp.Duration
 import org.threeten.bp.Instant
 import org.threeten.bp.temporal.ChronoUnit
 import java.util.UUID
+import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 
 @RunWith(AndroidJUnit4::class)
@@ -132,5 +133,42 @@ class BloodPressureRepositoryAndroidTest {
     val savedBloodPressure = appDatabase.bloodPressureDao().getOne(expected.uuid)!!
 
     assertThat(savedBloodPressure).isEqualTo(expected)
+  }
+
+  @Test
+  fun observing_a_deleted_blood_pressure_as_a_stream_should_work_correctly() {
+    val now = Instant.now(clock)
+    val bloodPressure = testData.bloodPressureMeasurement(
+        createdAt = now,
+        updatedAt = now,
+        deletedAt = null,
+        syncStatus = SyncStatus.DONE)
+    appDatabase.bloodPressureDao().save(listOf(bloodPressure))
+
+    val deletedMeasurementObserver = repository
+        .deletedMeasurements(bloodPressure.uuid)
+        .test()
+
+    val fifteenSeconds = Duration.ofSeconds(15L)
+    val tenHours = Duration.ofHours(10L)
+    val twentyDays = Duration.ofDays(20L)
+
+    listOf(fifteenSeconds, tenHours, twentyDays)
+        .map { now.plus(it) }
+        .map { bloodPressure.copy(deletedAt = it, updatedAt = it) }
+        .forEach { appDatabase.bloodPressureDao().save(listOf(it)) }
+
+    // We cannot verify we received the values exactly because Room does not always emit all the
+    // intermediate changes. So we'll either have to add a delay between each update to guarantee
+    // that the change notification will be published, or just verify that the last event received
+    // is what we expect.
+    deletedMeasurementObserver.await(1L, TimeUnit.SECONDS)
+    val receivedValues = deletedMeasurementObserver.values()
+
+    assertThat(receivedValues.size).isAtLeast(1)
+    assertThat(receivedValues.last())
+        .isEqualTo(bloodPressure.copy(deletedAt = now.plus(twentyDays), updatedAt = now.plus(twentyDays)))
+
+    deletedMeasurementObserver.dispose()
   }
 }
--- a/app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
@@ -17,7 +17,6 @@ import org.threeten.bp.Duration
 import org.threeten.bp.Instant
 import org.threeten.bp.temporal.ChronoUnit
 import java.util.UUID
-import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 
 @RunWith(AndroidJUnit4::class)
@@ -133,42 +132,5 @@ class BloodPressureRepositoryAndroidTest {
     val savedBloodPressure = appDatabase.bloodPressureDao().getOne(expected.uuid)!!
 
     assertThat(savedBloodPressure).isEqualTo(expected)
-  }
-
-  @Test
-  fun observing_a_deleted_blood_pressure_as_a_stream_should_work_correctly() {
-    val now = Instant.now(clock)
-    val bloodPressure = testData.bloodPressureMeasurement(
-        createdAt = now,
-        updatedAt = now,
-        deletedAt = null,
-        syncStatus = SyncStatus.DONE)
-    appDatabase.bloodPressureDao().save(listOf(bloodPressure))
-
-    val deletedMeasurementObserver = repository
-        .deletedMeasurements(bloodPressure.uuid)
-        .test()
-
-    val fifteenSeconds = Duration.ofSeconds(15L)
-    val tenHours = Duration.ofHours(10L)
-    val twentyDays = Duration.ofDays(20L)
-
-    listOf(fifteenSeconds, tenHours, twentyDays)
-        .map { now.plus(it) }
-        .map { bloodPressure.copy(deletedAt = it, updatedAt = it) }
-        .forEach { appDatabase.bloodPressureDao().save(listOf(it)) }
-
-    // We cannot verify we received the values exactly because Room does not always emit all the
-    // intermediate changes. So we'll either have to add a delay between each update to guarantee
-    // that the change notification will be published, or just verify that the last event received
-    // is what we expect.
-    deletedMeasurementObserver.await(1L, TimeUnit.SECONDS)
-    val receivedValues = deletedMeasurementObserver.values()
-
-    assertThat(receivedValues.size).isAtLeast(1)
-    assertThat(receivedValues.last())
-        .isEqualTo(bloodPressure.copy(deletedAt = now.plus(twentyDays), updatedAt = now.plus(twentyDays)))
-
-    deletedMeasurementObserver.dispose()
   }
 }
--- a/app/src/main/java/org/simple/clinic/bp/BloodPressureMeasurement.kt
+++ b/app/src/main/java/org/simple/clinic/bp/BloodPressureMeasurement.kt
@@ -71,6 +71,9 @@ data class BloodPressureMeasurement (
     @Query("SELECT * FROM bloodpressuremeasurement WHERE uuid = :uuid LIMIT 1")
     fun getOne(uuid: UUID): BloodPressureMeasurement?
 
+    @Query("SELECT * FROM bloodpressuremeasurement WHERE uuid = :uuid AND deletedAt IS NOT NULL")
+    fun getOneWhichIsDeleted(uuid: UUID): Flowable<BloodPressureMeasurement>
+
     @Query("SELECT COUNT(*) FROM bloodpressuremeasurement")
     fun count(): Flowable<Int>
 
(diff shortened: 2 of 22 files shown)


app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
```diff
@@ -115,4 +115,14 @@ class BloodPressureRepository @Inject constructor(
   }
 
   fun measurement(uuid: UUID): Single<BloodPressureMeasurement> = Single.fromCallable { dao.getOne(uuid) }
+
+  fun markBloodPressureAsDeleted(bloodPressureMeasurement: BloodPressureMeasurement): Completable {
+    return Completable.fromAction {
+      val deletedBloodPressureMeasurement = bloodPressureMeasurement.copy(
+          deletedAt = Instant.now(clock),
+          syncStatus = SyncStatus.PENDING)
```
saket: Can you please confirm with backend if we need to also bump the `updateAt` timestamp?

app/src/main/java/org/simple/clinic/bp/entry/confirmremovebloodpressure/ConfirmRemoveBloodPressureDialogController.kt
```diff
@@ -0,0 +1,43 @@
...
+        .replay()
+
+    return markBloodPressureAsDeleted(replayedEvents)
+  }
+
+  private fun markBloodPressureAsDeleted(events: Observable<UiEvent>): Observable<UiChange> {
+    val savedBloodPressureMeasurementStream = events
+        .ofType<ConfirmRemoveBloodPressureDialogCreated>()
+        .flatMapSingle { bloodPressureRepository.measurement(it.bloodPressureMeasurementUuid) }
+
```
saket: Feel free to use `Bp` instead of `BloodPressure` in your code. We're okay with using acronyms.

app/src/test/java/org/simple/clinic/bp/entry/confirmremovebloodpressure/ConfirmRemoveBloodPressureDialogControllerTest.kt
```diff
@@ -0,0 +1,49 @@
...
+  lateinit var bloodPressureRepository: BloodPressureRepository
+  lateinit var controller: ConfirmRemoveBloodPressureDialogController
+  lateinit var dialog: ConfirmRemoveBloodPressureDialog
+
+  val uiEvents = PublishSubject.create<UiEvent>()
+
+  @Before
+  fun setUp() {
+    bloodPressureRepository = mock()
+    dialog = mock()
```
saket: New pattern for creating mocks in `setUp()`?

app/src/main/java/org/simple/clinic/bp/entry/ConfirmRemoveBloodPressureDialog.kt
```diff
@@ -39,17 +53,60 @@ class ConfirmRemoveBloodPressureDialog : AppCompatDialogFragment() {
...
+  }
+
+  override fun onStart() {
+    super.onStart()
+    onStarts.onNext(Any())
+  }
+
+  override fun onDestroyView() {
+    super.onDestroyView()
+    screenDestroys.onNext(ScreenDestroyed())
```
saket: Wanna just use `RxView.detaches(view)` instead of maintaining this extra Subject?

app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
```diff
@@ -127,6 +128,41 @@ class BloodPressureRepositoryAndroidTest {
...
+    val now = Instant.now(clock)
+    val bloodPressure = testData.bloodPressureMeasurement(
+        createdAt = now,
+        updatedAt = now,
+        deletedAt = null,
+        syncStatus = SyncStatus.DONE)
+    appDatabase.bloodPressureDao().save(listOf(bloodPressure))
+
+    val deletedMeasurementObserver = repository.deletedMeasurementAsStream(bloodPressure.uuid)
+        .test()
```
saket: nitpick: can you please move `deletedMeasurementAsStream()` to the next line?

```kotlin
    val deletedMeasurementObserver = repository
        .deletedMeasurementAsStream(bloodPressure.uuid)
        .test()
```

saket: Fluent APIs look good when they are of >= 3 lines

app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
```diff
@@ -127,6 +128,41 @@ class BloodPressureRepositoryAndroidTest {
...
+
+    val fifteenSeconds = Duration.ofSeconds(15L)
+    val tenHours = Duration.ofHours(10L)
+    val twentyDays = Duration.ofDays(20L)
+
+    listOf(fifteenSeconds, tenHours, twentyDays)
+        .map { now.plus(it) }
+        .map { bloodPressure.copy(deletedAt = it) }
+        .forEach { appDatabase.bloodPressureDao().save(listOf(it)) }
```
saket: Wanna remove `forEach`?

```kotlin
    listOf(fifteenSeconds, tenHours, twentyDays)
        .map { now.plus(it) }
        .map { bloodPressure.copy(deletedAt = it) }
        .apply {
          appDatabase.bloodPressureDao().save(it)
        }
```

app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
```diff
@@ -115,4 +115,18 @@ class BloodPressureRepository @Inject constructor(
...
+
+      dao.save(listOf(deletedBloodPressureMeasurement))
+    }
+  }
+
+  fun deletedMeasurementAsStream(bloodPressureMeasurementUuid: UUID): Observable<BloodPressureMeasurement> {
```
saket: Most of our repository functions are streams, there is no need to mention it in the function name here.

Let's just use `deletedMeasurements()`?

app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
```diff
@@ -127,6 +128,41 @@ class BloodPressureRepositoryAndroidTest {
...
+
+    // We cannot verify we received the values exactly because Room does not always emit all the
+    // intermediate changes. So we'll either have to add a delay between each update to guarantee
+    // that the change notification will be published, or just verify that the last event received
+    // is what we expect.
+    deletedMeasurementObserver.await(1L, TimeUnit.SECONDS)
```
saket: Alternatively, can we do a blocking get which will give us all three items?

app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
```diff
@@ -127,4 +127,8 @@ class BloodPressureRepository @Inject constructor(
       dao.save(listOf(deletedBloodPressureMeasurement))
     }
   }
+
+  fun deletedMeasurementAsStream(bloodPressureMeasurementUuid: UUID): Observable<BloodPressureMeasurement> {
+    return dao.getOneWhichIsDeleted(bloodPressureMeasurementUuid).toObservable()
```
saket: Wow, `getOneWhichIsDeleted()` sounds like English. What do you think about just `getOneDeleted()`? :p

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -292,4 +294,17 @@ class BloodPressureEntrySheetController @Inject constructor(
         .withLatestFrom(bloodPressureMeasurementUuidStream) { _, bloodPressureMeasurementUuid -> bloodPressureMeasurementUuid }
         .map { uuid -> { ui: Ui -> ui.showConfirmRemoveBloodPressureDialog(uuid) } }
   }
+
+  private fun closeSheetWhenEditedBpIsDeleted(events: Observable<UiEvent>): Observable<UiChange>? {
+    val bloodPressureMeasurementUuidStream = events
+        .ofType<BloodPressureEntrySheetCreated>()
+        .map { it.openAs }
+        .filter { it is OpenAs.Update }
+        .map { (it as OpenAs.Update).bpUuid }
```
saket: Same `ofType<>()` question here

app/src/main/java/org/simple/clinic/bp/BloodPressureMeasurement.kt
```diff
@@ -71,6 +71,9 @@ data class BloodPressureMeasurement (
     @Query("SELECT * FROM bloodpressuremeasurement WHERE uuid = :uuid LIMIT 1")
     fun getOne(uuid: UUID): BloodPressureMeasurement?
 
+    @Query("SELECT * FROM bloodpressuremeasurement WHERE uuid = :uuid AND deletedAt IS NOT NULL")
+    fun getOneWhichIsDeleted(uuid: UUID): Flowable<BloodPressureMeasurement>
```
saket: Wait, the reason we use `getOne()` is to get around the fact that we have two functions for fetching BP measurements -- one immediately and one through Rx. The Rx ones aren't called `get…` like this one.

saket: Do you want to avoid checking `deletedAt` in sql and do it in code instead? It'll help us avoid the confusion with the naming because I have multiple feedback on the names in this PR.

```kotlin
@Query("SELECT * FROM bloodpressuremeasurement WHERE uuid = :uuid")
fun bp(uuid: UUID): Flowable<BloodPressureMeasurement>
```

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTest.kt
```diff
@@ -263,4 +266,35 @@ class BloodPressureEntrySheetControllerTest {
         listOf(OpenAs.Update(bpUuid), PatientMocker.bp(uuid = bpUuid, patientUuid = patientUuid))
     )
   }
+
+  /**
+  * **Note:** When removing the feature flag, this test should **NOT** be removed, but modified
+   * to test that the button is shown only when the OpenAs is [OpenAs.Update].
```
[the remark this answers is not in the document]
vinaysshenoy: I've added a test that hardcodes the feature flag to `true`. This is essentially a duplicate of this test, but this one will get removed when we lift the feature flag.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -278,4 +278,18 @@ class BloodPressureEntrySheetController @Inject constructor(
 
     return showEnterBloodPressureTitle.mergeWith(showEditBloodPressureTitle)
   }
+
+  private fun showConfirmRemoveBloodPressureDialog(events: Observable<UiEvent>): Observable<out UiChange>? {
+    val bloodPressureMeasurementUuidStream = events
+        .ofType<BloodPressureEntrySheetCreated>()
+        .map { it.openAs }
+        .filter { it is OpenAs.Update }
+        .map { (it as OpenAs.Update).bpUuid }
```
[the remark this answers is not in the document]
vinaysshenoy: Oh yes! I'll do this.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -278,4 +278,18 @@ class BloodPressureEntrySheetController @Inject constructor(
...
+
+    val removeClicks = events.ofType<BloodPressureRemoveClicked>()
+
+    return removeClicks
+        .withLatestFrom(bloodPressureMeasurementUuidStream) { _, bloodPressureMeasurementUuid -> bloodPressureMeasurementUuid }
+        .map { uuid -> { ui: Ui -> ui.showConfirmRemoveBloodPressureDialog(uuid) } }
```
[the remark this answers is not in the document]
vinaysshenoy: Yeah, I'll do this.

app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
vinaysshenoy: Good catch; we are supposed to do it. I have fixed the PR.

app/src/main/java/org/simple/clinic/bp/entry/confirmremovebloodpressure/ConfirmRemoveBloodPressureDialogController.kt
vinaysshenoy: It's alright, I'm not a huge fan of acronyms. 😌

app/src/test/java/org/simple/clinic/bp/entry/confirmremovebloodpressure/ConfirmRemoveBloodPressureDialogControllerTest.kt
vinaysshenoy: As in? I thought we'd agreed to create mocks in `setUp` because they don't automatically get reset after every test.

app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
vinaysshenoy: We really should get an autoformatter setup sometime.

vinaysshenoy: Fixed.

app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
vinaysshenoy: Techincally, this is the wrong usage of `.apply` since that is meant to make things like setting parameters on an object easier.

But I always want to simulate the blood pressure emitting multiple change notifications so I have to save them individually instead of as a single list.

app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
vinaysshenoy: Done.

app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
vinaysshenoy: We can't use a `blocking` call here because the source never completes and remains alive until it is manually disposed.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
vinaysshenoy: Changed!

app/src/main/java/org/simple/clinic/bp/BloodPressureMeasurement.kt
vinaysshenoy: Good idea, I'll make this change.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -7,6 +7,7 @@ import android.content.Intent
 import android.os.Bundle
```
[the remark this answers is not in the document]
vinaysshenoy: Done

app/src/main/java/org/simple/clinic/bp/entry/confirmremovebloodpressure/ConfirmRemoveBloodPressureDialogController.kt
```diff
@@ -27,13 +27,14 @@ class ConfirmRemoveBloodPressureDialogController @Inject constructor(
   }
 
   private fun markBloodPressureAsDeleted(events: Observable<UiEvent>): Observable<UiChange> {
-    val savedBloodPressureMeasurementStream = events
+    val savedBloodPressureMeasurementUuidStream = events
         .ofType<ConfirmRemoveBloodPressureDialogCreated>()
-        .flatMapSingle { bloodPressureRepository.measurement(it.bloodPressureMeasurementUuid) }
```
vinaysshenoy: FYI: This was done because we use a `withLatestFrom` on this stream later on which subscribes to this immediately which causes the database operation to execute on the main thread. This is probably the same reason why we have seen this happening in some other screens, which we'll have to investigate later.

vinaysshenoy:
@saket PR feedback addressed.

app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
saket: What do you think of doing this?

```kotlin
    val deletedMeasurements = repository
        .deletedMeasurements(bloodPressure.uuid)
        .blockingFirst()
```

app/src/main/java/org/simple/clinic/bp/entry/confirmremovebloodpressure/ConfirmRemoveBloodPressureDialogController.kt
saket: 👍

app/src/androidTest/java/org/simple/clinic/bp/BloodPressureRepositoryAndroidTest.kt
vinaysshenoy: This looks good actually. I'll switch to this.

vinaysshenoy: The actual problem with this test is that it is no longer essential for it to be an android test since  I moved the logic of filtering deleted blood pressures to the repository. I'll make this a JVM test instead and control the emissions myself to test it instead of working around Room's problems.

saket: Even better!

saket: Even better!

app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
```diff
@@ -129,6 +129,9 @@ class BloodPressureRepository @Inject constructor(
   }
 
   fun deletedMeasurements(bloodPressureMeasurementUuid: UUID): Observable<BloodPressureMeasurement> {
-    return dao.getOneWhichIsDeleted(bloodPressureMeasurementUuid).toObservable()
+    return dao
+        .bloodPressure(bloodPressureMeasurementUuid)
+        .filter { it.deletedAt != null }
+        .toObservable()
   }
```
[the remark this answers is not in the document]
saket: We discussed in person that we already have a `measurement(UUID)` function that the controller can use and filter deleted measurements.

saket:
@vinaysshenoy The `Edit blood pressure` text may overlap with `Remove`. Can you also improve this in your next PR?

PR #435, opened 2019-01-02
Exclude deleted blood pressure measurements from certain queries

vinaysshenoy:


vinaysshenoy:
Closing because it's a part of #439

PR #436, opened 2019-01-03
Allow BPs to be editable irrespective of the recording date

sanchita:


app/src/main/java/org/simple/clinic/summary/SummaryBloodPressureListItem.kt
```diff
@@ -59,19 +58,14 @@ data class SummaryBloodPressureListItem(
...
-          .append(resources.getString(R.string.patientsummary_bp_edit))
-          .popSpan()
-          .append(" ${Unicode.bullet} ")
-          .append(timestamp.displayText(context))
-          .build()
-
-    } else {
-      timestamp.displayText(context)
-    }
+    val colorSpanForEditLabel = ForegroundColorSpan(ResourcesCompat.getColor(resources, R.color.patientsummary_edit_label_color, context.theme))
```
vinaysshenoy: We no longer need to do this programmatically, correct? The reason this was being done programmatically earlier is because the `Edit` label (and the bullet) were bring conditionally shown. Now since they are always visible, we can move them to the XML and statically render them.

sanchita: I think this is being done since there are two different colors being used within the same text view i.e `holder.timestampTextView`.
"Edit" is of a different color than the `timestamp`

PR #437, opened 2019-01-03
Ignore patients without phone numbers when showing update phone dialog

saket:


vinaysshenoy:
> @vinaysshenoy I'm unable to write a failing test for this bug. Help me out tomorrow?

On second thought, do we really need to write a failing test for this. If this had been caught during review, we'd have just changed the operator and left it at that. I feel this can be merged in as is.

PR #438, opened 2019-01-04
Catch uncaught RxJava errors in unit tests

saket:
Starting from RxJava2, uncaught errors are not thrown but instead sent to the thread's default error handler. This results in a situation where JUnit would always pass even if any RxJava chain encounters an error.

More information here: https://github.com/ReactiveX/RxJava/issues/5234

This commit introduces a JUnit rule that observes uncaught errors, but it needs to be manually added to every test. The rule was copied from [AutoDispose](https://github.com/uber/AutoDispose).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -146,7 +146,7 @@ android {
 dependencies {
   implementation project(path: ':router')
 
-  testImplementation "junit:junit:$versions.junit"
+  implementation "junit:junit:$versions.junit"
   testImplementation "org.mockito:mockito-core:$versions.mockito"
   testImplementation "com.nhaarman:mockito-kotlin:$versions.mockitoKotlin"
   testImplementation "pl.pragmatists:JUnitParams:$versions.junitParams"
--- a/app/src/androidTest/java/org/simple/clinic/MigrationAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/MigrationAndroidTest.kt
@@ -39,12 +39,10 @@ import java.util.UUID
 @RunWith(AndroidJUnit4::class)
 class MigrationAndroidTest {
 
-  @Rule
-  @JvmField
+  @get:Rule
   val helper = MigrationTestHelperWithForeignConstraints()
 
-  @Rule
-  @JvmField
+  @get:Rule
   val expectedException = ExpectedException.none()
 
   @Test
--- a/app/src/androidTest/java/org/simple/clinic/drugs/sync/PrescriptionSyncAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/drugs/sync/PrescriptionSyncAndroidTest.kt
@@ -7,7 +7,6 @@ import org.junit.Rule
 import org.junit.rules.RuleChain
 import org.junit.runner.RunWith
 import org.simple.clinic.AuthenticationRule
-import org.simple.clinic.sync.RegisterPatientRule
 import org.simple.clinic.TestClinicApp
 import org.simple.clinic.TestData
 import org.simple.clinic.drugs.PrescribedDrug
@@ -15,9 +14,11 @@ import org.simple.clinic.drugs.PrescriptionRepository
 import org.simple.clinic.facility.FacilityRepository
 import org.simple.clinic.patient.SyncStatus
 import org.simple.clinic.sync.BaseSyncCoordinatorAndroidTest
+import org.simple.clinic.sync.RegisterPatientRule
 import org.simple.clinic.user.User
 import org.simple.clinic.user.UserSession
 import org.simple.clinic.util.Optional
+import org.simple.clinic.util.RxErrorsRule
 import java.util.UUID
 import javax.inject.Inject
 import javax.inject.Named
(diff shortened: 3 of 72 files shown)


app/src/test/java/org/simple/clinic/util/RxErrorsRule.kt
```diff
@@ -0,0 +1,53 @@
+package org.simple.clinic.util
```
vinaysshenoy: Do we not need to add this to any of the instrumented tests?

vinaysshenoy: I also found a couple of test classes where this should have been added, but wasn't.

1. `BruteForceProtectionTest`
2. `RecyclerViewUserScrollDetectorTest`

Did you intentionally not add this rule to this classes?

saket: Yes, we should. I've added the rule to all instrumentation tests. Please check if the rule chains are setup correctly?

saket: No, I think I missed these two classes. Fixing this.

app/build.gradle
```diff
@@ -146,7 +146,7 @@ android {
 dependencies {
   implementation project(path: ':router')
 
-  testImplementation "junit:junit:$versions.junit"
+  implementation "junit:junit:$versions.junit"
```
vinaysshenoy: Should we change this to `debugImplementation` instead?

app/src/androidTest/java/org/simple/clinic/MigrationAndroidTest.kt
```diff
@@ -43,9 +45,14 @@ class MigrationAndroidTest {
...
-  @Rule
-  @JvmField
-  val expectedException = ExpectedException.none()
+  private val rxErrorsRule = RxErrorsRule()
+
+  private val expectedException = ExpectedException.none()
+
+  @get:Rule
+  val ruleChain = RuleChain
+      .outerRule(rxErrorsRule)
```
vinaysshenoy: There's no RxJava usage in this class, right? Using this rule here is unnecessary.

app/src/androidTest/java/org/simple/clinic/drugs/sync/PrescriptionSyncAndroidTest.kt
```diff
@@ -53,13 +54,16 @@ class PrescriptionSyncAndroidTest : BaseSyncCoordinatorAndroidTest<PrescribedDru
...
+
+  private val registerPatientRule = RegisterPatientRule(UUID.randomUUID())
 
-  val registerPatientRule = RegisterPatientRule(UUID.randomUUID())
+  private val rxErrorsRule = RxErrorsRule()
 
   @get:Rule
   val ruleChain = RuleChain
       .outerRule(authenticationRule)
+      .around(rxErrorsRule)
```
vinaysshenoy: Should we make the `rxErrors` rule always the innermost one in the rule chain? It seems odd to break `authenticationRule` and `registerPatientRule`.

app/src/androidTest/java/org/simple/clinic/medicalhistory/MedicalHistorySyncAndroidTest.kt
```diff
@@ -44,11 +45,14 @@ class MedicalHistorySyncAndroidTest : BaseSyncCoordinatorAndroidTest<MedicalHist
...
   val authenticationRule = AuthenticationRule()
 
+  val rxErrorsRule = RxErrorsRule()
+
   val registerPatientRule = RegisterPatientRule(patientUuid = UUID.randomUUID())
 
   @get:Rule
   val ruleChain = RuleChain
       .outerRule(authenticationRule)
+      .around(rxErrorsRule)
```
vinaysshenoy: Same as earlier comment about `rxErrorsRule` being the innermost rule.

app/src/test/java/org/simple/clinic/util/RxErrorsRule.kt
vinaysshenoy: Done, and left feedback.

app/build.gradle
saket: Done

app/src/androidTest/java/org/simple/clinic/MigrationAndroidTest.kt
saket: Fixed

app/src/androidTest/java/org/simple/clinic/drugs/sync/PrescriptionSyncAndroidTest.kt
saket: Agreed. Fixing this.

app/src/androidTest/java/org/simple/clinic/medicalhistory/MedicalHistorySyncAndroidTest.kt
saket: Same answer as earlier comment about `rxErrorsRule` being the innermost rule.

saket:
@vinaysshenoy feedback addressed.

PR #441, opened 2019-01-07
Add LeakCanary +  Extract syncing of data from SyncWorker into DataSync

saket:


app/src/main/java/org/simple/clinic/sync/DataSync.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.simple.clinic.patient.sync.PatientSync
+import org.simple.clinic.protocol.sync.ProtocolSync
+import org.simple.clinic.user.UserSession
+import org.simple.clinic.util.ErrorResolver
+import org.simple.clinic.util.ResolvedError
+import org.simple.clinic.util.exhaustive
+import timber.log.Timber
+import javax.inject.Inject
+
+class DataSync @Inject constructor(
```
vinaysshenoy: Should this be `@AppScope` scoped?

saket: It's not maintaining any mutable state, so I don't see a reason for retaining this in memory throughout the app's life.

saket: Hrm, I don't think so. It's not maintaining any mutable state, so I don't see a reason for retaining this in memory throughout the app's life.

vinaysshenoy: 👍

PR #443, opened 2019-01-08
New Dosage Bottom Sheet for Prescriptions

sanchita:
This is the first part of the story to add the New Prescription Screen.

This PR only adds a new bottom sheet activity to show a list of dosages.
Click events on selecting a dosage will be delivered in the second part as well. 

Since this sheet will open when a drug is selected from the new prescription Screen (separate PR), hence to see this bottom sheet appear replace the following code in `PrescribedDrugsScreen`:

`fun showNewPrescriptionEntrySheet(patientUuid: UUID) {
    activity.startActivity(CustomPrescriptionEntrySheet.intent(context, patientUuid))
  }`

with 

`fun showNewPrescriptionEntrySheet(patientUuid: UUID) {
    activity.startActivity(PrescribedDrugWithDosagesSheet.intent(context, "Amlodipine"))
  }`


app/src/androidTest/java/org/simple/clinic/protocolv2/ProtocolRepositoryAndroidTest.kt
```diff
@@ -147,4 +147,36 @@ class ProtocolRepositoryAndroidTest {
...
+  @Test
+  fun when_fetching_dosages_for_a_drug_and_protocol_id_is_null_default_dosages_should_be_returned() {
+    val protocol = testData.protocol()
+    database.protocolDao().save(listOf(protocol))
+
+    val amlodipine20mg = testData.protocolDrug(name = "Amlodipine", dosage = "20mg", protocolUuid = protocol.uuid)
+    database.protocolDrugDao().save(listOf(amlodipine20mg))
+
+    val dosages = protocolRepository.dosagesForDrug(drugName = "Amlodipine", protocolUuid = null).blockingFirst()
+    assertThat(dosages).containsAllOf("5mg", "10mg")
```
saket: The default drugs are hardcoded in this test function. We're coupling this test with the implementation. Considering that `defaultProtocolDrugs()` is public for tests, let's use that instead?

app/src/androidTest/java/org/simple/clinic/protocolv2/ProtocolRepositoryAndroidTest.kt
```diff
@@ -147,4 +147,36 @@ class ProtocolRepositoryAndroidTest {
...
+    val telmisartan40mg = testData.protocolDrug(name = "Telmisartan", dosage = "40mg", protocolUuid = protocol1.uuid)
+    val telmisartan80mg = testData.protocolDrug(name = "Telmisartan", dosage = "80mg", protocolUuid = protocol2.uuid)
+    database.protocolDrugDao().save(listOf(amlodipine5mg, amlodipine10mg, amlodipine20mg, telmisartan40mg, telmisartan80mg))
+
+    val dosages = protocolRepository.dosagesForDrug(drugName = "Amlodipine", protocolUuid = protocol1.uuid).blockingFirst()
+    assertThat(dosages).containsAllOf("5mg", "10mg")
+  }
+
+  @Test
+  fun when_fetching_dosages_for_a_drug_and_protocol_id_is_null_default_dosages_should_be_returned() {
```
saket: Protocol ID could be non-null, but we may not have any drugs for that protocol. Do we still need to send default drugs?

saket: nitcpik: can we move `defaultProtocolDrugs()` after this function?

app/src/main/java/org/simple/clinic/protocol/ProtocolRepository.kt
```diff
@@ -163,4 +163,28 @@ class ProtocolRepository @Inject constructor(
     )
   }
 
+  fun dosagesForDrug(drugName: String, protocolUuid: UUID?): Observable<List<String>> {
```
saket: Just like how we've suffixed other functions in this class with `orDefault()`, can we do the same thing here?

app/src/main/java/org/simple/clinic/protocol/ProtocolRepository.kt
```diff
@@ -163,4 +163,28 @@ class ProtocolRepository @Inject constructor(
...
+        .dosagesForDrug(drugName, protocolUuid)
+        .replay()
+        .refCount()
+
+    val notEmptyDosages = dosages
+        .filter { it.isNotEmpty() }
+
+    val emptyDosages = dosages
+        .filter { it.isEmpty() }
+        .map { defaultDosages }
```
saket: Just like how you've used the names `syncedDrugs` and `defaultDrugs`, can we do the same thing here? That is, use `syncedDosages` and `defaultDosages`?

saket: Additionally, the name here does not indicate the data, it indicates what the filter operator is

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugWithDosagesEvents.kt
```diff
@@ -0,0 +1,7 @@
+package org.simple.clinic.drugs.selectionv2
+
+import org.simple.clinic.widgets.UiEvent
+
+data class PrescribedDrugsWithDosagesSheetCreated(val drugName: String) : UiEvent
+
+object ScreenDestroyed: UiEvent
```
saket: We already have a `ScreenDestroyed` object (org.simple.clinic.widgets). Can you use that?

If you don't use that event, then your controller stream will never get disposed. This is a hidden behavior, but unfortunately `ReplayUntilScreenIsDestroyed` depends on the other event to dispose itself. We should fix this in the future.

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugWithDosagesSheet.kt
```diff
@@ -23,6 +37,44 @@ class PrescribedDrugWithDosagesSheet : BottomSheetActivity() {
 
     recyclerView.adapter = adapter
     recyclerView.layoutManager = LinearLayoutManager(this)
+    displayDrugName()
+
+    Observable.mergeArray(sheetCreates())
+        .observeOn(io())
+        .compose(controller)
+        .observeOn(AndroidSchedulers.mainThread())
```
saket: `AndroidSchedulers.mainThread()` -> `mainThread()`?

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugWithDosagesSheet.kt
```diff
@@ -23,6 +37,44 @@ class PrescribedDrugWithDosagesSheet : BottomSheetActivity() {
...
+  }
+
+  private fun sheetCreates(): Observable<UiEvent> {
+    val drugName = intent.getStringExtra(KEY_DRUG_NAME)
+    return Observable.just(PrescribedDrugsWithDosagesSheetCreated(drugName))
+  }
+
+  private fun displayDrugName() {
+    val drugName = intent.getStringExtra(KEY_DRUG_NAME)
+    if (drugName.isNullOrEmpty().not()) {
```
saket: You can use `drugName.isNotBlank()` instead here. :)

saket: Wait, why is this condition even present? Can the name ever be empty? In that case, will we hide the name TextView altogether? 

If you find out that the name can be empty then let's move the logic of hiding the title or using a different title to the controller.

app/src/main/java/org/simple/clinic/drugs/DosageType.kt
```diff
@@ -0,0 +1,6 @@
+package org.simple.clinic.drugs
+
+sealed class DosageType {
+  data class Dosage(val dosage: String) : DosageType()
+  object None: DosageType()
+}
```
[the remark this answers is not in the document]
saket: Update: I'm not very convinced by `DosageType` because it's not very descriptive of where it's being used. What do you think of `DosagePickerOption`?

app/src/test/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugWithDosagesSheetControllerTest.kt
```diff
@@ -0,0 +1,61 @@
...
+class PrescribedDrugWithDosagesSheetControllerTest {
+
+  private val protocolRepository = mock<ProtocolRepository>()
+  private val userSession = mock<UserSession>()
+  private val screen = mock<PrescribedDrugWithDosagesSheet>()
+  private val facilityRepository = mock<FacilityRepository>()
+
+  private val uiEvents = PublishSubject.create<UiEvent>()
+  private lateinit var controller: PrescribedDrugWithDosagesSheetController
+
```
saket: I just realized that this test is missing RxErrorsRule

app/src/androidTest/java/org/simple/clinic/protocolv2/ProtocolRepositoryAndroidTest.kt
sanchita: I will have to check this with Deepa.

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDosageAdapter.kt
```diff
@@ -0,0 +1,64 @@
...
+  override fun onBindViewHolder(holder: PrescribedDosageViewHolder, position: Int) {
+    holder.dosageItem = getItem(position)
+    holder.render()
+  }
+}
+
+class PrescribedDosageViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
+
+  private val dosageTextView by bindView<TextView>(R.id.prescribed_drug_list_item_dosage_name)
+  private val divider by bindView<View>(R.id.prescribed_drug_list_item_divider)
```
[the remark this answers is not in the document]
sanchita: That way every UI element is a `View`. I did not suffix this one because of that. It is understood.
Also, it is being used as both a `divider` and a `dividerView` elsewhere.

app/src/main/java/org/simple/clinic/protocol/ProtocolRepository.kt
sanchita: Adding `orDefault` here can be misleading.
We are not going to return default dosages here. There is no concept of default dosages.

We are just checking that if `protocolUuid` is `null` that means the drugs shown are default.
Hence, we just want to get the dosages for those drugs from the default list.

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugWithDosagesSheet.kt
sanchita: You are right. `drugName` cannot be empty.

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDosageAdapter.kt
```diff
@@ -0,0 +1,64 @@
...
+      }
+      is DosageType.Dosage -> {
+        dosageTextView.text = dosageType.dosage
+        divider.visibility = View.VISIBLE
+      }
+    }
+  }
+}
+
+data class PrescribedDosageListItem(val dosageType: DosageType)
```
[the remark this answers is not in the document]
sanchita: I do not understand this. Can you please elaborate?

PR #446, opened 2019-01-09
Exclude deleted blood pressures (Part 2)

vinaysshenoy:
https://www.pivotaltracker.com/story/show/163046626

# Changes
- Excluded deleted blood pressures from overdue appointments.
- Removed feature flag for toggling the "Remove Blood Pressure" feature.

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
@@ -34,6 +34,7 @@ import org.simple.clinic.patient.Patient
 import org.simple.clinic.patient.PatientAddress
 import org.simple.clinic.patient.PatientPhoneNumber
 import org.simple.clinic.patient.PatientPhoneNumberType
+import org.simple.clinic.patient.PatientProfile
 import org.simple.clinic.patient.PatientRepository
 import org.simple.clinic.patient.PatientStatus
 import org.simple.clinic.patient.SyncStatus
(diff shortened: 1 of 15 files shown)


saket:
@vinaysshenoy I'll review this today

vinaysshenoy:
@saket I've fixed the failing tests.

vinaysshenoy:
> We are not handling the case where all BPs are deleted or no BPs are added for a patient.
> We still continue to show the `ScheduleAppointmentSheet` in these cases.
> As discussed before, we should not show the sheet if this happens.

@sanchita Done.

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
```diff
@@ -417,6 +416,55 @@ class PatientSummaryScreenControllerTest {
...
 
+  @Test
+  @Parameters(value = ["SEARCH", "NEW_PATIENT"])
+  fun `when all BPs for the patient are deleted and back is clicked, the schedule appointment sheet must not be shown`(
+      patientSummaryCaller: PatientSummaryCaller
+  ) {
+    whenever(bpRepository.bloodPressureCount(patientUuid)).thenReturn(Observable.just(0))
+
+    uiEvents.onNext(PatientSummaryScreenCreated(patientUuid, caller = patientSummaryCaller, screenCreatedTimestamp = Instant.now(clock)))
+    uiEvents.onNext(PatientSummaryBloodPressureClosed(true))
```
sanchita: Can we also pass `wasBloodPressureSaved` in the params?

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
```diff
@@ -417,6 +416,55 @@ class PatientSummaryScreenControllerTest {
...
+
+  @Test
+  @Parameters(value = ["SEARCH", "NEW_PATIENT"])
+  fun `when all BPs for the patient are deleted and done is clicked, the schedule appointment sheet must not be shown`(
+      patientSummaryCaller: PatientSummaryCaller
+  ) {
+    whenever(bpRepository.bloodPressureCount(patientUuid)).thenReturn(Observable.just(0))
+
+    uiEvents.onNext(PatientSummaryScreenCreated(patientUuid, caller = patientSummaryCaller, screenCreatedTimestamp = Instant.now(clock)))
+    uiEvents.onNext(PatientSummaryBloodPressureClosed(true))
```
sanchita: Same as above.

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
vinaysshenoy: Done.

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/bp/BloodPressureMeasurement.kt
```diff
@@ -77,6 +77,13 @@ data class BloodPressureMeasurement (
     @Query("SELECT COUNT(uuid) FROM bloodpressuremeasurement")
     fun count(): Flowable<Int>
 
+    @Query("""
+      SELECT COUNT(uuid)
+      FROM bloodpressuremeasurement
+      WHERE patientUuid = :patientUuid AND deletedAt IS NULL
+    """)
+    fun recordedBloodPressureCountForPatient(patientUuid: UUID): Flowable<Int>
```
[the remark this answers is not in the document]
vinaysshenoy: When we "delete" a BP, we are basically indicating that the BP does not exist and it's an implementation detail that for us, deleting a BP means setting a property (`deletedAt`) on a record. This implementation detail should not be exposed externally in any way. For instance, we wouldn't be talking about renaming this method if instead of setting a property, we were actually deleting the record.

The assumption is that all data access methods will exclude deleted records unless and until we specifically need to access deleted records (say, for sync) and these methods should be named appropriately. This is what we have done so far for all our models.

vinaysshenoy:
@sanchita feedback has been addressed.

app/src/main/java/org/simple/clinic/bp/BloodPressureMeasurement.kt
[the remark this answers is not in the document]
sanchita: Got it!

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -292,7 +294,14 @@ class PatientSummaryScreenController @Inject constructor(
         .ofType<PatientSummaryRestoredWithBPSaved>()
         .map { it.wasBloodPressureSaved }
 
-    val mergedBpSaves = Observable.merge(bloodPressureSaves, bloodPressureSaveRestores)
+    val allBpsForPatientDeleted = patientUuids
+        .switchMap(bpRepository::bloodPressureCount)
+        .map { recordedBpCount -> recordedBpCount == 0 }
+
+    val shouldShowAppointmentSheet = Observable.merge(bloodPressureSaves, bloodPressureSaveRestores)
```
sanchita: There is a bug in this logic here.
If the stream `Observable.merge(bloodPressureSaves, bloodPressureSaveRestores)` does not emit at all then this entire chain will become inactive because you are using `withLatestFrom`. This will lead to save and back clicks not being handled at all.

To reproduce this case:
- Open a patient. 
- Open and close the BP sheet without making any changes.
Now if you try to press `Save` or `Back`, it won't work.
 
I tried to fix this but this chain will have to be rewritten.

saket: 🥇

vinaysshenoy: Alright, I took a look at it and trying to fix the current stream seems a little complex. But we anyway have to make it so that the appointment sheet has to be shown always, regardless of whether the BP was saved or not.

I'll make this change in this PR and remove this problem in one stroke.

vinaysshenoy: Never mind, I found an easier fix. Pushing the fix now.

saket:
@vinaysshenoy I'll test this today post lunch

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
```diff
@@ -715,10 +766,14 @@ class PatientSummaryScreenControllerTest {
   }
 
   @Test
-  fun `when screen is opened and a canceled appointment with the patient does not exist then update phone dialog should not be shown`() {
```
saket: It took me a LOT of time to realize that you changed the ordering of these tests. The diff shows that many random changes were made to many functions and I was trying very hard to wrap my head around this.

vinaysshenoy: I actually just moved the `appointment cancelation reasons()` params provider method right below the test method where it was being used.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -292,7 +294,16 @@ class PatientSummaryScreenController @Inject constructor(
...
 
-    val mergedBpSaves = Observable.merge(bloodPressureSaves, bloodPressureSaveRestores)
+    val mergedBloodPressureSaves = bloodPressureSaves.mergeWith(bloodPressureSaveRestores)
+
+    val allBpsForPatientDeleted = patientUuids
+        .switchMap(bpRepository::bloodPressureCount)
+        .map { recordedBpCount -> recordedBpCount == 0 }
+
+    val shouldShowAppointmentSheet = Observables
+        .combineLatest(mergedBloodPressureSaves, allBpsForPatientDeleted) { wasBloodPressureSaved, recordedBpsAreDeleted ->
```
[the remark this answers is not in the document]
vinaysshenoy: This chain will be removed in the future when we show the appointment sheet all the time, so I won't make this change.

See https://www.pivotaltracker.com/story/show/163080975

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -338,6 +339,107 @@ class AppointmentRepositoryAndroidTest {
...
+      return testData.patientProfile(generatePhoneNumber = true)
+          .let { patientProfile ->
+            patientProfile.copy(patient = patientProfile.patient.copy(fullName = fullName))
+          }
+    }
+
+    val patients = listOf(
+        createPatient(fullName = "No BPs are deleted"),
+        createPatient(fullName = "Latest BP is deleted"),
+        createPatient(fullName = "Oldest BP is not deleted"),
```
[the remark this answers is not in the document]
vinaysshenoy: Yup, I'll do this in a later PR.

PR #449, opened 2019-01-15
Create user states document

vinaysshenoy:
https://www.pivotaltracker.com/story/show/163079776

This maintains a matrix of all the possible states a user can be in, as well as how they can get there and some extra information about how the application should behave when the user is in a particular state.


vinaysshenoy:
The markdown version of this is unreadable, but if you open the whole file, it should render the table properly on the GitHub viewer.

User States.md
```diff
@@ -0,0 +1,19 @@
+# User States
+
+| Logged In Status (Local to App) | User Status (Server and App) | Can Sync Data | Can Register Patients | How does the user get to this state                                                                                                                                                                                                                                                                     | Will Access Token Be Present | Notification to show on Home Screen | Initial screen to show when opened | Should Show Lock Screen |
```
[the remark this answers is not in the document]
vinaysshenoy: Done.

pratul:
Why hasn't this been merged yet?

User States.md
```diff
@@ -0,0 +1,19 @@
...
+|---------------------------------|------------------------------|---------------|-----------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------|-------------------------------------|------------------------------------|-------------------------|
+| Not Logged In                   | Waiting For Approval         | No            | Yes                   | The user can be in this state in two ways:<br> 1. When they select a facility during registration and the user is stored locally before scheduling the background registration call<br> 2. When a user number match is found on the server during login, but the request OTP call has not yet been made | No                           | Waiting for approval                | Home                               | Yes                     |
```
saket: I was wondering if we should remove "The user will be in this state when…" prefix for every `How does the user get to this state`.  

> The user will be in this state when a phone number match has been found on the server during login, but the request OTP call has not been made yet

to

>A phone number match has been found on the server during login, but the request OTP call has not been made yet.

Rest everything else looks good to me.

vinaysshenoy: Done.

PR #447, opened 2019-01-09
Retro active BP entry (v1)

saket:
https://www.pivotaltracker.com/story/show/163052502

Large-ish PR because the new (and incomplete) date entry functionality is behind a feature flag and it involved duplicating a few classes.

PR #451, opened 2019-01-16
Retro active bp entry (v2)

saket:
https://www.pivotaltracker.com/n/projects/2184102/stories/163052502

- [ ] Avoid hardcoding date pattern
- [x] Explore having a super type for incorrect dates to leverage `Observable#ofType()`.
- [x] Tests for validate2()
- [ ] Find a better name for `BloodPressureDateValidated` or revert to storing result as constructor param
- [x] Consider adding a `BpValidated` event
- [x] Rename `DateOfBirthFormatValidator` -> `UserInputDateValidator`
- [ ] Add a Pivotal story for displaying granular date errors (e.g., Day cannot be greater than 28 for February, Month cannot be greater than 12, etc.)
- [ ] Copy changes made in https://github.com/simpledotorg/simple-android/pull/446/commits/fccc709d7145a4cee5b18e63cb64098cc7e29b0d

### DIFF
--- a/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
+++ b/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
@@ -10,6 +10,7 @@ import android.view.inputmethod.EditorInfo
 import android.widget.Button
 import android.widget.EditText
 import android.widget.TextView
+import android.widget.ViewFlipper
 import com.jakewharton.rxbinding2.view.RxView
 import com.jakewharton.rxbinding2.widget.RxTextView
 import io.reactivex.Observable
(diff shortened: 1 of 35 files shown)


app/src/main/res/layout/sheet_blood_pressure_entry.xml
```diff
@@ -79,6 +79,7 @@
           android:importantForAutofill="no"
```
saket: @rakshakhegde How is this commit different from https://github.com/simpledotorg/simple-android/pull/447/commits/ba4c01d82a5269897f46a7ca522b66e3c89a7057?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -157,6 +182,28 @@ class BloodPressureEntrySheet : BottomSheetActivity() {
     }
   }
 
+  private fun screenTypeChanges(): Observable<UiEvent> {
+    // RxViewFlipper
+    //     .displayedChildChanges
+    //     .map()
+    return Observable.just(BloodPressureScreenChanged(ScreenType.BP_ENTRY))
```
saket: @rakshakhegde I moved the initial event of `BloodPressureScreenChanged(ScreenType.BP_ENTRY)` to this function. It's very common to emit initial events and it's in fact the job of each ui event function to ensure an initial event is emitted. 

If you take a look at RxBinding's source, you'll find that every observer emits a default value on subscription. This makes it easy to setup UI on start.

app/src/main/java/org/simple/clinic/widgets/ViewFlipperWithDebugPreview.kt
```diff
@@ -3,20 +3,26 @@ package org.simple.clinic.widgets
...
+import io.reactivex.subjects.BehaviorSubject
+import io.reactivex.subjects.Subject
 import org.simple.clinic.R
 
 /** Exposes a way to set the displayed child in layout preview. */
 class ViewFlipperWithDebugPreview(context: Context, attrs: AttributeSet) : ViewFlipper(context, attrs) {
 
   private var childToDisplayPostInflate: Int = 0
 
+  private val displayedChildChangesSubject: Subject<Int> = BehaviorSubject.create()
```
saket: @rakshakhegde I updated this subject's type to `Behavior` so that the last value is always replayed on subscribe. This is in line with how RxBinding works.

saket:
Closing in favor of #460

PR #453, opened 2019-01-17
Refactor sync into multiple groups (Part 1)

vinaysshenoy:
https://www.pivotaltracker.com/story/show/163217054

# Changes
- Inject sync configs in sync classes directly instead of into the `SyncCoordinator`
- Remove the appointments v1 and v2 feature flag


app/src/androidTest/java/org/simple/clinic/TestClinicApp.kt
```diff
@@ -81,7 +82,7 @@ class TestClinicApp : ClinicApp() {
                 SyncConfig(
                     frequency = Duration.ofMinutes(16),
                     backOffDelay = Duration.ofMinutes(5),
-                    batchSize = 10))
+                    batchSizeEnum = BatchSize.VERY_SMALL))
```
saket: `batchSizeEnum` -> `batchSize`?

app/src/main/java/org/simple/clinic/sync/SyncConfig.kt
```diff
@@ -8,5 +8,7 @@ data class SyncConfig(
 
     val backOffDelay: Duration,
 
-    val batchSize: Int // Number of records to sync in per network call
-)
+    val batchSizeEnum: BatchSize
```
saket: `batchSizeEnum` -> `batchSize`? 

Why did you suffix this with the type?

vinaysshenoy: Oh,  there was a lot of code that was still referring to the older `batchSize` property and I didn't want to break that. So I named this `batchSizeEnum` so I could expose the older property without breaking existing code. This will be fixed in the PR that completes sync refactoring.

app/src/androidTest/java/org/simple/clinic/TestClinicApp.kt
vinaysshenoy: Closing this since it seems to be a duplicate of the other comment.

saket:
@vinaysshenoy I'll review this today

vinaysshenoy:
@saket Feedback resolved.

PR #460, opened 2019-01-21
Retro active bp entry (part 2)

saket:
https://www.pivotaltracker.com/n/projects/2184102/stories/163052502

This feature is continued in #465


app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -77,16 +77,15 @@ class BloodPressureEntrySheet : BottomSheetActivity() {
...
-        .observeOn(Schedulers.io())
+    Observable.mergeArray(
+    sheetCreates(),
+    screenDestroys,
+    systolicTextChanges(),
+    diastolicTextChanges(),
+    diastolicImeOptionClicks(),
+    diastolicBackspaceClicks(),
+    removeClicks())
+    .observeOn(Schedulers.io())
```
saket: This weird formatting gets fixed in a later commit.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
```diff
@@ -449,7 +449,7 @@ class BloodPressureEntrySheetControllerV2 @Inject constructor(
         .withLatestFrom(newBpDataStream) { _, newBp -> newBp }
         .ofType<NewBpData.ReadyToSave>()
         .flatMapSingle { (date, bp, patientUuid) ->
-          val dateAsInstant = date.atStartOfDay(UTC).toInstant()
+          val dateAsInstant = date.atTime(OffsetTime.now(clock)).toInstant()
```
saket: Just realized that I cannot use the injected clock here. Need to inject `ZoneId` separately.

app/src/main/java/org/simple/clinic/widgets/ageanddateofbirth/DateOfBirthFormatValidator.kt
```diff
@@ -17,27 +17,42 @@ class DateOfBirthFormatValidator @Inject constructor(
...
   fun validate(dateText: String, nowDate: LocalDate = LocalDate.now(UTC)): Result {
-    return try {
+    return when (validate2(dateText, nowDate)) {
+      is Result2.Valid -> Result.VALID
+      is Result2.Invalid.InvalidPattern -> Result.INVALID_PATTERN
+      is Result2.Invalid.DateIsInFuture -> Result.DATE_IS_IN_FUTURE
+    }
+  }
+
+  fun validate2(dateText: String, nowDate: LocalDate = LocalDate.now(UTC)): Result2 {
```
vinaysshenoy: I'm not too sure we should be defaulting to the UTC date here. Since this is used for validation of user input, user will enter a date in local time, while here we'll compare it against the UTC time. This might lead to subtle bugs.

What do you think about either:
- Injecting the system time zone and use that to generate the `nowDate`.
- Removing the default parameter value and forcing users to provide the `nowDate` all the time?

app/src/test/java/org/simple/clinic/newentry/DateOfBirthFormatValidatorTest.kt
```diff
@@ -23,18 +25,46 @@ class DateOfBirthFormatValidatorTest {
...
     val futureDateResult = dobValidator.validate("01/01/3000", nowDate = LocalDate.parse("2018-07-16"))
 
     assertThat(futureDateResult).isEqualTo(Result.DATE_IS_IN_FUTURE)
   }
+
+  @Test
+  fun `validate future date (v2)`() {
+    val format = DateTimeFormatter.ofPattern("dd/MM/yyyy", Locale.ENGLISH)
+    val dobValidator = DateOfBirthFormatValidator(format)
+    val futureDateResult = dobValidator.validate2("01/01/3000", nowDate = LocalDate.parse("2018-07-16"))
```
vinaysshenoy: Can you make this a parameterized test and provide a couple more combinations of dates to verify this behaviour?

app/src/main/java/org/simple/clinic/widgets/ageanddateofbirth/DateOfBirthFormatValidator.kt
```diff
@@ -17,27 +17,42 @@ class DateOfBirthFormatValidator @Inject constructor(
     DATE_IS_IN_FUTURE,
   }
 
+  sealed class Result2 {
```
vinaysshenoy: Instead of adding `Result2`, `validate2` methods in the same class, what do you think about creating another class `DateOfBithFormatValidator2` that then has nicely named methods? The older validator could get a reference to the new one via injection and classes could inject the newer class easily.

saket:
@vinaysshenoy a few layout related changes will come in another PR.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
```diff
@@ -365,12 +364,35 @@ class BloodPressureEntrySheetControllerV2 @Inject constructor(
...
+        .flatMap { (date, newBp, existingBpUuid) ->
+          bloodPressureRepository.measurement(existingBpUuid)
+              .take(1)
+              .map { existingBp ->
+                val dateAsInstant = date.atStartOfDay(UTC).toInstant()
+                existingBp.copy(
+                    systolic = newBp.systolic,
+                    diastolic = newBp.diastolic,
+                    createdAt = dateAsInstant,
+                    updatedAt = dateAsInstant)
```
vinaysshenoy: There's no need to set the `updatedAt` timestamp again here since `BloodPressureRepository#updateMeasurement` does it for us.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTestV2.kt
```diff
@@ -482,4 +482,66 @@ class BloodPressureEntrySheetControllerTestV2 {
...
     verify(sheet).setBpSavedResultAndFinish()
   }
+
+  @Test
+  @Suppress("IMPLICIT_CAST_TO_ANY")
+  @Parameters(method = "params for showing date validation errors")
+  fun `when save is clicked, date entry is active and input is invalid then validation errors should be shown`(
+      openAs: OpenAs,
+      errorResult: DateOfBirthFormatValidator.Result2,
+      errorUiChangeVerifications: UiChange
```
vinaysshenoy: Since this is not exactly a `UiChange`, can we explicitly declare the type as `(BloodPressureEntrySheet) -> Unit` instead?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
```diff
@@ -47,14 +47,15 @@ class BloodPressureEntrySheetControllerV2 @Inject constructor(
     val replayedEvents = ReplayUntilScreenIsDestroyed(events)
         .compose(ReportAnalyticsEvents())
         .compose(combineDateInputs())
+        .compose(validateBpInput())
```
vinaysshenoy: I'm really happy we took the time to standardize this now.

vinaysshenoy: I also think we should rename `UiEvent` -> `Event` at this point.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTestV2.kt
```diff
@@ -528,8 +528,26 @@ class BloodPressureEntrySheetControllerTestV2 {
   }
 
   @Test
-  fun `when BP entry is active, BP readings are valid and next arrow is pressed then date entry should be shown`() {
-    // TODO
+  @Parameters(method = "params for OpenAs types")
+  fun `when BP entry is active, BP readings are valid and next arrow is pressed then date entry should be shown`(
```
vinaysshenoy: Should this test be combined with `when save is clicked, BP entry is active and input is valid then show date entry`?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
```diff
@@ -363,19 +367,10 @@ class BloodPressureEntrySheetControllerV2 @Inject constructor(
...
-
-    val diastolicChanges = events
-        .ofType<BloodPressureDiastolicTextChanged>()
-        .map { it.diastolic }
-
-    val bpChanges = validDates
-        .withLatestFrom(systolicChanges, diastolicChanges)
-        .map { (_, systolic, diastolic) -> BP(systolic.toInt(), diastolic.toInt()) }
+    val validBps = events
+        .ofType<BloodPressureBpValidated>()
```
vinaysshenoy: The name `BloodPressureBpValidated` is a little redundant, don't you think? If the events had been prefixed with `BloodPressureEntry`, it might have made sense but it feels a little weird now.

app/src/main/java/org/simple/clinic/bp/entry/BpValidator.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.simple.clinic.bp.entry.BpValidator.Validation.ErrorSystolicEmpty
+import org.simple.clinic.bp.entry.BpValidator.Validation.ErrorSystolicLessThanDiastolic
+import org.simple.clinic.bp.entry.BpValidator.Validation.ErrorSystolicTooHigh
+import org.simple.clinic.bp.entry.BpValidator.Validation.ErrorSystolicTooLow
+import org.simple.clinic.bp.entry.BpValidator.Validation.Success
+import javax.inject.Inject
+
+class BpValidator @Inject constructor() {
+
+  // TODO: Rename to "Result".
```
vinaysshenoy: Do you want to complete these `TODO`s? They seem to be trivial.

app/src/test/java/org/simple/clinic/bp/entry/BpValidatorTest.kt
```diff
@@ -0,0 +1,72 @@
+package org.simple.clinic.bp.entry
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+
+class BpValidatorTest {
```
vinaysshenoy: Can we make this a parameterized test and add a couple more test cases to each test?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
```diff
@@ -430,31 +429,59 @@ class BloodPressureEntrySheetControllerV2 @Inject constructor(
...
               }
               .flatMapCompletable { bloodPressureRepository.updateMeasurement(it) }
-              .andThen(Observable.just({ ui: Ui -> ui.setBpSavedResultAndFinish() }))
+              .andThen(Single.just({ ui: Ui -> ui.setBpSavedResultAndFinish() }))
         }
 
     return saveNewBp.mergeWith(updateExistingBp)
   }
+
+  sealed class NewBpData {
```
vinaysshenoy: What do you think about combining these two sealed classes into one?

```kotlin
sealed class SaveBpData {
    data class ReadyToCreateBp(val parsedDate: LocalDate, val bpResult: Success, val patientUuid: PatientUuid) : SaveBpData()
    data class ReadyToUpdateBp(val parsedDate: LocalDate, val bpResult: Success, val bpUuid: UUID) : SaveBpData()
    object NeedsCorrection : SaveBpData()
}
```

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
vinaysshenoy: We're using the clock in a lot of places where we'll need to inject the zone separately.

app/src/main/res/values/strings.xml
```diff
@@ -131,6 +131,8 @@
   <string name="bloodpressureentry_error_diastolic_180">Diastolic cannot be more than 180</string>
   <string name="bloodpressureentry_error_systolic_empty">Enter systolic and diastolic pressures</string>
   <string name="bloodpressureentry_error_diastolic_empty">Enter systolic and diastolic pressures</string>
+  <string name="bloodpressureentry_error_date_invalid_pattern">Enter date in DD/MM/YYYY</string>
```
vinaysshenoy: Shouldn't this now be `DD/MM/YY`?

app/src/main/java/org/simple/clinic/bp/BloodPressureModule.kt
```diff
@@ -50,7 +50,7 @@ open class BloodPressureModule {
   open fun provideBloodPressureEntryConfig(): BloodPressureConfig {
     return BloodPressureConfig(
         deleteBloodPressureFeatureEnabled = false,
-        dateEntryEnabled = false)
+        dateEntryEnabled = true)
```
vinaysshenoy: Can we leave this disabled for now until we make the changes to inject the system time zone?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
saket: Doing this

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -22,12 +22,15 @@ import org.simple.clinic.bp.entry.BloodPressureEntrySheetController.Validation.E
...
+import org.threeten.bp.Clock
+import org.threeten.bp.Instant
 import javax.inject.Inject
 
 @Deprecated(message = "Use BloodPressureEntrySheetControllerV2 instead")
 class BloodPressureEntrySheetController @Inject constructor(
     private val bloodPressureRepository: BloodPressureRepository,
-    private val configProvider: Single<BloodPressureConfig>
+    private val configProvider: Single<BloodPressureConfig>,
+    private val clock: Clock
```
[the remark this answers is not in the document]
saket: The default argument uses `Instant.now(clock)` and because a clock isn't present for `BloodPressureRepository`, it ends up in an NPE.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
saket: On a second thought, this is correct, right? We don't need to care about the user's time zone. The user's timezone is only needed when we need to display the time.

app/src/main/java/org/simple/clinic/bp/BloodPressureModule.kt
saket: See my comment?

https://github.com/simpledotorg/simple-android/pull/460/commits/e9722f3bb9718e481c0b49a4f354f1b66dc46f17#r251286760

app/src/test/java/org/simple/clinic/bp/BloodPressureRepositoryTest.kt
```diff
@@ -50,7 +51,7 @@ class BloodPressureRepositoryTest {
     whenever(facilityRepository.currentFacility(userSession)).thenReturn(Observable.just(facility))
 
     val patientUuid = UUID.randomUUID()
-    repository.saveMeasurement(patientUuid, systolic = 120, diastolic = 65).subscribe()
+    repository.saveMeasurement(patientUuid, systolic = 120, diastolic = 65, createdAt = Instant.now()).subscribe()
```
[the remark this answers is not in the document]
saket: What do you mean by this? There is no `updateAt` param in `saveMeasurement()`

app/src/main/java/org/simple/clinic/widgets/ageanddateofbirth/DateOfBirthFormatValidator.kt
saket: Good catch. Because the class is now called `UserInputDateValidator`, I think we can just inject the user's timezone. What do you think?

app/src/test/java/org/simple/clinic/newentry/DateOfBirthFormatValidatorTest.kt
saket: Can I do this separately? Asking because this change wasn't a part of this feature. I simply duplicated the logic for using sealed classes.

app/src/main/java/org/simple/clinic/widgets/ageanddateofbirth/DateOfBirthFormatValidator.kt
saket: I don't see the benefit 🤔 

> that then has nicely named methods

what do you mean by this?

saket: One benefit of keeping both the functions in the same class is discoverability during code completion.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
saket: Fixing this

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTestV2.kt
saket: Done

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
saket: Agreed

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTestV2.kt
saket: I intentionally kept them separate to indicate that the tests treat next arrow and save as separate events even though they do the same thing right now. That said, I should name them similarly. Fixing this.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
saket: I know 😅 I did this to keep it consistent with `BloodPressureDateValidated`. Otherwise we could have `BloodPressureValidated` and `BloodPressureDateValidated`.

What do you think about `BloodPressureReadingsValidated`?

app/src/main/java/org/simple/clinic/bp/entry/BpValidator.kt
saket: I'll do this in my next PR

app/src/test/java/org/simple/clinic/bp/entry/BpValidatorTest.kt
saket: Good idea, we should do this. I'll do this in a separate PR, again because this change wasn't a part of this feature. I'm adding these TODOs to https://github.com/simpledotorg/simple-android/pull/465's description but will move them to their pivotal stories in the end.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
```diff
@@ -140,6 +144,34 @@ class BloodPressureEntrySheetControllerV2 @Inject constructor(
...
+          { ui: Ui ->
+            val currentDate = LocalDate.now(clock)
+            ui.setDate(currentDate.dayOfMonth, currentDate.monthValue, currentDate.year)
+          }
+        }
+
+    val prefillForExistingBp = openAsStream
+        .ofType<OpenAs.Update>()
+        .flatMap { bloodPressureRepository.measurement(it.bpUuid) }
+        .take(1)
```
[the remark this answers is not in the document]
saket: That is correct, pre-filling the UI should only happen once

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTestV2.kt
```diff
@@ -519,8 +519,43 @@ class BloodPressureEntrySheetControllerTestV2 {
   }
 
   @Test
-  fun `when BP entry is active, BP readings are invalid and next arrow is pressed then date entry should not be shown`() {
-    // TODO
+  @Parameters(method = "params for OpenAs and bp validation errors")
+  fun `when BP entry is active, BP readings are invalid and next arrow is pressed then date entry should not be shown`(
```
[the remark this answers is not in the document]
saket: Right, fixing this

app/src/main/java/org/simple/clinic/widgets/ViewFlipperWithDebugPreview.kt
```diff
@@ -3,20 +3,26 @@ package org.simple.clinic.widgets
...
+  private val displayedChildChangesSubject: Subject<Int> = BehaviorSubject.create()
+  val displayedChildChanges: Observable<Int> = displayedChildChangesSubject.hide()
+
   init {
     if (isInEditMode) {
       val attributes = context.obtainStyledAttributes(attrs, R.styleable.ViewFlipperWithDebugPreview)
       childToDisplayPostInflate = attributes.getInt(R.styleable.ViewFlipperWithDebugPreview_debug_displayedChild, 0)
       attributes.recycle()
     }
+    displayedChildChangesSubject.onNext(displayedChild)
```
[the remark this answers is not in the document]
saket: That'd be nice, but I wrote this in a way that we should be able to drop-in RxBinding if they ever add support for ViewFlipper. `displayedChildResId` is our extension value.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
```diff
@@ -148,24 +148,29 @@ class BloodPressureEntrySheetControllerV2 @Inject constructor(
         .ofType<BloodPressureEntrySheetCreated>()
         .map { it.openAs }
 
+    val showDateOnUi = { date: LocalDate ->
+      { ui: Ui ->
+        val dayString = date.dayOfMonth.toString().padStart(length = 2, padChar = '0')
```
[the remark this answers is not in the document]
saket: Ooh right! I'll inject this padding character from the module in the meanwhile.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
vinaysshenoy: Yes, you're correct. This specific instance needs the UTC time.

app/src/test/java/org/simple/clinic/bp/BloodPressureRepositoryTest.kt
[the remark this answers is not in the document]
vinaysshenoy: Oh, I mean in the verification block. We don't verify whether the BP was created with the passed in timestamp for both `createdAt` and `updatedAt`.

app/src/main/java/org/simple/clinic/widgets/ageanddateofbirth/DateOfBirthFormatValidator.kt
vinaysshenoy: Works for me.

app/src/test/java/org/simple/clinic/newentry/DateOfBirthFormatValidatorTest.kt
vinaysshenoy: Sure.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
```diff
@@ -316,11 +321,19 @@ class BloodPressureEntrySheetControllerV2 @Inject constructor(
...
     val yearChanges = events
         .ofType<BloodPressureYearChanged>()
-        .map { it.year }
+        .map { it.twoDigitYear }
 
     val combinedDates = Observables
         .combineLatest(dayChanges, monthChanges, yearChanges)
-        .map { (d, m, y) -> BloodPressureDateChanged(date = "$d/$m/$y") }
+        .map { (dd, mm, yy) ->
+          val paddedDd = dd.padStart(length = 2, padChar = '0')
```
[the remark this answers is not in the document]
saket: Injecting through DI

app/src/sharedTest/java/org/simple/clinic/util/TestClock.kt
```diff
@@ -19,7 +19,7 @@ class TestClock : Clock() {
 
   fun setYear(year: Int) {
     if (year < 1970) {
-      throw AssertionError("year should be >= 1970!")
+      throw AssertionError("twoDigitYear should be >= 1970!")
```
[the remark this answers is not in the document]
saket: Whoops, fixng

app/src/main/res/drawable/ic_round_arrow_forward_ios_24px.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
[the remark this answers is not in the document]
saket: I knew something looked odd, but couldn't spot what. It's `px`, right? Fixing this.

app/src/main/res/values/dimens.xml
```diff
@@ -41,4 +41,6 @@
   <dimen name="patientsummary_bp_list_item_last_in_group_bottom_padding">10dp</dimen>
   <dimen name="patientsummary_bp_list_item_multiple_in_group_bp_top_padding">0dp</dimen>
   <dimen name="patientsummary_bp_list_item_multiple_in_group_bp_bottom_padding">@dimen/spacing_12</dimen>
+
+  <dimen name="bloodpressureentry_navigation_arrow">48dp</dimen>
```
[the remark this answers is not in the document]
saket: Whoops, fixing

app/src/main/java/org/simple/clinic/widgets/ageanddateofbirth/DateOfBirthFormatValidator.kt
vinaysshenoy: > I don't see the benefit 🤔
> 
> > that then has nicely named methods
> 
> what do you mean by this?

What I mean is there won't be a `Result2`, `validate2`, etc. This will only be used when declaring the type `DateOfBirthFormatValidator2` at the injection site. The actual usages of the instance will be cleaner, like

```kotlin
dateValidator.validate()
```

However, we will be removing this soon anyway once we replace the deprecated usages with the newer ones. We can leave this in for now, and fix the method naming once the migration completes.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
vinaysshenoy: Works.

app/src/main/res/values/dimens.xml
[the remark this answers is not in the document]
saket: Done

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -149,10 +152,10 @@ class BloodPressureEntrySheet : BottomSheetActivity() {
       .map(::BloodPressureDiastolicTextChanged)
 
   private fun diastolicImeOptionClicks(): Observable<BloodPressureSaveClicked> {
-    return Observable
-        .merge(
-            RxTextView.editorActions(systolicEditText) { actionId -> actionId == EditorInfo.IME_ACTION_DONE },
-            RxTextView.editorActions(diastolicEditText) { actionId -> actionId == EditorInfo.IME_ACTION_DONE })
+    return listOf(systolicEditText, diastolicEditText, dayEditText, monthEditText, yearEditText)
+        .map { RxTextView.editorActions(it) { actionId -> actionId == EditorInfo.IME_ACTION_DONE } }
```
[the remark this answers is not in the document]
saket: 💯

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
```diff
@@ -263,7 +263,7 @@ class BloodPressureEntrySheetControllerV2 @Inject constructor(
```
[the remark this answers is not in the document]
saket: Sorry I can't see the commit message with your commit 😅

app/src/main/java/org/simple/clinic/bp/BloodPressureModule.kt
vinaysshenoy: > See my comment?
> 
> [e9722f3#r251286760](https://github.com/simpledotorg/simple-android/commit/e9722f3bb9718e481c0b49a4f354f1b66dc46f17#r251286760)

This doesn't link to any comment.

app/src/main/res/drawable/ic_round_arrow_forward_ios_24px.xml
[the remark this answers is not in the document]
vinaysshenoy: It's not just that, it's got `_ios_` in the name too. 😅

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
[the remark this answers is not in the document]
vinaysshenoy: The commit message I see is `Fix: save BP button doesn't for the first time`.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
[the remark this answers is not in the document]
vinaysshenoy: The original comment got duplicated for some reason. 😓

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
```diff
@@ -263,7 +263,7 @@ class BloodPressureEntrySheetControllerV2 @Inject constructor(
 
     val validations = Observables.combineLatest(systolicChanges, diastolicChanges, screenChanges)
         .filter { (_, _, screen) -> screen == BP_ENTRY }
-        .map { (systolic, diastolic, _) -> bpValidator.validate(systolic, diastolic) }
```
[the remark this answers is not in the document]
saket: Sure

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
saket: Even better. I'll make this change 👍.

app/src/main/res/values/strings.xml
saket: Whoops, fixing this

app/src/main/java/org/simple/clinic/bp/BloodPressureModule.kt
saket: Do you still want me to do this? Considering using the user's timezone in `UserInputDateValidator` can go in a separate PR?

app/src/test/java/org/simple/clinic/bp/BloodPressureRepositoryTest.kt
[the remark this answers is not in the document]
saket: Got it. Sure, doing this.

app/src/main/res/drawable/ic_round_arrow_forward_ios_24px.xml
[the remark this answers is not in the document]
saket: Oh I thought that was intentional to shame people because they're using iOS icons :p I'll remove this because the name already says `round_arrow`.

app/src/main/java/org/simple/clinic/bp/BloodPressureModule.kt
vinaysshenoy: Are we going to make that change immediately? If we are, then we can skip this for now.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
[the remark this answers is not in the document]
saket: Changed to `Fix: save BP button work doesn't on the first click`

saket:
@vinaysshenoy fyi the commit, `97922a3f: Enable retro-active BP entry feature` has been re-ordered to the bottom of the commit list but Github is showing the old ordering as usual.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerV2.kt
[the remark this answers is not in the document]
vinaysshenoy: Shouldn't that be `Fix: save BP button doesn't work on the first click`?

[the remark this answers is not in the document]
vinaysshenoy: Unless you've somehow become Master Yoda over the weekend.

saket:
@vinaysshenoy typo in the commit fixed

PR #462, opened 2019-01-23
Custom prescription screen

sanchita:


app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntrySheetv2.kt
```diff
@@ -0,0 +1,105 @@
...
+      super.onBackgroundClick()
+    }
+  }
+
+  private fun sheetCreates(): Observable<UiEvent> {
+    val patientUuid = intent.getSerializableExtra(KEY_PATIENT_UUID) as UUID
+    return Observable.just(CustomPrescriptionSheetCreated(patientUuid))
+  }
+
+  private fun drugNameChanges() = drugNameEditText.textChanges(::CustomPrescriptionDrugNameTextChanged)
```
saket: @sanchita I'm trying to review this PR commit-wise. I'm on the first commit and it looks like this commit is doing multiple things:
1. Adding V2 classes so that the new feature can be put behind a flag
2. Adding other changes like this one:

```kotlin
70: private fun drugDosageChanges() = drugDosageEditText.textChanges(::CustomPrescriptionDrugDosageTextChanged)
```

Because the entire file was duplicated, it's now difficult to track what was changed.

app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntrySheetv2.kt
```diff
@@ -94,11 +94,17 @@ class CustomPrescriptionEntrySheetv2 : BottomSheetActivity() {
   }
 
   companion object {
-    private const val KEY_PATIENT_UUID = "patientUuid"
+    private const val KEY_OPEN_AS = "openAs"
 
-    fun intent(context: Context, patientUuid: UUID): Intent {
+    fun intentForAddNewPrescription(context: Context, patientUuid: UUID): Intent {
```
saket: `intentForAddNewPrescription` -> `intentForNewPrescription`? Sounds more natural.

app/src/test/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntryControllerv2Test.kt
```diff
@@ -0,0 +1,96 @@
...
+    controller = CustomPrescriptionEntryControllerv2(prescriptionRepository)
+
+    uiEvents
+        .compose(controller)
+        .subscribe { uiChange -> uiChange(sheet) }
+  }
+
+  @Test
+  fun `save should remain disabled while drug name is empty`() {
+    uiEvents.onNext(CustomPrescriptionDrugNameTextChanged(""))
```
saket: These tests don't look like they should be a part of this commit.

app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntrySheetv2.kt
sanchita: This change is only because of the layout changes.
I have changed `EditText` -> `TextInputEditText`

saket: That shouldn't affect the existing code. They're both `EditText`, right?

saket: Additionally, you're not even making use of any APIs of `TextInputLayout`. You could continue using `EditText` as the type for `drugNameEditText` and `drugDosageEditText` while keeping `TextInputEditText` in the xml.

app/src/main/java/org/simple/clinic/drugs/selection/entry/CustomPrescriptionEntrySheet.kt
```diff
@@ -50,8 +65,8 @@ class CustomPrescriptionEntrySheet : BottomSheetActivity() {
   }
 
   override fun onBackgroundClick() {
-    val drugNameEmpty = drugNameEditText.text.isEmpty()
-    val dosageEmpty = drugDosageEditText.text.isEmpty()
+    val drugNameEmpty = drugNameEditText.text.isNullOrEmpty()
```
saket: `isNullOrEmpty` -> `isBlank()`?

PR #465, opened 2019-01-24
Retro active bp entry (part 3)

saket:


saket:
@vinaysshenoy FYI: the merge destination of this PR will need to be updated once #460 gets merged

vinaysshenoy:
> @vinaysshenoy FYI: the merge destination of this PR will need to be updated once #460 gets merged

Sound we just merge it into `sn/21jan/retro-active-bp-entry` and merge that into `master` in one go? We can avoid having to run checks again that way.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -184,14 +175,11 @@ class BloodPressureEntrySheet : BottomSheetActivity() {
 
   private fun hardwareBackPresses(): Observable<UiEvent> {
     return Observable.create { emitter ->
-      val interceptor = object : BackPressInterceptor {
-        override fun onInterceptBackPress(callback: BackPressInterceptCallback) {
-          emitter.onNext(BloodPressureBackPressed)
-          callback.markBackPressIntercepted()
```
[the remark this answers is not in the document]
vinaysshenoy: Never mind, I see that this was changed to use a different interceptor than the one in `ScreenRouter`.

app/src/main/java/org/simple/clinic/bp/entry/LinearLayoutWithPreImeKeyEventListener.kt
```diff
@@ -7,13 +7,11 @@ import android.widget.LinearLayout
 
 class LinearLayoutWithPreImeKeyEventListener(context: Context, attrs: AttributeSet) : LinearLayout(context, attrs) {
 
-  var backKeyPressInterceptor: () -> Unit = {
-    throw AssertionError("Back-press interceptor not set")
-  }
+  var backKeyPressInterceptor: (() -> Unit)? = null
```
vinaysshenoy: Why was this made `nullable`? Didn't we add that assertion so that you are expected to override that property if you use this widget?

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTestV2.kt
```diff
@@ -615,6 +615,18 @@ class BloodPressureEntrySheetControllerTestV2 {
...
+
+  @Test
+  fun `when BP entry is active and back is pressed then the sheet should be closed`() {
+    uiEvents.run {
+      onNext(BloodPressureScreenChanged(BP_ENTRY))
+      onNext(BloodPressureBackPressed)
+    }
+
+    verify(sheet).finish()
+    verify(sheet, never()).showBpEntryScreen()
```
vinaysshenoy: Is this verification necessary?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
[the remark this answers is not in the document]
saket: `LinearLayoutWithPreImeKeyEventListener` intercepts all back presses anyway so this wasn't required anymore

app/src/main/java/org/simple/clinic/bp/entry/LinearLayoutWithPreImeKeyEventListener.kt
saket: I made this nullable as part of this change:

https://github.com/simpledotorg/simple-android/blob/5583e830ea9a475c586f7b7f31f960a69499d414/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt#L181

saket: Essentially clearing any reference from Rx to avoid leaks

PR #466, opened 2019-01-24
New Custom prescription screen

sanchita:
Pivotol story -> https://www.pivotaltracker.com/story/show/163017678


app/src/main/AndroidManifest.xml
```diff
@@ -37,6 +37,11 @@
       android:theme="@style/Clinic.V2.Theme.BottomSheetActivity"
```
sanchita: The next commit is `f1a9748b Change Text -> TextInputLayout for labels in CustomPrescriptionEntrySheet`.
Please go to that after this commit is reviewed.

saket:
@sanchita can you add a link to the pivotal story here?

sanchita:
> @sanchita can you add a link to the pivotal story here?

Done!

app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntrySheetv2.kt
```diff
@@ -5,7 +5,7 @@ import android.content.Intent
 import android.os.Bundle
```
saket: `Change Text -> TextInputLayout for labels in CustomPrescriptionEntrySheet` -> `Use TextInputLayout for labels in  CustomPrescriptionEntrySheet`?

app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntrySheetv2.kt
```diff
@@ -50,8 +51,8 @@ class CustomPrescriptionEntrySheetv2 : BottomSheetActivity() {
   }
 
   override fun onBackgroundClick() {
-    val drugNameEmpty = drugNameEditText.text.isEmpty()
-    val dosageEmpty = drugDosageEditText.text.isEmpty()
+    val drugNameEmpty = drugNameEditText.text.isNullOrEmpty()
+    val dosageEmpty = drugDosageEditText.text.isNullOrEmpty()
```
saket: Do you want to use `isBlank()` instead?

app/src/main/java/org/simple/clinic/drugs/PrescriptionRepository.kt
```diff
@@ -22,7 +23,8 @@ class PrescriptionRepository @Inject constructor(
     private val database: AppDatabase,
     private val dao: PrescribedDrug.RoomDao,
     private val facilityRepository: FacilityRepository,
-    private val userSession: UserSession
+    private val userSession: UserSession,
+    val clock: Clock
```
saket: Missing `private` modifier?

app/src/androidTest/java/org/simple/clinic/drugs/PrescriptionRepositoryAndroidTest.kt
```diff
@@ -3,6 +3,7 @@ package org.simple.clinic.drugs
 import androidx.test.runner.AndroidJUnit4
```
saket: `Provide update prescription` doesn't… make sense. Did you miss some words in the middle?

app/src/test/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntryControllerv2Test.kt
```diff
@@ -93,4 +96,59 @@ class CustomPrescriptionEntryControllerv2Test {
...
+  @Suppress("Unused")
+  private fun `params for showing title`(): List<List<Any>> {
+    return listOf(
+        listOf(OpenAs.New(patientUuid), true),
+        listOf(OpenAs.Update(prescriptionUuid), false)
+    )
+  }
+
+  @Parameters(method = "params for showing remove button")
+  @Test
```
saket: nitpick: can you move `@Parameters()` annotation below `@Test` like the other functions?

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugEvents.kt
```diff
@@ -7,3 +7,5 @@ import java.util.UUID
 data class ProtocolDrugSelected(val drugName: String, val prescribedDrug: PrescribedDrug?) : UiEvent
```
saket: Can `Edit` in the commit name be in lowercase?

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugEvents.kt
```diff
@@ -7,3 +7,5 @@ import java.util.UUID
...
```
saket: Can we rename this event to `ProtocolDrugClicked`? Because the user doesn't "select" anything yet. It just sounds more correct to me.

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugEvents.kt
```diff
@@ -7,3 +7,5 @@ import java.util.UUID
...
 
 data class PrescribedDrugsScreenCreated(val patientUuid: UUID) : UiEvent
+
+data class UpdateCustomPrescription(val prescribedDrug: PrescribedDrug) : UiEvent
```
saket: We suffix events by user actions e.g., `Clicked`, `Selected`, `Created`, etc. Can we do the same here? Otherwise, I can't tell when this event gets emitted just by looking at the name.

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugScreenV2.kt
```diff
@@ -122,4 +123,8 @@ class PrescribedDrugScreenV2(context: Context, attrs: AttributeSet) : LinearLayo
   fun showDosageSelectionSheet(drugName: String, patientUuid: UUID, prescribedDrugUuid: UUID?) {
     activity.startActivity(DosagePickerSheet.intent(context, drugName, patientUuid, prescribedDrugUuid))
   }
+
+  fun showUpdateCustomPrescriptionScreen(prescribedDrug: PrescribedDrug) {
```
saket: `showUpdateCustomPrescriptionScreen()` -> `showUpdateCustomPrescriptionSheet()`?

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugEvents.kt
saket: I spent a lot of time staring at `val prescribedDrug: PrescribedDrug?` because something felt wrong about its nullable nature and I realized it's because of the name!

Because the event is named as `ProtocolDrugSelected`, it looks wrong that the `drug` is nullable. How can a drug be selected if it's null?

I then realized that this param is the previously selected drug. Let's improve this name? Something along the lines of `previouslySelectedDosage`.

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugScreenV2.kt
```diff
@@ -122,4 +123,8 @@ class PrescribedDrugScreenV2(context: Context, attrs: AttributeSet) : LinearLayo
...
+    activity.startActivity(CustomPrescriptionEntrySheetv2.intentForUpdatePrescription(context, prescribedDrug.uuid))
```
saket: What do you think about renaming `intentForUpdatePrescription()` -> `intentForUpdatingPrescription()`? Or `intentForPrescriptionUpdate`?

I don't have strong thoughts about this so it's okay if you want to skip.

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugsScreenControllerV2.kt
```diff
@@ -35,7 +35,8 @@ class PrescribedDrugsScreenControllerV2 @Inject constructor(
         handleDoneClicks(replayedEvents),
         populateDrugsList(replayedEvents),
         addNewPrescription(replayedEvents),
-        selectDosage(replayedEvents))
+        selectDosage(replayedEvents),
```
saket: `selectDosage()` -> `showDosagePicker`?

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugsScreenControllerV2.kt
```diff
@@ -35,7 +35,8 @@ class PrescribedDrugsScreenControllerV2 @Inject constructor(
...
+        updateCustomPrescription(replayedEvents))
```
saket: This function name suggests that it is going to also update, which is not true. Let's rename this to `showUpdateCustomPrescription()` or something?

app/src/test/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntryControllerv2Test.kt
```diff
@@ -151,4 +151,20 @@ class CustomPrescriptionEntryControllerv2Test {
...
+    whenever(prescriptionRepository.prescription(prescriptionUuid)).thenReturn(Observable.just(prescribedDrug))
+    whenever(prescriptionRepository.updatePrescription(any())).thenReturn(Completable.complete())
+
+    uiEvents.onNext(CustomPrescriptionSheetCreated(OpenAs.Update(prescriptionUuid)))
+    uiEvents.onNext(CustomPrescriptionDrugNameTextChanged("Atenolol"))
+    uiEvents.onNext(CustomPrescriptionDrugDosageTextChanged("5mg"))
+    uiEvents.onNext(SaveCustomPrescriptionClicked())
+
+    verify(prescriptionRepository).updatePrescription(prescribedDrug.copy(name = "Atenolol", dosage = "5mg"))
+    verify(sheet).finish()
```
saket: Let's also add assertions for checking the inverse functionality? That is `prescriptionRepository.savePrescription()` never gets called here and vice versa.

app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntryControllerv2.kt
```diff
@@ -173,4 +176,32 @@ class CustomPrescriptionEntryControllerv2 @Inject constructor(
...
+  private fun closeSheetWhenPrescriptionIsDeleted(events: Observable<UiEvent>): Observable<UiChange> {
+    val prescribedDrugUuuids = events
+        .ofType<CustomPrescriptionSheetCreated>()
+        .filter { it.openAs is OpenAs.Update }
+        .map { it.openAs as OpenAs.Update }
+        .map { it.prescribedDrugUuid }
+
+    return prescribedDrugUuuids
+        .flatMap(prescriptionRepository::prescription)
+        .filter { it.isDeleted }
```
saket: Nice! This sheet gets finished as soon as the observed item is removed. Persistence makes state management so easy!

app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntrySheetv2.kt
```diff
@@ -44,7 +45,13 @@ class CustomPrescriptionEntrySheetv2 : BottomSheetActivity() {
...
 
-    Observable.mergeArray(sheetCreates(), drugNameChanges(), drugDosageChanges(), drugDosageFocusChanges(), saveClicks())
+    Observable.mergeArray(
+        sheetCreates(),
+        drugNameChanges(),
+        drugDosageChanges(),
+        drugDosageFocusChanges(),
+        saveClicks(),
+        removeClicks())
         .observeOn(io())
```
saket: The arguments to `mergeArray()` and operators on the Observable are on the same indentation. Can you instead try moving `mergeArray()` to a next line?

```kotlin
Observable
    .mergeArray(
        sheetCreates(),
        drugNameChanges(),
        drugDosageChanges(),
        drugDosageFocusChanges(),
        saveClicks(),
        removeClicks())
    .observeOn(io())
    .compose(controller)
    .observeOn(mainThread())
    .takeUntil(onDestroys)
```

app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntrySheetv2.kt
```diff
@@ -84,9 +91,11 @@ class CustomPrescriptionEntrySheetv2 : BottomSheetActivity() {
 
     return RxView.clicks(saveButton)
         .mergeWith(dosageImeClicks)
-        .map { SaveCustomPrescriptionClicked() }
+        .map { SaveCustomPrescriptionClicked }
   }
 
+  private fun removeClicks(): Observable<UiEvent> = RxView.clicks(removeMedicineButton).map { RemoveCustomPrescriptionClicked }
```
saket: Can you break this line into multiple lines? Multiple operators on different lines please.

```kotlin
  private fun removeClicks(): Observable<UiEvent> =
      RxView
          .clicks(removeMedicineButton)
          .map { RemoveCustomPrescriptionClicked }
```

app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntrySheetv2.kt
```diff
@@ -121,7 +130,11 @@ class CustomPrescriptionEntrySheetv2 : BottomSheetActivity() {
   }
 
   fun setDosage(dosage: String?) {
-    drugDosageEditText.setTextAndCursor(dosage?:"")
+    drugDosageEditText.setTextAndCursor(dosage ?: "")
```
saket: Not important, but `EditText` accepts null strings so this isn't needed

app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/confirmremovedialog/ConfirmRemovePrescriptionDialog.kt
```diff
@@ -0,0 +1,109 @@
...
+    return Observable.just(ConfirmRemovePrescriptionDialogCreated(prescriptionUuidToDelete))
+  }
+
+  private fun removeClicks(): Observable<UiEvent> {
+    val button = (dialog as AlertDialog).getButton(DialogInterface.BUTTON_POSITIVE)
+    return RxView.clicks(button).map { RemovePrescriptionClicked }
+  }
+
+  companion object {
+    private const val TAG = "CustomPrescriptionEntryScreenFragment"
```
saket: Copy paste error? :)

app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/confirmremovedialog/ConfirmRemovePrescriptionDialogController.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.simple.clinic.widgets.UiEvent
+import javax.inject.Inject
+
+typealias Ui = ConfirmRemovePrescriptionDialog
+typealias UiChange = (Ui) -> Unit
+
+class ConfirmRemovePrescriptionDialogController @Inject constructor(
+    private val prescriptionRepository: PrescriptionRepository
+) : ObservableTransformer<UiEvent, UiChange> {
+  override fun apply(events: Observable<UiEvent>): ObservableSource<UiChange> {
```
saket: Line break before apply please

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugEvents.kt
sanchita: The event is `ProtocolDrugClicked`(after renaming) which refers to clicking a _protocol_ drug and the argument being passed is a `prescribedDrug` which is a _prescription_ and not the protocol drug itself. 
The protocol drug cannot be `null` but a prescription for a protocol drug can be. This relationship is already established in our app. 
Also, its not a dosage that we are passing but the prescription itself. Hence naming it on the lines of dosage can be misleading. 
Does this make sense?

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugsScreenControllerV2.kt
sanchita: This name is along the same lines as `addNewPrescription` in this file.

app/src/main/java/org/simple/clinic/drugs/selectionv2/PrescribedDrugEvents.kt
saket: Whoops sorry, I meant prescription and not dosage. Got a bit confused here. `forProtocolDrug` suffix isn't necessary because that's implied by the event name. I am trying to make it easy to understand why the item is nullable through the name itself. Let's just go with `prescription`. I think the larger problem here is the name `PrescribedDrug`. It should be renamed to `Prescription`.

app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntrySheetv2.kt
saket: ~rule~ guideline 😅

saket: I just think there's too much happening here in a single line.

app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/confirmremovedialog/ConfirmRemovePrescriptionDialog.kt
saket: Ohh. That doesn't sound right. The screen can potentially have multiple fragments. The tag is always supposed to identify the fragment and not the caller. That's why the API is also called `FragmentManager.findFragmentByTag()`.

app/src/test/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntryControllerv2Test.kt
sanchita: Duplicate comment.

app/src/main/java/org/simple/clinic/drugs/selectionv2/entry/CustomPrescriptionEntrySheetv2.kt
sanchita: `isBlank()` does not check for `null`.

PR #473, opened 2019-01-25
Create CONTRIBUTING.md

saket:


README.md
```diff
@@ -4,55 +4,29 @@
...
 - Go back to phone settings. A new setting group called  `Developer options `will now be available.
 - Open `Developer options` and enable `USB debugging`.
 
-The binary can now be installed on the phone by running:
+If `adb` fails with a `no devices/emulators found` error, it is possible that the device is connected to the computer in charging only mode. In this case, you should see a notification on the device to change this to debugging.
 
-```
-~/Library/Android/sdk/platform-tools/adb install {path to apk}
-```
+## Downloading a build from CI
```
pratul: We should remove this section because Bitrise is accessible only by a subset of people. We can't keep adding external contributors to Bitrise. The way systems like Jenkins solve this is by having publicly accessible running builds and downloads.

README.md
```diff
@@ -62,7 +36,7 @@ Follow the [simple-server instructions](https://github.com/simpledotorg/simple-s
 
 #### Java 8's date and time
 
-Clinic uses [lazythreetenbp](https://github.com/gabrielittner/lazythreetenbp) for working with date and time. Due to some limitations, the IDE does not know how to download its sources. As a work around, the sources can be downloaded from [the maven repository](http://search.maven.org/#search%7Cga%7C1%7Cthreetenbp) and manually attached to Android Studio.
+Simple uses [lazythreetenbp](https://github.com/gabrielittner/lazythreetenbp) for working with date and time. Due to some limitations, the IDE does not understand how to download its sources. As a work around, the sources can be downloaded from [the maven repository](http://search.maven.org/#search%7Cga%7C1%7Cthreetenbp) and manually attached to Android Studio.
```
pratul: Let's not mix styles -- either use `we` everywhere or `Simple` everywhere.

README.md
```diff
@@ -4,55 +4,29 @@
...
 
-```
-~/Library/Android/sdk/platform-tools/adb install {path to apk}
-```
+## Downloading a build from CI
 
-If `adb` fails with a `no devices/emulators found` error, it is possible that the device is connected to the computer in charging only mode. In this case, you should see a notification on the device to change this to debugging.
+We use Bitrise for continuous integration. It is configured to build a binary every time a commit is pushed to the repository. These binaries can be found by navigating [to this page](https://www.bitrise.io/app/db9b195f645cfed7#/builds_) and opening the details of any build. The details page will contain a generated APK called `app-debug.apk` under the "Apps & Artifacts" tab.
 
 ## Build and deploy Simple Server
```
pratul: We should clarify that the server is not _really_ needed to run the app, and is only needed to sync data across devices.

README.md
pratul: Actually, on second thought, we should remove the entire Miscellaneous section from the README, and move them into the code as comments.
- lazythreetenbp info can be mentioned in the code as a comment 
- workmanager debugging info can be mentioned as a comment 
- androidTest failure workaround can also be mentioned somewhere as a comment. Maybe in the base-class of the android tests?

saket: Comments in the code would have been nice, but dates are used everywhere in the app. There is no starting point. Same for `androidTest`. 

We can move these to separate wiki pages.

pratul: How about the place where lazythreetenbp is being initialised?

saket: Hrm, if you're a new developer and you encounter one of the date classes, will you be able to figure out that you should open the place where `lazythreetenbp` is being initialized?

pratul:
No, but if you're a new developer you'd be too inexperienced to read any
documentation or follow anything. An experienced developer will just grep
their way through things!

README.md
pratul: No, but if you're a new developer you'd be too inexperienced to read any documentation or follow anything. An experienced developer will just grep their way through things!

saket: Sorry I meant "new to this project".

Even as an experienced dev, it took me a while to understand why am I unable to add sources. Had to go through some pages on Github and some other issue tracking website. I don't think it has to do anything with your experience.

PR #476, opened 2019-01-29
Location based facility suggestion (part 1)

saket:

This PR,
- adds `Coordinates` to Facility schema + database migration
- adds Play Services library

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
@@ -5,6 +5,7 @@ import android.database.sqlite.SQLiteConstraintException
 import androidx.room.migration.Migration
 import androidx.sqlite.db.SupportSQLiteDatabase
 import androidx.test.runner.AndroidJUnit4
+import com.f2prateek.rx.preferences2.Preference
 import com.google.common.truth.Truth
 import com.google.common.truth.Truth.assertThat
 import com.google.common.truth.Truth.assertWithMessage
@@ -14,8 +15,12 @@ import org.junit.Test
 import org.junit.rules.ExpectedException
 import org.junit.runner.RunWith
 import org.simple.clinic.storage.inTransaction
+import org.simple.clinic.util.Just
+import org.simple.clinic.util.None
+import org.simple.clinic.util.Optional
 import java.util.UUID
 import javax.inject.Inject
+import javax.inject.Named
 
 @Suppress("LocalVariableName")
 @RunWith(AndroidJUnit4::class)
@@ -29,6 +34,10 @@ class DatabaseMigrationAndroidTest {
 
   @Inject
   lateinit var migrations: ArrayList<Migration>
+
+  @Inject
+  @field:Named("last_facility_pull_token")
+  lateinit var lastFacilityPullToken: Preference<Optional<String>>
 
   @Before
   fun setup() {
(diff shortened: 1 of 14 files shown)


saket:
@vinaysshenoy re-rolled with master and fixed the failing build

app/src/androidTest/java/org/simple/clinic/TestData.kt
```diff
@@ -241,6 +242,9 @@ class TestData @Inject constructor(
       pinCode: String? = null,
       protocolUuid: UUID? = UUID.randomUUID(),
       groupUuid: UUID? = UUID.randomUUID(),
+      location: Coordinates = Coordinates(
```
vinaysshenoy: Shouldn't this parameter be nullable?

app/src/main/java/org/simple/clinic/facility/Facility.kt
```diff
@@ -41,6 +43,9 @@ data class Facility(
     // have group UUID until they're synced again.
     val groupUuid: UUID?,
 
+    @Embedded(prefix = "location_")
+    val location: Coordinates?,
```
vinaysshenoy: The only reason this is nullable is because of backward-compatibility, correct?

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
```diff
@@ -117,6 +118,7 @@ object PatientMocker {
       pinCode: String? = null,
       protocolUuid: UUID? = UUID.randomUUID(),
       groupUuid: UUID? = UUID.randomUUID(),
+      location: Coordinates = Coordinates(latitude = 1.908537, longitude = 73.537524),
```
vinaysshenoy: This property should be nullable, right?

vinaysshenoy:
@saket Your tests are still failing. This seems to be because the facilities API seems to be sending `NULL` as the latitude and longitude for some facilities. This either needs to be fixed on the backend, or we need to make these properties nullable as well.

In addition, do you want to clear the facilities pull token in this PR itself since this is where the migration is happening?

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -799,6 +799,43 @@ class DatabaseMigrationAndroidTest {
...
+        '2018-09-25T11:20:42.008Z',
+        '2018-09-25T11:20:42.008Z',
+        'PENDING',
+        '2018-09-25T11:20:42.008Z')
+    """)
+
+    val db_v28 = helper.migrateTo(28)
+    db_v28.query("""SELECT * FROM Facility""").use {
+      it.moveToNext()
+      assertThat(it.columnCount).isEqualTo(17)
```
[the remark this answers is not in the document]
saket: Should I just assert the existence of each column one by one in 15 lines?

app/src/androidTest/java/org/simple/clinic/TestData.kt
saket: Right, fixing this

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
saket: Yep, fixing this

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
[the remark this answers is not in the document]
vinaysshenoy: Yes, that works.

app/src/main/java/org/simple/clinic/facility/Facility.kt
saket: Two reasons:
1. Backward compatibility
2. We don't have the location of every facility yet

saket:

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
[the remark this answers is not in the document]
saket: Done

PR #477, opened 2019-01-30
Refactor syncs into separate sync groups

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/163217054

### DIFF
--- a/app/src/main/java/org/simple/clinic/sync/SyncWorker.kt
+++ b/app/src/main/java/org/simple/clinic/sync/SyncWorker.kt
@@ -1,15 +1,28 @@
 package org.simple.clinic.sync
 
 import android.content.Context
+import androidx.work.Data
 import androidx.work.Worker
 import androidx.work.WorkerParameters
 import org.simple.clinic.ClinicApp
 import javax.inject.Inject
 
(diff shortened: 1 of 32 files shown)


app/src/main/java/org/simple/clinic/sync/SyncWorker.kt
```diff
@@ -1,23 +1,45 @@
...
 import androidx.work.WorkerParameters
 import org.simple.clinic.ClinicApp
 import javax.inject.Inject
 
-class SyncWorker(context: Context, workerParams: WorkerParameters) : Worker(context, workerParams) {
+class SyncWorker(context: Context, private val workerParams: WorkerParameters) : Worker(context, workerParams) {
 
   companion object {
     const val TAG = "patient-sync"
+    private const val NO_GROUP_ID = "no-group-id"
```
saket: Can we rename this to `NO_GROUP_ID_FOR_BACKWARD_COMPATIBILITY` or something? I wasn't able to understand on my own why we would need this until you explained the reason.

saket: Feel free to do this in a later commit

PR #480, opened 2019-02-05
Change rules for showing schedule appointment sheet

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/163080975

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
@@ -66,6 +66,8 @@ class PatientSummaryScreenController @Inject constructor(
     val replayedEvents = ReplayUntilScreenIsDestroyed(events)
         .compose(ReportAnalyticsEvents())
         .compose(mergeWithPatientSummaryChanges())
+        .compose(mergeWithBloodPressureSaves())
+        .compose(mergeWithAllBloodPressuresDeleted())
         .replay()
 
     return Observable.mergeArray(
@@ -75,11 +77,13 @@ class PatientSummaryScreenController @Inject constructor(
         updateMedicalHistory(replayedEvents),
         openBloodPressureBottomSheet(replayedEvents),
         openPrescribedDrugsScreen(replayedEvents),
-        handleBackAndDoneClicks(replayedEvents),
         exitScreenAfterSchedulingAppointment(replayedEvents),
         openBloodPressureUpdateSheet(replayedEvents),
         patientSummaryResultChanged(replayedEvents),
-        showUpdatePhoneDialogIfRequired(replayedEvents))
+        showUpdatePhoneDialogIfRequired(replayedEvents),
+        showScheduleAppointmentSheet(replayedEvents),
+        goBackWhenBackClicked(replayedEvents),
+        goToHomeOnDoneClick(replayedEvents))
   }
 
   private fun reportViewedPatientEvent(events: Observable<UiEvent>): Observable<UiChange> {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -271,69 +305,104 @@ class PatientSummaryScreenController @Inject constructor(
         }
   }
 
-  private fun handleBackAndDoneClicks(events: Observable<UiEvent>): Observable<UiChange> {
-    val callers = events
-        .ofType<PatientSummaryScreenCreated>()
-        .map { it.caller }
-
+  private fun showScheduleAppointmentSheet(events: Observable<UiEvent>): Observable<UiChange> {
```
vinaysshenoy: The older method was handling all possible chains from `Back` and `Done` and was getting quite unruly, so I broke it up as a part of this PR. As a consequence, this diff is difficult to parse. Might be easier to open this file separately.

vinaysshenoy: Same goes for the `PatientSummaryScreenControllerTest` as well.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -271,69 +305,104 @@ class PatientSummaryScreenController @Inject constructor(
...
-        .switchMap(bpRepository::bloodPressureCount)
-        .map { recordedBpCount -> recordedBpCount == 0 }
+    val allBpsForPatientDeletedStream = events
+        .ofType<PatientSummaryAllBloodPressuresDeleted>()
+        .map { it.allBloodPressuresDeleted }
 
-    val shouldShowAppointmentSheet = Observables
-        .combineLatest(mergedBloodPressureSaves, allBpsForPatientDeleted) { wasBloodPressureSaved, recordedBpsAreDeleted ->
-          if (recordedBpsAreDeleted) false else wasBloodPressureSaved
+    val shouldShowScheduleAppointmentSheetOnBackClicksStream = Observables
```
saket: `shouldShowScheduleAppointmentSheetOnBackClicksStream` -> `showAppointmentSheetOnBackClicksStream`?

Sorry I'm just trying to reduce the words here; this name is so long 😅

saket: Oh and your other streams following this don't start with `should` so this becomes consistent

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
```diff
@@ -346,74 +346,92 @@ class PatientSummaryScreenControllerTest {
...
-      wasBloodPressureSaved: Boolean,
+  @Parameters(method = "patient summary callers")
+  fun `when there are no patient summary changes and all bps are not deleted, clicking on back must go back`(
       patientSummaryCaller: PatientSummaryCaller
   ) {
-    uiEvents.onNext(PatientSummaryScreenCreated(patientUuid, caller = patientSummaryCaller, screenCreatedTimestamp = Instant.now(clock)))
-    uiEvents.onNext(PatientSummaryRestoredWithBPSaved(wasBloodPressureSaved))
-    uiEvents.onNext(PatientSummaryDoneClicked())
+    val patientSummaryItems = mock<PatientSummaryItems>()
+    whenever(patientSummaryItems.hasItemChangedSince(any())).thenReturn(false)
```
saket: nitpick: you've grouped these two lines together in some functions and kept a line break in between in some functions. Please remove this break if you get a chance to touch this class again?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -66,6 +66,8 @@ class PatientSummaryScreenController @Inject constructor(
     val replayedEvents = ReplayUntilScreenIsDestroyed(events)
         .compose(ReportAnalyticsEvents())
         .compose(mergeWithPatientSummaryChanges())
+        .compose(mergeWithBloodPressureSaves())
+        .compose(mergeWithAllBloodPressuresDeleted())
```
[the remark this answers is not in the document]
vinaysshenoy: Yes, to me this indicates that the the transformer is going to add new events to the stream. Otherwise, I assume it's going to transform the stream and drop some of the source events in some fashion. Does that make sense?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -179,6 +183,36 @@ class PatientSummaryScreenController @Inject constructor(
     }
   }
 
+  private fun mergeWithBloodPressureSaves(): ObservableTransformer<UiEvent, UiEvent> {
+    return ObservableTransformer { events ->
+      val bloodPressureSaves = events
+          .ofType<PatientSummaryBloodPressureClosed>()
+          .startWith(PatientSummaryBloodPressureClosed(false))
```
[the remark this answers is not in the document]
vinaysshenoy: This event has been removed.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
vinaysshenoy: This stream actually computes a `Boolean` on whether we should show the sheet or not. The actual showing of the sheet is handled by another stream, named `showScheduleAppointmentSheetOnBackClicks`.

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
vinaysshenoy: There's no line break in the code. Is this related to GitHub's diff viewer?

saket: Yes, my bad sorry

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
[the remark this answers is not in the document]
saket: Agreed. But we should follow this convention elsewhere too. We'll now have a mix of multiple conventions and that's not good.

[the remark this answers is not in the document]
vinaysshenoy: True. In this case, I basically just followed the same convention which we had with `mergeWithPatientSummaryChanges`. `mergeWithBloodPressureSaves` has been deleted in a later commit. I can rename `mergeWithAllBloodPressuresDeleted` to something like `checkWhetherAllBloodPressuresAreDeleted` or something similar.

For what it's worth, I checked for usages of the `.compose` method on `ReplayUntilScreenIsDestroyed`, and I see that we are using a mix of "mergeWith" prefixed and non-prefixed throughout the codebase.

PR #487, opened 2019-02-11
Read user's location when displaying facilities

saket:

This feature isn't complete yet. This PR introduces,

- displaying the progress indicator in facility selection screen until location is received or it times out
- calculating facilities nearby user (the calculation formula will be added in a later PR)

### DIFF
--- a/app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
+++ b/app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
@@ -4,7 +4,6 @@ import android.annotation.SuppressLint
 import android.app.Activity
 import com.facebook.stetho.Stetho
 import com.squareup.leakcanary.LeakCanary
-import dagger.Provides
 import io.github.inflationx.viewpump.ViewPump
 import io.reactivex.Single
 import org.simple.clinic.activity.TheActivity
@@ -19,8 +18,6 @@ import org.simple.clinic.di.DebugAppComponent
 import org.simple.clinic.facility.FacilityRepository
 import org.simple.clinic.login.LoginModule
 import org.simple.clinic.login.applock.AppLockConfig
-import org.simple.clinic.registration.RegistrationConfig
-import org.simple.clinic.registration.RegistrationModule
 import org.simple.clinic.security.pin.BruteForceProtectionConfig
 import org.simple.clinic.security.pin.BruteForceProtectionModule
 import org.simple.clinic.sync.SyncScheduler
@@ -90,12 +87,6 @@ class DebugClinicApp : ClinicApp() {
         .loginModule(object : LoginModule() {
           override fun appLockConfig(): Single<AppLockConfig> {
             return Single.just(AppLockConfig(lockAfterTimeMillis = TimeUnit.SECONDS.toMillis(4)))
-          }
-        })
-        .registrationModule(object : RegistrationModule() {
-          @Provides
-          override fun registrationConfig(): Single<RegistrationConfig> {
-            return Single.just(RegistrationConfig(retryBackOffDelayInMinutes = 1))
           }
         })
         .crashReporterModule(object : CrashReporterModule() {
--- a/app/src/main/java/org/simple/clinic/facility/change/FacilityListItemBuilder.kt
+++ b/app/src/main/java/org/simple/clinic/facility/change/FacilityListItemBuilder.kt
@@ -1,11 +1,57 @@
 package org.simple.clinic.facility.change
 
 import org.simple.clinic.facility.Facility
+import org.simple.clinic.location.Coordinates
+import org.simple.clinic.util.Distance
+import org.simple.clinic.util.Haversine
+import org.simple.clinic.util.Kilometers
 
 object FacilityListItemBuilder {
 
-  fun build(facilities: List<Facility>, searchQuery: String): List<FacilityListItem> {
-    return facilities.map { uiModel(it, searchQuery) }
+  fun build(
+      facilities: List<Facility>,
+      searchQuery: String,
+      userLocation: Coordinates?,
+      proximityThreshold: Distance
+  ): List<FacilityListItem> {
+    val nearbyFacilities = when {
+      userLocation != null -> facilitiesNearbyUser(facilities, userLocation, proximityThreshold)
+      else -> emptyList()
+    }
+
+    // TODO: Generate section headers
+    val nearbyFacilityListItems = nearbyFacilities.map { uiModel(it, searchQuery) }
+    val allFacilityListItems = facilities.map { uiModel(it, searchQuery) }
+
+    return nearbyFacilityListItems + allFacilityListItems
+  }
+
+  fun build(
+      facilities: List<Facility>,
+      searchQuery: String
+  ): List<FacilityListItem> {
+    return build(
+        facilities = facilities,
+        searchQuery = searchQuery,
+        userLocation = null,
+        proximityThreshold = Kilometers(0.0))
+  }
+
+  private fun facilitiesNearbyUser(
+      facilities: List<Facility>,
+      userLocation: Coordinates,
+      proximityThreshold: Distance
+  ): List<Facility> {
+    val facilitiesToDistance = facilities
+        .filter { it.location != null }
+        .map { it to Haversine.distance(userLocation, it.location!!) }
+
+    val nearbyFacilities = facilitiesToDistance
+        .filter { (_, distance) -> distance <= proximityThreshold }
+
+    return nearbyFacilities
+        .sortedBy { (_, distance) -> distance.kilometers() }
+        .map { (facility, _) -> facility }
   }
 
   private fun uiModel(facility: Facility, searchQuery: String): FacilityListItem {
@@ -23,7 +69,7 @@ object FacilityListItemBuilder {
     val fullAddress = if (facility.streetAddress.isNullOrBlank()) {
       FacilityListItem.Address.WithoutStreet(district = facility.district, state = facility.state)
     } else {
-      FacilityListItem.Address.WithStreet(street = facility.streetAddress!!, district = facility.district, state = facility.state)
+      FacilityListItem.Address.WithStreet(street = facility.streetAddress, district = facility.district, state = facility.state)
     }
 
     return FacilityListItem(facility = facility, name = highlightedName, address = fullAddress)
--- a/app/src/main/java/org/simple/clinic/facility/change/FacilityListItemBuilder.kt
+++ b/app/src/main/java/org/simple/clinic/facility/change/FacilityListItemBuilder.kt
@@ -4,7 +4,11 @@ import org.simple.clinic.facility.Facility
 
 object FacilityListItemBuilder {
 
-  fun build(facility: Facility, searchQuery: String): FacilityListItem {
+  fun build(facilities: List<Facility>, searchQuery: String): List<FacilityListItem> {
+    return facilities.map { uiModel(it, searchQuery) }
+  }
+
+  private fun uiModel(facility: Facility, searchQuery: String): FacilityListItem {
     val canHighlight = searchQuery.isNotBlank() && facility.name.contains(searchQuery, ignoreCase = true)
 
     val highlightedName = if (canHighlight) {
--- a/app/src/main/java/org/simple/clinic/facility/change/FacilityListItemBuilder.kt
+++ b/app/src/main/java/org/simple/clinic/facility/change/FacilityListItemBuilder.kt
@@ -1,0 +1,27 @@
+package org.simple.clinic.facility.change
+
+import org.simple.clinic.facility.Facility
+
+object FacilityListItemBuilder {
+
+  fun build(facility: Facility, searchQuery: String): FacilityListItem {
+    val canHighlight = searchQuery.isNotBlank() && facility.name.contains(searchQuery, ignoreCase = true)
+
+    val highlightedName = if (canHighlight) {
+      FacilityListItem.Name.Highlighted(
+          name = facility.name,
+          highlightStart = facility.name.indexOf(searchQuery, ignoreCase = true),
+          highlightEnd = facility.name.indexOf(searchQuery, ignoreCase = true) + searchQuery.length)
+    } else {
+      FacilityListItem.Name.Plain(facility.name)
+    }
+
+    val fullAddress = if (facility.streetAddress.isNullOrBlank()) {
+      FacilityListItem.Address.WithoutStreet(district = facility.district, state = facility.state)
+    } else {
+      FacilityListItem.Address.WithStreet(street = facility.streetAddress!!, district = facility.district, state = facility.state)
+    }
+
+    return FacilityListItem(facility = facility, name = highlightedName, address = fullAddress)
+  }
+}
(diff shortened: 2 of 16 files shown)


app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
```diff
@@ -92,12 +89,6 @@ class DebugClinicApp : ClinicApp() {
             return Single.just(AppLockConfig(lockAfterTimeMillis = TimeUnit.SECONDS.toMillis(4)))
           }
         })
-        .registrationModule(object : RegistrationModule() {
-          @Provides
-          override fun registrationConfig(): Single<RegistrationConfig> {
-            return Single.just(RegistrationConfig(retryBackOffDelayInMinutes = 1))
-          }
-        })
```
saket: This wasn't needed because the actual values are the same.

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenEvents.kt
```diff
@@ -14,3 +15,7 @@ data class RegistrationFacilityClicked(val facility: Facility) : UiEvent {
 data class RegistrationFacilitySearchQueryChanged(val query: String): UiEvent {
   override val analyticsName = "Registration:Facility Selection:Search Query Changed"
 }
+
+data class RegistrationUserLocationUpdated(val location: LocationUpdate): UiEvent {
+  override val analyticsName = "Registration:Facility Selection:Location Updated ${location.javaClass.simpleName}"
```
saket: I should add a `:` after `Location Updated`.

app/src/main/java/org/simple/clinic/util/Distance.kt
```diff
@@ -0,0 +1,15 @@
+package org.simple.clinic.util
+
+sealed class Distance {
+  abstract fun meters(): Double
+  abstract fun kilometers(): Double
```
saket: `meters()` isn't needed yet, but I wanted to show an example of what we can have in the future apart from just `kilometers()`.

vinaysshenoy: I actually think a better way to do it is how `java.util.time.Duration` is designed:

```kotlin
data class Distance(val distance: Long, val unit: DistanceUnit) {
  private val millimeters = distance * unit.millimeterConversionFactor

  companion object {
    fun ofMillimeters(millimeters: Long) = Distance(millimeters, MILLIMETERS)
    fun ofMeters(meters: Long) = Distance(millimeters, METERS)
  }

  enum class DistanceUnit(val millimeterConversionFactor: Long) {
    MILLIMETERS(1),
    METERS(1_000),
    KILOMETERS(1_000_000)
  }
}
```

This will enable us to do things like adding two `Distance` instances, converting from one to another easily, etc.

saket: That was exactly my intention, but I wanted to use sealed classes instead of enums. The units can still be converted from one to another this way. I'll try improving this.

vinaysshenoy: Why though? Sealed classes are not a good fit for this use case. They are good when you want to have a single way to define a limited set of types that are related, but this doesn't seem like one of those cases.

saket: I like using sealed classes even when the properties of each type is the same. If enums are used in this case, we'll have to use two different values for representing them:

```kotlin
value = 15
unit = DistanceUnit.KILOMETER
```

With classes,

```kotlin
distance = Kilometer(15.0)
```

With enums, when returning them from a function, we'll have to combine them in some manner:

```kotlin
fun haversine(): Pair<Double, DistanceUnit> {
  return 15.0 to DistanceUnit.KILOMETER
}
```

With classes,

```kotlin
fun haversine(): Distance {
  return Kilometer(15.0)
}
```

For converting them from one unit to another, you'll have to do:

```kotlin
val (value, unit) = haversine(...)
val meters = unit.toMeters(value)
```

With classes, 

```kotlin
val distance = haversine(...)
val meters = distance.meters()
```

I don't think `TimeUnit` is a good reference for design. Using them introduces a lot of verbosity. `Duration` from Java 8 date-time is neater.

saket: Wait, I looked at the source of `Duration`. Your example is not how it's designed. I actually thought you were referencing `TimeUnit` which maintains the unit and the value separately.

saket: Is there any specific reason why you prefer referencing distance units through `Distance.of...()`? I like the idea of being able to create objects of each distance unit separately

```kotlin
fun haversine(): Distance {
  return Distance.ofKilometers(15.0)
}
```

vs,

```kotlin
fun haversine(): Distance {
  return Kilometers(15.0)
}
```

I understand your reason for easier conversion. I'll add a `meters` field to the base class so each unit subclass doesn't have to deal with conversion.

vinaysshenoy: > Wait, I looked at the source of `Duration`. Your example is not how it's designed. I actually thought you were referencing `TimeUnit` which maintains the unit and the value separately.

Oh yeah, I just meant the entire api style, like using `Distance.ofMeters()`, and not the actual implementation itself.

vinaysshenoy: Another reason for having a single class that uses the same values to represent it is we can make this `Distance` class `Comparable`. Doing this with sealed classes will be cumbersome. Representing this in this way also makes (de)serializing `Distance` instances very easy. Otherwise, you'll need to have a polymorphic (de)serializer of some sort that will have a growing like of pointless `when` statements.

app/src/main/java/org/simple/clinic/util/Haversine.kt
```diff
@@ -0,0 +1,11 @@
+package org.simple.clinic.util
+
+import org.simple.clinic.location.Coordinates
+
+object Haversine {
```
vinaysshenoy: Why do we need a separate object for this when we can either have a top level function in, say, `LocationMath.kt`? Or have it as a function in the `Coordinates`class?

```kotlin
data class Coordinates(

    @ColumnInfo(name = "latitude")
    val latitude: Double,

    @ColumnInfo(name = "longitude")
    val longitude: Double
) {
  fun haversine(other: Coordinates): Distance
}
```

app/src/test/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenControllerTest.kt
```diff
@@ -134,17 +138,46 @@ class RegistrationFacilitySelectionScreenControllerTest {
...
-    // TODO
+    val facilities = listOf(
+        PatientMocker.facility(name = "Facility 1"),
+        PatientMocker.facility(name = "Facility 2"))
+    whenever(facilityRepository.facilities(any())).thenReturn(Observable.just(facilities))
+    whenever(facilityRepository.recordCount()).thenReturn(Observable.just(facilities.size))
+    whenever(facilitySync.pullWithResult()).thenReturn(Single.just(FacilityPullResult.Success()))
+
+    whenever(locationRepository.streamUserLocation(any())).thenReturn(
+        Observable
```
vinaysshenoy: Instead of doing this, why don't we use a `PublishSubject<LocationUpdate>` instead?

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenEvents.kt
saket: okay

app/src/main/java/org/simple/clinic/util/Distance.kt
saket: > Another reason for having a single class that uses the same values to represent it is we can make this `Distance` class `Comparable`.

That should still work with subclasses because the raw value will stay inside `Distance`. In fact, there's already a `compareTo()` function. We just need to mark the class as `Comparable`.

> Representing this in this way also makes (de)serializing `Distance` instances very easy. Otherwise, you'll need to have a polymorphic (de)serializer of some sort that will have a growing like of pointless `when` statements.

True, but let's worry about this when we reach there? I know that Moshi is already working on adding support for sealed classes so we won't have to deal with this very soon. https://github.com/square/moshi/pull/741

saket: For the record: I'll change `Distance` to use factory functions for creating different units instead of using subclasses. 

sad_parrot.gif

app/src/main/java/org/simple/clinic/facility/change/FacilityListItemBuilder.kt
```diff
@@ -1,11 +1,57 @@
...
+  ): List<Facility> {
+    val facilitiesToDistance = facilities
+        .filter { it.location != null }
+        .map { it to Haversine.distance(userLocation, it.location!!) }
+
+    val nearbyFacilities = facilitiesToDistance
+        .filter { (_, distance) -> distance <= proximityThreshold }
+
+    return nearbyFacilities
+        .sortedBy { (_, distance) -> distance.kilometers() }
```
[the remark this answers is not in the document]
saket: I realized I could just make it agnostic of the unit by making `Distance` implement `Comparable. So this operator becomes simpler:

```
sortedBy { (_, distance) -> distance }
```

app/src/main/java/org/simple/clinic/util/Haversine.kt
saket: `Coordinates#haversine()` sounds much better. I'll do this.

app/src/test/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenControllerTest.kt
saket: That is indeed more understandable. I'll do this.

app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
```diff
@@ -4,7 +4,6 @@ import android.annotation.SuppressLint
 import android.app.Activity
```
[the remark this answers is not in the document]
saket: Whoops, fixed

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenEvents.kt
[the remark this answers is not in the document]
saket: I see. I would keep the name local to this class though. What do you think?

```kotlin
override val analyticsName = "Registration:Facility Selection:Location Updated ${when (location) {
  is LocationUpdate.TurnedOff -> "TurnedOff"
  is LocationUpdate.Available -> "Available"
}}"
```

app/src/test/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenControllerTest.kt
```diff
@@ -40,13 +48,29 @@ class RegistrationFacilitySelectionScreenControllerTest {
...
+  private val configTemplate = RegistrationConfig(
+      retryBackOffDelayInMinutes = 0,
+      locationListenerExpiry = Duration.ofSeconds(0),
+      locationUpdateInterval = Duration.ofSeconds(0))
+  private val configProvider = BehaviorSubject.createDefault(configTemplate)
+
   @Before
   fun setUp() {
-    controller = RegistrationFacilitySelectionScreenController(facilitySync, facilityRepository, userSession, registrationScheduler)
+    RxJavaPlugins.setComputationSchedulerHandler { testComputationScheduler }
```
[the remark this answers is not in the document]
saket: TODO: Discuss this in person:
We're now starting to do this more often. It feels like it should be obvious by now that we want to control time for time related operators in the controller. Thoughts?

app/src/test/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenControllerTest.kt
```diff
@@ -59,26 +83,81 @@ class RegistrationFacilitySelectionScreenControllerTest {
...
+  }
+
+  @Test
+  fun `when screen is started then location should be fetched`() {
+    whenever(facilityRepository.recordCount()).thenReturn(Observable.never())
+    whenever(locationRepository.streamUserLocation(any())).thenReturn(Observable.never())
+
+    uiEvents.onNext(ScreenCreated())
+
+    verify(locationRepository).streamUserLocation(any())
```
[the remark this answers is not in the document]
saket: Makes sense. Updating this.

app/src/test/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenControllerTest.kt
```diff
@@ -59,26 +83,81 @@ class RegistrationFacilitySelectionScreenControllerTest {
...
+    val facilities = emptyList<Facility>()
+    whenever(facilityRepository.facilities()).thenReturn(Observable.just(facilities))
+    whenever(facilityRepository.recordCount()).thenReturn(Observable.just(facilities.size))
+    whenever(facilitySync.pullWithResult()).thenReturn(Single.just(FacilityPullResult.Success()))
+    whenever(locationRepository.streamUserLocation(any())).thenReturn(Observable.never())
+
+    uiEvents.onNext(ScreenCreated())
+    verify(screen).showProgressIndicator()
+
+    testComputationScheduler.advanceTimeBy(6, TimeUnit.SECONDS)
```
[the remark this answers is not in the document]
saket: TODO: Discuss in person: 
Sure, what kind of params do you suggest we use?

app/src/main/java/org/simple/clinic/util/Haversine.kt
saket: I just realized that I've implemented Haversine in a later PR. Do you mind if I do this in that PR?

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenEvents.kt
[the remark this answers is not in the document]
vinaysshenoy: Works.

app/src/test/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenControllerTest.kt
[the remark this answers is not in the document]
vinaysshenoy: Nothing specific; Just a couple of different values to ensure that the time limit is not hardcoded in the controller and is picked up from the config.

app/src/main/java/org/simple/clinic/util/Haversine.kt
vinaysshenoy: Yeah, feel free to do that.

app/src/test/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenControllerTest.kt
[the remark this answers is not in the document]
saket: Doen

saket:

vinaysshenoy:
@saket a9dacd4e8921683b76affab9de4687a0144845aa has broken tests. Please check?

saket:
@vinaysshenoy whoops I forgot to check in some changes resulting in the tests passing locally but not on CI. Fixing this.

app/src/main/java/org/simple/clinic/util/Haversine.kt
saket: This is done and you'll be able to see this in a future PR

PR #489, opened 2019-02-13
Remove IN_FLIGHT status from SyncStatus

rakshakhegde:
https://www.pivotaltracker.com/story/show/161802996

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
@@ -861,11 +861,481 @@ class DatabaseMigrationAndroidTest {
 
     assertThat(lastFacilityPullToken.get()).isEqualTo(None)
   }
+
+  @Test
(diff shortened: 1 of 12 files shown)


app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -861,11 +861,481 @@ class DatabaseMigrationAndroidTest {
...
+    val columnCount = 12
+
+    val db_v28 = helper.createDatabase(version = 28)
+    db_v28.assertColumnCount(tableName = tableName, expectedCount = columnCount)
+
+    db_v28.execSQL("""
+      INSERT INTO "MedicalHistory" VALUES(
+        'uuid',
+        'patientUuid',
+        0,
```
vinaysshenoy: We are using an enum `Answer` now instead of booleans. I assume this is still working because it is inserted as a string, but can you please change this to use the enum values to retain consistency between the database version and the schema?

PR #490, opened 2019-02-13
Initial implementation of SyncIndicatorView

sanchita:
**Pivotol story** -> https://www.pivotaltracker.com/story/show/163755918
**Part 1:**
This PR adds the `SyncIndicatorView` and `SyncIndicatorViewController`
Also, it does a basic implementation of accessing the `SyncProgress` and updating indicator state based on that.

### DIFF
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorState.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorState.kt
@@ -1,0 +1,8 @@
+package org.simple.clinic.sync.indicator
+
+sealed class SyncIndicatorState {
+  object ConnectToSync : SyncIndicatorState()
+  object Synced : SyncIndicatorState()
+  object SyncPending : SyncIndicatorState()
+  object Syncing : SyncIndicatorState()
+}
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorStatusCalculator.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorStatusCalculator.kt
@@ -1,38 +1,46 @@
 package org.simple.clinic.sync.indicator
 
+import android.annotation.SuppressLint
 import com.f2prateek.rx.preferences2.Preference
 import io.reactivex.schedulers.Schedulers.io
 import org.simple.clinic.di.AppScope
 import org.simple.clinic.sync.DataSync
+import org.simple.clinic.sync.LastSyncedState
 import org.simple.clinic.sync.SyncGroup
-import org.simple.clinic.sync.SyncProgress
+import org.simple.clinic.sync.SyncProgress.FAILURE
+import org.simple.clinic.sync.SyncProgress.SUCCESS
+import org.simple.clinic.sync.SyncProgress.SYNCING
 import org.simple.clinic.util.Just
-import org.simple.clinic.util.Optional
-import org.threeten.bp.Clock
+import org.simple.clinic.util.UtcClock
 import org.threeten.bp.Instant
 import javax.inject.Inject
-import javax.inject.Named
 
 @AppScope
 class SyncIndicatorStatusCalculator @Inject constructor(
     dataSync: DataSync,
-    clock: Clock,
-    @Named("last_frequent_sync_succeeded_timestamp") private val lastSyncSuccessTimestamp: Preference<Optional<Instant>>,
-    @Named("last_frequent_sync_result") private val lastSyncProgress: Preference<Optional<SyncProgress>>
+    utcClock: UtcClock,
+    private val lastSyncProgress: Preference<LastSyncedState>
 ) {
   init {
-    saveSyncResults(dataSync, clock)
+    saveSyncResults(dataSync, utcClock)
   }
 
-  private fun saveSyncResults(dataSync: DataSync, clock: Clock) {
-    dataSync
+  @SuppressLint("CheckResult")
+  private fun saveSyncResults(dataSync: DataSync, utcClock: UtcClock) {
+    val syncResultsStream = dataSync
         .streamSyncResults()
         .distinctUntilChanged()
+
+    syncResultsStream
         .subscribeOn(io())
         .filter { (syncGroup, _) -> syncGroup == SyncGroup.FREQUENT }
-        .doOnNext { (_, result) -> lastSyncProgress.set(Just(result)) }
-        .filter { (_, result) -> result == SyncProgress.SUCCESS }
-        .map { lastSyncSuccessTimestamp.set(Just(Instant.now(clock))) }
-        .subscribe()
+        .map { (_, progress) ->
+          var updatedState = lastSyncProgress.get().copy(lastSyncProgress = Just(progress))
+          when (progress) {
+            SUCCESS -> updatedState.copy(lastSyncSuccessTimestamp = Just(Instant.now(utcClock)))
+            FAILURE, SYNCING -> updatedState
+          }
+        }
+        .subscribe { lastSyncProgress.set(it) }
   }
 }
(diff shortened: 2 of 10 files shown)


app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorState.kt
```diff
@@ -0,0 +1,8 @@
+package org.simple.clinic.sync.indicator
```
saket: Unrelated to this PR, but I was looking at `DataSync` and I think we should consider using a default value for `syncProgress` with a `BehaviorSubject` instead of `Publish` because it looks like you're already changing the value to `SUCCESS` when sync completes. 

If you find using `SUCCESS` uncomfortable then we could rename it to `IDLE`.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -1,21 +1,53 @@
...
-    return Observable.never()
+    return updateIndicatorView(replayedEvents)
+  }
+
+  private fun updateIndicatorView(events: Observable<UiEvent>): Observable<UiChange> {
+    val syncResultStream = lastSyncGroupResult.asObservable().distinctUntilChanged()
+
+    return Observables.combineLatest(
+        events.ofType<SyncIndicatorViewCreated>(),
+        syncResultStream.filterAndUnwrapJust()) { _, result ->
```
saket: Beware of using a `filter` before `combineLatest`. You'll receive potentially stale values from `combineLatest` if the streams are filtered. 

Tell me if I'm wrong, but it looks like you want to filter after combining the streams here. What happens if the sync result is cleared? Here's my attempted visualization of the streams:

```
viewCreated: SyncIndicatorViewCreated
syncResultStream: None
combineLatest: --

viewCreated: SyncIndicatorViewCreated
syncResultStream: Just(SYNCING)
combineLatest: Pair(SyncIndicatorViewCreated, SYNCING)

viewCreated: SyncIndicatorViewCreated
syncResultStream: Just(SUCCESS)
combineLatest: Pair(SyncIndicatorViewCreated, SUCCESS)

viewCreated: SyncIndicatorViewCreated
syncResultStream: None
combineLatest: Pair(SyncIndicatorViewCreated, SUCCESS)
```

Was I able to explain this?

app/src/test/java/org/simple/clinic/sync/indicator/SyncIndicatorViewControllerTest.kt
```diff
@@ -0,0 +1,71 @@
...
+  private val syncResultPreference = mock<Preference<Optional<SyncProgress>>>()
+
+  lateinit var controller: SyncIndicatorViewController
+  private val syncProgressStream = PublishSubject.create<Optional<SyncProgress>>()
+  private val syncTimestampStream = PublishSubject.create<Optional<Instant>>()
+
+  private val uiEvents = PublishSubject.create<UiEvent>()
+
+  private val indicator = mock<SyncIndicatorView>()
+  @Before
```
saket: Line break please

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorState.kt
sanchita: Actually the default value for sync in terms of the indicator is not `SUCCESS`. It is `PENDING`.
Also, `DataSync` only comes into picture when we are actually syncing and not when we are scheduling syncs. So `SyncProgress` should not really have a default state. It should only be used to denote sync status when syncing starts.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
```diff
@@ -28,4 +28,8 @@ class SyncIndicatorView(context: Context, attrs: AttributeSet) : LinearLayout(co
   }
 
   private fun screenCreates() = Observable.just(SyncIndicatorViewCreated)
+
+  fun updateState(syncState: SyncIndicatorState) {
+    TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
```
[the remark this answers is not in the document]
sanchita: This will be removed int he next PR

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: This makes sense. I will do the necessary changes.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorStatusCalculator.kt
```diff
@@ -4,35 +4,39 @@ import com.f2prateek.rx.preferences2.Preference
...
         .streamSyncResults()
         .distinctUntilChanged()
+
+    syncResultsStream
         .subscribeOn(io())
         .filter { (syncGroup, _) -> syncGroup == SyncGroup.FREQUENT }
-        .doOnNext { (_, result) -> lastSyncProgress.set(Just(result)) }
-        .filter { (_, result) -> result == SyncProgress.SUCCESS }
-        .map { lastSyncSuccessTimestamp.set(Just(Instant.now(clock))) }
+        .doOnNext { (_, results) ->
```
saket: `doOnNext()` operators are used for side effects like logging, etc. You're performing the primary job of this class here. :)

Just move all of this to `subscribe`? Feel free to do this in your next PR.

saket: The indentation looks wrong. Maybe do an auto-format?

saket: Can `results` be called `progress` instead?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorStatusCalculator.kt
```diff
@@ -4,35 +4,39 @@ import com.f2prateek.rx.preferences2.Preference
...
+            var updatedState = lastSyncProgress.get().copy(lastSyncProgress = Just(results))
+            if (results == SyncProgress.SUCCESS) {
+              updatedState = updatedState.copy(lastSyncSuccessTimestamp = Just(Instant.now(clock)))
+            }
```
saket: I really prefer avoiding variables. It'll be nice if we can do `val` everywhere. 

```kotlin
val updatedState = when (progress) {
  SUCCESS ->
    lastSyncProgress.get().copy(
        lastSyncProgress = Just(progress),
        lastSyncSuccessTimestamp = Just(Instant.now(clock)))

  SYNCING, FAILURE ->
    lastSyncProgress.get().copy(lastSyncProgress = Just(progress))
}
```

Using an exhaustive `when` will also give us safety to enum changes.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorStatusCalculator.kt
```diff
@@ -4,35 +4,39 @@ import com.f2prateek.rx.preferences2.Preference
...
-import org.simple.clinic.util.Optional
 import org.threeten.bp.Clock
 import org.threeten.bp.Instant
 import javax.inject.Inject
-import javax.inject.Named
 
 @AppScope
 class SyncIndicatorStatusCalculator @Inject constructor(
     dataSync: DataSync,
     clock: Clock,
```
saket: We've replaced `Clock` with `UtcClock` and `UserClock`. I just realized that Dagger will throw an error once you start using `SyncIndicatorStatusCalculator` somewhere.

PR #493, opened 2019-02-15
Setup location access using FusedLocationProvider

saket:

### DIFF
--- a/app/src/main/java/org/simple/clinic/location/LocationRepository.kt
+++ b/app/src/main/java/org/simple/clinic/location/LocationRepository.kt
@@ -1,13 +1,84 @@
 package org.simple.clinic.location
 
+import android.annotation.SuppressLint
+import android.app.Application
+import android.location.Location
+import com.google.android.gms.location.FusedLocationProviderClient
+import com.google.android.gms.location.LocationAvailability
+import com.google.android.gms.location.LocationCallback
+import com.google.android.gms.location.LocationRequest
+import com.google.android.gms.location.LocationResult
+import com.google.android.gms.location.LocationServices
 import io.reactivex.Observable
 import org.threeten.bp.Duration
 import javax.inject.Inject
 
-class LocationRepository @Inject constructor() {
+typealias LocationProvider = FusedLocationProviderClient
+
+@SuppressLint("MissingPermission")
+class LocationRepository @Inject constructor(private val appContext: Application) {
+
+  data class LocationStatus(val isUsable: Boolean)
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/simple/clinic/location/LocationRepository.kt
```diff
@@ -1,13 +1,84 @@
...
+  private fun locationUpdates(provider: LocationProvider, request: LocationRequest): Observable<Coordinates> {
+    val locationChanges = Observable.create<LocationResult> { emitter ->
+      val locationCallback = object : LocationCallback() {
+        override fun onLocationResult(result: LocationResult) {
+          emitter.onNext(result)
+        }
+      }
+
+      emitter.setCancellable { provider.removeLocationUpdates(locationCallback) }
+      provider.requestLocationUpdates(request, locationCallback, appContext.mainLooper)
```
vinaysshenoy: Because of this, all emissions from here will be emitted on the main thread. This is not evident to users of this class. Is there something we can do to make this more evident?

app/src/main/java/org/simple/clinic/location/LocationRepository.kt
```diff
@@ -1,13 +1,84 @@
...
+
+    return locationChanges
+        .filter { it.lastLocation != null }
+        .map { Coordinates(it.lastLocation.latitude, it.lastLocation.longitude) }
+        .startWith(lastKnownLocation(provider))
+  }
+
+  private fun lastKnownLocation(provider: LocationProvider): Observable<Coordinates> {
+    return Observable.create<Coordinates> { emitter ->
+      provider.lastLocation.addOnSuccessListener { location: Location? ->
```
vinaysshenoy: Do we need to remove this listener?

saket: I tried, but there's no way to remove a listener or cancel a Task.

saket: Do you have other ideas?

vinaysshenoy: No, let's leave it as is for now.

app/src/main/java/org/simple/clinic/location/LocationRepository.kt
saket: Discussed in person: I'm doing this in an upcoming PR

saket:

PR #494, opened 2019-02-16
 Bugfix: mark patients as pending-synced when they're marked as dead

saket:


app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -125,9 +124,13 @@ class PatientRepository @Inject constructor(
 
   fun updatePatientStatusToDead(patientUuid: UUID): Completable {
     return Completable.fromAction {
-      database.patientDao()
-          .updatePatientStatus(patientUuid, PatientStatus.DEAD)
-      setSyncStatus(listOf(patientUuid), PENDING)
+      database
```
vinaysshenoy: Do you want to add a test for this?

saket: Completely forgot this. I'll do this.

saket: Done

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -488,19 +491,49 @@ class PatientRepositoryAndroidTest {
...
-    assertThat(searchResult).isEmpty()
+    assertThat(deadPatient.status).isEqualTo(PatientStatus.DEAD)
+  }
+
+  @Test
+  fun when_patient_is_marked_dead_they_should_marked_as_pending_sync() {
+    val timeOfCreation = Instant.now(testClock)
+
+    val patient = patientRepository
+        .saveOngoingEntry(testData.ongoingPatientEntry(fullName = "Ashok Kumar"))
```
vinaysshenoy: We already have a `TestData#patientProfile` which we should be using instead of creating it via an `OngoingPatientEntry`. Should we use that instead?

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -488,19 +491,49 @@ class PatientRepositoryAndroidTest {
...
+    testClock.advanceBy(Duration.ofDays(365))
+    val timeOfDeath = Instant.now(testClock)
+
+    patientRepository.updatePatientStatusToDead(patient.uuid).blockingAwait()
+    val deadPatient: Patient = patientRepository.patient(patient.uuid)
+        .unwrapJust()
+        .blockingFirst()
+
+    assertThat(deadPatient.syncStatus).isEqualTo(SyncStatus.PENDING)
+    assertThat(deadPatient.updatedAt).isNotEqualTo(timeOfCreation)
```
vinaysshenoy: Is this assertion really necessary since we are already asserting that it is equal to the `timeOfDeath` immediately after?

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
saket: I considered doing that but it only saves 1 line:

```kotlin
val patientProfile = testData.patientProfile(patientName = "Ashok Kumar")
val patient = patientProfile.patient
patientRepository.save(listOf(patientProfile)).blockingAwait()
```

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
saket: Discussed in person that `timeOfCreation` can be equal to `timeOfDeath` if we forget to advance time.

PR #495, opened 2019-02-19
Add analytics event for reporting timing events

vinaysshenoy:
Pivotal Story: https://www.pivotaltracker.com/story/show/164005123

### DIFF
--- a/app/src/main/java/org/simple/clinic/analytics/Analytics.kt
+++ b/app/src/main/java/org/simple/clinic/analytics/Analytics.kt
@@ -1,7 +1,14 @@
 package org.simple.clinic.analytics
 
 import android.net.NetworkCapabilities
-import android.net.NetworkCapabilities.*
+import android.net.NetworkCapabilities.TRANSPORT_BLUETOOTH
+import android.net.NetworkCapabilities.TRANSPORT_CELLULAR
+import android.net.NetworkCapabilities.TRANSPORT_ETHERNET
+import android.net.NetworkCapabilities.TRANSPORT_LOWPAN
+import android.net.NetworkCapabilities.TRANSPORT_VPN
+import android.net.NetworkCapabilities.TRANSPORT_WIFI
+import android.net.NetworkCapabilities.TRANSPORT_WIFI_AWARE
+import org.threeten.bp.Duration
 import java.util.UUID
 
 object Analytics {
@@ -93,6 +100,17 @@ object Analytics {
             "transport" to networkTransportType,
             "downstreamKbps" to downstreamBandwidthKbps,
             "upstreamKbps" to upstreamBandwidthKbps
+        ))
+      }
+    }
+  }
+
+  fun reportTimeTaken(operationName: String, timeTaken: Duration) {
+    reporters.forEach {
+      it.safely("Error reporting time taken event") {
+        createEvent("TimeTaken", mapOf(
+            "operationName" to operationName,
+            "timeTakenInMillis" to timeTaken.toMillis()
         ))
       }
     }
--- a/app/src/test/java/org/simple/clinic/analytics/AnalyticsTest.kt
+++ b/app/src/test/java/org/simple/clinic/analytics/AnalyticsTest.kt
@@ -4,6 +4,7 @@ import com.google.common.truth.Truth.assertThat
 import org.junit.After
 import org.junit.Test
 import org.simple.clinic.analytics.MockAnalyticsReporter.Event
+import org.threeten.bp.Duration
 import java.util.UUID
 
 class AnalyticsTest {
@@ -60,6 +61,13 @@ class AnalyticsTest {
   }
 
   @Test
+  fun `when reporting a time taken event without any reporters, no error should be thrown`() {
+    Analytics.reportTimeTaken(
+        operationName = "test",
+        timeTaken = Duration.ofMillis(500L))
+  }
+
+  @Test
   fun `when a reporter fails when sending interaction events, no error should be thrown`() {
     Analytics.addReporter(FailingAnalyticsReporter())
     Analytics.reportUserInteraction("Test")
@@ -108,6 +116,12 @@ class AnalyticsTest {
   }
 
   @Test
+  fun `when a reporter fails sending a time taken event, no error should be thrown`() {
+    Analytics.addReporter(FailingAnalyticsReporter())
+    Analytics.reportTimeTaken("test", Duration.ofMillis(100L))
+  }
+
+  @Test
   fun `when setting the user id, the property must also be set on the reporters`() {
     val reporter = MockAnalyticsReporter()
     Analytics.addReporter(reporter)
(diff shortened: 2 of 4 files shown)


app/src/test/java/org/simple/clinic/analytics/TimingCollectorTest.kt
```diff
@@ -0,0 +1,306 @@
...
+    clock.advanceBy(advanceClockBy)
+    timingCollector.start(stageName)
+
+    assertThat(timingCollector.completed).isEmpty()
+    assertThat(timingCollector.ongoing["$name:$stageName"])
+        .isEqualTo(now.plus(advanceClockBy))
+  }
+
+  @Test
+  fun `starting and stopping a stage that has already completed must override it`() {
```
sanchita: AFAIK, both the duration for that stage will be reported to Heap and not overridden. Is that right?
Hence the word `override` is misleading here.

app/src/main/java/org/simple/clinic/analytics/TimingCollector.kt
```diff
@@ -0,0 +1,98 @@
...
+ * // Once the entire operation is complete, push higher level events to analytics
+ * collector.finish()
+ * ```
+ *
+ * **Note:** Since an operation can happen across multiple threads, this class has been designed
+ * with thread-safety in mind. So it uses immutable maps for tracking operations, and atomic
+ * primitives where necessary.
+ **/
+class TimingCollector(
+    val name: String,
```
sanchita: I am not sure what this `name` means?
Is it the `operationName` or `eventName`? 
Can we call it that if so?

app/src/test/java/org/simple/clinic/analytics/TimingCollectorTest.kt
```diff
@@ -0,0 +1,306 @@
...
+    clock.advanceBy(Duration.ofMinutes(1L))
+    timingCollector.start(stage2)
+
+    timingCollector.finish()
+
+    assertThat(reporter.receivedEvents).isEmpty()
+  }
+
+  @Test
+  fun `starting a stage after finishing the collector must have no effect`() {
```
sanchita: Does this mean that if in the same class we need to report multiple operations, we should create a new instance of `TimeCollector` each time?
I would like to understand this in person.

app/src/main/java/org/simple/clinic/analytics/TimingCollector.kt
sanchita: I also want to do this to have a clear indication of what an `operation` is and what a `stage` is in context to this class.

app/src/test/java/org/simple/clinic/analytics/TimingCollectorTest.kt
sanchita: Can we also add an assertion to check what the `totalTimeTakenForOperation` would be when this operation is finished?

app/src/main/java/org/simple/clinic/analytics/TimingCollector.kt
```diff
@@ -0,0 +1,98 @@
+package org.simple.clinic.analytics
```
sanchita: We could reword this commit. Its not very informative.
How about `Add a TimingCollector to report time taken for an operation`?

app/src/main/java/org/simple/clinic/analytics/TimingCollector.kt
```diff
@@ -0,0 +1,98 @@
...
+      ongoing += "$name:$stageName" to Instant.now(clock)
+    }
+  }
+
+  fun stop(stageName: String) {
+    if(finished.get().not()) {
+      val analyticsEventName = "$name:$stageName"
+      ongoing[analyticsEventName]?.let { stageStartedAt ->
+        val now = Instant.now(clock)
+        val timeTakenToCompleteStage = Duration.between(stageStartedAt, now)
```
sanchita: `Duration.between()` can return negative values.
Should we calculate absolute value here?

app/src/main/java/org/simple/clinic/analytics/TimingCollector.kt
```diff
@@ -0,0 +1,98 @@
...
+ * collector.stop("stop-by-facility")
+ *
+ * // Once the entire operation is complete, push higher level events to analytics
+ * collector.finish()
+ * ```
+ *
+ * **Note:** Since an operation can happen across multiple threads, this class has been designed
+ * with thread-safety in mind. So it uses immutable maps for tracking operations, and atomic
+ * primitives where necessary.
+ **/
```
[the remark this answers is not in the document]
vinaysshenoy: This documentation is written in [`KDoc`](https://kotlinlang.org/docs/reference/kotlin-doc.html), which is the standard for documenting Kotlin code. It follows Markdown syntax, so enclosing text in three backticks renders it as a code-formatted block in the IDE.

app/src/main/java/org/simple/clinic/analytics/TimingCollector.kt
vinaysshenoy: Good catch. I'll change it to an absolute duration.

app/src/main/java/org/simple/clinic/analytics/Analytics.kt
```diff
@@ -98,6 +105,18 @@ object Analytics {
     }
   }
 
+  fun reportTimeTaken(operationName: String, runId: String, timeTaken: Duration) {
+    reporters.forEach {
+      it.safely("Error reporting time taken event") {
+        createEvent("TimeTaken", mapOf(
+            "operationName" to operationName,
+            "runId" to runId,
+            "timeTakenMillis" to timeTaken.toMillis()
```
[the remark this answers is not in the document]
vinaysshenoy: Done.

app/src/test/java/org/simple/clinic/analytics/TimingCollectorTest.kt
vinaysshenoy: > AFAIK, both the duration for that stage will be reported to Heap and not overridden. Is that right?
> Hence the word `override` is misleading here.

Yes, I'll rework this a bit.

app/src/main/java/org/simple/clinic/analytics/TimingCollector.kt
```diff
@@ -0,0 +1,98 @@
...
+class TimingCollector(
```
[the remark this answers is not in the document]
vinaysshenoy: I'll call it `OperationTimeTracker`.

app/src/main/java/org/simple/clinic/analytics/TimingCollector.kt
vinaysshenoy: Done.

app/src/test/java/org/simple/clinic/analytics/TimingCollectorTest.kt
vinaysshenoy: I have removed the `.finish()` operation because it was getting hard to reason about what it's behaviour should be. I examined the reason why I added it, and I realised it can be replaced with the usage of another `OperationTimingTracker` and this makes things a lot simpler.

app/src/test/java/org/simple/clinic/analytics/TimingCollectorTest.kt
vinaysshenoy: Yes, let's talk about this in person.

vinaysshenoy: Discussed in person: Since the `finish()` operation has been removed, this test has been removed as well. The entire notion of "finishing" a timing tracker has been removed.

app/src/test/java/org/simple/clinic/analytics/AnalyticsTest.kt
```diff
@@ -59,6 +60,14 @@ class AnalyticsTest {
         upstreamBandwidthKbps = 100)
   }
 
+  @Test
+  fun `when reporting a time taken event without any reporters, no error should be shown`() {
```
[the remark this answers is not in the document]
vinaysshenoy: Done.

PR #503, opened 2019-02-21
Update sync indicator based on last synced state

sanchita:


saket:
@sanchita can you add a link to the pivotal story?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -40,18 +41,39 @@ class SyncIndicatorViewController @Inject constructor(
...
+            SUCCESS -> syncedSuccessState(stateStream.lastSyncSuccessTimestamp)
+            FAILURE -> SyncPending
+            SYNCING -> Syncing
+          }
+          { ui: Ui -> ui.updateState(indicatorState) }
+        }
+  }
+
+  private fun syncedSuccessState(timestamp: Instant?): SyncIndicatorState {
+    if (timestamp == null) {
```
vinaysshenoy: Can `timestamp` ever be `null` when this is invoked?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -40,18 +41,39 @@ class SyncIndicatorViewController @Inject constructor(
...
+      return Synced
+    }
+
+    val now = Instant.now(utcClock)
+    val difference = Math.max(0, now.toEpochMilli() - timestamp.toEpochMilli())
+    val timeSinceLastSync = Duration.ofMillis(difference)
+    val mostFrequentSyncInterval = enumValues<SyncInterval>().map { it.frequency }.min()!!
```
vinaysshenoy: Can you add these chained method calls on different lines?

```kotlin
val mostFrequentSyncInterval = enumValues<SyncInterval>()
        .map { it.frequency }
        .min()!!
```

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -40,18 +41,39 @@ class SyncIndicatorViewController @Inject constructor(
...
+
+    val maxDuration = Duration.ofHours(12)
```
vinaysshenoy: Rename `maxDuration` -> `maxIntervalForPendingSync` or something similar?

vinaysshenoy: Magic values like this should be injected via a `Config` object. Can you move this to a config object?

app/src/test/java/org/simple/clinic/sync/indicator/SyncIndicatorViewControllerTest.kt
```diff
@@ -54,19 +54,22 @@ class SyncIndicatorViewControllerTest {
...
   @Suppress("Unused")
   private fun `params for testing sync status indicator update`(): List<List<Any>> {
     return listOf(
-        listOf(SYNCING, Syncing),
-        listOf(SUCCESS, Synced),
-        listOf(FAILURE, SyncPending)
+        listOf(LastSyncedState(SYNCING), Syncing),
+        listOf(LastSyncedState(SUCCESS), Synced),
+        listOf(LastSyncedState(FAILURE), SyncPending),
+        listOf(LastSyncedState(SUCCESS, Instant.now(utcClock).minus(20, ChronoUnit.MINUTES)), SyncPending),
```
vinaysshenoy: Another reason why the 12 hour duration should be injected via a config object, is that this test is written under the assumption that the duration is 12 hours. If we change that duration in the future, we will also have to change these tests, which doesn't make sense.

Injecting it via a configuration will let us separate what we are testing from what the production configuration is.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -41,38 +41,57 @@ class SyncIndicatorViewController @Inject constructor(
...
+    val lastSyncedStateStream = lastSyncState
         .asObservable()
         .distinctUntilChanged()
-        .filter { it.lastSyncProgress != null }
 
-    return Observables
-        .combineLatest(screenCreated, syncStateStream)
+    val syncProgress = lastSyncedStateStream.filter { it.lastSyncProgress != null }
+
+    val defaultSyncIndicatorState = lastSyncedStateStream
```
vinaysshenoy: Can you move this chain above the `syncProgress` chain because the two aren't related.

vinaysshenoy: Rename `defaultSyncIndicatorState` -> `showDefaultSyncIndicatorState`?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -41,38 +41,57 @@ class SyncIndicatorViewController @Inject constructor(
...
+        .filter { it.lastSyncProgress == null }
+        .map { { ui: Ui -> ui.updateState(SyncPending) } }
+
+    val syncIndicatorState = Observables
```
vinaysshenoy: Rename `syncIndicatorState` -> `showSyncIndicatorState`?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: I realized this later and made the change in my next PR.
I will do it here itself.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: Good idea!

sanchita: Do you mind if I do this in the next PR?
It will involve adding a new `SyncIndicatorConfig`.

app/src/test/java/org/simple/clinic/sync/indicator/SyncIndicatorViewControllerTest.kt
sanchita: This will be done in the next PR.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: Done.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
vinaysshenoy: `maxIntervalSinceLastSync` still doesn't sound too obvious to me. What do you think about `intervalSinceLastSyncToConsiderFailed`? We can update this in the next PR.

It's okay to have long names. We have auto-suggestions, so having a longer name that is more *obvious* is preferable.

sanchita: With the help of @saket , we came up with `syncFailureThreshold`.

PR #504, opened 2019-02-21
Show recent patients UI on home screen

rakshakhegde:
https://www.pivotaltracker.com/story/show/164004483

This UI work is incomplete, so it's hidden behind a feature flag. So testing is required to check if it shows up as expected in QA build.

### DIFF
--- a/app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
+++ b/app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
@@ -114,7 +114,7 @@ class DebugClinicApp : ClinicApp() {
         .patientModule(object : PatientModule() {
           override fun providePatientConfig(): Observable<PatientConfig> {
             return super.providePatientConfig()
-                .map { it.copy(scanSimpleCardFeatureEnabled = true, showRecentPatients = true) }
+                .map { it.copy(scanSimpleCardFeatureEnabled = true) }
           }
         })
         .build()
--- a/app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
+++ b/app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
@@ -114,7 +114,7 @@ class DebugClinicApp : ClinicApp() {
         .patientModule(object : PatientModule() {
           override fun providePatientConfig(): Observable<PatientConfig> {
             return super.providePatientConfig()
-                .map { it.copy(scanSimpleCardFeatureEnabled = true) }
+                .map { it.copy(scanSimpleCardFeatureEnabled = true, showRecentPatients = true) }
           }
         })
         .build()
--- a/app/src/main/java/org/simple/clinic/patient/PatientConfig.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientConfig.kt
@@ -7,7 +7,11 @@ package org.simple.clinic.patient
  * */
 const val MAXIMUM_SQLITE_QUERY_LIMIT = 1000
 
-data class PatientConfig(val limitOfSearchResults: Int, val scanSimpleCardFeatureEnabled: Boolean) {
+data class PatientConfig(
+    val limitOfSearchResults: Int,
+    val scanSimpleCardFeatureEnabled: Boolean,
+    val showRecentPatients: Boolean = false
+) {
   init {
     if (limitOfSearchResults !in 1 until MAXIMUM_SQLITE_QUERY_LIMIT) {
       throw AssertionError("limit of search results must be within [1, 999]")
--- a/app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
+++ b/app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
@@ -6,21 +6,19 @@ import kotlinx.android.synthetic.main.recent_patient_item_view.view.*
 import org.simple.clinic.R
 import org.simple.clinic.patient.Gender
 
-class RecentPatientItem(private val data: Data) : Item<ViewHolder>() {
+class RecentPatientItem(
+    val title: String,
+    val lastBp: String,
+    val gender: Gender
+) : Item<ViewHolder>() {
 
   override fun getLayout(): Int = R.layout.recent_patient_item_view
 
   override fun bind(viewHolder: ViewHolder, position: Int) {
     viewHolder.itemView.apply {
-      recentpatient_item_title.text = data.title
-      recentpatient_item_last_bp.text = data.lastBp
-      recentpatient_item_gender.setImageResource(data.gender.displayIconRes)
+      recentpatient_item_title.text = title
+      recentpatient_item_last_bp.text = lastBp
+      recentpatient_item_gender.setImageResource(gender.displayIconRes)
     }
   }
-
-  data class Data(
-      val title: String,
-      val lastBp: String,
-      val gender: Gender
-  )
 }
--- a/app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
+++ b/app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
@@ -1,0 +1,12 @@
+package org.simple.clinic.patient.recent
+
+import com.xwray.groupie.Item
+import com.xwray.groupie.ViewHolder
+import org.simple.clinic.R
+
+class RecentPatientItem : Item<ViewHolder>() {
+  override fun getLayout(): Int = R.layout.recent_patient_item_view
+
+  override fun bind(viewHolder: ViewHolder, position: Int) {
+  }
+}
--- a/app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
+++ b/app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
@@ -2,11 +2,25 @@ package org.simple.clinic.patient.recent
 
 import com.xwray.groupie.Item
 import com.xwray.groupie.ViewHolder
+import kotlinx.android.synthetic.main.recent_patient_item_view.view.*
 import org.simple.clinic.R
+import org.simple.clinic.patient.Gender
 
-class RecentPatientItem : Item<ViewHolder>() {
+class RecentPatientItem(private val data: Data) : Item<ViewHolder>() {
+
   override fun getLayout(): Int = R.layout.recent_patient_item_view
 
   override fun bind(viewHolder: ViewHolder, position: Int) {
+    viewHolder.itemView.apply {
+      recentpatient_item_title.text = data.title
+      recentpatient_item_last_bp.text = data.lastBp
+      recentpatient_item_gender.setImageResource(data.gender.displayIconRes)
+    }
   }
+
+  data class Data(
+      val title: String,
+      val lastBp: String,
+      val gender: Gender
+  )
 }
--- a/app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
+++ b/app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
@@ -1,5 +1,6 @@
 package org.simple.clinic.patient.recent
 
+import android.annotation.SuppressLint
 import com.xwray.groupie.Item
 import com.xwray.groupie.ViewHolder
 import kotlinx.android.synthetic.main.recent_patient_item_view.view.*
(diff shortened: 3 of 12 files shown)


app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
```diff
@@ -15,12 +16,40 @@ class RecentPatientsView(context: Context, attrs: AttributeSet) : RecyclerView(c
...
     adapter = groupAdapter
 
-    groupAdapter.add(RecentPatientItem())
-    groupAdapter.add(RecentPatientItem())
-    groupAdapter.add(RecentPatientItem())
-    groupAdapter.add(RecentPatientItem())
-    groupAdapter.add(RecentPatientItem())
-    groupAdapter.add(RecentPatientItem())
-    groupAdapter.add(RecentPatientItem())
+    groupAdapter.add(RecentPatientItem(RecentPatientItem.Data(
```
sanchita: Please move adding this data for testing to a separate method.
You can also add a comment to it saying that this is only for testing until the controller is ready.

app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
```diff
@@ -114,7 +114,7 @@ class DebugClinicApp : ClinicApp() {
         .patientModule(object : PatientModule() {
           override fun providePatientConfig(): Observable<PatientConfig> {
             return super.providePatientConfig()
-                .map { it.copy(scanSimpleCardFeatureEnabled = true) }
+                .map { it.copy(scanSimpleCardFeatureEnabled = true, showRecentPatients = true) }
```
sanchita: We should keep this flag disabled everywhere until the development finishes.
Hence, this should be removed.

app/src/main/res/layout/screen_patients.xml
```diff
@@ -63,21 +63,6 @@
         android:src="@drawable/illustrations_homescreen"
         tools:ignore="ContentDescription" />
 
-      <FrameLayout
```
sanchita: As discussed, once you will use a duplicate xml for `screen_patients` for your development, this commit won't be necessary.

app/src/main/res/layout/screen_patients.xml
```diff
@@ -7,68 +7,70 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_below="@id/patients_scan_simple_card"
+        android:layout_centerHorizontal="true"
+        android:adjustViewBounds="true"
+        android:scaleType="fitXY"
+        android:src="@drawable/illustrations_homescreen"
+        tools:ignore="ContentDescription" />
+
+      <org.simple.clinic.patient.recent.RecentPatientsView
```
saket: There are two major problems here.

1. Having nested scrollable views that scroll in the same direction/axis is never a good idea.

2. A `RecyclerView` with height set to `wrap_content` is as good as a `LinearLayout` because there's no recycling happening here. 

A `RecyclerView` is an extremely complex layout that is built to specially handle recycling of items in a performant manner using various optimizations . By using `LinearLayout`, you're forcing RV to perform those optimizations and end up wasting them. Plus, if I see a `RecyclerView`, I expect things to recycle. I'm not sure how to put this in better words, but it's _wrong_.

Just use a `LinearLayout` and plug it with your adapter. Even if you plan to reuse this View in multiple screens, you're actually only going to reuse the adapter. There's no advantage of reusing a `RecyclerView` in two places.

Using a LinearLayout with an adapter requires just 4 lines. 

```kotlin
recentPatientItems.map { item ->
  val viewHolder = adapter.onCreateViewHolder(...)
  linearLayout.addView(viewHolder.itemView)
  viewHolder.render(item)
}
```

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
```diff
@@ -5,12 +5,30 @@ import android.util.AttributeSet
...
 class RecentPatientsView(context: Context, attrs: AttributeSet) : RecyclerView(context, attrs) {
 
+  @Inject
+  lateinit var config: Observable<PatientConfig>
+
   override fun onFinishInflate() {
     super.onFinishInflate()
+    TheActivity.component.inject(this)
+
+    config.subscribeBy(onNext = {
```
sanchita: Move the config to either `HomeScreen` or `HomePagerAdapter`

app/src/main/res/layout/screen_patients.xml
saket: I was thinking that it's a good idea that you want to reuse one View in both `patients screen` and `full list of recent patients` screen, but I think it's okay to duplicate some logic. I'm sure in the future we'll get different kind of requirements for behaviors in both the screens and you'll find better about it. 

This has already happened with facility selection in the past. There's one screen that's used during nurse registration and one post login (from home screen). They had exactly the same behavior in the beginning, but they diverged later. They're merging again now, but my point is that it's okay to not reuse recent patients list.

rakshakhegde: As discussed in person, we'll keep this as RecyclerView itself.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
```diff
@@ -2,11 +2,25 @@ package org.simple.clinic.patient.recent
...
 
   override fun bind(viewHolder: ViewHolder, position: Int) {
+    viewHolder.itemView.apply {
+      recentpatient_item_title.text = data.title
+      recentpatient_item_last_bp.text = data.lastBp
+      recentpatient_item_gender.setImageResource(data.gender.displayIconRes)
+    }
   }
+
+  data class Data(
```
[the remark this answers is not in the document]
saket: Alternatively, I think this inner class can be removed entirely. The suggested way by `Groupie` is to keep the values as constructor params of the item class (which is `RecentPatientItem` in this case). 

```kotlin
class RecentPatientItem(
    val title: String,
    val lastBp: String,
    val gender: Gender
) : Item<ViewHolder>() {
```

If the number of params start becoming larger, you can always break them down into more abstractions. See `FacilityListItem`.

app/src/main/res/layout/screen_patients.xml
saket: ^ that's not what we discussed in person, but it's okay if we're in disagreement.

rakshakhegde: My reasons for keeping this: it's more idiomatic to use a component that's built for templatized rows, i.e., RecyclerView, instead of shoehorning `LinearLayout` and making it work with RecyclerView's adapters.

There's no degradation of performance as you've stated. Once we state RecyclerView's height as wrap_content, it does not try to needlessly recycle the views. All the child views are created in one shot when the view is created.

Also, if we start using LinearLayout and RecyclerView for the same purposes, we lose consistency across the project. Breaking this does not make sense for this use case as recyclerview can handle all those scenarios.

I realize the change to LinearLayout is just addition of couple of lines, but this breaks the obvious way of implementing an obvious UI component.

rakshakhegde: Also, this is what I thought we had discussed in person. I am sorry, if I didn't convey this properly.

app/src/main/res/layout/screen_patients.xml
```diff
@@ -70,6 +70,12 @@
       android:text="@string/patients_record_every_patient" />
   </FrameLayout>
 
+  <org.simple.clinic.patient.recent.RecentPatientsView
```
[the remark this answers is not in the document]
rakshakhegde: Thank you for the support. Got this config injected the `PatientsScreenKey` and it's working out :)

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
```diff
@@ -2,11 +2,25 @@ package org.simple.clinic.patient.recent
 
 import com.xwray.groupie.Item
 import com.xwray.groupie.ViewHolder
+import kotlinx.android.synthetic.main.recent_patient_item_view.view.*
 import org.simple.clinic.R
+import org.simple.clinic.patient.Gender
+
+class RecentPatientItem(private val data: Data) : Item<ViewHolder>() {
```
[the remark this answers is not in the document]
rakshakhegde: Will be adding this in a future PR when click events will be handled.

app/src/main/res/layout/screen_patients.xml
```diff
@@ -7,74 +7,84 @@
...
-    android:layout_below="@id/patients_scan_simple_card"
-    android:layout_centerHorizontal="true"
-    android:adjustViewBounds="true"
-    android:scaleType="fitXY"
-    android:src="@drawable/illustrations_homescreen"
-    tools:ignore="ContentDescription" />
+      <org.simple.clinic.widgets.OmegaCenterIconButton
+        android:id="@+id/patients_search_patients"
+        android:layout_width="match_parent"
+        android:layout_height="56dp"
```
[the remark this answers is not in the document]
rakshakhegde: this comes from existing code. I've changed 56dp but not 20dp for now.

app/src/main/res/layout/screen_patients.xml
```diff
@@ -7,74 +7,84 @@
...
+        android:layout_marginEnd="@dimen/spacing_16"
+        android:layout_marginStart="@dimen/spacing_16"
+        android:layout_marginTop="@dimen/spacing_16"
+        android:background="@drawable/button_outline_ripple"
+        android:drawablePadding="@dimen/spacing_8"
+        android:drawableStart="@drawable/ic_search_24dp"
+        android:fontFamily="sans-serif"
+        android:gravity="center"
+        android:paddingStart="20dp"
```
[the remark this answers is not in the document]
rakshakhegde: I've left this out for now.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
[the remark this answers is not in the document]
rakshakhegde: @saket This looks good.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
rakshakhegde: Added the config check to `PatientsScreenKey`.

app/src/main/java/org/simple/clinic/patient/PatientConfig.kt
```diff
@@ -7,7 +7,11 @@ package org.simple.clinic.patient
  * */
 const val MAXIMUM_SQLITE_QUERY_LIMIT = 1000
 
-data class PatientConfig(val limitOfSearchResults: Int, val scanSimpleCardFeatureEnabled: Boolean) {
+data class PatientConfig(
+    val limitOfSearchResults: Int,
+    val scanSimpleCardFeatureEnabled: Boolean,
+    val showRecentPatients: Boolean = false
```
sanchita: Set the value of the config `showRecentPatients` in `PatientModule` from where this config is provided.

app/src/main/res/layout/screen_patients.xml
```diff
@@ -10,7 +10,7 @@
   <org.simple.clinic.widgets.OmegaCenterIconButton
     android:id="@+id/patients_search_patients"
     android:layout_width="match_parent"
-    android:layout_height="56dp"
```
sanchita: You can leave this file unedited for the time being.
Once your feature completes these changes will anyway be merged from your `screen_patients_v2`

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
```diff
@@ -0,0 +1,24 @@
+package org.simple.clinic.patient.recent
+
+import com.xwray.groupie.Item
+import com.xwray.groupie.ViewHolder
+import kotlinx.android.synthetic.main.recent_patient_item_view.view.*
+import org.simple.clinic.R
+import org.simple.clinic.patient.Gender
+
+class RecentPatientItem(
+    val title: String,
```
sanchita: Is this the patient name?
Can you rename this to `patientName` or even `name`

app/src/main/res/layout/recent_patient_item_view.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.cardview.widget.CardView xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:app="http://schemas.android.com/apk/res-auto"
+  xmlns:tools="http://schemas.android.com/tools"
+  android:layout_width="match_parent"
+  android:layout_height="wrap_content"
+  android:layout_marginBottom="8dp"
```
sanchita: There is hard-coding across this file.
Please use `dimens` wherever you can.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
rakshakhegde: This includes patient name and age. Let's keep it as is?

rakshakhegde: I've split name and age into two different fields.

PR #505, opened 2019-02-21
Send timing events for patient search to analytics

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/164005123

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
@@ -74,7 +74,9 @@ class PatientRepository @Inject constructor(
 
   // TODO: Get user from caller.
   private fun sortByCurrentFacility(): ObservableTransformer<List<PatientSearchResult>, List<PatientSearchResult>> {
-    return ObservableTransformer { searchResults ->
+    return ObservableTransformer { upstream ->
+      val searchResults = upstream.replay().refCount()
+
       val currentFacilityUuidStream = userSession
           .requireLoggedInUser()
           .switchMap { facilityRepository.currentFacility(it) }
--- a/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
@@ -6,6 +6,7 @@ import io.reactivex.ObservableTransformer
 import io.reactivex.Single
 import io.reactivex.rxkotlin.Observables
 import org.simple.clinic.AppDatabase
+import org.simple.clinic.analytics.OperationTimingTracker
 import org.simple.clinic.di.AppScope
 import org.simple.clinic.facility.FacilityRepository
 import org.simple.clinic.patient.SyncStatus.DONE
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -49,10 +50,22 @@ class PatientRepository @Inject constructor(
   private var ongoingNewPatientEntry: OngoingNewPatientEntry = OngoingNewPatientEntry()
 
   fun search(name: String): Observable<List<PatientSearchResult>> {
+    val timingTracker = OperationTimingTracker("Search Patient", utcClock)
```
saket: There are some IDE inspection warnings in `OperationTimingTracker`. Can you please resolve them?

build.gradle
```diff
@@ -46,7 +46,7 @@ buildscript {
       playServicesAuth    : '16.0.1',
       playServicesVision  : '16.2.0',
       viewPump            : '1.0.0',
-      heap                : '0.7.1',
+      heap                : '0.8.3',
```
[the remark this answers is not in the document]
vinaysshenoy: I am assuming they mean 3.2.1+. I will create a separate PR for migrating the gradle plugin and confirm.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
saket: I was reading `OperationTimingTracker`'s documentation and found this:

> **Note:** Since an operation can happen across multiple threads, this class has been designed
> with thread-safety in mind. So it uses immutable maps for tracking operations, and atomic
> primitives where necessary.

Using immutable values doesn't entirely offer thread safety right? It's still possible for two threads to change a value at the same time and result in a race condition unless you synchronize the function calls. Does this make sense?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -49,10 +50,22 @@ class PatientRepository @Inject constructor(
...
+    val fuzzyFilterPatientNameAnalytics = "Fuzzy Filtering By Name"
+    val fetchPatientDetailsAnalytics = "Fetch Patient Details"
+    val sortByVisitedFacilityAnalytics = "Sort By Visited Facility"
+
     val patientUuidsMatchingName = database.patientSearchDao()
         .nameAndId(PatientStatus.ACTIVE)
         .toObservable()
-        .switchMapSingle { searchPatientByName.search(name, it) }
+        .switchMapSingle {
+          timingTracker.stop(fetchPatientNameAnalytics)
```
saket: I was getting confused why are we stopping before starting this stage until I realized the start is happening later in this function. Do you think we can move them closer? What if we move the `doOnSubscribe()` operator here, right before `switchMapSingle`?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -49,10 +50,22 @@ class PatientRepository @Inject constructor(
...
+          timingTracker.start(fuzzyFilterPatientNameAnalytics)
+          searchPatientByName.search(name, it)
+              .doOnSuccess { timingTracker.stop(fuzzyFilterPatientNameAnalytics) }
+        }
```
saket: Can this `doOnSuccess` be moved outside `switchMapSingle`? The indentation will look a bit cleaner.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: Yes, that was intentional. It isn't crucial for us report these events to Heap, so I felt it was alright that some events might occasionally not get reported to Heap while we avoid the expense of thread synchronisation.

My intention in making these immutable was only to avoid concurrent modification exceptions  for things that don't affect the actual functioning of the app. Does that make sense?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: I tried doing this, but because of the way we combine the streams (which involve early subscriptions) the timings were not being captured properly.

I'll take another stab at it.

saket: Ah right, the point of subscriptions are different.

PR #506, opened 2019-02-22
Suggest nearby facilities when changing facilities post-login

saket:
https://www.pivotaltracker.com/story/show/164135362

Visual changes:

1. Updated color of facility names to blue. This introduces a bug because highlighting of search query uses the same color and we did not realize that when building this. This has been fixed in a later commit.

2. Copy location based facility suggestions feature from `RegistrationFacilitySelectionScreen` into `FacilityChangeScreen`. This is not finished yet. I decided to not include the remaining commits to keep this PR small. 

This PR is slightly large because a lot of things (that have already been reviewed) were copy pasted from `RegistrationFacilitySelectionScreen` to `FacilityChangeScreen`.

@sanchita if you feel this PR should be shorter, I can remove the last 3 commits. I tried including only the 2nd & 3rd last commits and removing the last commit, but it's difficult to test them on device independently.


app/src/main/java/org/simple/clinic/location/Coordinates.kt
```diff
@@ -9,4 +10,34 @@ data class Coordinates(
...
+    private const val EARTH_RADIUS = 6371.0088
+
+    /**
+     * Haversine formula determines the great-circle distance between two points
+     * on a sphere given their longitudes and latitudes.
+     *
+     * Source: https://github.com/jasonwinn/haversine
+     * Modified to use with Kotlin.
+     */
+    fun haversineDistance(start: Coordinates, end: Coordinates): Distance {
```
saket: Deleting `Haversine` object and moving the calculation to `Coordinates.haversineDistance()` was @vinaysshenoy's feedback here: https://github.com/simpledotorg/simple-android/pull/487#discussion_r256693845

Not important for this PR, but we later realized that using this function directly in `FacilityListItemBuilder` is not a good idea from testability POV so we ended up moving this again to a class called `DistanceCalculator`.

app/src/main/java/org/simple/clinic/facility/change/FacilityChangeScreen.kt
```diff
@@ -140,11 +140,9 @@ class FacilityChangeScreen(context: Context, attrs: AttributeSet) : RelativeLayo
 
   fun showProgressIndicator() {
     progressView.visibility = VISIBLE
-    facilityRecyclerView.visibility = GONE
```
saket: The `RecyclerView` will now only get populated once the progress indicator is hidden from the controller so this is no longer needed.

app/src/main/java/org/simple/clinic/location/Coordinates.kt
sanchita: Calculating distance should not be a part of `Coordinates`.
Since you will do this as part of the next PR, I will leave the review of this.

app/src/main/java/org/simple/clinic/facility/change/FacilityChangeScreenController.kt
```diff
@@ -96,6 +98,23 @@ class FacilityChangeScreenController @Inject constructor(
...
   }
 
+  private fun showProgressForReadingLocation(events: Observable<UiEvent>): Observable<UiChange> {
+    val screenCreations = events.ofType<ScreenCreated>()
+    val locationPermissionChanges = events
+        .ofType<FacilityChangeLocationPermissionChanged>()
+        .map { it.result }
+
+    val showProgress = Observables.zip(screenCreations, locationPermissionChanges)
+        .filter { (_, permissionResult) -> permissionResult == GRANTED }
```
sanchita: Can't the `filter` happen before we `zip` i.e. in `locationPermissionChanges`?

sanchita: Also, that will mean that `showProgress` will always happen if the stream is active.
Does it make sense?

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
```diff
@@ -273,4 +273,65 @@ class FacilityChangeScreenControllerTest {
   fun `params for permission denials`(): List<RuntimePermissionResult> {
     return listOf(DENIED, NEVER_ASK_AGAIN)
   }
+
+  @Test
+  @Parameters("6", "7", "8")
```
sanchita: Can you also add `"5"` as a param here?
Just to make the test watertight

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
```diff
@@ -273,4 +273,65 @@ class FacilityChangeScreenControllerTest {
...
+    uiEvents.onNext(ScreenCreated())
+    uiEvents.onNext(FacilityChangeLocationPermissionChanged(GRANTED))
+    uiEvents.onNext(FacilityChangeLocationPermissionChanged(GRANTED))
+
+    verify(screen, times(1)).showProgressIndicator()
+  }
+
+  @Test
+  @Parameters(method = "params for permission denials")
+  fun `when screen starts and location permission is unavailable then progress indicator should not be shown`(
```
sanchita: This test checks if location permission was `denied` and not `unavailable` whereas the test name says "unavailable".
Can you check and correct this?

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
```diff
@@ -273,4 +273,65 @@ class FacilityChangeScreenControllerTest {
...
+    verify(screen).hideProgressIndicator()
+  }
+
+  @Test
+  fun `when screen starts and location permission is available then progress indicator should be shown`() {
+    whenever(locationRepository.streamUserLocation(any(), any())).thenReturn(Observable.never())
+
+    uiEvents.onNext(ScreenCreated())
+    uiEvents.onNext(FacilityChangeLocationPermissionChanged(GRANTED))
+    uiEvents.onNext(FacilityChangeLocationPermissionChanged(GRANTED))
```
sanchita: I think there should also be a test for when the permission changes from one state to another i.e. `DENIED` -> `GRANTED`.
What do you think?

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
```diff
@@ -331,7 +337,51 @@ class FacilityChangeScreenControllerTest {
...
   @Suppress("unused")
   fun `params for location updates`() = listOf(
-      LocationUpdate.Unavailable,
-      LocationUpdate.Available(location = Coordinates(0.0, 0.0), timeSinceBootWhenRecorded = Duration.ofNanos(0))
+      Unavailable,
+      Available(location = Coordinates(0.0, 0.0), timeSinceBootWhenRecorded = Duration.ofNanos(0))
   )
+
+  @Test
+  fun `when both facilities and location are fetched only then should facilities be shown`() {
```
sanchita: Does this mean if location is not fetched then facilities should not be shown?

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
```diff
@@ -138,4 +175,102 @@ class FacilityChangeScreenControllerTest {
...
+    val locationNewerThanStaleThreshold = Available(
+        location = Coordinates(0.0, 0.0),
+        timeSinceBootWhenRecorded = Duration.ofMillis(elapsedRealtimeClock.millis()))
+
+    locationUpdates.onNext(locationNewerThanStaleThreshold)
+    verify(screen).updateFacilities(any(), any())
+  }
+
+  @Suppress("unused")
+  fun `params for permission denials`(): List<RuntimePermissionResult> {
```
sanchita: Move this function to below the test it is being used in.

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
saket: Yes

saket: The spec can be found in the description box on zeplin: https://zpl.io/2vWEqR5

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
saket: The test will fail for `5`. The only values that can be used are > threshold (5)

sanchita: It passes for `"5"`.
I tested this.

app/src/test/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenControllerTest.kt
```diff
@@ -183,7 +183,7 @@ class RegistrationFacilitySelectionScreenControllerTest {
     whenever(facilityRepository.recordCount()).thenReturn(Observable.just(facilities.size))
```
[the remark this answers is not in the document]
saket: Whoops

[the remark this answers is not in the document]
saket: Fixing this

app/src/main/java/org/simple/clinic/facility/change/FacilityChangeModule.kt
```diff
@@ -0,0 +1,20 @@
...
+import dagger.Provides
+import io.reactivex.Single
+import org.simple.clinic.util.Distance
+import org.threeten.bp.Duration
+
+@Module
+class FacilityChangeModule {
+
+  @Provides
+  fun config(): Single<FacilityChangeConfig> {
```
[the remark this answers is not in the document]
saket: Sure, done

app/src/main/java/org/simple/clinic/facility/change/FacilityChangeScreenController.kt
saket: No, the `filter` should happen after zipping or else we can see stale values. However it does not matter in the current situation because both streams will only emit once. We aren't observing changes to location permissions and only picking up the latest value.

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
sanchita: The description on Zeplin says: 
> If location detection fails within 5 seconds, this state will be resolved and the user would see the list of facilities
This test name seems to contradict that or maybe I don't understand what do you mean by "location is fetched"
Does it mean `LocationUpdate` is received?

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
saket: Whoops, you're right. The timer expires right after 5s. I'll update this.

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
saket: I wanted to refer to `denied` and `rejected permanently` with one word. `unavailable` is what came to my mind. Do you have a better suggestion?

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
saket: I actually started with that, but realized that according to the spec, we will not observe changes in permission. Only the first result will be picked so it's not required. I'm emitting multiple `GRANTED` values here just as a sanity check to ensure `showProgressIndicator()` doesn't get called twice.

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
saket: Aye

app/src/main/java/org/simple/clinic/facility/change/FacilityChangeScreenController.kt
sanchita: If that is the case, then please consider using `withLatestFrom` instead.

saket: Discussed externally: I've updated the commit to use `combineLatest()` along with `take(1)` so it's easier to understand.

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
sanchita: Both `DENIED` and `NEVER ASK AGAIN` point to permission being denied.
`Unavailable` sounds like misleading in this context.

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
saket: > Does it mean `LocationUpdate` is received?

yes! 

This line is perhaps more relevant from the spec:

> The loading screen is resolved when 1) the list of facilities has been fetched AND 2) the rough location of the user has been detected.

app/src/test/java/org/simple/clinic/facility/change/FacilityChangeScreenControllerTest.kt
saket: Renamed `is unavailable` -> `was denied`.

PR #509, opened 2019-02-22
Add Sync Indicator UI

sanchita:
Pivotol story -> https://www.pivotaltracker.com/story/show/163755918

### DIFF
--- a/app/src/main/java/org/simple/clinic/activity/TheActivityComponent.kt
+++ b/app/src/main/java/org/simple/clinic/activity/TheActivityComponent.kt
@@ -52,6 +52,7 @@ import org.simple.clinic.search.results.PatientSearchResultsScreen
 import org.simple.clinic.security.pin.PinEntryCardView
 import org.simple.clinic.summary.PatientSummaryScreen
 import org.simple.clinic.summary.updatephone.UpdatePhoneNumberDialog
+import org.simple.clinic.sync.indicator.SyncIndicatorView
 import org.simple.clinic.util.InstantRxPreferencesConverter
 import org.simple.clinic.widgets.RxTheActivityLifecycle
 import org.simple.clinic.widgets.TheActivityLifecycle
@@ -103,6 +104,7 @@ interface TheActivityComponent {
   fun inject(target: QrCodeScannerView)
   fun inject(target: RecentPatientsView)
   fun inject(patientsScreenKey: PatientsScreenKey)
+  fun inject(target: SyncIndicatorView)
 
   @Subcomponent.Builder
   interface Builder {
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorModule.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorModule.kt
@@ -5,7 +5,10 @@ import com.f2prateek.rx.preferences2.RxSharedPreferences
 import com.squareup.moshi.Moshi
 import dagger.Module
 import dagger.Provides
+import io.reactivex.Observable
 import org.simple.clinic.sync.LastSyncedState
+import org.threeten.bp.Duration
+import org.threeten.bp.temporal.ChronoUnit
 
 @Module
 class SyncIndicatorModule {
@@ -14,5 +17,10 @@ class SyncIndicatorModule {
   fun lastFrequentSyncResult(rxSharedPreferences: RxSharedPreferences, moshi: Moshi): Preference<LastSyncedState> {
     val typeConverter = LastSyncedState.RxPreferenceConverter(moshi)
     return rxSharedPreferences.getObject("last_frequent_sync_result_v1", LastSyncedState(), typeConverter)
+  }
+
+  @Provides
+  fun syncIndicatorConfig(): Observable<SyncIndicatorConfig> {
+    return Observable.just(SyncIndicatorConfig(syncFailureThreshold = Duration.of(12, ChronoUnit.HOURS)))
   }
 }
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
@@ -3,14 +3,26 @@ package org.simple.clinic.sync.indicator
 import android.annotation.SuppressLint
 import android.content.Context
 import android.util.AttributeSet
+import android.view.LayoutInflater
 import android.widget.LinearLayout
+import android.widget.TextView
 import com.jakewharton.rxbinding2.view.RxView
 import io.reactivex.Observable
 import io.reactivex.android.schedulers.AndroidSchedulers.mainThread
 import io.reactivex.schedulers.Schedulers.io
+import kotterknife.bindView
+import org.simple.clinic.R
+import org.simple.clinic.activity.TheActivity
+import org.simple.clinic.sync.indicator.SyncIndicatorState.ConnectToSync
+import org.simple.clinic.sync.indicator.SyncIndicatorState.SyncPending
+import org.simple.clinic.sync.indicator.SyncIndicatorState.Synced
+import org.simple.clinic.sync.indicator.SyncIndicatorState.Syncing
+import org.simple.clinic.widgets.setCompoundDrawableStart
 import javax.inject.Inject
 
 class SyncIndicatorView(context: Context, attrs: AttributeSet) : LinearLayout(context, attrs) {
+
+  val syncStatusTextView by bindView<TextView>(R.id.sync_indicator_status_text)
 
   @Inject
   lateinit var controller: SyncIndicatorViewController
@@ -18,6 +30,10 @@ class SyncIndicatorView(context: Context, attrs: AttributeSet) : LinearLayout(co
   @SuppressLint("CheckResult")
   override fun onFinishInflate() {
     super.onFinishInflate()
+
+    LayoutInflater.from(context).inflate(R.layout.sync_indicator, this, true)
+
+    TheActivity.component.inject(this)
 
     screenCreates()
         .observeOn(io())
@@ -30,6 +46,23 @@ class SyncIndicatorView(context: Context, attrs: AttributeSet) : LinearLayout(co
   private fun screenCreates() = Observable.just(SyncIndicatorViewCreated)
 
   fun updateState(syncState: SyncIndicatorState) {
-    TODO("not implemented") //Will be implemented when the UI is added.
+    when (syncState) {
+      ConnectToSync -> {
+        syncStatusTextView.text = context.getString(R.string.sync_indicator_status_failed)
+        syncStatusTextView.setCompoundDrawableStart(R.drawable.ic_warning_24dp)
+      }
+      is Synced -> {
+        syncStatusTextView.text = context.resources.getQuantityString(R.plurals.sync_indicator_status_synced, syncState.minAgo.toInt(), syncState.minAgo)
+        syncStatusTextView.setCompoundDrawableStart(R.drawable.ic_cloud_done_24dp)
+      }
+      SyncPending -> {
+        syncStatusTextView.text = context.getString(R.string.sync_indicator_status_pending)
+        syncStatusTextView.setCompoundDrawableStart(R.drawable.ic_cloud_upload_24dp)
+      }
+      Syncing -> {
+        syncStatusTextView.text = context.getString(R.string.sync_indicator_status_syncing)
+        syncStatusTextView.setCompoundDrawableStart(R.drawable.ic_sync_24dp)
+      }
+    }
   }
 }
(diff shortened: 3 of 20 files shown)


app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorModule.kt
```diff
@@ -15,4 +18,9 @@ class SyncIndicatorModule {
     val typeConverter = LastSyncedState.RxPreferenceConverter(moshi)
     return rxSharedPreferences.getObject("last_frequent_sync_result_v1", LastSyncedState(), typeConverter)
   }
+
+  @Provides
+  fun syncIndicatorModule(): Single<SyncIndicatorConfig> {
```
saket: Did you copy the class name as this functions name? 😄

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
```diff
@@ -31,7 +31,7 @@ class SyncIndicatorView(context: Context, attrs: AttributeSet) : LinearLayout(co
   override fun onFinishInflate() {
     super.onFinishInflate()
 
-    LayoutInflater.from(context).inflate(R.layout.sync_indicator, this, true)
+    LayoutInflater.from(context).inflate(R.layout.sync_indicator, this, false)
```
saket: When passing `attachToRoot = false`, you're supposed to receive the result of this function and add the views manually. This is effectively doing nothing. Is this intentional?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -52,11 +54,11 @@ class SyncIndicatorViewController @Inject constructor(
     val syncProgress = lastSyncedStateStream.filter { it.lastSyncProgress != null }
 
     val showSyncIndicatorState = Observables
-        .combineLatest(screenCreated, syncProgress)
-        { _, stateStream ->
+        .combineLatest(screenCreated, syncProgress, configProvider.toObservable())
```
saket: Can I propose renaming `SyncProgress` to `SyncResult` or something? It doesn't feel like we're capturing the progress. What do you think?

saket: Vinay has started using `Observable` instead of `Single` for configs because we anyway convert them to observables using `configProvider.toObservable()`. Do you want to do the same thing?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -52,11 +54,11 @@ class SyncIndicatorViewController @Inject constructor(
...
+        { _, stateStream, config ->
           val indicatorState = when (stateStream.lastSyncProgress!!) {
-            SUCCESS -> syncIndicatorState(stateStream.lastSyncSuccessTimestamp)
-            FAILURE -> syncedFailureState(stateStream.lastSyncSuccessTimestamp)
+            SUCCESS -> syncIndicatorState(stateStream.lastSyncSuccessTimestamp!!, config.syncFailureThreshold)
+            FAILURE -> syncedFailureState(stateStream.lastSyncSuccessTimestamp, config.syncFailureThreshold)
```
saket: What do you think about renaming `lastSyncSuccessTimestamp` -> `lastSuccessfulSyncTime`? Or `lastSyncSucceededAt`? 

`lastSyncSuccessTimestamp` doesn't sound very natural.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -65,25 +67,24 @@ class SyncIndicatorViewController @Inject constructor(
     return showSyncIndicatorState.mergeWith(showDefaultSyncIndicatorState)
   }
 
-  private fun syncedFailureState(timestamp: Instant?): SyncIndicatorState {
+  private fun syncedFailureState(timestamp: Instant?, maxIntervalSinceLastSync: Duration): SyncIndicatorState {
```
saket: Did you intentionally not rename `maxIntervalSinceLastSync` to `syncFailureThreshold?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -65,25 +67,24 @@ class SyncIndicatorViewController @Inject constructor(
...
     if (timestamp == null) {
       return SyncPending
     }
 
-    return syncIndicatorState(timestamp)
+    return syncIndicatorState(timestamp, maxIntervalSinceLastSync)
   }
 
-  private fun syncIndicatorState(timestamp: Instant?): SyncIndicatorState {
+  private fun syncIndicatorState(timestamp: Instant, syncFailureThreshold: Duration): SyncIndicatorState {
```
saket: It feels strange that we have two functions here that only differ in nullability of one param. 

@vinaysshenoy can you take a look at this? Does this look strange to you as well?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorModule.kt
sanchita: No I did not. I will rename it if that's what you mean though.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: We are capturing the sync progress and not the result. 
We already renamed this from `syncResult` -> `syncProgress`
Refer https://github.com/simpledotorg/simple-android/pull/485#discussion_r256263171

vinaysshenoy: The initial PR did have it named `SyncResult`. I didn't feel that was right because it doesn't actually record the result of a sync, but whether a sync was `syncing`, `done` or `failed`, and I suggested `SyncProgress` as a name for it. I would've suggested `SyncStatus`, but since we already use that for our DB models, I felt it would not be a good choice.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: I wanted to separate the nullability check because this will happen only in case of failure.
I could move the check to inside the `when` block itself.
I just felt this was more readable.

vinaysshenoy: You're right, it does seem a little weird. Maybe we should consider merging the two methods.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: Its renamed in the later commit.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
sanchita: Done.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
saket: Heh okay

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
saket: > I wanted to separate the nullability check because this will happen only in case of failure.

I see. In this specific case, it feels like merging sounds better as @vinaysshenoy said

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorModule.kt
sanchita: Done

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: Done

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: Renamed.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: Done

saket:
@sanchita I had one more comment but I can't find it anymore. I'll post it again in office tomorrow. If you get time, please re-roll this branch with master?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -65,28 +65,21 @@ class SyncIndicatorViewController @Inject constructor(
...
-
-    val maxIntervalSinceLastSync = Duration.ofHours(12)
     val mostFrequentSyncInterval = enumValues<SyncInterval>()
         .map { it.frequency }
         .min()!!
-
     return when {
       timeSinceLastSync > maxIntervalSinceLastSync -> ConnectToSync
       timeSinceLastSync > mostFrequentSyncInterval -> SyncPending
       timeSinceLastSync.isNegative -> SyncPending
```
saket: It's a bit confusing to read that `timeSinceLastSync` can be negative until you realize that this can happen if the user manually changes the device time. In fact, it looks baffling on the first look. :P 

Can this be made more obvious? It could be as simple as just extracting this to a variable and documenting the code by the variable's name:

```kotlin
val syncHappenedInTheFuture = timestamp > now

return when {
  syncHappenedInTheFuture -> SyncPending
  timeSinceLastSync > syncFailureThreshold -> ConnectToSync
  timeSinceLastSync > mostFrequentSyncInterval -> SyncPending
  else -> Synced(timeSinceLastSync.toMinutes())
}
```

What do you think?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
```diff
@@ -31,7 +31,7 @@ class SyncIndicatorView(context: Context, attrs: AttributeSet) : LinearLayout(co
   override fun onFinishInflate() {
     super.onFinishInflate()
 
-    LayoutInflater.from(context).inflate(R.layout.sync_indicator, this, false)
+    LayoutInflater.from(context).inflate(R.layout.sync_indicator, this, true)
```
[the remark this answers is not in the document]
sanchita: It is a change to make `SyncIndicatorView` complete.
Does this also mean that `TheActivity.component.inject(this)` should not be part of this commit?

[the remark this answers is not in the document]
sanchita: If you want, I could rename the commit?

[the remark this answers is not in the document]
saket: I think we should instead revert this change in the commit, `Add SyncIndicatorConfig for injecting syncFailureThreshold`. `attachToRoot=false` should have never been done so let's just remove it from the history?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: Makes sense. Will make this change.

app/src/main/res/layout/sync_indicator.xml
```diff
@@ -9,8 +9,6 @@
     android:layout_width="wrap_content"
     android:layout_height="wrap_content"
     android:drawablePadding="@dimen/spacing_8"
-    android:drawableStart="@drawable/ic_cloud_upload_24dp"
-    android:gravity="center_vertical"
-    android:text="@string/sync_indicator_status_pending" />
+    android:gravity="center_vertical" />
```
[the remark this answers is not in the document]
sanchita: Removed this change altogether.
I do not need to do this anymore.

app/src/main/java/org/simple/clinic/activity/TheActivityComponent.kt
```diff
@@ -103,6 +104,7 @@ interface TheActivityComponent {
   fun inject(target: QrCodeScannerView)
   fun inject(target: RecentPatientsView)
   fun inject(patientsScreenKey: PatientsScreenKey)
+  fun inject(target: SyncIndicatorView)
```
saket: sidenote: @rakshakhegde can you rename `patientsScreenKey` to `target`? Asking you because this looks like a part of your feature.

PR #512, opened 2019-02-25
Rename Staging to Demo

pratul:
We keep package name and build variant as Staging for now. Only external things get renamed to Demo.


saket:
This looks good. I've started a build on Bitrise to generate an APK so that I can test it on a device: https://app.bitrise.io/build/d17d54199d3f6c54

vinaysshenoy:
The adaptive icon foreground and the logos for the app bar have not been updated. I'll add this to a checklist of things to do whenever adding/renaming a new build flavour.

PR #514, opened 2019-02-26
Fetch recent patients

rakshakhegde:
https://www.pivotaltracker.com/story/show/163976144

The SQL query is incomplete. I still need to add queries for remaining tables. That will be done once I write full test suite.

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -4,7 +4,6 @@ import androidx.arch.core.executor.testing.InstantTaskExecutorRule
 import com.google.common.truth.Truth.assertThat
 import io.reactivex.Completable
 import io.reactivex.Observable
-import io.reactivex.Single
 import org.junit.After
 import org.junit.Before
 import org.junit.Rule
--- a/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
@@ -408,8 +408,8 @@ class PatientRepository @Inject constructor(
         .andThen(reportsRepository.deleteReportsFile().toCompletable())
   }
 
-  fun recentPatients(facilityUuid: UUID): Observable<List<RecentPatient>> =
+  fun recentPatients(facilityUuid: UUID, limit: Int): Observable<List<RecentPatient>> =
       database.recentPatientDao()
-          .recentPatients(facilityUuid)
+          .recentPatients(facilityUuid, limit)
           .toObservable()
 }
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/simple/clinic/patient/recent/RecentPatient.kt
```diff
@@ -0,0 +1,67 @@
...
+              FROM PrescribedDrug T
+			        WHERE facilityUuid = :facilityUuid
+              GROUP BY patientUuid
+          ) PD ON P.uuid = PD.patientUuid
+        WHERE (
+          BP.facilityUuid = :facilityUuid OR
+          PD.facilityUuid = :facilityUuid
+        )
+        ORDER BY latestUpdatedAt DESC
+        LIMIT 10
```
vinaysshenoy: Can you pass this limit as a parameter to the method instead of hard-coding it in the query? The number `10` here is a presentation detail and not specific to the query, so we'd want to have it configurable. Ideally, this number will be configured via the remote config.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -4,7 +4,6 @@ import androidx.arch.core.executor.testing.InstantTaskExecutorRule
 import com.google.common.truth.Truth.assertThat
```
vinaysshenoy: We avoid using the `.kt` extensions in commit messages. In addition, instead of writing the method name, we should prefer writing a short description of what the method does.

Can you reword this commit message to something like `Add method to fetch recent patients to PatientRepository`?

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
```diff
@@ -25,51 +33,21 @@ class RecentPatientsView(context: Context, attrs: AttributeSet) : RecyclerView(c
...
-        lastBp = "141/90",
-        gender = Gender.FEMALE
-    ))
-    groupAdapter.add(RecentPatientItem(
-        name = "Anish7 Acharya",
-        age = 43,
-        lastBp = "142/90",
-        gender = Gender.TRANSGENDER
-    ))
+    userSession.requireLoggedInUser()
```
vinaysshenoy: This should be done in a controller. Are you going to be doing this in the PR that completes this feature?

rakshakhegde: Yup. This story is essentially to ensure that fetching recent patients is working as expected. The UI parts will be done in the associated UI story.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -406,4 +407,9 @@ class PatientRepository @Inject constructor(
         }
         .andThen(reportsRepository.deleteReportsFile().toCompletable())
   }
+
+  fun recentPatients(facilityUuid: UUID): Observable<List<RecentPatient>> =
+      database.recentPatientDao()
+          .recentPatients(facilityUuid, limit = 10)
```
vinaysshenoy: The `limit` here is again presentation logic, and should be decided by the user of this method, and not this class itself. This is something that we would want to change via remote config, so it should injected via a config object.

PR #515, opened 2019-02-26
Add timer to sync indicator

sanchita:
Pivotol story -> https://www.pivotaltracker.com/story/show/163755918
To test this, you can enable `showRecentPatients` flag in `PatientConfig`

### DIFF
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
@@ -53,7 +53,11 @@ class SyncIndicatorView(context: Context, attrs: AttributeSet) : LinearLayout(co
       }
       is Synced -> {
         val durationToMinsAgo = syncState.durationSince.toMinutes().toInt()
-        syncStatusTextView.text = context.resources.getQuantityString(R.plurals.sync_indicator_status_synced, durationToMinsAgo, durationToMinsAgo)
+        syncStatusTextView.text = if (durationToMinsAgo == 0) {
+          context.getString(R.string.sync_indicator_status_synced_just_now)
+        } else {
+          context.getString(R.string.sync_indicator_status_synced, durationToMinsAgo)
+        }
         syncStatusTextView.setCompoundDrawableStart(R.drawable.ic_cloud_done_24dp)
       }
       SyncPending -> {
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
@@ -21,6 +21,7 @@ import org.simple.clinic.util.UtcClock
 import org.simple.clinic.widgets.UiEvent
 import org.threeten.bp.Duration
 import org.threeten.bp.Instant
+import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 
 typealias Ui = SyncIndicatorView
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
```diff
@@ -53,7 +53,11 @@ class SyncIndicatorView(context: Context, attrs: AttributeSet) : LinearLayout(co
       }
       is Synced -> {
         val durationToMinsAgo = syncState.durationSince.toMinutes().toInt()
-        syncStatusTextView.text = context.resources.getQuantityString(R.plurals.sync_indicator_status_synced, durationToMinsAgo, durationToMinsAgo)
+        if (durationToMinsAgo == 0) {
```
rakshakhegde: We could convert this block to:
```kotlin
syncStatusTextView.text = if (durationToMinsAgo == 0) {
  context.getString(R.string.sync_indicator_status_synced_just_now)
} else {
  context.getString(R.string.sync_indicator_status_synced, durationToMinsAgo)
}
```

This will ensure that `syncStatusTextView.text` is always set and not forgotten to assign a value if a new conditional branch gets added.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -55,11 +56,20 @@ class SyncIndicatorViewController @Inject constructor(
     val showSyncIndicatorState = Observables
         .combineLatest(screenCreated, syncProgress, configProvider)
         { _, stateStream, config ->
-          val indicatorState = when (stateStream.lastSyncProgress!!) {
+          when (stateStream.lastSyncProgress!!) {
```
rakshakhegde: Are we sure that `lastSyncProgress` will not be null here? Maybe we could add a null branch to this when block and set some default UI state.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -55,11 +56,20 @@ class SyncIndicatorViewController @Inject constructor(
...
             SUCCESS, FAILURE -> syncIndicatorState(stateStream.lastSyncSucceededAt, config.syncFailureThreshold)
             SYNCING -> Syncing
           }
-          { ui: Ui -> ui.updateState(indicatorState) }
+        }
+        .switchMap {
+          if (it is Synced) {
+            val durationSince = it.durationSince
```
rakshakhegde: I don't understand this `durationSince` logic here. Let's discuss?

app/src/main/res/layout/screen_patients_v2.xml
```diff
@@ -64,11 +64,22 @@
         android:src="@drawable/illustrations_homescreen"
         tools:ignore="ContentDescription" />
 
-      <org.simple.clinic.patient.recent.RecentPatientsView
-        android:id="@+id/patients_record_recent_patients"
+      <LinearLayout
```
rakshakhegde: We could continue using RelativeLayout itself without introducing LinearLayout here? Or you may wanna do this in next PR.

rakshakhegde:
@sanchita In which story are we handling the "click on Sync Pending to trigger syncing"?

app/src/main/res/layout/screen_patients_v2.xml
sanchita: This is just a temporary code until we do our integration.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: The `null` check already happens in the `syncProgress` stream.
Check this:

`    val syncProgress = lastSyncedStateStream.filter { it.lastSyncProgress != null }
`

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
sanchita: Sure.

sanchita:
> @sanchita In which story are we handling the "click on Sync Pending to trigger syncing"?

Its this story-> https://www.pivotaltracker.com/story/show/164243208

PR #518, opened 2019-02-28
Fetch recent patients Part 1

rakshakhegde:
https://www.pivotaltracker.com/story/show/163976144

Database fetching is completely working!

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -878,26 +878,6 @@ class PatientRepositoryAndroidTest {
 
   @Test
   fun verify_recent_patients_are_retrieved_as_expected() {
-
-    fun savePatientWithBp(): RecentPatient {
-      val patientProfile = testData.patientProfile()
-      patientRepository.save(listOf(patientProfile)).blockingAwait()
-
-      val bpMeasurement = testData.bloodPressureMeasurement(patientUuid = patientProfile.patient.uuid)
-      database.bloodPressureDao().save(listOf(bpMeasurement))
-      return patientProfile.patient.toRecentPatient(bpMeasurement)
-    }
-
-    val facilityUuid = testData.qaUserFacilityUuid()
-
-    fun verifyRecentPatientOrder(vararg expectedRecentPatients: RecentPatient) {
-      patientRepository
-          .recentPatients(facilityUuid, limit = 3)
-          .test()
-          .assertValue(expectedRecentPatients.toList())
-          .dispose()
-    }
-
     val recentPatient1 = savePatientWithBp()
 
     verifyRecentPatientOrder(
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -874,4 +875,104 @@ class PatientRepositoryAndroidTest {
...
+  }
+
+  private fun Patient.toRecentPatient(bpMeasurement: BloodPressureMeasurement) = RecentPatient(
+      uuid = uuid,
+      fullName = fullName,
+      gender = gender,
+      dateOfBirth = dateOfBirth,
+      age = age,
+      bpUpdatedAt = bpMeasurement.updatedAt,
+      lastBp = RecentPatient.LastBp(bpMeasurement.systolic.toString(), bpMeasurement.diastolic.toString())
```
sanchita: `LastBp` class should also carry the date when BP was taken.
You will have to display that in `RecentPatientItem`

sanchita: Also, I think `LastBp` should take `systolic` and `diastolic` as `Int`.
The conversion to `String` should happen in the views.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -21,6 +21,7 @@ import org.simple.clinic.facility.FacilityRepository
 import org.simple.clinic.medicalhistory.MedicalHistoryRepository
```
sanchita: I can see that this commit is doing more than adding a test.
Can you separate out the changes to the query in a different commit?

app/src/main/java/org/simple/clinic/patient/recent/RecentPatient.kt
```diff
@@ -32,6 +32,11 @@ data class RecentPatient(
   @Dao
   interface RoomDao {
 
+    /**
```
sanchita: Can you mention what "RecentPatient" means first and then describe the query?

sanchita: Also, it is alright if the query is not described at all. That is more of a good to have.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatient.kt
```diff
@@ -38,7 +38,10 @@ data class RecentPatient(
         MAX(
```
sanchita: Typo?
`BP.updatedAt bpUpdatedAt` -> `bp_updatedAt`

app/src/main/java/org/simple/clinic/patient/recent/RecentPatient.kt
```diff
@@ -53,9 +56,26 @@ data class RecentPatient(
...
+              GROUP BY appointmentUuid
+          ) COMM ON AP.uuid = COMM.appointmentUuid
+          LEFT JOIN (
+            SELECT MAX(updatedAt) latestUpdatedAt, T.*
+              FROM MedicalHistory T
+              GROUP BY patientUuid
+          ) MH ON P.uuid = MH.patientUuid
         WHERE (
           BP.facilityUuid = :facilityUuid OR
-          PD.facilityUuid = :facilityUuid
```
sanchita: Do we even need this `WHERE` clause ?
We are already checking `facilituUuid` in individual `LEFT JOIN` queries.
Please add a test to verify this.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
rakshakhegde: Will be displaying the date as part of next PR.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatient.kt
rakshakhegde: Added test, seems like this WHERE clause is required. Need to investigate why.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatient.kt
rakshakhegde: I've updated the comment.

PR #519, opened 2019-02-28
Add manual sync to Sync Indicator

sanchita:
Pivotol story -> https://www.pivotaltracker.com/story/show/164243208
Part 1 of the story above.
This PR adds manual sync to sync indicator.

### DIFF
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
@@ -17,6 +17,7 @@ import org.simple.clinic.sync.indicator.SyncIndicatorState.ConnectToSync
 import org.simple.clinic.sync.indicator.SyncIndicatorState.SyncPending
 import org.simple.clinic.sync.indicator.SyncIndicatorState.Synced
 import org.simple.clinic.sync.indicator.SyncIndicatorState.Syncing
+import org.simple.clinic.util.ResolvedError
 import org.simple.clinic.widgets.setCompoundDrawableStart
 import javax.inject.Inject
 
@@ -72,5 +73,9 @@ class SyncIndicatorView(context: Context, attrs: AttributeSet) : LinearLayout(co
         syncStatusTextView.setCompoundDrawableStart(R.drawable.ic_round_sync_16px)
       }
     }
+  }
+
+  fun showErrorDialog(errorType: ResolvedError) {
+    //This will be implemented in a later commit
   }
 }
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
@@ -23,6 +23,7 @@ import javax.inject.Inject
 class SyncIndicatorView(context: Context, attrs: AttributeSet) : LinearLayout(context, attrs) {
 
   private val syncStatusTextView by bindView<TextView>(R.id.sync_indicator_status_text)
+  private val syncIndicatorLayout by bindView<LinearLayout>(R.id.sync_indicator_root_layout)
 
   @Inject
   lateinit var controller: SyncIndicatorViewController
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -95,4 +99,19 @@ class SyncIndicatorViewController @Inject constructor(
       else -> Synced(timeSinceLastSync)
     }
   }
+
+  private fun startSync(events: Observable<UiEvent>): Observable<UiChange> {
+    fun errorsStream() = dataSync
```
sanchita: This is a `fun` and not a `val` because while writing tests for this controller, I realized that if it is a `val` then `dataSync.streamSyncErrors()` will have to be configured in `setUp()` for every test by returning `Observable.never()`
And in doing so, it will not be overridden when we actually want to test this function.

saket: Sidenote:

Is my understanding correct that you wanted the `val`'s value to be lazily executed? In my experience, lambdas work nicely for this case.

sanchita: Yes. You are right. That is what I want.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
```diff
@@ -73,4 +74,8 @@ class SyncIndicatorView(context: Context, attrs: AttributeSet) : LinearLayout(co
       }
     }
   }
+
+  fun showErrorDialog(errorType: ResolvedError) {
+    TODO("not implemented")
```
vinaysshenoy: Can you change this to a comment instead?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
vinaysshenoy: I am assuming you will make this change?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -101,9 +101,17 @@ class SyncIndicatorViewController @Inject constructor(
   }
 
   private fun startSync(events: Observable<UiEvent>): Observable<UiChange> {
+    fun errorsStream() = dataSync
+        .streamSyncErrors()
+        .take(1)
+        .map { { ui: Ui -> ui.showErrorDialog(it) } }
+
+    val syncStream = dataSync
+        .sync(null)
```
vinaysshenoy: We should not trigger a sync if a sync is already ongoing. This will trigger a new sync every time a click happens, right?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewEvents.kt
```diff
@@ -3,3 +3,5 @@ package org.simple.clinic.sync.indicator
 import org.simple.clinic.widgets.UiEvent
 
 object SyncIndicatorViewCreated : UiEvent
+
+object SyncIndicatorViewClicked : UiEvent
```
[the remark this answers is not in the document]
sanchita: Yes. I should.

app/src/main/res/layout/sync_indicator.xml
```diff
@@ -1,7 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+  android:id="@+id/sync_indicator_root_layout"
   android:layout_width="match_parent"
-  android:layout_height="match_parent">
+  android:layout_height="match_parent"
+  android:clickable="true"
```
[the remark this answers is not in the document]
sanchita: It is a view. It won't be clickable until this is added explicitly.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: Yes I will.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: Done.

PR #520, opened 2019-02-28
Gradle caching, dependency updates

pratul:


rakshakhegde:
@pratul I don't see much incremental-build time difference with caching enabled or not. It's almost always 4 seconds on my MBP. Regardless, I'm accepting the PR because it's not negatively affecting the build times too.

PR #526, opened 2019-03-05
Bypass bug in Proguard 6.0.3 

pratul:


pratul:
Created all release builds and they're building fine now -- https://app.bitrise.io/build/38f3de2340d9427b

PR #534, opened 2019-03-11
Phone number masking

rakshakhegde:
https://www.pivotaltracker.com/story/show/164462722

### DIFF
--- a/app/src/main/java/org/simple/clinic/phone/MaskedPhoneCaller.kt
+++ b/app/src/main/java/org/simple/clinic/phone/MaskedPhoneCaller.kt
@@ -3,7 +3,7 @@ package org.simple.clinic.phone
 import io.reactivex.Completable
 
 /**
- * See [TwilioMaskedPhoneCaller].
+ * See [OfflineMaskedPhoneCaller].
  */
 interface MaskedPhoneCaller {
   fun maskAndCall(numberToMask: String, caller: Caller): Completable
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -72,6 +72,7 @@ android {
       versionNameSuffix "-qa"
       buildConfigField "String", "API_ENDPOINT", "\"" + qaApiEndpoint + "\""
       buildConfigField "String", "HEAP_ID", "\"" + qaHeapId + "\""
+      buildConfigField "String", "MASKED_PHONE_NUMBER", "\"" + qaMaskedPhoneNumber + "\""
       buildConfigField "boolean", "DISABLE_SCREENSHOT", "false"
     }
 
@@ -81,6 +82,7 @@ android {
       versionNameSuffix "-demo"
       buildConfigField "String", "API_ENDPOINT", "\"" + stagingApiEndpoint + "\""
       buildConfigField "String", "HEAP_ID", "\"" + stagingHeapId + "\""
+      buildConfigField "String", "MASKED_PHONE_NUMBER", "\"" + stagingMaskedPhoneNumber + "\""
       buildConfigField "boolean", "DISABLE_SCREENSHOT", "false"
     }
 
@@ -90,6 +92,7 @@ android {
       versionNameSuffix "-sandbox"
       buildConfigField "String", "API_ENDPOINT", "\"" + sandboxApiEndpoint + "\""
       buildConfigField "String", "HEAP_ID", "\"" + sandboxHeapId + "\""
+      buildConfigField "String", "MASKED_PHONE_NUMBER", "\"" + sandboxMaskedPhoneNumber + "\""
       buildConfigField "boolean", "DISABLE_SCREENSHOT", "false"
     }
 
@@ -97,6 +100,7 @@ android {
       dimension 'track'
       buildConfigField "String", "API_ENDPOINT", "\"" + productionApiEndpoint + "\""
       buildConfigField "String", "HEAP_ID", "\"" + productionHeapId + "\""
+      buildConfigField "String", "MASKED_PHONE_NUMBER", "\"" + productionMaskedPhoneNumber + "\""
       buildConfigField "boolean", "DISABLE_SCREENSHOT", "true"
     }
   }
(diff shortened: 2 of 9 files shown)


app/src/main/java/org/simple/clinic/phone/MaskedPhoneCallerImplementation.kt
```diff
@@ -1,12 +1,11 @@
 package org.simple.clinic.phone
 
-import android.net.Uri
 import io.reactivex.Completable
 import io.reactivex.Single
 import org.simple.clinic.activity.TheActivity
 import javax.inject.Inject
 
-class TwilioMaskedPhoneCaller @Inject constructor(
+class MaskedPhoneCallerImplementation @Inject constructor(
```
saket: Just a suggestions: should we name this `OfflineMaskedPhoneCaller` instead?

app/src/main/java/org/simple/clinic/phone/TwilioMaskedPhoneCaller.kt
```diff
@@ -19,13 +19,11 @@ class TwilioMaskedPhoneCaller @Inject constructor(
 
   private fun maskNumber(config: PhoneNumberMaskerConfig, numberToMask: String): String {
     return if (config.maskingEnabled) {
-      // TODO: This number is temporary. Every nurse will
-      // be assigned a number during login/registration.
-      val twilioNumber = "+1 111 111 1111"
+      val number = "07930684724"
```
vinaysshenoy: Can we remove this number from the source code? Since this is an actual number that people can call?

vinaysshenoy: Instead of hardcoding this number, we should provide it via the `PhoneNumberMaskerConfig`.

app/src/main/java/org/simple/clinic/phone/MaskedPhoneCaller.kt
```diff
@@ -3,7 +3,7 @@ package org.simple.clinic.phone
 import io.reactivex.Completable
 
 /**
- * See [TwilioMaskedPhoneCaller].
+ * See [MaskedPhoneCallerImplementation].
```
vinaysshenoy: We should rename this to something more similar to what it actually does, like `OfflineMaskedPhoneCaller` or something.

app/src/test/java/org/simple/clinic/phone/MaskedPhoneCallerImplementationTest.kt
```diff
@@ -34,6 +37,18 @@ class MaskedPhoneCallerImplementationTest {
     assertThat(caller.calledNumber).isEqualTo(plainNumber)
   }
 
+  @Test
```
vinaysshenoy: We should also add a test that the masked phone number is not used when the flag is unset.

app/src/test/java/org/simple/clinic/phone/MaskedPhoneCallerImplementationTest.kt
```diff
@@ -34,6 +37,18 @@ class MaskedPhoneCallerImplementationTest {
...
+  fun `when masking is enabled then masked phone number should be called`() {
+    config = PhoneNumberMaskerConfig(maskingEnabled = true)
+
+    val caller: Caller = mock()
+    val plainNumber = "123"
+
+    maskedPhoneCaller.maskAndCall(plainNumber, caller = caller).blockingAwait()
+
+    verify(caller).call(activity, "07930684724,123%23")
```
vinaysshenoy: We should check this number into the source code.

app/src/main/java/org/simple/clinic/phone/MaskedPhoneCallerImplementation.kt
```diff
@@ -21,7 +20,7 @@ class MaskedPhoneCallerImplementation @Inject constructor(
     return if (config.maskingEnabled) {
       val number = "07930684724"
 
-      val stopCharacter = Uri.encode("#")
+      val stopCharacter = "%23" // encoding for '#'
```
vinaysshenoy: Can we revert to using `Uri.encode` instead? There are many types of encoding, and using `Uri.encode` is a lot more explicit about what kind of encoding we are doing.

If you did this because we can't use the `Uri` class in the test, then we should create an interface, say `PhoneNumberEncoder` and then replace it in the test with something that can be used in the tests.

app/src/main/java/org/simple/clinic/phone/TwilioMaskedPhoneCaller.kt
saket: 😵

app/src/main/java/org/simple/clinic/phone/MaskedPhoneCallerImplementation.kt
rakshakhegde: Done

app/src/main/java/org/simple/clinic/phone/MaskedPhoneCaller.kt
rakshakhegde: Done

app/src/main/java/org/simple/clinic/phone/MaskedPhoneCallerImplementation.kt
rakshakhegde: Turns out we do not need encoding anymore. Plain "#" is what's required.

app/src/test/java/org/simple/clinic/phone/MaskedPhoneCallerImplementationTest.kt
rakshakhegde: This test is already there as the first test.

app/src/main/java/org/simple/clinic/phone/TwilioMaskedPhoneCaller.kt
rakshakhegde: Extracted this into `PhoneNumberMaskerConfig` as part of a new commit.

rakshakhegde: Extracted this into `PhoneNumberMaskerConfig` as part of a new commit.

app/src/test/java/org/simple/clinic/phone/MaskedPhoneCallerImplementationTest.kt
rakshakhegde: This is removed due to `PhoneNumberMaskerConfig` as part of a new commit.

PR #535, opened 2019-03-11
Add MissingPhoneReminder table to database

saket:
Part 1/2 of https://www.pivotaltracker.com/story/show/163448748

@vinaysshenoy gigantic PR because DB migration

### DIFF
--- a/app/schemas/org.simple.clinic.AppDatabase/31.json
+++ b/app/schemas/org.simple.clinic.AppDatabase/31.json
@@ -1,0 +1,1238 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 31,
+    "identityHash": "57705d2baef816ecd17eef09bc13ce20",
+    "entities": [
+      {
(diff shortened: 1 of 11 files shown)


vinaysshenoy:
@saket I was a little unsure of the notion of a "visit", so I asked Akshay to clarify. Can you check the notes on https://zpl.io/25GDqyn and confirm if this PR is still valid before I pick this up for review?

vinaysshenoy:
> @saket I was a little unsure of the notion of a "visit", so I asked Akshay to clarify. Can you check the notes on https://zpl.io/25GDqyn and confirm if this PR is still valid before I pick this up for review?

Looking at the comment made by Akshay, it seems like we need to show the dialog when the first BP is recorded. It seems to me that we can show the dialog when the count of recorded BPs on the patient summary screen goes from 0 to 1 and avoid needing to create a new table at all.

saket:
@vinaysshenoy true. I'll open a new PR with the changes.

PR #536, opened 2019-03-12
Mark appointment as visited

sanchita:
Pivotol story -> https://www.pivotaltracker.com/story/show/164200231

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
@@ -775,4 +775,55 @@ class AppointmentRepositoryAndroidTest {
     val (appointment) = appointmentRepository.lastCreatedAppointmentForPatient(patientId).blockingFirst()
     assertThat(appointment!!).isEqualTo(secondAppointment)
   }
+
+  @Test
+  fun marking_appointment_older_than_current_date_should_work_correctly() {
+    val patientId = UUID.randomUUID()
+    val appointmentUuid1 = UUID.randomUUID()
+    val appointmentUuid2 = UUID.randomUUID()
+    val scheduleDate = LocalDate.now(testClock).plusMonths(1)
+
+    testClock.advanceBy(Duration.ofHours(1))
+    database
+        .appointmentDao()
+        .save(listOf(testData.appointment(
+            uuid = appointmentUuid1,
+            patientUuid = patientId,
+            status = SCHEDULED,
+            syncStatus = SyncStatus.DONE,
+            scheduledDate = scheduleDate,
+            createdAt = Instant.now(testClock),
+            updatedAt = Instant.now(testClock)
+        )))
+    val firstAppointment = database.appointmentDao().getOne(appointmentUuid1)
+    testClock.advanceBy(Duration.ofHours(1))
+
+    appointmentRepository.markAppointmentsCreatedBeforeTodayAsVisited(patientId).blockingAwait()
+
+    assertThat(database.appointmentDao().getOne(appointmentUuid1)).isEqualTo(firstAppointment)
+
+    testClock.advanceBy(Duration.ofDays(1))
+
+    database
+        .appointmentDao()
+        .save(listOf(testData.appointment(
+            uuid = appointmentUuid2,
+            patientUuid = patientId,
+            scheduledDate = scheduleDate,
+            status = SCHEDULED,
+            syncStatus = SyncStatus.PENDING,
+            createdAt = Instant.now(testClock),
+            updatedAt = Instant.now(testClock)
+        )))
+
+    val secondAppointment = database.appointmentDao().getOne(appointmentUuid2)
+    appointmentRepository.markAppointmentsCreatedBeforeTodayAsVisited(patientId).blockingAwait()
+
+    database.appointmentDao().getOne(firstAppointment!!.uuid)!!.run {
+      assertThat(status).isEqualTo(VISITED)
+      assertThat(syncStatus).isEqualTo(SyncStatus.PENDING)
+      assertThat(updatedAt).isEqualTo(Instant.now(testClock))
+    }
+    assertThat(database.appointmentDao().getOne(appointmentUuid2)).isEqualTo(secondAppointment)
+  }
 }
--- a/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
+++ b/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
@@ -21,6 +21,7 @@ import org.simple.clinic.bp.entry.BpValidator.Validation.ErrorSystolicLessThanDi
 import org.simple.clinic.bp.entry.BpValidator.Validation.ErrorSystolicTooHigh
 import org.simple.clinic.bp.entry.BpValidator.Validation.ErrorSystolicTooLow
 import org.simple.clinic.bp.entry.BpValidator.Validation.Success
+import org.simple.clinic.overdue.AppointmentRepository
 import org.simple.clinic.patient.PatientUuid
 import org.simple.clinic.util.UserClock
 import org.simple.clinic.util.UserInputDatePaddingCharacter
@@ -44,6 +45,7 @@ typealias UiChange = (Ui) -> Unit
  */
 class BloodPressureEntrySheetController @Inject constructor(
     private val bloodPressureRepository: BloodPressureRepository,
+    private val appointmentRepository: AppointmentRepository,
     private val dateValidator: UserInputDateValidator,
     private val bpValidator: BpValidator,
     private val userClock: UserClock,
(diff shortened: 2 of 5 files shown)


app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -775,4 +775,34 @@ class AppointmentRepositoryAndroidTest {
...
+    testClock.advanceBy(Duration.ofHours(1))
+
+    appointmentRepository.markAppointmentsCreatedBeforeTodayAsVisited(patientId).blockingAwait()
+
+    val (lastAppointment) = appointmentRepository.lastCreatedAppointmentForPatient(patientId).blockingFirst()
+    assertThat(lastAppointment).isEqualTo(firstAppointment)
+
+    testClock.advanceBy(Duration.ofDays(1))
+
+    appointmentRepository.schedule(patientId, scheduleDate).blockingGet()
```
vinaysshenoy: `AppointmentRepository#schedule` internally marks all older appointments as visited, so this test does not verify whether `markAppointmentCreatedBeforeTodayAsVisited` works as expected.

vinaysshenoy: Ideally, when testing a class, we should not invoke other methods in the class except the one that we are testing and use other means to verify the method behaviour.

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
```diff
@@ -138,5 +138,18 @@ data class Appointment(
 
     @Query("DELETE FROM Appointment")
     fun clear()
+
+    @Query("""
+      UPDATE Appointment SET status = :updatedStatus, syncStatus = :newSyncStatus, updatedAt = :newUpdatedAt
+      WHERE patientUuid = :patientUuid AND status = :scheduledStatus AND createdAt < :olderThan
```
vinaysshenoy: When I see the name `olderThan`, I expect the property to be a duration. If I were to write this behaviour down in english, it would be "Mark all scheduled appointments for a given patient and **older than** 13th March 2019 as visited", which sounds incorrect.

I would expect it to be more like "Mark all scheduled appointments for a given patient and **created before** 13th March 2019 as visited"

So, can you rename `olderThan` to something more appropriate like `createdBefore`?

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
sanchita: Good catch. I will use the `dao` directly.

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -775,4 +775,54 @@ class AppointmentRepositoryAndroidTest {
...
+    testClock.advanceBy(Duration.ofHours(1))
+
+    appointmentRepository.markAppointmentsCreatedBeforeTodayAsVisited(patientId).blockingAwait()
+
+    val lastAppointment = database.appointmentDao().getOne(appointmentUuid)!!
+    assertThat(lastAppointment).isEqualTo(firstAppointment)
+
+    testClock.advanceBy(Duration.ofDays(1))
+
+    database
```
vinaysshenoy: Is there a need to save this appointment at all?

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -775,4 +775,54 @@ class AppointmentRepositoryAndroidTest {
...
+            updatedAt = Instant.now(testClock)
+        )))
+
+    val firstAppointment = database.appointmentDao().getOne(appointmentUuid)
+    testClock.advanceBy(Duration.ofHours(1))
+
+    appointmentRepository.markAppointmentsCreatedBeforeTodayAsVisited(patientId).blockingAwait()
+
+    val lastAppointment = database.appointmentDao().getOne(appointmentUuid)!!
+    assertThat(lastAppointment).isEqualTo(firstAppointment)
```
vinaysshenoy: We can inline this instead of saving the `lastAppointment` as a value.

```kotlin
assertThat(database.appointmentDao().getOne(appointmentUuid))
  .isEqualTo(firstAppointment)
```

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -775,4 +775,54 @@ class AppointmentRepositoryAndroidTest {
...
+            status = SCHEDULED,
+            syncStatus = SyncStatus.PENDING,
+            createdAt = Instant.now(testClock),
+            updatedAt = Instant.now(testClock)
+        )))
+    appointmentRepository.markAppointmentsCreatedBeforeTodayAsVisited(patientId).blockingAwait()
+
+    val oldAppointment = database.appointmentDao().getOne(lastAppointment.uuid)
+
+    oldAppointment?.apply {
```
vinaysshenoy: Instead of saving this as a separate `oldAppointment`, why can't we just inline it?

```kotlin
database.appointmentDao().getOne(appointment.uuid)!!.let { oldAppointment ->
  // Run assertions
}
```

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
sanchita: Yes. There is. To know that the newer appointment is not affected by `markAppointmentsCreatedBeforeTodayAsVisited`

vinaysshenoy: But we aren't verifying this in the test.

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
vinaysshenoy: We should be using a `!!.let {}` instead of a `?.let {}` because if the appointment gets deleted for whatever reason, these assertions will never run and the tests will pass.

sanchita: Sure we can. Although we are creating the appointment in this test itself. So I can't think of how will that happen.

vinaysshenoy: It's a safety net. It's unlikely something will happen, but it will fail the test if it does.

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
sanchita: Oh yes. I will add that assertion.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -488,6 +490,7 @@ class BloodPressureEntrySheetController @Inject constructor(
           bloodPressureRepository
               .saveMeasurement(newBp.patientUuid, newBp.systolic, newBp.diastolic, toUtcInstant(newBp.date))
               .map { BloodPressureSaved(wasDateChanged = prefilledDate != newBp.date) }
+              .doOnSuccess { appointmentRepository.markAppointmentsCreatedBeforeTodayAsVisited(newBp.patientUuid) }
```
vinaysshenoy: Since `AppointmentRepository#markAppointmentsCreatedBeforeTodayAsVisited` returns a `Completable`, invoking this in a `.doOnSuccess` is not the right solution since the `Completable` is never subscribed to. In addition, this is not a side effect, but a flow that must be executed as a part of this chain. What you need is a `.flatMap` instead:

```kotlin
.flatMap { bpSaved -> 
  appointmentRepository
    .markAppointmentsCreatedBeforeTodayAsVisited(newBp.patientUuid)
    .toSingleDefault(bpSaved)
}
```

sanchita: I guess I am still not sure what side effects are for. The reason I put this as a side effect was because it is not the main job in this chain. I will change this then.

vinaysshenoy: I've found that an easy way is to ask yourself, "if this fails, does it impact the behaviour of the app in a way that the user should be informed?". If yes, it should not be a side effect.

sanchita: Oh! Let's talk about this in person.

PR #539, opened 2019-03-12
Add non-reactive pop with result functionality to ScreenRouter

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/164565554


router/src/main/java/org/simple/clinic/router/screen/ScreenRouter.kt
```diff
@@ -78,16 +80,15 @@ class ScreenRouter(
...
+  fun popWithResult(key: String, result: Any?) {
+    screenResults.put(key, result)
     pop()
+  }
 
-    resultBus.send(result)
+  @Suppress("UNCHECKED_CAST")
+  @CheckResult
+  fun <T> retrieveResult(key: String): T? {
+    return screenResults.get(key) as T?
```
saket: Hrm, one thing I don't understand is how do you plan to remove `streamScreenResults()`? Did you intentionally leave it?

saket:
Thanks for doing this @vinaysshenoy. This change was long pending.

router/src/main/java/org/simple/clinic/router/screen/ScreenRouter.kt
vinaysshenoy: Yeah, I was planning to remove it, but I saw its being used for sending permission grant results. Once this gets merged in, I'll create a story for changing permission results to the new flow and remove the streamed results.

saket: Do you plan to do it right away? It won't be nice to have two different ways of exchanging screen results. 

I can sit with you and come up with a solution instead of deferring it to later because it won't happen anytime soon otherwise.

saket: Off the top of my head, I can think of a solution where `ScreenResults` piggybacks on `ScreenResultBus`. Any screen result sent through `ScreenRouter` is available via both `retrieveResult()` and `streamScreenResults()`. The only difference would be that the latter would be an observable infinite stream. 

Any value sent to `ScreenRouter` could be first sent to `ScreenResultBus` and it could later be observed by `ScreenResults` to hold it.

saket:
@vinaysshenoy Hrm, on a second thought I would like to think a bit more on this. The way `ScreenRouter` helps exchange screen results is no different than how it works with Activities and `startActivityForResult()`. If we were using Activities, we would still have to deal with the homescreen ViewPager instantiating its children asynchronously. 

I am wondering if it's more correct to remove asynchronousity (if that's a word) from the homescreen ViewPager. Thoughts?

router/src/main/java/org/simple/clinic/router/screen/ScreenRouter.kt
vinaysshenoy: > Do you plan to do it right away? It won't be nice to have two different ways of exchanging screen results.

 I can do this in this PR, actually. It will be nice to have it implemented and functional when we merge this in.

> Any value sent to ScreenRouter could be first sent to ScreenResultBus and it could later be observed by ScreenResults to hold it.

I don't want to do this because then we will run into the same problem that we have now. Late subscribers will not receive the screen results, and if we leave the results in memory we will run into the same issues that we'll have with a `BehaviourSubject`, i.e, screens might receive stale results.

I'd rather that the intended consumers use the result once and convert into a stream before forwarding it to the controllers instead of the screen router exposing a stream.

vinaysshenoy:
> @vinaysshenoy Hrm, on a second thought I would like to think a bit more on this. The way `ScreenRouter` helps exchange screen results is no different than how it works with Activities and `startActivityForResult()`. If we were using Activities, we would still have to deal with the homescreen ViewPager instantiating its children asynchronously.
> 
> I am wondering if it's more correct to remove asynchronousity (if that's a word) from the homescreen ViewPager. Thoughts?

We technically can't remove the async view inflation from the `ViewPager` since that's how its designed to function. In addition, if we were using activities, we would work around this by providing this result to the `PagerAdapter` as a stream and then pass it down to the actual screen. I feel this is an ugly solution that exists only because of how `ViewPager` and the android platform works.

router/src/main/java/org/simple/clinic/router/screen/ScreenRouter.kt
saket: Let's discuss this in person tomorrow

saket:
@vinaysshenoy we could remove asynchronous inflation of ViewPager children by placing the child Views in the xml directly and getting rid of the adapter entirely. Having synchronous code is always better. What do you think?

router/src/main/java/org/simple/clinic/router/screen/ScreenRouter.kt
vinaysshenoy: If I make the changes, can we merge this in today? The rest of the simple ID feature is blocked on this.

vinaysshenoy: We can always discuss and figure out a better solution and implement it later.

vinaysshenoy:
> @vinaysshenoy we could remove asynchronous inflation of ViewPager children by placing the child Views in the xml directly and getting rid of the adapter entirely. Having synchronous code is always better. What do you think?

Won't we have to implement the swipe functionality ourselves then?

router/src/main/java/org/simple/clinic/router/screen/ScreenRouter.kt
saket: > The rest of the simple ID feature is blocked on this.

This is why I felt we should implement a stopgap solution in simple ID for now because this would have required a good amount of discussion. I'd hate to merge this under pressure :( 

> We can always discuss and figure out a better solution and implement it later.

I don't want to put in a half baked solution because maintenance stories like this never get done unless something is blocked on it.

saket: Your call. I'd recommend placing a stopgap solution in favor of properly improving the app's ui navigation, but I can understand if that's too much an efort.

vinaysshenoy: Technically, `Flow` itself is a half-baked solution. :P

IMO, this is a better solution than what we have now because this is an imperative solution that will support both synchronous as well as asynchronous view inflation. Adding in a workaround or a hack for simple ID would be more cumbersome than using this.

vinaysshenoy: I'd like to get this in without changing how permission results are delivered and finish up the Simple ID v1 flow. Then we can figure out what is a good solution for doing a `startActivityWithResult` equivalent for our use cases.

saket: > Technically, Flow itself is a half-baked solution. :P

Why? What we're trying to solve in this PR is entirely independent of Flow. We wrote the results bus along the lines of how Activities exchange data. We would have still encountered this problem even if we weren't using Flow.

> Adding in a workaround or a hack for simple ID would be more cumbersome than using this.

Not really. We're using `SharedPreferences` for showing the `patient-saved` snackbar and it's working just fine. There's nothing cumbersome about it.

Instead of calling `screenRouter.popWithResult()`, you'd call `preference.set()` and `screenRouter.retrieveResult()` will get replaced by `preference.get()` or `preference.asObservable()`. 

> Then we can figure out what is a good solution for doing a startActivityWithResult equivalent for our use cases.

We already have a `startActivityWithResult()` equivalent in place. The current implementation works exactly like it. :p

saket:
@vinaysshenoy I think if we wrap them inside `ViewPager` we'll still get support for swipe behavior. Can you please test this? If it works we'll be able to get rid of this whole discussion.

router/src/main/java/org/simple/clinic/router/screen/ScreenRouter.kt
vinaysshenoy: > > Technically, Flow itself is a half-baked solution. :P
> 
> Why? What we're trying to solve in this PR is entirely independent of Flow. We wrote the results bus along the lines of how Activities exchange data. We would have still encountered this problem even if we weren't using Flow.

The reason I am bringing this is up is that Flow is a half-baked solution that doesn't provide a good way to start a screen with results which is, IMO, an essential feature for any navigation library.

> > Adding in a workaround or a hack for simple ID would be more cumbersome than using this.
> 
> Not really. We're using `SharedPreferences` for showing the `patient-saved` snackbar and it's working just fine. There's nothing cumbersome about it.
> 
> Instead of calling `screenRouter.popWithResult()`, you'd call `preference.set()` and `screenRouter.retrieveResult()` will get replaced by `preference.get()` or `preference.asObservable()`.

I don't see any difference between using shared preferences and what this solution is since shared preferences is a key value store on the disk, and what we have here is an in-memory key-value store. We used shared preferences in the patients screen to work around this problem in the exact same way, except that this is in-memory.

> > Then we can figure out what is a good solution for doing a startActivityWithResult equivalent for our use cases.
> 
> We already have a `startActivityWithResult()` equivalent in place. The current implementation works exactly like it. :p

Not really. We use the permission changes in the `PatientsScreen` and the streaming of permission results works fine there. I guess activity results are being delivered at a later point in time after the `PatientsScreen` is inflated in some way.

saket: > The reason I am bringing this is up is that Flow is a half-baked solution that doesn't provide a good way to start a screen with results which is, IMO, an essential feature for any navigation library.

Sure. Flow is just a library for managing backstack. That said, we already have an alternative in place and it works exactly like how `Activity.startActivityForResult()` works.

> I don't see any difference between using shared preferences and what this solution is since shared preferences is a key value store on the disk, and what we have here is an in-memory key-value store. 

Awesome. If you don't see any difference then let's go with SharedPreferences and work on the Screen Router APIs properly in the future 😄 

> We use the permission changes in the PatientsScreen and the streaming of permission results works fine there. I guess activity results are being delivered at a later point in time after the PatientsScreen is inflated in some way.

Correct, `PatientsScreen` is already inflated when the user grants permission so it works out fine in that case.

saket: On a second thought, we've spent a lot of time discussing this. I don't think we're going to reach an agreement. Let's go ahead with whatever you feel like doing right now. 

Btw, if ViewPager works out without an adapter then that will be the best.

vinaysshenoy:
> @vinaysshenoy I think if we wrap them inside `ViewPager` we'll still get support for swipe behavior. Can you please test this? If it works we'll be able to get rid of this whole discussion.

I don't want to do this now for multiple reasons:
- I am pretty sure `ViewPager` will not support statically adding pages via XML. Even if it does, I don't want to do this because it's not a scalable solution because there's no guarantee that we won't have to support dynamically loading pages in the future.
- I do not want to spend more time on testing other workarounds when one workaround is already done. In addition, the method that I replaced wasn't being used anywhere and no behaviour has changed, nor are we going to replace any of the older calls to the permission results stream with this new synchronous result call. I'd like to revisit this once we have some more breathing room after getting Simple ID out.

router/src/main/java/org/simple/clinic/router/screen/ScreenResults.kt
```diff
@@ -0,0 +1,25 @@
...
+
+class ScreenResults {
+
+  private val results = mutableMapOf<String, Any?>()
+
+  fun put(key: String, value: Any?) {
+    results[key] = value
+  }
+
+  fun get(key: String): Any? {
```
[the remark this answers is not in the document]
vinaysshenoy: Done.

PR #542, opened 2019-03-13
Add clicks to Recent patients

sanchita:
Pivotol story -> 
https://www.pivotaltracker.com/story/show/164533339
https://www.pivotaltracker.com/story/show/164564290

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/recent/RecentPatientEvents.kt
+++ b/app/src/main/java/org/simple/clinic/patient/recent/RecentPatientEvents.kt
@@ -1,0 +1,8 @@
+package org.simple.clinic.patient.recent
+
+import org.simple.clinic.widgets.UiEvent
+import java.util.UUID
+
+data class RecentPatientItemClicked(val patientUuid: UUID) : UiEvent {
+  override val analyticsName = "Recent Patients: Item clicked"
+}
--- a/app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
+++ b/app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
@@ -1,5 +1,6 @@
 package org.simple.clinic.patient.recent
 
+import android.annotation.SuppressLint
 import android.content.Context
 import android.util.AttributeSet
 import android.widget.FrameLayout
@@ -8,11 +9,19 @@ import com.jakewharton.rxbinding2.view.RxView
 import com.xwray.groupie.GroupAdapter
 import io.reactivex.Observable
 import io.reactivex.android.schedulers.AndroidSchedulers
+import io.reactivex.rxkotlin.ofType
+import io.reactivex.subjects.PublishSubject
 import kotlinx.android.synthetic.main.recent_patients.view.*
 import org.simple.clinic.R
 import org.simple.clinic.activity.TheActivity
+import org.simple.clinic.summary.PatientSummaryCaller
+import org.simple.clinic.summary.PatientSummaryScreenKey
+import org.simple.clinic.util.UtcClock
 import org.simple.clinic.widgets.ScreenCreated
+import org.simple.clinic.widgets.UiEvent
 import org.simple.clinic.widgets.visibleOrGone
+import org.threeten.bp.Instant
+import java.util.UUID
 import javax.inject.Inject
 
 class RecentPatientsView(context: Context, attrs: AttributeSet) : FrameLayout(context, attrs) {
@@ -20,8 +29,17 @@ class RecentPatientsView(context: Context, attrs: AttributeSet) : FrameLayout(co
   @Inject
   lateinit var controller: RecentPatientsViewController
 
+  @Inject
+  lateinit var activity: TheActivity
+
+  @Inject
+  lateinit var utcClock: UtcClock
+
   private val groupAdapter = GroupAdapter<com.xwray.groupie.ViewHolder>()
 
+  private val adapterEvents = PublishSubject.create<UiEvent>()
+
+  @SuppressLint("CheckResult")
   override fun onFinishInflate() {
     super.onFinishInflate()
     TheActivity.component.inject(this)
@@ -38,11 +56,17 @@ class RecentPatientsView(context: Context, attrs: AttributeSet) : FrameLayout(co
         .takeUntil(RxView.detaches(this))
         .observeOn(AndroidSchedulers.mainThread())
         .subscribe { uiChange -> uiChange(this) }
+
+    adapterEvents
+        .ofType<RecentPatientItemClicked>()
+        .takeUntil(RxView.detaches(this))
+        .subscribe { openPatientSummary(it.patientUuid) }
   }
 
   private fun screenCreates() = Observable.just(ScreenCreated())
 
   fun updateRecentPatients(recentPatients: List<RecentPatientItem>) {
+    recentPatients.forEach { it.uiEvents = adapterEvents }
     groupAdapter.update(recentPatients)
   }
 
@@ -56,5 +80,14 @@ class RecentPatientsView(context: Context, attrs: AttributeSet) : FrameLayout(co
 
   fun clearRecentPatients() {
     groupAdapter.clear()
+  }
+
+  private fun openPatientSummary(patientUuid: UUID) {
+    activity.screenRouter.push(
+        PatientSummaryScreenKey(
+            patientUuid = patientUuid,
+            caller = PatientSummaryCaller.SEARCH,
+            screenCreatedTimestamp = Instant.now(utcClock)
+        ))
   }
 }
--- a/app/src/main/res/layout/recent_patient_item_view.xml
+++ b/app/src/main/res/layout/recent_patient_item_view.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.cardview.widget.CardView xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.cardview.widget.CardView style="@style/Clinic.V2.CardView"
+  xmlns:android="http://schemas.android.com/apk/res/android"
   xmlns:app="http://schemas.android.com/apk/res-auto"
   xmlns:tools="http://schemas.android.com/tools"
   android:layout_width="match_parent"
--- a/app/src/main/res/layout/recent_patients.xml
+++ b/app/src/main/res/layout/recent_patients.xml
@@ -7,7 +7,8 @@
   <androidx.recyclerview.widget.RecyclerView
     android:id="@+id/recent_patient_recyclerview"
     android:layout_width="match_parent"
-    android:layout_height="wrap_content" />
+    android:layout_height="wrap_content"
+    android:layout_marginBottom="@dimen/spacing_96" />
 
   <TextView
     android:id="@+id/recent_patient_no_recent_patients"
--- a/app/src/main/res/layout/recent_patients.xml
+++ b/app/src/main/res/layout/recent_patients.xml
@@ -8,7 +8,8 @@
     android:id="@+id/recent_patient_recyclerview"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:layout_marginBottom="@dimen/spacing_96" />
+    android:clipToPadding="false"
+    android:paddingBottom="@dimen/spacing_96" />
 
   <TextView
     android:id="@+id/recent_patient_no_recent_patients"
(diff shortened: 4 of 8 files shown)


app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
```diff
@@ -33,16 +51,22 @@ class RecentPatientsView(context: Context, attrs: AttributeSet) : FrameLayout(co
...
+    Observable.mergeArray(screenCreates(), adapterEvents)
         .compose(controller)
         .takeUntil(RxView.detaches(this))
         .observeOn(AndroidSchedulers.mainThread())
         .subscribe { uiChange -> uiChange(this) }
+
+    adapterEvents
+        .ofType<RecentPatientItemClicked>()
+        .takeUntil(RxView.detaches(this))
+        .subscribe { openPatientSummary(it.patientUuid) }
```
rakshakhegde: Handling events should be done in controller

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
```diff
@@ -33,16 +51,22 @@ class RecentPatientsView(context: Context, attrs: AttributeSet) : FrameLayout(co
...
   }
 
   private fun screenCreates() = Observable.just(ScreenCreated())
 
   fun updateRecentPatients(recentPatients: List<RecentPatientItem>) {
+    recentPatients.forEach { it.uiEvents = adapterEvents }
```
rakshakhegde: this too should be part of the controller maybe?

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
```diff
@@ -57,4 +81,13 @@ class RecentPatientsView(context: Context, attrs: AttributeSet) : FrameLayout(co
   fun clearRecentPatients() {
     groupAdapter.clear()
   }
+
+  private fun openPatientSummary(patientUuid: UUID) {
+    activity.screenRouter.push(
+        PatientSummaryScreenKey(
+            patientUuid = patientUuid,
+            caller = PatientSummaryCaller.SEARCH,
```
rakshakhegde: Is this Caller valid? Aren't we opening this from home screen/recent patients?

rakshakhegde: Maybe we need to add a new entry in `PatientSummaryCaller`.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
sanchita: The reason I am not using controller for this case is because there is no decision-making required. It is just one `UiEvent` leading to a `UiChange` directly.
@saket and I have already discussed this. We are doing something similar in a few more places where a back click would lead to a screen change directly.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
sanchita: This cannot be a part of the controller as this a UI related logic. Also, the controller is agnostic to the `adapterEvents`.
Refer `PrescribedDrugScreen#populateDrugsList()` for this.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
sanchita: I thought about this but right now this `caller` has the same behavior as `SEARCH`.
The `caller` is used in `PatientSummaryScreen` to distinguish between a new and an existing patient.
@saket What do you think we should do here?

app/src/main/res/layout/recent_patient_item_view.xml
```diff
@@ -6,7 +6,8 @@
   android:layout_height="wrap_content"
   android:layout_marginBottom="@dimen/spacing_8"
   android:layout_marginEnd="@dimen/spacing_8"
-  android:layout_marginStart="@dimen/spacing_8">
+  android:layout_marginStart="@dimen/spacing_8"
+  android:foreground="?selectableItemBackground">
```
saket: We aren't using ripples as selectors for cards in the app. Can you check what other layouts are using?

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
saket: You're right. We're going to see this problem when we open patient summary by scanning BP passports. We should consider renaming them to `NEW_PATIENT` and `EXISTING_PATIENT`.

sanchita: Yes I was thinking the same. Since `SEARCH` is not very meaningful here.
I will rename it then.

saket: We'll also require renaming test functions for patient summary. Please do that in a separate PR?

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
rakshakhegde: Alright, can you change `Observable.mergeArray(screenCreates(), adapterEvents)` back to `Observable.mergeArray(screenCreates())`?

app/src/main/res/layout/recent_patient_item_view.xml
sanchita: Fixed

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
sanchita: I have already made this change in this PR.

saket: I see you've already renamed `SEARCH` to `EXISTING_PATIENT`. Don't forget to update test names in `PatientSummaryScreenControllerTest` and `PatientSummaryScreenKey#caller` to `type` or something.

saket: While you're at it, can you please also rename `PatientSummaryScreenKey#screenCreatedTimestamp` to `screenCreatedAt`?

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
sanchita: Done

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
sanchita: I could not find any test names specific to `caller`.

sanchita: > I see you've already renamed `SEARCH` to `EXISTING_PATIENT`. Don't forget to update test names in `PatientSummaryScreenControllerTest` and `PatientSummaryScreenKey#caller` to `type` or something.

Does that mean `PatientSummaryCaller` should be renamed as well?

saket: Yes. 

Let's do that in a later PR? We should get this PR merged first.

app/src/main/res/layout/recent_patients.xml
```diff
@@ -7,7 +7,8 @@
   <androidx.recyclerview.widget.RecyclerView
     android:id="@+id/recent_patient_recyclerview"
     android:layout_width="match_parent"
-    android:layout_height="wrap_content" />
+    android:layout_height="wrap_content"
+    android:layout_marginBottom="@dimen/spacing_96" />
```
saket: Hey, I don't think this is what we wanted. This margin will be persistent. We want the extra spacing to be present after the last item.

saket: Btw, did you intentionally capitalize `Recent` in this commit message? Looks… odd one out

`Add bottom margin to Recent patient recycler view`

saket: Oh wait, I just realized that because this list isn't scrollable, margins == spacing after the last item. 🤦‍♂️ 

Can we still do it the correct way please? 

```xml
android:paddingBottom="@dimen/spacing_96"
android:clipChildrenToPadding=false
```

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientEvents.kt
```diff
@@ -0,0 +1,8 @@
+package org.simple.clinic.patient.recent
+
+import org.simple.clinic.widgets.UiEvent
+import java.util.UUID
+
+data class RecentPatientItemClicked(val patientUuid: UUID) : UiEvent {
+  override val analyticsName = "Recent Patients: Item clicked"
```
saket: `Item clicked` -> `Patient clicked`? 

Item isn't natural to read when you're looking at analytics.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
```diff
@@ -8,20 +9,37 @@ import com.jakewharton.rxbinding2.view.RxView
...
 
+  @Inject
+  lateinit var activity: TheActivity
+
   private val groupAdapter = GroupAdapter<com.xwray.groupie.ViewHolder>()
 
+  private val adapterEvents = PublishSubject.create<UiEvent>()
+
+  @Inject
+  lateinit var utcClock: UtcClock
```
saket: Can you group all properties that are annotated with `@Inject`?

app/src/main/res/layout/recent_patients.xml
sanchita: Yes I did.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientEvents.kt
sanchita: `Recent Patients: Patient clicked`?

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
sanchita: If it is absolutely necessary.

app/src/main/res/layout/recent_patients.xml
sanchita: I don't understand. How is it any different?

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
saket: That's the convention we've followed _everywhere_

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientEvents.kt
saket: yep?

app/src/main/res/layout/recent_patients.xml
saket: oh okay

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientEvents.kt
sanchita: Reworded this commit.

app/src/main/res/layout/recent_patients.xml
saket: Marging: spacing between the RecyclerView and the screen bottom. This spacing will always be visible.
Padding + clipChildren=false: extra spacing *after the last item*. This spacing will only be visible when you reach the bottom of the list.

Of course, this difference gets nullified because of an external factor i.e., the recent patient list doesn't scroll in home screen. It's height is set to `wrap_content`.

saket: If it helps: `Padding + clipChildren=false` will act as a footer.

saket: Let me try giving you some screenshots for comparison.

saket: Here you go. With margin: 

Screenshot_1552564740

sanchita: Can we do this offline instead of here?
This looks strange for a PR feedback.

sanchita: Also, I have made the changes you asked.

PR #544, opened 2019-03-15
Add support for Meta models in BusinessId

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/164629790

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/patient/businessid/BusinessIdMetaAdapterAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/businessid/BusinessIdMetaAdapterAndroidTest.kt
@@ -1,0 +1,47 @@
+package org.simple.clinic.patient.businessid
+
+import androidx.test.runner.AndroidJUnit4
+import com.google.common.truth.Truth.assertThat
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.simple.clinic.TestClinicApp
+import java.util.UUID
+import javax.inject.Inject
+
+@RunWith(AndroidJUnit4::class)
+class BusinessIdMetaAdapterAndroidTest {
+
+  @field:Inject
+  lateinit var businessIdMetaAdapter: BusinessIdMetaAdapter
+
+  @Before
+  fun setUp() {
+    TestClinicApp.appComponent().inject(this)
+  }
+
+  @Test
+  fun all_known_mappings_of_business_meta_version_must_be_de_serialized_correctly() {
+    // This test is to ensure that whenever a new meta version is added, its corresponding
+    // json adapter is also added to the MoshiBusinessIdMetaAdapter.
+    val businessIdMetas = BusinessId.MetaVersion
+        .values()
+        .map { metaVersion ->
+          metaVersion to when (metaVersion) {
+            is BusinessId.MetaVersion.Unknown -> throw RuntimeException("$metaVersion should not be a part of known mappings!")
+            is BusinessId.MetaVersion.BpPassportV1 -> BusinessIdMeta.BpPassportV1(
+                assigningUserUuid = UUID.randomUUID(),
+                assigningFacilityUuid = UUID.randomUUID()
+            )
+            else -> throw RuntimeException("$metaVersion has not been handled!")
+          }
+        }
+
+    businessIdMetas
+        .forEach { (metaVersion, meta) ->
+          val serialized = businessIdMetaAdapter.serialize(meta, metaVersion)
+          val deserialized = businessIdMetaAdapter.deserialize(serialized, metaVersion)
+          assertThat(deserialized).isEqualTo(meta)
+        }
+  }
+}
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/simple/clinic/patient/businessid/BusinessId.kt
```diff
@@ -84,17 +85,20 @@ data class BusinessId(
...
-      @VisibleForTesting
+      @VisibleForTesting(otherwise = VisibleForTesting.NONE)
       fun random() = TypeAdapter.knownMappings.keys.shuffled().first()
+
+      @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+      fun values() = TypeAdapter.knownMappings.keys
     }
 
-    object BpPassportMetaV1 : MetaVersion()
+    object BpPassportMetaVersionV1 : MetaVersion()
```
rakshakhegde: `VersionV1`? V=Version, so can we rename this to something better? Maybe leave it as V1 or Version1?

rakshakhegde: Might also require commit reword

app/src/androidTest/java/org/simple/clinic/patient/businessid/BusinessIdMetaAdapterAndroidTest.kt
```diff
@@ -0,0 +1,56 @@
...
+
+  @Test
+  fun all_known_mappings_of_business_meta_version_must_be_de_serialized_correctly() {
+    // This test is to ensure that whenever a new meta version is added, its corresponding
+    // json adapter is also added to the MoshiBusinessIdMetaAdapter.
+    val businessIdMetas = BusinessId.MetaVersion
+        .values()
+        .map { metaVersion ->
+          metaVersion to when (metaVersion) {
+            is Unknown -> throw RuntimeException("$metaVersion should not be a part of known mappings!")
```
rakshakhegde: Do we need `Unknown` check? It's handled in else block. Also I'm not very comfortable with logic (loops in this case) in tests. Can we look into making this test more straightforward?

app/src/androidTest/java/org/simple/clinic/patient/businessid/BusinessIdMetaAdapterAndroidTest.kt
```diff
@@ -0,0 +1,56 @@
...
+            else -> throw RuntimeException("$metaVersion has not been handled!")
+          }
+        }
+
+    businessIdMetas
+        .map { (metaVersion, meta) ->
+          Triple(
+              metaVersion,
+              meta,
+              businessIdMetaAdapter.serialize(meta, metaVersion)
```
rakshakhegde: Let's simplify this block by serializing in the following `forEach` block instead of creating an additional `Triple` in between? It also makes it easier to follow, wdyt?

app/src/main/java/org/simple/clinic/patient/businessid/BusinessId.kt
saket: `VersionV1` 😂

app/src/androidTest/java/org/simple/clinic/patient/businessid/BusinessIdMetaAdapterAndroidTest.kt
vinaysshenoy: Discussed offline: This was intentionally added because we don't want to forget to add an adapter for (de)serialising new type of `BusinessIdMeta` objects. The error message is meant to be different for both the `Unknown` case and the other types.

app/src/androidTest/java/org/simple/clinic/patient/businessid/BusinessIdMetaAdapterAndroidTest.kt
vinaysshenoy: Makes sense, I'll do this.

app/src/main/java/org/simple/clinic/patient/businessid/BusinessId.kt
vinaysshenoy: > `VersionV1` 😂

This is intentional, since we have an enum `meta_version` which represents the type of metadata that is serialised in the `BusinessId` model.

So a model of type `BpPassport` might have three different types of encoded metadata, `v1`, `v2`, `v3`, which will be represent as `BpPassportMetaVersionV1`, `BpPassportMetaVersionV2`, `BpPassportMetaVersionV3` and so on.

vinaysshenoy: I have renamed the enum values to remove the `Version` string. This makes the enum values a little easier to read, but it also means that statically importing both enum types of `BusinessIdMeta` and `BusinessIdMetaVersion` in a single file is not possible since the values in both enums are not `BpPassportV1`.

vinaysshenoy:
@rakshakhegde feedback addressed.

PR #545, opened 2019-03-15
Add missing images for ADR #002

saket:
🤦‍♂️ this PR will fix the broken images in https://github.com/simpledotorg/simple-android/blob/master/doc/arch/002-bottom-sheets.md


rakshakhegde:
@saket Please change the url in the MD file to relative links

PR #550, opened 2019-03-18
"Patient has already visited" feature

rakshakhegde:
https://www.pivotaltracker.com/story/show/164068865

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/removepatient/RemoveAppointmentSheet.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/removepatient/RemoveAppointmentSheet.kt
@@ -41,6 +41,7 @@ class RemoveAppointmentSheet : BottomSheetActivity() {
   @Inject
   lateinit var controller: RemoveAppointmentSheetController
 
+  private val patientAlreadyVisitedRadioButton by bindView<RadioButton>(R.id.removeappointment_reason_patient_already_visited)
   private val notRespondingRadioButton by bindView<RadioButton>(R.id.removeappointment_reason_patient_not_responding)
   private val invalidPhoneNumberRadioButton by bindView<RadioButton>(R.id.removeappointment_reason_invalid_phone_number)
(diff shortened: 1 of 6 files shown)


app/src/test/java/org/simple/clinic/home/overdue/removepatient/RemoveAppointmentSheetControllerTest.kt
```diff
@@ -114,4 +114,20 @@ class RemoveAppointmentSheetControllerTest {
         MovedToPrivatePractitioner,
         Other)
   }
+
+  @Test
+  fun `when done is clicked, and a "Patient has already visited" reason is selected, then mark as already visited should be called`() {
```
vinaysshenoy: Rename test to `when done is clicked, and a "Patient has already visited" reason is selected, then appointment should be marked as visited`?

app/src/main/java/org/simple/clinic/home/overdue/removepatient/RemoveAppointmentSheet.kt
```diff
@@ -41,6 +41,7 @@ class RemoveAppointmentSheet : BottomSheetActivity() {
   @Inject
```
vinaysshenoy: Reword commit to `Add "Patient already visited" to remove appointment options`?

app/src/test/java/org/simple/clinic/home/overdue/removepatient/RemoveAppointmentSheetControllerTest.kt
rakshakhegde: Done

app/src/main/java/org/simple/clinic/home/overdue/removepatient/RemoveAppointmentSheet.kt
rakshakhegde: Reworded

PR #552, opened 2019-03-19
Hide sync indicator when the user is waiting for approval

sanchita:
Pivotol story -> https://www.pivotaltracker.com/story/show/164630895

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
@@ -4,6 +4,7 @@ import android.Manifest
 import android.annotation.SuppressLint
 import android.content.Context
 import android.util.AttributeSet
+import android.view.View
 import android.view.animation.AnimationUtils
 import android.widget.Button
 import android.widget.RelativeLayout
@@ -26,6 +27,7 @@ import org.simple.clinic.router.screen.ScreenRouter
 import org.simple.clinic.scanid.ScanSimpleIdScreenKey
 import org.simple.clinic.scanid.ScanSimpleIdScreenResult
 import org.simple.clinic.search.PatientSearchScreenKey
+import org.simple.clinic.sync.indicator.SyncIndicatorView
 import org.simple.clinic.util.RuntimePermissions
 import org.simple.clinic.widgets.ScreenCreated
 import org.simple.clinic.widgets.ScreenDestroyed
@@ -64,6 +66,7 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
   private val nameInAppointmentSavedText by bindView<TextView>(R.id.patients_summary_appointment_saved_name)
   private val dateInAppointmentSavedText by bindView<TextView>(R.id.patients_summary_appointment_saved_date)
   private val scanSimpleCardButton by bindView<Button>(R.id.patients_scan_simple_card)
+  private val syncIndicatorView by bindView<SyncIndicatorView>(R.id.patients_sync_indicator)
 
   @IdRes
   private var currentStatusViewId: Int = R.id.patients_user_status_hidden
@@ -217,10 +220,10 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
   }
 
   fun hideSyncIndicator() {
-    //TODO handle visibility
+    syncIndicatorView.visibility = View.GONE
   }
 
   fun showSyncIndicator() {
-    //TODO handle visibility
+    syncIndicatorView.visibility = View.VISIBLE
   }
 }
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
@@ -215,4 +215,12 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
   fun requestCameraPermissions() {
     RuntimePermissions.request(activity, CAMERA_PERMISSION, REQUESTCODE_CAMERA_PERMISSION)
   }
+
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
```diff
@@ -226,4 +227,20 @@ class PatientsScreenController @Inject constructor(
...
+  private fun handleVisibilityOfSyncIndicator(events: Observable<UiEvent>): Observable<UiChange> {
+    val screenCreated = events.ofType<ScreenCreated>()
+
+    val canUserSync = { userSession.canSyncData().distinctUntilChanged() }
+
+    val userApprovedForSync = Observables.combineLatest(screenCreated, canUserSync())
+        .filter { (_, canUserSync) -> canUserSync }
+        .map { { ui: Ui -> ui.showSyncIndicator() } }
+
+    val userWaitingForApproval = Observables.combineLatest(screenCreated, canUserSync())
```
saket: `canUserSync()` gets called immediately when this function is called so the lambda can be removed. By doing so we can also share the stream between `userApprovedForSync` and `userWaitingForApproval`.

app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
```diff
@@ -226,4 +227,20 @@ class PatientsScreenController @Inject constructor(
...
+        .filter { (_, canUserSync) -> canUserSync.not() }
+        .map { { ui: Ui -> ui.hideSyncIndicator() } }
+
+    return userApprovedForSync.mergeWith(userWaitingForApproval)
```
saket: Hrm, what was the decision behind splitting these two streams? 

I can't see any advantages and maybe we can simplify them:

```kotlin
val screenCreated = events.ofType<ScreenCreated>()

val canUserSync = userSession
    .canSyncData()
    .distinctUntilChanged()

return Observables.combineLatest(screenCreated, canUserSync)
    .map { (_, canSync) ->
      when {
        canSync -> { ui: Ui -> ui.showSyncIndicator() }
        else -> { ui: Ui -> ui.hideSyncIndicator() }
      }
    }
```

app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
```diff
@@ -226,4 +227,20 @@ class PatientsScreenController @Inject constructor(
         .filter { it.permissionResult == RuntimePermissionResult.GRANTED }
         .map { Ui::openScanSimpleIdCardScreen }
   }
+
+  private fun handleVisibilityOfSyncIndicator(events: Observable<UiEvent>): Observable<UiChange> {
```
saket: What do you think of `handleVisibilityOfSyncIndicator` -> `showSyncIndicatorIfUserCanSyncData` ?

The term `handle` is often abused because it's very generic

app/src/test/java/org/simple/clinic/home/patients/PatientsScreenControllerTest.kt
```diff
@@ -473,4 +479,23 @@ class PatientsScreenControllerTest {
...
+    whenever(userSession.refreshLoggedInUser()).thenReturn(Completable.never())
+    whenever(hasUserDismissedApprovedStatus.asObservable()).thenReturn(Observable.just(false))
+    whenever(hasUserDismissedApprovedStatus.get()).thenReturn(false)
+    whenever(patientSummaryResult.get()).thenReturn(PatientSummaryResult.NotSaved)
+
+    uiEvents.onNext(ScreenCreated())
+
+    canSyncStream.onNext(false)
+    canSyncStream.onNext(true)
+    canSyncStream.onNext(true)
```
saket: Nice! You're also verifying that multiple emissions with the same value do not have any effect.

app/src/test/java/org/simple/clinic/home/patients/PatientsScreenControllerTest.kt
```diff
@@ -473,4 +479,23 @@ class PatientsScreenControllerTest {
         listOf(RuntimePermissionResult.DENIED, false),
         listOf(RuntimePermissionResult.NEVER_ASK_AGAIN, false))
   }
+
+  @Test
+  fun `sync indicator should be visible only when user is approved for syncing`() {
+    whenever(userSession.loggedInUser()).thenReturn(Observable.just(PatientMocker.loggedInUser(status = UserStatus.WAITING_FOR_APPROVAL).toOptional()))
```
saket: There's a lot happening in this line. Let's break this?

```kotlin
val user = Just(PatientMocker.loggedInUser(status = WAITING_FOR_APPROVAL))
whenever(userSession.loggedInUser()).thenReturn(Observable.just(user))
```

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
```diff
@@ -217,10 +221,10 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
   }
 
   fun hideSyncIndicator() {
-    //TODO handle visibility
```
saket: We should toggle visibility changes with animation. A subtle fade in/out will look good.

app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
sanchita: Just to avoid using imperative code in RxStreams.
Maybe it is not that relevant here.

app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
sanchita: Right!

PR #556, opened 2019-03-20
Add help screen

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/164758290

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/HomeScreen.kt
+++ b/app/src/main/java/org/simple/clinic/home/HomeScreen.kt
@@ -6,6 +6,7 @@ import android.util.AttributeSet
 import android.view.ViewGroup
 import android.widget.Button
 import android.widget.RelativeLayout
+import androidx.appcompat.widget.Toolbar
 import androidx.viewpager.widget.ViewPager
 import com.jakewharton.rxbinding2.view.RxView
 import io.reactivex.Observable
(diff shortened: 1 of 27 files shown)


vinaysshenoy:
@rakshakhegde This is a large PR. I'm sorry about this.

This is mostly a clone of the Reports feature which you worked on, so it should hopefully be familiar enough that the bulk of it shouldn't be a problem to read.

app/src/main/res/values/colors_v2.xml
```diff
@@ -32,4 +32,5 @@
 
   <color name="launcher_background">@color/white100</color>
   <color name="facility_search_query_highlight">@color/blue2</color>
+  <color name="status_bar">#0a2e52</color>
```
rakshakhegde: Is this status bar's color not a standard color used across other screens?

app/src/main/java/org/simple/clinic/home/HomeScreen.kt
```diff
@@ -89,4 +93,20 @@ class HomeScreen(context: Context, attrs: AttributeSet) : RelativeLayout(context
...
+
+  fun showHelpButton() {
+    showHelpMenuItem
+        .apply {
+          isEnabled = true
+          isVisible = true
+        }
+  }
+
+  fun hideHelpButton() {
```
rakshakhegde: Can we combine these two show/hide functions by passing in a boolean, just the way we've done at `HelpScreen#showWebview`?

app/src/main/res/values/colors_v2.xml
vinaysshenoy: Nope, there's a separate color in Zeplin only for this: https://zpl.io/aR1dWmp

app/src/main/java/org/simple/clinic/home/HomeScreen.kt
vinaysshenoy: Sure.

PR #557, opened 2019-03-20
Clean up recent patients

saket:
I was trying to fix a bug in recent patients, but I'll need admin access to sandbox before I can continue my investigation. This PR contains changes I made in the meanwhile to clean up some code.

https://www.pivotaltracker.com/n/projects/2184102/stories/164720063

### DIFF
--- a/app/src/main/res/layout/recent_patients.xml
+++ b/app/src/main/res/layout/recent_patients.xml
@@ -22,5 +22,4 @@
     android:background="@drawable/recent_patient_no_recent_patients_background"
     android:padding="@dimen/spacing_16"
     android:text="@string/recent_patient_no_recent_patients" />
-
 </FrameLayout>
--- a/app/src/main/res/layout/recent_patients.xml
+++ b/app/src/main/res/layout/recent_patients.xml
@@ -5,21 +5,21 @@
   android:orientation="vertical">
 
   <androidx.recyclerview.widget.RecyclerView
-    android:id="@+id/recent_patient_recyclerview"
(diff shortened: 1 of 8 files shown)


saket:
@vinaysshenoy the last commit's description contains screenshots of before/after the commit is applied.

app/src/main/res/layout/recent_patients.xml
```diff
@@ -16,7 +16,7 @@
     style="@style/Clinic.V2.TextAppearance.Subtitle1Center.Grey1"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:layout_marginBottom="@dimen/spacing_96"
+    android:layout_marginBottom="104dp"
```
vinaysshenoy: I die a little inside every time I see a hardcoded value. 😢

vinaysshenoy:
@saket Looks good to me, but one of the tests (`verify_recent_patients_are_retrieved_as_expected`) seems to be failing. Looks like flakiness caused by generating timestamps for test data since the only thing that's unexpected is the order of the results. Take a look?

app/src/main/res/layout/recent_patients.xml
saket: Ohh but why? This value isn't shared with any screen. What do you suggest?

vinaysshenoy: Oh, I just meant in general. :P

saket:
@rakshakhegde maybe you should also take a look at the changes, because you originally wrote most of the changed code.

PR #561, opened 2019-03-22
Fix recent patient SQL query to pick up BP from all facilities

rakshakhegde:
https://www.pivotaltracker.com/story/show/164824779


saket:
@rakshakhegde I'm splitting this commit into two because it fixes two separate bugs with recent patients.

saket:
@rakshakhegde I'm splitting this commit into two because it fixes two separate bugs with recent patients. Can you take a look?

PR #563, opened 2019-03-23
Adds a check for evaluating whether to load Heap in classpath based on a gradle param

### LINKED ISSUE #347: Create a build configuration to disable the Heap Analytics integration across the app
App cannot compile on windows, any quick fix ?

aldefy:
(Attempting to fix #347) 

### Run  

To run with heap disabled : 
```./gradlew :app:assembleDebug``` or any build variant configuration

To run with heap enabled : 
```./gradlew -PheapEnabled=true :app:assembleDebug```

### Nice to have : 
Create an install script -  `../gradle/build_init.gradle` file and apply plugin checks for the same in the `app` module and add install tasks may be

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -6,11 +6,10 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 apply plugin: 'kotlin-kapt'
 apply plugin: 'io.sentry.android.gradle'
-
+if (hasProperty('heapEnabled'))
+  apply plugin: 'com.heapanalytics.android'
 
 apply from: "../quality/install-git-hook.gradle"
-if (hasProperty('heapEnabled'))
-    apply plugin: 'com.heapanalytics.android'
 
 repositories {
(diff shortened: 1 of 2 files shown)


aldefy:
@pratul hope this could be used to solve the problem

pratul:
@aldefy I made the mistake of clicking "Update branch" which made github add a merge commit to your branch. Please delete that commit and rebase your branch with master manually.

aldefy:
@pratul updated

pratul:
We had a discussion about this internally, and here are our thoughts on how we want to approach this:

- We would like this (and all other flags which will disable external services) to be enabled by default. So, this flag should probably be `enableHeapAnalytics` and should ship with the default as `true`. 
- Please document this flag with a explanatory comment in the settings.gradle file.

pratul:
Also, is there a specific reason why your flag functions the way it does? We have another build flag `runProguard` which works slightly differently. Unless you have a specific reason for your method, please switch to the method that already exists in the code.

PR #566, opened 2019-03-25
Fix: Hide "free call" callout if masking is disabled

rakshakhegde:


saket:
@rakshakhegde this PR contains untested logic. Please add a test?

rakshakhegde:
@saket The logic is in UI. Should the part about the header also be in the controller? How to write the test?

saket:
@rakshakhegde feature flags can live in the app for quite long, especially once we make them remote configurable. We should aim for always testing them to ensure we don't accidentally break the UI.

We could send the header item from the controller and ensure it only gets sent if the flag is enabled.

vinaysshenoy:
@rakshakhegde There's a conflict here. Fix it?

saket:
@rakshakhegde can we get this done today?

rakshakhegde:
@saket This is what I'm working on right now.

saket:
@rakshakhegde 🙂:+1:

rakshakhegde:
@saket This PR has been updated. Please review.

PR #571, opened 2019-03-26
Add query to check if a patient is defaulter

sanchita:

Part 1:
This PR just adds a query to check if a patient is defaulter.

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -19,6 +19,9 @@ import org.simple.clinic.bp.BloodPressureRepository
 import org.simple.clinic.drugs.PrescriptionRepository
 import org.simple.clinic.facility.Facility
 import org.simple.clinic.facility.FacilityRepository
+import org.simple.clinic.medicalhistory.MedicalHistory
+import org.simple.clinic.medicalhistory.MedicalHistory.Answer.NO
+import org.simple.clinic.medicalhistory.MedicalHistory.Answer.YES
 import org.simple.clinic.medicalhistory.MedicalHistoryRepository
 import org.simple.clinic.overdue.AppointmentRepository
 import org.simple.clinic.overdue.communication.CommunicationRepository
@@ -29,6 +32,7 @@ import org.simple.clinic.patient.businessid.BusinessIdMeta
 import org.simple.clinic.patient.businessid.BusinessIdMetaAdapter
 import org.simple.clinic.patient.recent.RecentPatient
 import org.simple.clinic.patient.sync.PatientPayload
+import org.simple.clinic.protocol.ProtocolDrug
 import org.simple.clinic.reports.ReportsRepository
 import org.simple.clinic.user.UserSession
 import org.simple.clinic.util.Just
@@ -40,6 +44,7 @@ import org.simple.clinic.util.unwrapJust
 import org.threeten.bp.Duration
 import org.threeten.bp.Instant
 import org.threeten.bp.LocalDate
+import org.threeten.bp.temporal.ChronoUnit
 import java.util.UUID
 import javax.inject.Inject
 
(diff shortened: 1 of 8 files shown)


app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -1416,4 +1421,177 @@ class PatientRepositoryAndroidTest {
...
+
+      protocolDrug?.let {
+        prescriptionRepository.savePrescription(patientUuid = patientUuid, drug = it).blockingAwait()
+      }
+
+      appointmentDate?.let {
+        appointmentRepository.schedule(patientUuid = patientUuid, appointmentDate = it).blockingGet()
+      }
+
+      val isDefaulter = patientRepository.isPatientDefaulter(patientUuid).blockingFirst()
```
vinaysshenoy: This is a side effect, and should not be done here. What should be done instead is that this method should return the generated `uuid`, and we should retain them and run the assertions after saving all the patients in the database.

vinaysshenoy: This test is also not readable since the test setup (saving the patient) is mixed with running the actual method that is being tested (`PatientRepository#isPatientADefaulter`).

In general, most tests should follow a structure like

```kotlin
fun test() {
  fun savePatient() {}

  // given
 val uuid1 = savePatientRecord()
 val uuid2 = savePatientRecord()
 ..

 // when
 val isPatient1ADefaulter = patientRepository.isPatientDefaulter(uuid1)
 val isPatient2ADefaulter = patientRepository.isPatientDefauler(uuid2)
 ..

 // expect
 assertThat(isPatient1ADefaulter.isTrue()
 assertThat(isPatient2ADefaulter).isFalse()
}
```

Following this rule makes the test a lot more readable and gives it a logical flow.

app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -200,5 +201,36 @@ data class Patient(
...
+              ELSE 0
+            END
+          ) AS isPatientDefaulter
+          FROM Patient P
+          LEFT JOIN (SELECT * FROM BloodPressureMeasurement BP WHERE BP.deletedAt IS NULL AND BP.patientUuid = :patientUuid ORDER BY BP.createdAt DESC LIMIT 1) BP
+          LEFT JOIN (SELECT * FROM MedicalHistory MH WHERE MH.deletedAt IS NULL AND MH.patientUuid = :patientUuid ORDER BY MH.updatedAt DESC LIMIT 1) MH
+          LEFT JOIN (SELECT * FROM PrescribedDrug PD WHERE PD.deletedAt IS NULL AND PD.patientUuid = :patientUuid ORDER BY PD.updatedAt DESC LIMIT 1) PD
+          LEFT JOIN Appointment A ON(A.patientUuid = P.uuid AND A.deletedAt IS NULL AND A.status = :scheduled)
+          WHERE P.uuid = :patientUuid AND P.deletedAt IS NULL
+
```
vinaysshenoy: Extra line?

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -1416,4 +1421,177 @@ class PatientRepositoryAndroidTest {
     assertThat(patientRepository.findPatientWithBusinessId(deletedBusinessIdentifier).blockingFirst()).isEqualTo(None)
     assertThat(patientRepository.findPatientWithBusinessId("missing_identifier").blockingFirst()).isEqualTo(None)
   }
+  
+  @Test
+  fun if_patient_is_defaulter_then_isPatientDefaulter_should_return_true() {
+
```
vinaysshenoy: Extra line?

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -1416,4 +1421,177 @@ class PatientRepositoryAndroidTest {
...
+        hasHadKidneyDisease: MedicalHistory.Answer = NO,
+        protocolDrug: ProtocolDrug?,
+        appointmentDate: LocalDate?
+    ) {
+      val patientUuid = patientRepository.saveOngoingEntry(testData.ongoingPatientEntry(fullName = fullName))
+          .andThen(patientRepository.saveOngoingEntryAsPatient())
+          .blockingGet()
+          .uuid
+
+      bpMeasurement?.forEach {
```
vinaysshenoy: `BloodPressureRepository` already has a method to save a list of `BloodPressureMeasurement` records. Do you want to use that instead?

app/src/main/java/org/simple/clinic/patient/Patient.kt
vinaysshenoy: Instead of using `*` to fetch all the fields from the tables, can we just fetch the ones that are necessary to compute whether a patient is a defaulter or not?

app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -200,5 +201,36 @@ data class Patient(
...
+  """)
+    abstract fun isPatientDefaulter(
```
vinaysshenoy: I am not sure how this will be used. If we are going to be showing a list of defaulters, it essentially means we will be fetching a list of all patient UUIDs and then running a query on each of them to check if they are a defaulter or not. Since we will have tens of thousands of records, showing the entire list of defaulter patients will be extremely slow and inefficient.

Do you want to change this query to return a list of defaulter patients instead?

app/src/main/java/org/simple/clinic/patient/PatientConfig.kt
```diff
@@ -10,7 +12,8 @@ const val MAXIMUM_SQLITE_QUERY_LIMIT = 1000
 data class PatientConfig(
     val limitOfSearchResults: Int,
     val scanSimpleCardFeatureEnabled: Boolean,
-    val recentPatientLimit: Int
+    val recentPatientLimit: Int,
+    val dueDaysToDefaulterPatient: Period
```
vinaysshenoy: `dueDaysToDefaulterPatient` doesn't seem to be very descriptive.

Change this to `daysSinceLastVisitToConsiderDefaulter` instead? It's a bit longer, but a lot more clearer IMO.

vinaysshenoy: In addition, the type is a `Period` but the name of the property starts with `days`. If we change the instantiation of the period to be months instead, this name is then incorrect.

Change it `timeSinceLastVisitToConsiderDefaulter`?

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
sanchita: This is added because there  `fun` here.
I found it more readable this way.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
sanchita: I agree that I should not combine the two operations.
Let me try to refactor this.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
sanchita: I need to add `patientUuid` to every `BloodPressureMeasurement`. Since I won't have that value outside this method `savePatientRecord`, I will have to update the `bpMeasurement` for each record to add `patientUuid`.

app/src/main/java/org/simple/clinic/patient/Patient.kt
sanchita: This query will be used to check if a particular patient is defaulter or not.
This will happen after we save a new BP for a patient. So I do not need a list of defaulters anywhere.

app/src/main/java/org/simple/clinic/patient/PatientConfig.kt
sanchita: I want to specific about the smallest unit here.
Since I do not care about the `time` but about `days`, I would rather use `periodToDefaulterPatient` instead of `time`

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
sanchita: Refactored this.

app/src/main/java/org/simple/clinic/patient/Patient.kt
sanchita: Done.

app/src/main/java/org/simple/clinic/patient/PatientConfig.kt
sanchita: Renamed to `periodToDefaulterPatientSinceLastVisit`

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -508,4 +508,17 @@ class PatientRepository @Inject constructor(
...
+
+  fun isPatientDefaulter(patientUuid: UUID): Observable<Boolean> {
+    return configProvider
+        .map { it.periodSinceLastVisitToDefaultPatient }
+        .switchMap {
+          database
+              .patientDao()
+              .isPatientDefaulter(
+                  patientUuid = patientUuid,
+                  lastRecordedBpThreshold = LocalDate.now(utcClock).minusDays(it.days.toLong())
```
vinaysshenoy: FYI: The `java.time` classes let you work directly with `Duration` and `Period`. You can do

```kotlin
LocalDate.now(utcClock).minus(periodSinceLastVisitToDefaultPatient)
```

instead of converting it to days and subtracting it as a primitive.

vinaysshenoy: Feel free to do this in the PR that completes this feature.

PR #572, opened 2019-03-28
Add isDefaulter to Appointment data model

sanchita:

Part 2: In this part we add `isDefaulter` flag to `Appointment` data model.


app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
```diff
@@ -210,6 +211,7 @@ class AppointmentRepository @Inject constructor(
           cancelReason = cancelReason,
           remindOn = remindOn,
           agreedToVisit = agreedToVisit,
+          isDefaulter = null,
```
saket: TODO: discuss strategy for adding `isDefaulter` to the payload.

app/src/main/java/org/simple/clinic/overdue/AppointmentPullResponse.kt
```diff
@@ -5,7 +5,7 @@ import com.squareup.moshi.JsonClass
 import org.simple.clinic.sync.DataPullResponse
 
 @JsonClass(generateAdapter = true)
-data class AppointmentPullResponseV2(
+data class AppointmentPullResponse(
```
saket: Shouldn't this be called `AppointmentPullResponseV3`?

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
```diff
@@ -242,6 +242,7 @@ object PatientMocker {
       syncStatus: SyncStatus = SyncStatus.PENDING,
       agreedToVisit: Boolean? = null,
       remindOn: LocalDate? = LocalDate.now(UTC).minusDays(2),
+      isDefaulter: Boolean? = null,
```
saket: Let's use a random value so that `isDefaulter` is used in tests? 

```kotlin
isDefaulter: Boolean? = listOf(true, false, null).shuffled().first()
```

saket:
@sanchita PR looks good overall

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
sanchita: I will write separate tests for checking random values of `isDefaulter` later.
Since I am not using this field at the moment.

saket: I'm sure you will, the randomness will be useful to other tests that are using `PatientMocker.appointment()`

PR #576, opened 2019-03-28
Use AppointmentSyncApiV3

sanchita:

### DIFF
--- a/app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
+++ b/app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
@@ -32,7 +32,7 @@ class AppointmentRepository @Inject constructor(
     private val appointmentConfigProvider: Single<AppointmentConfig>
 ) : SynceableRepository<Appointment, AppointmentPayload> {
 
-  fun schedule(patientUuid: UUID, appointmentDate: LocalDate): Single<Appointment> {
+  fun schedule(patientUuid: UUID, appointmentDate: LocalDate, isDefaulter: Boolean?): Single<Appointment> {
     val newAppointmentStream = facilityRepository
         .currentFacility(userSession)
         .firstOrError()
@@ -46,7 +46,7 @@ class AppointmentRepository @Inject constructor(
               cancelReason = null,
               remindOn = null,
               agreedToVisit = null,
-              isDefaulter = null,
+              isDefaulter = isDefaulter,
               syncStatus = SyncStatus.PENDING,
               createdAt = Instant.now(utcClock),
               updatedAt = Instant.now(utcClock),
@@ -57,6 +57,10 @@ class AppointmentRepository @Inject constructor(
         }
 
     return markOlderAppointmentsAsVisited(patientUuid).andThen(newAppointmentStream)
+  }
+
+  fun schedule(patientUuid: UUID, appointmentDate: LocalDate): Single<Appointment> {
+    return schedule(patientUuid, appointmentDate, null)
   }
 
   private fun markOlderAppointmentsAsVisited(patientUuid: UUID): Completable {
--- a/app/src/main/java/org/simple/clinic/overdue/AppointmentSync.kt
+++ b/app/src/main/java/org/simple/clinic/overdue/AppointmentSync.kt
@@ -3,6 +3,7 @@ package org.simple.clinic.overdue
 import com.f2prateek.rx.preferences2.Preference
 import io.reactivex.Completable
 import io.reactivex.Single
+import io.reactivex.rxkotlin.Observables
 import org.simple.clinic.sync.ModelSync
 import org.simple.clinic.sync.SyncConfig
 import org.simple.clinic.sync.SyncCoordinator
@@ -14,9 +15,12 @@ import javax.inject.Named
 class AppointmentSync @Inject constructor(
     private val syncCoordinator: SyncCoordinator,
     private val repository: AppointmentRepository,
-    private val api: AppointmentSyncApiV2,
+    private val apiV2: AppointmentSyncApiV2,
+    private val apiV3: AppointmentSyncApiV3,
     private val userSession: UserSession,
-    @Named("last_appointment_pull_token") private val lastPullToken: Preference<Optional<String>>,
+    private val appointmentConfig: Single<AppointmentConfig>,
+    @Named("last_appointment_pull_token") private val lastPullTokenV2: Preference<Optional<String>>,
+    @Named("last_appointment_pull_token_v3") private val lastPullTokenV3: Preference<Optional<String>>,
     @Named("sync_config_frequent") private val configProvider: Single<SyncConfig>
 ) : ModelSync {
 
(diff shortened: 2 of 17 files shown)


vinaysshenoy:
@sanchita I have marked this as a WIP PR again because of the data type change of the `is_defaulter` property.

app/src/main/java/org/simple/clinic/overdue/AppointmentSync.kt
```diff
@@ -14,9 +15,12 @@ import javax.inject.Named
 class AppointmentSync @Inject constructor(
     private val syncCoordinator: SyncCoordinator,
     private val repository: AppointmentRepository,
-    private val api: AppointmentSyncApiV2,
+    private val apiV2: AppointmentSyncApiV2,
+    private val apiV3: AppointmentSyncApiV3,
     private val userSession: UserSession,
-    @Named("last_appointment_pull_token") private val lastPullToken: Preference<Optional<String>>,
+    private val appointmentConfig: Single<AppointmentConfig>,
```
vinaysshenoy: We have started representing the config provider as an `Observable` instead of a `Single` in new places because we found that we are converting it to an `Observable` wherever we are using it anyway. Feel free to make this change if you like.

app/src/test/java/org/simple/clinic/sync/ModelSyncTest.kt
```diff
@@ -82,10 +84,16 @@ class ModelSyncTest {
               AppointmentSync(
                   syncCoordinator = syncCoordinator,
                   repository = mock(),
-                  api = mock(),
+                  apiV2 = mock(),
```
vinaysshenoy: Can you also add another instance of `AppointmentSync` where the v3 flag is enabled so that the code is exercised as a part of this test?

app/src/test/java/org/simple/clinic/sync/ModelSyncTest.kt
```diff
@@ -269,10 +277,16 @@ class ModelSyncTest {
...
```
vinaysshenoy: Can you also add another instance of AppointmentSync where the v3 flag is enabled so that the code is exercised as a part of this test?

app/src/test/java/org/simple/clinic/sync/ModelSyncTest.kt
```diff
@@ -398,10 +412,16 @@ class ModelSyncTest {
...
+                  apiV3 = mock(),
```
vinaysshenoy: Can you also add another instance of AppointmentSync where the v3 flag is enabled so that the code is exercised as a part of this test?

app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
```diff
@@ -32,7 +32,7 @@ class AppointmentRepository @Inject constructor(
     private val appointmentConfigProvider: Single<AppointmentConfig>
 ) : SynceableRepository<Appointment, AppointmentPayload> {
 
-  fun schedule(patientUuid: UUID, appointmentDate: LocalDate): Single<Appointment> {
+  fun schedule(patientUuid: UUID, appointmentDate: LocalDate, isDefaulter: Boolean?): Single<Appointment> {
```
vinaysshenoy: I know that this will change, but I'd like to give feedback on this design.

The `isDefaulter` flag might be `nullable` when defined in the entity, but that is because we have a limitation when migrating. However methods we expose from the API should behave as if the flag was not nullable, so this should be a `Boolean`.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -88,13 +95,44 @@ class ScheduleAppointmentSheetController @Inject constructor(
...
+    val patientUuidStream = events
+        .ofType<ScheduleAppointmentSheetCreated>()
+        .map { it.patientUuid }
+
+    val combinedStreams = Observables.combineLatest(
+        events.ofType<SchedulingSkipped>(),
+        patientUuidStream,
+        configProvider.toObservable()
+    )
+    val defaulterFeatureEnabledStream = combinedStreams
```
vinaysshenoy: This name implies that this is a stream that emits whether the feature is enabled or not. Rename this stream to `closeSheetWithDefaulterFeatureEnabled` and the other `closeSheetWithDefaulterFeatureDisabled`?

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -88,13 +95,44 @@ class ScheduleAppointmentSheetController @Inject constructor(
...
+        .filter { (_, _, config) -> config.isApiV3Enabled }
+        .switchMap { (_, patientUuid, config) ->
+          patientRepository
+              .isPatientDefaulter(patientUuid)
+              .flatMapSingle { isPatientDefaulter ->
+                when {
+                  isPatientDefaulter -> appointmentRepository
```
vinaysshenoy: What do you think of

```kotlin
val defaulterFeatureEnabledStream = combinedStreams
        .filter { (_, _, config) -> config.isApiV3Enabled }
        .switchMap { (_, patientUuid, config) ->
          val isPatientDefaulterStream = patientRepository
              .isPatientDefaulter(patientUuid)
              .replay()
              .refCount()

          val saveAppointmentAndCloseSheet = isPatientDefaulterStream
              .filter { isPatientADefaulter -> isPatientADefaulter }
              .flatMapSingle {
                appointmentRepository
                    .schedule(
                        patientUuid = patientUuid,
                        appointmentDate = LocalDate.now(utcClock).plus(config.appointmentDuePeriodForDefaulters),
                        isDefaulter = true)
              }
              .map { { ui: Ui -> ui.closeSheet() } }


          val closeSheetWithoutSavingAppointment = isPatientDefaulterStream
              .filter { isPatientADefaulter -> isPatientADefaulter.not() }
              .map { { ui: Ui -> ui.closeSheet() } }

          Observable.merge(saveAppointmentAndCloseSheet, closeSheetWithoutSavingAppointment)
        }
```

vinaysshenoy: It's more verbose than what was there earlier, but I also feel it's more readable. This might also fix the code complexity issue raised by Code Climate.

vinaysshenoy: There's a failing test which is also fixed by my implementation.

vinaysshenoy: > There's a failing test which is also fixed by my implementation.

Never mind, there's no failing test.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
sanchita: These stream names are created to distinguish between the two streams. 
The stream name `defaulterFeatureEnabledStream` here means that this stream deals with defaulter feature being enabled whereas the other one deals with the feature being disabled.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
sanchita: This is more or less the same implementation. Instead of using `when`, this implementation uses `filter`.
Although, I will test these changes and see if it fixes the Code climate issue.

app/src/main/java/org/simple/clinic/overdue/AppointmentSync.kt
sanchita: Oh yes. I knew this. Sorry I forgot.

sanchita: Since `AppointmentConfig` was not introduced in this PR, I would like to make this change in a separate story.

sanchita:
> @sanchita I have marked this as a WIP PR again because of the data type change of the `is_defaulter` property.

I will do this change as part of a separate PR.

PR #580, opened 2019-04-03
Add a mechanism to format an Identifier for display

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/165080249

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/shortcode/UuidShortCodeCreatorModule.kt
+++ b/app/src/main/java/org/simple/clinic/patient/shortcode/UuidShortCodeCreatorModule.kt
@@ -1,0 +1,18 @@
+package org.simple.clinic.patient.shortcode
+
+import dagger.Module
+import dagger.Provides
+
+@Module
+class UuidShortCodeCreatorModule {
+
+  @Provides
+  fun provideUuidShortCodeCreator(): UuidShortCodeCreator {
+    val requiredShortCodeLength = 7
+
+    return UuidShortCodeCreator(
+        requiredShortCodeLength = requiredShortCodeLength,
+        characterFilter = DigitFilter()
+    )
+  }
+}
--- a/app/src/main/java/org/simple/clinic/util/identifierdisplay/IdentifierDisplayAdapter.kt
+++ b/app/src/main/java/org/simple/clinic/util/identifierdisplay/IdentifierDisplayAdapter.kt
@@ -1,0 +1,33 @@
+package org.simple.clinic.util.identifierdisplay
+
+import org.simple.clinic.patient.businessid.Identifier
+import org.simple.clinic.patient.businessid.Identifier.IdentifierType.Unknown
+
+/**
+ * We need to render items of [Identifier] type in the UI, sometimes differently from how
+ * they are represented in code.
+ *
+ * For example, the short code of a [Identifier.IdentifierType.BpPassport] is represented as a
+ * 7-digit string, ex ("1234567"), but when rendering it in the UI we need to display it as
+ * "123 4567". We do not want to store these strings with the spaces in business layer because this
+ * is a presentation detail.
+ *
+ * This class is a helper class that is meant to convert [Identifier] instances to their display
+ * formats.
+ **/
+class IdentifierDisplayAdapter(
+    private val converters: Map<Identifier.IdentifierType, IdentifierToTextConverter>
+) {
+
+  fun toCharSequence(identifier: Identifier): CharSequence {
+    return when (identifier.type) {
+      is Unknown -> identifier.value
+      else -> converters[identifier.type]?.convert(identifier) ?: identifier.value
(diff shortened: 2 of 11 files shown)


app/src/main/java/org/simple/clinic/patient/shortcode/FilterDigits.kt
```diff
@@ -0,0 +1,8 @@
+package org.simple.clinic.patient.shortcode
+
+class FilterDigits: UuidShortCodeCreator.CharacterFilter {
```
sanchita: This class name should be on the same lines as `CharacterFilter`.
Can you rename it to `DigitFilter`

app/src/main/java/org/simple/clinic/patient/shortcode/UuidShortCodeCreatorModule.kt
```diff
@@ -0,0 +1,18 @@
...
+
+@Module
+class UuidShortCodeCreatorModule {
+
+  @Provides
+  fun provideUuidShortCodeCreator(): UuidShortCodeCreator {
+    val requiredShortCodeLength = 7
+
+    return UuidShortCodeCreator(
+        requiredShortCodeLength,
```
sanchita: Can you add named arguments here?

app/src/main/java/org/simple/clinic/util/IdentifierDisplayAdapter.kt
```diff
@@ -0,0 +1,44 @@
...
+ * is a presentation detail.
+ *
+ * This class is a helper class that is meant to convert [Identifier] instances to their display
+ * formats.
+ **/
+class IdentifierDisplayAdapter @Inject constructor(private val uuidShortCodeCreator: UuidShortCodeCreator) {
+
+  fun toCharSequence(identifier: Identifier): CharSequence {
+    return when (identifier.type) {
+      Identifier.IdentifierType.BpPassport -> bpPassportToCharSequence(identifier.value)
```
sanchita: Creating a method for each type of `Identifier` will quickly lead to one method for each type.
Eventually the churn of this file would be more.

I think a better way to implement this would be to create separate adapters for each `Identifier` and then create a map of `IdentifierType` to adapter.
We can then inject this map to the screens that need the identifier display.

sanchita: Also, I just realized that we are injecting `UuidShortCodeCreator` which is specific to `BpPassport` whereas this class is common to all `Identifier`s

app/src/main/java/org/simple/clinic/patient/shortcode/FilterDigits.kt
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/patient/shortcode/UuidShortCodeCreatorModule.kt
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/util/IdentifierDisplayAdapter.kt
vinaysshenoy: Sounds good. Done.

app/src/main/java/org/simple/clinic/util/identifierdisplay/IdentifierDisplayAdapter.kt
```diff
@@ -0,0 +1,33 @@
...
+) {
+
+  fun toCharSequence(identifier: Identifier): CharSequence {
+    return when (identifier.type) {
+      is Unknown -> identifier.value
+      else -> converters[identifier.type]?.convert(identifier) ?: identifier.value
+    }
+  }
+
+  interface IdentifierToCharSequenceConverter {
```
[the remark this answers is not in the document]
vinaysshenoy: Sounds good. I'll change it,

app/src/main/java/org/simple/clinic/util/identifierdisplay/BpPassportCharSequenceConverter.kt
```diff
@@ -0,0 +1,27 @@
+package org.simple.clinic.util.identifierdisplay
+
+import org.simple.clinic.patient.businessid.Identifier
+import org.simple.clinic.patient.shortcode.UuidShortCode
+import org.simple.clinic.patient.shortcode.UuidShortCodeCreator
+import java.util.UUID
+import javax.inject.Inject
+
+class BpPassportCharSequenceConverter @Inject constructor(
```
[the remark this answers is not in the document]
vinaysshenoy: Done.

vinaysshenoy:
Closing because of Bitrise.

PR #583, opened 2019-04-05
Add `appointment_type` to  Appointment

sanchita:
Pivotol story -> https://www.pivotaltracker.com/story/show/165143985

This story is in continuation to the story for adding defaulters to Overdue:
https://www.pivotaltracker.com/story/show/164655585

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
@@ -1410,6 +1410,55 @@ class DatabaseMigrationAndroidTest {
       assertThat(it.string("deletedAt")).isEqualTo("deleted-at")
     }
   }
+
+  @Test
+  fun migration_32_to_33() {
+    val tableName = "Appointment"
+    val columnCount = 13
+
+    val db_v32 = helper.createDatabase(version = 32)
+    db_v32.assertColumnCount(tableName, columnCount)
+
+    db_v32.execSQL("""
+            INSERT INTO $tableName VALUES(
+            'uuid',
+            'patientUuid',
+            'facility-uuid',
+            'scheduled-date',
+            'status',
+            'cancel-reason',
+            'remind-on',
+            '1',
+            null,
+            'PENDING',
+            'created-at',
+            'updated-at',
+            null)
+        """)
+
+    val db_v33 = helper.migrateTo(version = 33)
+
+    db_v33.query("""
+           SELECT * FROM $tableName
+        """).use {
+      it.moveToNext()
+      assertThat(it.columnCount).isEqualTo(columnCount)
+
+      assertThat(it.string("uuid")).isEqualTo("uuid")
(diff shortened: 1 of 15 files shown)


rakshakhegde:
@sanchita The spacing has been changed from 2 to 4 spaces in the entire file in `DatabaseMigrationAndroidTest.kt`. Can you please fix that?

sanchita:
> @sanchita The spacing has been changed from 2 to 4 spaces in the entire file in `DatabaseMigrationAndroidTest.kt`. Can you please fix that?

Done.

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -1410,6 +1410,55 @@ class DatabaseMigrationAndroidTest {
...
+
+      assertThat(it.string("uuid")).isEqualTo("uuid")
+      assertThat(it.string("patientUuid")).isEqualTo("patientUuid")
+      assertThat(it.string("facilityUuid")).isEqualTo("facility-uuid")
+      assertThat(it.string("scheduledDate")).isEqualTo("scheduled-date")
+      assertThat(it.string("status")).isEqualTo("status")
+      assertThat(it.string("cancelReason")).isEqualTo("cancel-reason")
+      assertThat(it.string("remindOn")).isEqualTo("remind-on")
+      assertThat(it.boolean("agreedToVisit")).isTrue()
+      assertThat(it.string("appointmentType")).isEqualTo(null)
```
rakshakhegde: Instead of saving "appointmentType" as null, can we save a valid value?

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
```diff
@@ -45,6 +49,42 @@ data class Appointment(
...
+
+    object Manual : AppointmentType()
+
+    object Automatic : AppointmentType()
+
+    data class Unknown(val actual: String) : AppointmentType()
+
+    object TypeAdapter : SafeEnumTypeAdapter<AppointmentType>(
+        knownMappings = mapOf(
+            Manual to "manual",
```
rakshakhegde: Are there tests for checking if the conversion from Enum to string and string to enum is working as expected?

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
sanchita: We are not inserting this value explicitly. 
When we are creating `appointment_type` column, we expect this value to be set a `null`.
This assertion is to test that.

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
sanchita: This will be covered as part of the `AppointmentSyncAndroidTest`.

Whenever we add a new field to any of the payloads, we test syncing using the `*SyncAndroidTest` which internally calls the respective `SynceableRepository` methods to merge payload data to database models.
This will automatically get tested as part of that flow.

PR #585, opened 2019-04-08
Update CHANGELOG

rakshakhegde:


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # CHANGELOG
 
-## xxxxxxxxx
-- ?
+## Upcoming release
+- Deleted BP, MedicalHistory, Communication, etc. entries are not picked up while displaying Recent Patients
```
pratul: It would be nice to have the language be pleasing to an end-user. Try not to be rigid in the tone. Here's what I'd suggest... 

```suggestion
- Recent Patients list will not get affected by blood pressures and medical history entries that have been deleted
```

rakshakhegde: Sure

PR #586, opened 2019-04-08
Add ADR for single-activity Flow-based navigation

pratul:


doc/arch/005-single-activity-view-based-navigation.md
```diff
@@ -0,0 +1,33 @@
...
+We will use a single `Activity` to hold all our “screens”, and entirely avoid using `Fragment` classes. To switch between “screens”, we will use [Flow](https://github.com/square/flow) — a library that enables navigating between different UI states. 
+
+Flow makes it easy to construct screens using just `View` classes. Views are inflated synchronously, which drastically reduces complexity and eliminates race conditions. When we inflate a layout XML using `LayoutInflater.inflate()`, we are guaranteed that the View will be ready for making any modifications. A single `Activity` also allows for easy animations, because inter-activity transitions are harder to get right. 
+
+
+## Consequences 
+
+Using a single `Activity` to hold all views, and using Flow for navigation is quite unusual. Flow is an external library, written and maintained by a third-party — it is open source, but if the library gets neglected, we will have to take up the task of maintaining it. 
+
+Using `Fragment` classes and `Activity` classes is so common in the community that help is easy to find and obtain. If we run into any roadblocks with Flow, we will not find the same degree of support.
```
vinaysshenoy: IMO, we can drop the 'classes' after `Fragment`, so the sentence just becomes "Using `Fragment` and `Activity` classes is so common...". What do you think?

vinaysshenoy: We might also want to add that we will need to maintain our own mechanism for starting a screen which has to complete with a result, and how to deliver that result to the invoking screen.

vinaysshenoy: @pratul This is marked as "Resolved", but there's no change in the ADR.

PR #593, opened 2019-04-11
Add capability to render Identifier types to the IdentifierDisplayAdapter

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/165271096

### DIFF
--- a/app/src/main/java/org/simple/clinic/util/identifierdisplay/IdentifierDisplayAdapter.kt
+++ b/app/src/main/java/org/simple/clinic/util/identifierdisplay/IdentifierDisplayAdapter.kt
@@ -18,7 +18,7 @@ import org.simple.clinic.patient.businessid.Identifier.IdentifierType.Unknown
  **/
 class IdentifierDisplayAdapter(
     @get:VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-    val converters: Map<Identifier.IdentifierType, IdentifierToTextConverter>,
+    val converters: Map<Identifier.IdentifierType, IdentifierDisplayFormatter>,
     private val unknownValueFallback: (Identifier) -> String,
     private val unknownTypeFallback: (Identifier) -> String
 ) {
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/simple/clinic/util/identifierdisplay/IdentifierDisplayAdapter.kt
```diff
@@ -19,7 +19,8 @@ import org.simple.clinic.patient.businessid.Identifier.IdentifierType.Unknown
 class IdentifierDisplayAdapter(
     @get:VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     val converters: Map<Identifier.IdentifierType, IdentifierToTextConverter>,
-    private val unknownValueFallback: (Identifier) -> String
+    private val unknownValueFallback: (Identifier) -> String,
```
sanchita: I was just thinking why can't we create an `UnknownIdentifierFormatter`?
Rather than treating `Unknown` as a special case entirely, we could use this formatter.

Also, another advantage of doing this would be that we don't need to keep adding constructor parameters as we add more formatting methods.

Wdyt?

app/src/test/java/org/simple/clinic/util/identifierdisplay/IdentifierDisplayAdapterTest.kt
```diff
@@ -91,11 +93,81 @@ class IdentifierDisplayAdapterTest {
...
+
+    val convertedForDisplay = listOf(knownIdentifier, unknownIdentifier)
+        .map(identifierDisplayAdapter::typeAsText)
+
+    val expected = listOf("bp_passport", expectedConvertedUnknownType)
+    assertThat(convertedForDisplay).isEqualTo(expected)
+  }
+
+  @Suppress("Unused")
+  private fun `params for unknown identifier type fallback`(): List<List<Any>> {
```
sanchita: The fallback for all `Unknown` identifiers will always be the same.
Hence, I think, keeping the `fallback` configurable is not adding any value.

To use `UnknownIdentifierFormatter` also makes sense for testing and we don't need to do parametrized testing for different fallbacks for `Unknown` anymore.

app/src/main/java/org/simple/clinic/util/identifierdisplay/IdentifierDisplayAdapter.kt
vinaysshenoy: I did think of this at first, but `Unknown` is a data class and not an `object`. So doing this will mean that I'll have to change the type of `converters` to `Map<Class<IdentifierType>, IdentifierDisplayFormatter>`.

I'll explore this and see what it looks like.

vinaysshenoy: I've done this in the last commit. Please check?

app/src/test/java/org/simple/clinic/util/identifierdisplay/IdentifierDisplayAdapterTest.kt
vinaysshenoy: I've done this in the last commit. Please check?

PR #594, opened 2019-04-11
More documentation updates

pratul:


doc/adding-changing-build-variant.md
```diff
@@ -0,0 +1,19 @@
...
+
+## For adding a new build variant
+
+- Add the product flavour in the `productFlavours` closure in `app/build.gradle`.
+- Update the `variantFilter` closure to remove the unnecessary build for the new variant.
+- Add the API endpoint in `gradle.properties` and add this as the `API_ENDPOINT` build config field when defining the product flavour in `app/build.gradle`. This might need to be overriden on Bitrise based on the needs of the build flavour.
+- Add the Heap ID in `gradle.properties` and add this as the `HEAP_ID` build config field when defining the product flavour in `app/build.gradle`. This might need to be overriden on the Bitrise based on the needs of the build flavour.
+- Update the `applicationIdSuffix` and `versionNameSuffix` when defining the product flavour.
+- Update the `afterEvaluate` closure at the end of the `android` block in `app/build.gradle` to include the build tasks for the new variant (only needed if you have enabled proguard for the build).
+
```
sanchita: Do you think it would would be a good idea to add a snapshot of sample code here to make it easier to understand?

pratul: IMO that might tend towards "over-documentation". You raise a good point though, so I've clarified some of the steps with examples of current variants, and updated the wording.

PR #598, opened 2019-04-13
Add link id with patient screen

sanchita:
Pivotol story -> https://www.pivotaltracker.com/story/show/165211601

Part 1 of the story above. This PR creates a new bottom sheet `LinkIdWithPatient`

### DIFF
--- a/app/src/test/java/org/simple/clinic/summary/linkId/LinkIdWithPatientSheetControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/summary/linkId/LinkIdWithPatientSheetControllerTest.kt
@@ -1,6 +1,8 @@
 package org.simple.clinic.summary.linkId
 
+import com.nhaarman.mockito_kotlin.any
 import com.nhaarman.mockito_kotlin.mock
+import com.nhaarman.mockito_kotlin.never
 import com.nhaarman.mockito_kotlin.verify
 import com.nhaarman.mockito_kotlin.whenever
 import io.reactivex.Single
@@ -17,7 +19,6 @@ import java.util.UUID
 
 class LinkIdWithPatientSheetControllerTest {
 
-
   @get:Rule
   val rxErrorsRule = RxErrorsRule()
 
@@ -27,22 +28,17 @@ class LinkIdWithPatientSheetControllerTest {
 
   private val uiEvents = PublishSubject.create<UiEvent>()
 
-  private lateinit var controller: LinkIdWithPatientSheetController
+  private val controller = LinkIdWithPatientSheetController(patientRepository)
 
   private val patientUuid = UUID.randomUUID()
 
-  private lateinit var identifier: Identifier
+  private val identifier = Identifier(
+      value = patientUuid.toString(),
+      type = Identifier.IdentifierType.random()
+  )
 
   @Before
   fun setUp() {
-
-    controller = LinkIdWithPatientSheetController(patientRepository)
-
-    identifier = Identifier(
-        value = patientUuid.toString(),
-        type = Identifier.IdentifierType.random()
-    )
-
     uiEvents
         .compose(controller)
         .subscribe { uiChange -> uiChange(sheet) }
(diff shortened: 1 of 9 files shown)


app/src/test/java/org/simple/clinic/summary/linkId/LinkIdWithPatientSheetControllerTest.kt
```diff
@@ -0,0 +1,72 @@
...
+  @get:Rule
+  val rxErrorsRule = RxErrorsRule()
+
+  private val patientRepository = mock<PatientRepository>()
+
+  private val sheet = mock<LinkIdWithPatientSheet>()
+
+  private val uiEvents = PublishSubject.create<UiEvent>()
+
+  private lateinit var controller: LinkIdWithPatientSheetController
```
rakshakhegde: Instead of using `lateinit var`, we can use `val` here. JUnit recreates the entire class along with the properties after every test.
So instead let's do this?
`private val controller = LinkIdWithPatientSheetController(patientRepository )`

app/src/test/java/org/simple/clinic/summary/linkId/LinkIdWithPatientSheetControllerTest.kt
```diff
@@ -0,0 +1,72 @@
...
+
+  private val patientUuid = UUID.randomUUID()
+
+  private lateinit var identifier: Identifier
```
rakshakhegde: Similarly here, identifier can be turned into `val`.

app/src/test/java/org/simple/clinic/summary/linkId/LinkIdWithPatientSheetControllerTest.kt
```diff
@@ -0,0 +1,72 @@
...
+    uiEvents.onNext(LinkIdWithPatientSheetCreated(patientUuid, identifier))
+    uiEvents.onNext(LinkIdWithPatientAddClicked)
+
+    verify(patientRepository).addIdentifierToPatient(patientUuid, identifier)
+    verify(sheet).closeSheet()
+  }
+
+  @Test
+  fun `when cancel is clicked, the sheet should close without saving id`() {
+    uiEvents.onNext(LinkIdWithPatientCancelClicked)
```
rakshakhegde: This test should actually be failing right? Even without "created" event, the actions are being performed.

Please add another test to ensure that sending `LinkIdWithPatientCancelClicked` before `LinkIdWithPatientSheetCreated` does not trigger `closeSheet`.

sanchita: Cancel cannot be clicked until the sheet is created. This is not possible in the actual UI flow.
Hence I did not add `LinkIdWithPatientSheetCreated`.
Having said that I think that from the point of the tests, it does make sense to emit the created event first.
I will add this.

app/src/test/java/org/simple/clinic/summary/linkId/LinkIdWithPatientSheetControllerTest.kt
sanchita: The reason for initializing `LinkIdWithPatientSheetController` in every test is so that when we run test on the controller, it creates a clean instance every time. Here it may not make a difference right now but if for instance we start using a `clock`, we would want it to be reset with every test.

app/src/test/java/org/simple/clinic/summary/linkId/LinkIdWithPatientSheetControllerTest.kt
sanchita: I will do this.

app/src/test/java/org/simple/clinic/summary/linkId/LinkIdWithPatientSheetControllerTest.kt
rakshakhegde: You're right, we need clean instance every time. That's precisely what JUnit does, it recreates the entire class for every single test. So we can actually just use `val` and it will work.

sanchita: Ok. I did not know this. I will make the change.

sanchita: Done this change.

PR #601, opened 2019-04-16
Add support for registering a new patient with an Identifier

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/165344520

This is part 1 of the story which adds the necessary changes in the repository to create a new patient with a `BusinessId`. This feature will be completed in the next PR which makes the changes to the `PatientEntryScreen`.

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/Patient.kt
+++ b/app/src/main/java/org/simple/clinic/patient/Patient.kt
@@ -8,6 +8,7 @@ import androidx.room.Index
 import androidx.room.PrimaryKey
 import androidx.room.Query
 import io.reactivex.Flowable
+import org.intellij.lang.annotations.Language
 import org.simple.clinic.medicalhistory.MedicalHistory
 import org.simple.clinic.overdue.Appointment
 import org.simple.clinic.patient.businessid.BusinessId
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -229,5 +216,34 @@ data class Patient(
...
+
+          PPN.uuid phone_uuid, PPN.patientUuid phone_patientUuid, PPN.number phone_number,
+          PPN.phoneType phone_phoneType, PPN.active phone_active,
+          PPN.createdAt phone_createdAt, PPN.updatedAt phone_updatedAt, PPN.deletedAt phone_deletedAt,
+
+          BI.uuid businessid_uuid, BI.patientUuid businessid_patientUuid, BI.identifier businessid_identifier,
+          BI.identifierType businessid_identifierType, BI.meta businessid_meta, BI.metaVersion businessid_metaVersion,
+          BI.createdAt businessid_createdAt, BI.updatedAt businessid_updatedAt, BI.deletedAt businessid_deletedAt
+        FROM Patient P
+        INNER JOIN PatientAddress PA ON P.addressUuid == PA.uuid
```
rakshakhegde: Is there a difference between single equals vs double equals? We've used single equals everywhere else.

vinaysshenoy: In SQLite, both are acceptable and they are the same thing. I think @sanchita looked this up recently.

sanchita: Both are essentially the same in SQLite.

PR #603, opened 2019-04-16
Open LinkIdWithPatientSheet from PatientSummaryScreen

sanchita:
Pivotol story -> https://www.pivotaltracker.com/story/show/165211601

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
@@ -34,6 +34,7 @@ import org.simple.clinic.patient.Gender
 import org.simple.clinic.patient.Patient
 import org.simple.clinic.patient.PatientAddress
 import org.simple.clinic.patient.PatientPhoneNumber
+import org.simple.clinic.patient.businessid.Identifier
 import org.simple.clinic.router.screen.ActivityResult
 import org.simple.clinic.router.screen.BackPressInterceptCallback
 import org.simple.clinic.router.screen.BackPressInterceptor
@@ -42,6 +43,7 @@ import org.simple.clinic.router.screen.SCREEN_CHANGE_ANIMATION_DURATION
 import org.simple.clinic.router.screen.ScreenRouter
 import org.simple.clinic.scheduleappointment.ScheduleAppointmentSheet
 import org.simple.clinic.summary.addphone.AddPhoneNumberDialog
+import org.simple.clinic.summary.linkId.LinkIdWithPatientSheet
 import org.simple.clinic.summary.updatephone.UpdatePhoneNumberDialog
 import org.simple.clinic.util.Optional
 import org.simple.clinic.util.UtcClock
@@ -321,6 +323,10 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
 
   fun showUpdatePrescribedDrugsScreen(patientUuid: UUID) {
     screenRouter.push(PrescribedDrugsScreenKey(patientUuid))
+  }
+
+  fun openLinkIdWithPatientSheet(patientUuid: UUID, identifier: Identifier) {
+    activity.startActivity(LinkIdWithPatientSheet.intent(context, patientUuid, identifier))
   }
 }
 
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -332,6 +333,16 @@ class PatientSummaryScreenController @Inject constructor(
         .mergeWith(showScheduleAppointmentSheetOnDoneClicks)
   }
 
+  private fun openLinkIdWithPatientSheet(events: Observable<UiEvent>): Observable<UiChange> {
+    return events
+        .ofType<PatientSummaryScreenCreated>()
+        .filter { it.openIntention is OpenIntention.LinkIdWithPatient }
```
vinaysshenoy: This can also be written as

```kotlin
events
  .ofType<PatientSummaryScreenCreated>()
  .map { it.openIntention }
  .ofType<LinkIdWithPatient>()
```

This has the advantage of not requiring to cast it again further down the chain.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -322,7 +322,8 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
   }
 
   fun showUpdatePrescribedDrugsScreen(patientUuid: UUID) {
-    screenRouter.push(PrescribedDrugsScreenKey(patientUuid))
+    openLinkIdWithPatientSheet(patientUuid, Identifier(patientUuid.toString(), Identifier.IdentifierType.BpPassport))
```
vinaysshenoy: I'm assuming this was added for testing. 😅

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -322,7 +322,8 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
   }
```
vinaysshenoy: What do you think of squashing this commit with the first one?

app/src/main/res/layout/sheet_link_id_with_patient.xml
```diff
@@ -22,7 +22,6 @@
     android:layout_height="wrap_content"
```
vinaysshenoy: The `linkidwithpatient_text` TextView has the text appearance set with a `<style>` tag instead of a `<textAppearance>` tag. Please change?

app/src/main/res/layout/sheet_link_id_with_patient.xml
```diff
@@ -32,28 +31,26 @@
...
       android:layout_weight="1"
       android:background="@drawable/button_outline_ripple"
       android:backgroundTint="@color/grey1"
       android:text="@string/linkidwithpatient_button_cancel"
-      android:textAppearance="@style/Clinic.V2.TextAppearance.Button1.Grey1" />
+      android:textAppearance="@style/Clinic.V2.TextAppearance.Button1.Grey1"
+      tools:ignore="ButtonStyle" />
 
-    <org.simple.clinic.widgets.PrimarySolidButton
+    <Button
```
vinaysshenoy: This button is not the same size as in the design. Please fix?

app/src/main/java/org/simple/clinic/summary/linkId/LinkIdWithPatientSheet.kt
```diff
@@ -32,20 +38,31 @@ class LinkIdWithPatientSheet : BottomSheetActivity() {
   @Inject
   lateinit var controller: LinkIdWithPatientSheetController
 
+  @Inject
+  lateinit var idDisplayAdapter: IdentifierDisplayAdapter
+
+  lateinit var identifier: Identifier
```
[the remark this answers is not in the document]
sanchita: I will do this for `identifier`.
`patientUuid` is already accessed locally to a method.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
sanchita: This is not feasible since I also need `it.patientUuid`.

app/src/main/res/layout/sheet_link_id_with_patient.xml
sanchita: Fixed.

vinaysshenoy:
@sanchita LGTM. There are two parts which we can complete in the next PR:

- We need to handle the clicks on the "Cancel" and system back button to close both this screen and the summary screen.
- The ripple colour for the "Cancel" button needs to be changed to take into account the different outline colour of the button.

PR #605, opened 2019-04-16
Display Identifier in PatientEntryScreen

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/165344520

### DIFF
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
@@ -1,12 +1,7 @@
 package org.simple.clinic.newentry
 
+import android.annotation.SuppressLint
 import android.content.Context
-import com.google.android.material.textfield.TextInputLayout
-import androidx.transition.ChangeBounds
-import androidx.transition.Fade
-import androidx.transition.TransitionManager
-import androidx.transition.TransitionSet
-import androidx.interpolator.view.animation.FastOutSlowInInterpolator
 import android.util.AttributeSet
 import android.view.View
 import android.view.ViewGroup
@@ -17,6 +12,12 @@ import android.widget.RadioGroup
 import android.widget.RelativeLayout
 import android.widget.ScrollView
 import android.widget.TextView
+import androidx.interpolator.view.animation.FastOutSlowInInterpolator
+import androidx.transition.ChangeBounds
+import androidx.transition.Fade
+import androidx.transition.TransitionManager
+import androidx.transition.TransitionSet
+import com.google.android.material.textfield.TextInputLayout
 import com.jakewharton.rxbinding2.view.RxView
 import com.jakewharton.rxbinding2.widget.RxRadioGroup
 import com.jakewharton.rxbinding2.widget.RxTextView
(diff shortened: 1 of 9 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -66,6 +65,7 @@
   <string name="patiententry_error_dateofbirth_is_in_future">Fix patient’s date of birth</string>
   <string name="patiententry_error_phonenumber_length_less">Phone number cannot be less than 6 digits</string>
   <string name="patiententry_error_phonenumber_length_more">Phone number cannot be more than 12 digits</string>
+  <string name="patiententry_identifier">BP Passports</string>
```
sanchita: Shouldn't we be using `plurals` for this string?

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
```diff
@@ -126,32 +128,37 @@ class PatientEntryScreenController @Inject constructor(
...
+              personDetailChanges,
+              phoneNumberChanges,
+              addressChanges
+          ) { personal, phone, address ->
+            OngoingPatientEntryChanged(OngoingNewPatientEntry(personal, address, phone.toNullable()))
+          }
+          .flatMapSingle { changedPatientEntry ->
+            patientRepository.ongoingEntry()
+                .map { alreadyPresentEntry ->
+                  changedPatientEntry.copy(entry = changedPatientEntry.entry.copy(identifier = alreadyPresentEntry.identifier))
```
sanchita: I was trying to simplify this code further.
Right now this block of code is a little confusing to read. 
I tried doing something like following:
```
 val identifiers = events
          .ofType<ScreenCreated>()
          .flatMapSingle { patientRepository.ongoingEntry() }
          .map { it.identifier.toOptional() }
```
and then combining with the rest of it:
```
...
 identifiers
          ) { personal, phone, address, identifier ->
            OngoingPatientEntryChanged(OngoingNewPatientEntry(personal, address, phone.toNullable(), identifier.toNullable()))
          }
```
But the other tests are failing and it would require changing the tests.
Can you take a look at this if possible?

sanchita: Let me know if this requires pairing.

app/src/main/res/values/strings.xml
vinaysshenoy: Good point. Not that I think about it, when we are registering a patient, there will never be more than one passport. This should be a singular instead. I'll confirm with Akshay.

vinaysshenoy: Confirmed with Akshay offline, this will be a singular string.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
```diff
@@ -311,4 +312,27 @@ class PatientEntryScreenController @Inject constructor(
...
+    val savedOngoingEntry = patientRepository
+        .ongoingEntry()
+        .toObservable()
+        .replay()
+        .refCount()
+
+    val showIdentifiersSection = screenCreates
+        .withLatestFrom(savedOngoingEntry) { _, entry -> entry }
+        .filter { it.identifier != null }
+        .map { { ui: Ui -> ui.showIdentifierSection(true) } }
```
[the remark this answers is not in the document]
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
```diff
@@ -176,6 +187,14 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
       }
       genderButton.isChecked = true
     }
+
+    entry.identifier?.let { identifier ->
+      displayIdentifier(identifier)
+    }
```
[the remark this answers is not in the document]
vinaysshenoy: Looks good. I'll do this.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
vinaysshenoy: The tests are failing because you are not emitting the `ScreenCreated` event in the test. This change looks good right now, but I realised when working with this file that we need to move `OngoingNewPatientEntry` out of `PatientRepository` and persist it to disk as a `SharedPreference`.

This is because this is a model that needs to survive process restart. So this entire method and the `OngoingPatientEntryChanged` event are going to be removed in the near future. I'd like to leave this method as is for now.

vinaysshenoy:
@sanchita Feedback addressed.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
sanchita: https://www.pivotaltracker.com/story/show/165442090

PR #607, opened 2019-04-17
Make patient config remotely configurable

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/165406792

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/PatientConfig.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientConfig.kt
@@ -1,20 +1,47 @@
 package org.simple.clinic.patient
 
-/**
- * See the [SQLite Limits](https://www.sqlite.org/limits.html) doc for more information.
- *
- * See the section "maximum number of host parameters in a single SQL statement".
- * */
-const val MAXIMUM_SQLITE_QUERY_LIMIT = 1000
+import androidx.annotation.VisibleForTesting
+import io.reactivex.Observable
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/patient/PatientConfig.kt
```diff
@@ -1,20 +1,47 @@
...
+            .long("patients_numberofsearchresults", 100L)
+            .coerceAtLeast(1L)
+            .coerceAtMost(MAXIMUM_SQLITE_QUERY_LIMIT)
+            .toInt()
+
+        val bpPassportFeatureEnabled = configReader
+            .boolean("patients_bppassportfeature_enabled", false)
+
+        val numberOfRecentPatients = configReader
+            .long("patients_recentpatientsinhomescreen", 10L)
```
sanchita: The name for config here does not tell me what this number is?
So I was thinking that now that we have started using Remote configs, we should define a naming convention.
How about something like `<configType>_<configName>_<configUnitDesc>`. Eg: `patients_searchresults_limit`. 
Wdyt?
So you can you make this as `patients_recentpatients_limit`?

vinaysshenoy: Makes sense, I'll do this.

vinaysshenoy: Done.

PR #610, opened 2019-04-17
Add: Add ID to patient search results screen

rakshakhegde:


app/src/main/java/org/simple/clinic/addidtopatient/searchresults/AddIdToPatientSearchResultsScreen.kt
```diff
@@ -0,0 +1,134 @@
...
+    }
+    toolbar.setOnClickListener {
+      screenRouter.pop()
+    }
+
+    val identifierType = identifierDisplayAdapter.typeAsText(screenKey.identifier)
+    val identifierValue = identifierDisplayAdapter.valueAsText(screenKey.identifier)
+
+    val identifierTextAppearanceSpan = TextAppearanceSpan(context, R.style.Clinic_V2_TextAppearance_Body0Left_NumericBold_White100)
+    toolbar.title = Truss()
```
vinaysshenoy: The default title provided by the `Toolbar` is single-line, so setting the title on this ends up clipping the text. Can you look at how `AddIdToPatientSearchScreen` is doing this and do the same here?


Screenshot_1555502577

app/src/main/java/org/simple/clinic/addidtopatient/searchresults/AddIdPatientSearchResultsScreenEvents.kt
```diff
@@ -0,0 +1,14 @@
+package org.simple.clinic.addidtopatient.searchresults
+
+import org.simple.clinic.patient.PatientSearchResult
+import org.simple.clinic.widgets.UiEvent
+
+data class AddIdToPatientSearchResultsScreenCreated(val key: AddIdToPatientSearchResultsScreenKey) : UiEvent
```
vinaysshenoy: Please do not pass keys to the controllers. Keys are an implementation detail on the UI layer, and controllers do not need to know about keys. If you need to pass along all the information, in a key to the controller, please declare them as properties of the `ScreenCreated` event.

PR #612, opened 2019-04-18
Bp passport app bar display

sanchita:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/165378803

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
@@ -318,7 +318,7 @@ class PatientRepository @Inject constructor(
 
     val businessIdSave = cachedOngoingEntry
         .flatMapCompletable { entry ->
-          if(entry.identifier == null) {
+          if (entry.identifier == null) {
             Completable.complete()
           } else {
             addIdentifierToPatient(patientUuid, entry.identifier).toCompletable()
@@ -525,7 +525,6 @@ class PatientRepository @Inject constructor(
     return database
         .patientDao()
         .findPatientsWithBusinessId(identifier)
-        .toObservable()
         .map { patients ->
           if (patients.isEmpty()) {
             None
@@ -533,15 +532,22 @@ class PatientRepository @Inject constructor(
             patients.first().toOptional()
           }
         }
+        .toObservable()
+
   }
 
-  fun bpPassportForPatient(patientUuid: UUID): Single<Optional<BusinessId>> {
-    return Single.fromCallable {
-      database
-          .businessIdDao()
-          .latestIdForType(patientUuid, BpPassport)
-          .toOptional()
-    }
+  fun bpPassportForPatient(patientUuid: UUID): Observable<Optional<BusinessId>> {
+    return database
+        .businessIdDao()
+        .latestForPatientByType(patientUuid, BpPassport)
+        .map { bpPassports ->
+          if (bpPassports.isEmpty()) {
+            None
+          } else {
+            bpPassports.first().toOptional()
+          }
+        }
+        .toObservable()
   }
 
   fun isPatientDefaulter(patientUuid: UUID): Observable<Boolean> {
--- a/app/src/main/java/org/simple/clinic/patient/businessid/BusinessId.kt
+++ b/app/src/main/java/org/simple/clinic/patient/businessid/BusinessId.kt
@@ -111,8 +111,7 @@ data class BusinessId(
       WHERE patientUuid = :patientUuid AND identifierType = :identifierType AND deletedAt IS NULL
       ORDER BY createdAt DESC LIMIT 1
       """)
-    fun latestIdForType(patientUuid: UUID, identifierType: IdentifierType): BusinessId?
-
+    fun latestForPatientByType(patientUuid: UUID, identifierType: IdentifierType): Flowable<List<BusinessId>>
   }
 }
 
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
@@ -3,6 +3,7 @@ package org.simple.clinic.summary
 import android.annotation.SuppressLint
 import android.content.Context
 import android.os.Parcelable
+import android.text.style.TextAppearanceSpan
 import android.util.AttributeSet
 import android.view.ViewGroup
 import android.widget.Button
@@ -34,6 +35,7 @@ import org.simple.clinic.patient.Gender
 import org.simple.clinic.patient.Patient
 import org.simple.clinic.patient.PatientAddress
 import org.simple.clinic.patient.PatientPhoneNumber
+import org.simple.clinic.patient.businessid.BusinessId
 import org.simple.clinic.patient.businessid.Identifier
 import org.simple.clinic.router.screen.ActivityResult
 import org.simple.clinic.router.screen.BackPressInterceptCallback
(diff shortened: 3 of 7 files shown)


app/src/main/java/org/simple/clinic/patient/businessid/BusinessId.kt
```diff
@@ -111,7 +111,7 @@ data class BusinessId(
       WHERE patientUuid = :patientUuid AND identifierType = :identifierType AND deletedAt IS NULL
       ORDER BY createdAt DESC LIMIT 1
       """)
-    fun latestIdForType(patientUuid: UUID, identifierType: IdentifierType): BusinessId?
+    fun latestIdForType(patientUuid: UUID, identifierType: IdentifierType): Flowable<List<BusinessId>>
```
vinaysshenoy: Rename this method to `latestForPatientOfType`?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -536,12 +536,18 @@ class PatientRepository @Inject constructor(
...
+        .latestIdForType(patientUuid, BpPassport)
+        .toObservable()
+        .map { bpPassports ->
+          if (bpPassports.isEmpty()) {
+            None
+          } else {
+            bpPassports.first().toOptional()
+          }
+        }
+        .firstOrError()
```
vinaysshenoy: This method can return an `Observable<Optional<BusinessId>>` and then consumers can decide whether they need a `Single` or not. I personally prefer this approach, because it also allows the same method to publish change notifications downstream. What do you think?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -536,12 +536,18 @@ class PatientRepository @Inject constructor(
...
-    return Single.fromCallable {
-      database
-          .businessIdDao()
-          .latestIdForType(patientUuid, BpPassport)
-          .toOptional()
-    }
+    return database
+        .businessIdDao()
+        .latestIdForType(patientUuid, BpPassport)
+        .toObservable()
```
vinaysshenoy: Can you move this `.toObservable()` to after the `.map { }` operator?

vinaysshenoy: In fact, if you still need to publish it as a `Single`, you don't need a `.toObservable()` here, because `Flowable` also has a `firstOrError()` operator that converts it to a `Single`.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -227,6 +239,31 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
     fullNameTextView.text = resources.getString(R.string.patientsummary_name, name, genderLetter, age)
   }
 
+  private fun displayBpPassport(bpPassport: Optional<BusinessId>, isPhoneNumberVisible: Boolean) {
+    bpPassportTextView.visibility = if (bpPassport.isNotEmpty()) View.VISIBLE else View.GONE
```
vinaysshenoy: We already have a utility method `View.visibleOrGone(Boolean)` that can be used to do this sort of thing, FYI.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -227,6 +239,31 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
...
+    bpPassportTextView.text = when {
+      bpPassport.isNotEmpty() -> {
+        val bullet = if (isPhoneNumberVisible) Unicode.bullet else ""
+        val identifier = bpPassport.toNullable()?.identifier!!
+        val textSpan = TextAppearanceSpan(context, R.style.Clinic_V2_TextAppearance_Body2Left_White72)
+        val numericSpan = TextAppearanceSpan(context, R.style.Clinic_V2_TextAppearance_Body2Left_Numeric_White72)
+        Truss()
+            .pushSpan(textSpan)
+            .append(bullet.toString())
+            .append(" ")
```
vinaysshenoy: This is a problem because if the phone number is not present, the BP Passport will be displayed with an extra space and will not be aligned with the rest of the header, right?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -117,8 +117,14 @@ class PatientSummaryScreenController @Inject constructor(
     val phoneNumbers = patientUuid
         .flatMap { patientRepository.phoneNumber(it) }
 
-    return Observables.combineLatest(sharedPatients, addresses, phoneNumbers)
-        .map { (patient, address, phoneNumber) -> { ui: Ui -> ui.populatePatientProfile(patient, address, phoneNumber) } }
+    val bpPassport = patientUuid
+        .flatMapSingle { patientRepository.bpPassportForPatient(it) }
```
vinaysshenoy: If a sync happens in the background while this patient summary is open, and this sync saves a Bp Passport for the patient, this won't get updated.

The repository should return an `Observable` instead so that change notifications can be refreshed on the UI.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -227,6 +239,31 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
...
   }
 
+  private fun displayBpPassport(bpPassport: Optional<BusinessId>, isPhoneNumberVisible: Boolean) {
+    bpPassportTextView.visibility = if (bpPassport.isNotEmpty()) View.VISIBLE else View.GONE
+
+    bpPassportTextView.text = when {
+      bpPassport.isNotEmpty() -> {
+        val bullet = if (isPhoneNumberVisible) Unicode.bullet else ""
+        val identifier = bpPassport.toNullable()?.identifier!!
+        val textSpan = TextAppearanceSpan(context, R.style.Clinic_V2_TextAppearance_Body2Left_White72)
```
vinaysshenoy: We do not need to explicitly set this text appearance here since we have already set it in the layout.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: I took a stab at cleaning this method up. I added a string resource:
```xml
<string name="patientsummary_identifier_type">%1$s :\u0020</string>
```

And I then changed the method to
```kotlin
private fun displayBpPassport(bpPassport: BusinessId?, isPhoneNumberVisible: Boolean) {
    bpPassportTextView.visibility = bpPassport?.let { VISIBLE } ?: GONE

    bpPassportTextView.text = when(bpPassport) {
      null -> ""
      else -> {
        val identifier = bpPassport.identifier
        val numericSpan = TextAppearanceSpan(context, R.style.Clinic_V2_TextAppearance_Body2Left_Numeric_White72)
        val identifierType = resources.getString(R.string.patientsummary_identifier_type, identifierDisplayAdapter.typeAsText(identifier))

        val formattedIdentifier = Truss()
            .append(identifierType)
            .pushSpan(numericSpan)
            .append(identifierDisplayAdapter.valueAsText(identifier))
            .popSpan()
            .build()

        if(isPhoneNumberVisible) " ${Unicode.bullet} $formattedIdentifier" else formattedIdentifier
      }
    }
  }
```

What do you think? This will also fix the problem of the label not matching the alignment with the rest of the header text when there is no phone number.

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
```diff
@@ -130,23 +133,35 @@ class PatientSummaryScreenControllerTest {
...
   @Test
-  @Parameters(method = "patient summary open intentions")
-  fun `patient's profile should be populated`(intention: OpenIntention) {
+  @Parameters(method = "params for patient summary populating profile")
+  fun `patient's profile should be populated`(intention: OpenIntention, bpPassport: BusinessId?) {
     val addressUuid = UUID.randomUUID()
     val patient = PatientMocker.patient(uuid = patientUuid, addressUuid = addressUuid)
     val address = PatientMocker.address(uuid = addressUuid)
     val phoneNumber = None
+    val optionaBpPassport = bpPassport.toOptional()
```
vinaysshenoy: Typo? It should be `optionalBpPassport`.

app/src/main/java/org/simple/clinic/patient/businessid/BusinessId.kt
sanchita: This name seems like there is a type of patient.
Let me think of something else.

vinaysshenoy: Maybe just `latestForPatient`?

vinaysshenoy:
@sanchita LGTM. I've made a small change to make the code block that renders the identifier a little cleaner.

PR #613, opened 2019-04-19
Recent patient complete log

rakshakhegde:
Added UI to be able to show "See All" button but I'm not showing it right away. Will include the button once the entire flow works.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -123,6 +123,7 @@
   <string name="patients_scan_simple_card">Scan Simple Card</string>
   <string name="patients_recentpatients_title">Recent Patients</string>
   <string name="patients_recentpatients_no_recent_patients">No recent patients</string>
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -123,6 +123,7 @@
   <string name="patients_scan_simple_card">Scan Simple Card</string>
   <string name="patients_recentpatients_title">Recent Patients</string>
   <string name="patients_recentpatients_no_recent_patients">No recent patients</string>
+  <string name="patients_recentpatients_see_all">SEE ALL</string>
```
vinaysshenoy: Can we not use full capitalization in the string resources? Since these are buttons, we should be defining these buttons in title case and then use `textAppCaps=true` on the button/textview to capitalize them.

vinaysshenoy: I'll make this change and push.

PR #617, opened 2019-04-22
Disable Kotlin view extensions

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/165490945

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -160,6 +160,7 @@ android {
 
   androidExtensions {
     experimental = true
+    features = ["parcelize"]
   }
 
   sourceSets {
--- a/app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientItem.kt
+++ b/app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientItem.kt
@@ -5,7 +5,8 @@ import android.content.Context
 import android.view.View
 import com.xwray.groupie.ViewHolder
 import io.reactivex.subjects.Subject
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
```diff
@@ -23,37 +26,56 @@ data class RecentPatientItem(
...
+    }
+
+    private fun lastBpText(context: Context, lastBp: LastBp?): String {
+      return lastBp?.run { "$systolic/$diastolic, ${updatedAtRelativeTimestamp.displayText(context)}" } ?: ""
+    }
+  }
 }
 
-object SeeAllItem : RecentPatientItemType(0) {
+object SeeAllItem : RecentPatientItemType<ViewHolder>(0) {
```
sanchita: Can you fix the `0` being sent as the `adapterId`?
Can you create a constant instead and pass that?

sanchita:
@vinaysshenoy Please resolve conflicts on this branch.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientItem.kt
vinaysshenoy: I want to consider how we generate adapter IDs across the board since we don't do it very well at the moment. I'll do this when we get around to that?

sanchita: Sure.

PR #618, opened 2019-04-22
BP Passport Api integration

sanchita:
Pivotol story -> https://www.pivotaltracker.com/story/show/165464580


app/src/main/java/org/simple/clinic/patient/sync/PatientSyncApiV3.kt
```diff
@@ -0,0 +1,27 @@
...
+  fun push(
+      @Body body: PatientPushRequest
+  ): Single<DataPushResponse>
+
+  @GET("$version/patients/sync")
+  fun pull(
+      @Query("limit") recordsToPull: Int,
+      @Query("process_token") lastPullTimestamp: String? = null
+  ): Single<PatientPullResponse>
+
```
vinaysshenoy: Extra line?

app/src/main/java/org/simple/clinic/patient/sync/PatientSync.kt
```diff
@@ -50,11 +50,14 @@ class PatientSync @Inject constructor(
...
   private fun toRequest(patients: List<PatientProfile>): PatientPushRequest {
     return PatientPushRequest(
-        patients.map { (patient, address, phoneNumbers) ->
+        patients.map { (patient, address, phoneNumbers, businessIds) ->
           val numberPayloads = phoneNumbers
               .map(::phoneNumberPayload)
               .let { payloads -> if (payloads.isEmpty()) null else payloads }
 
+          val businessIdPayloads = businessIds
+              .map { it.toPayload() }
```
vinaysshenoy: FYI, this lambda can also be represented as a method reference like so:

```kotlin
val businessIdPayloads = businessIds.map(BusinessId::toPayload)
```

Not necessary to change this, but I feel this looks a little cleaner.

app/src/androidTest/java/org/simple/clinic/TestData.kt
```diff
@@ -195,7 +195,7 @@ class TestData @Inject constructor(
       deletedAt: Instant? = null,
       address: PatientAddressPayload = addressPayload(),
       phoneNumbers: List<PatientPhoneNumberPayload>? = listOf(phoneNumberPayload()),
-      businessIds: List<BusinessIdPayload>? = listOf(businessIdPayload(patientUuid = uuid))
+      businessIds: List<BusinessIdPayload> = listOf(businessIdPayload(patientUuid = uuid))
```
vinaysshenoy: I've realised that when we are generating test business IDs, we don't generate metadata for it. In the test, we set the meta version to `org.simple.clinic.bppassport.meta.v1`, but set the meta to an empty string. Since this is meant to be an integration test, do you feel that we should generate metadata appropriately?

app/src/androidTest/java/org/simple/clinic/TestData.kt
```diff
@@ -195,7 +195,7 @@ class TestData @Inject constructor(
       deletedAt: Instant? = null,
```
vinaysshenoy: `PatientRepositoryAndroidTest` has a test `verify_saving_busines_ids_works_as_expected` which can be deleted once we enable this API integration.

app/src/androidTest/java/org/simple/clinic/TestData.kt
sanchita: I have seen that we are not using `metadata` anywhere.
It is a `String` and we are assigning it `""` which seems fine for now. 

If we ever we start using the `metadata`, we will change this then.

vinaysshenoy: Integration tests are not just for us, they're also for the server so that they get feedback if any contracts are broken. Our contract is basically that if we set a meta version, there also needs to be a corresponding meta data, which is why I believe we should send it. Does that make sense?

sanchita: Oh ok. I did not know that.
I guess we should do it then.

vinaysshenoy:
@sanchita Please review the final commit?

PR #624, opened 2019-04-24
Recent patients screen

rakshakhegde:
Pending tasks:
- ~Add tests for `RecentPatientsScreenController`~
- Segregate patients by day

### DIFF
--- a/app/src/main/java/org/simple/clinic/recentpatient/RecentPatientsScreen.kt
+++ b/app/src/main/java/org/simple/clinic/recentpatient/RecentPatientsScreen.kt
@@ -1,22 +1,20 @@
-package org.simple.clinic.patient.recent
+package org.simple.clinic.recentpatient
 
-import android.annotation.SuppressLint
 import android.content.Context
 import android.util.AttributeSet
-import android.view.View
-import android.widget.FrameLayout
+import android.widget.LinearLayout
+import androidx.appcompat.widget.Toolbar
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import com.jakewharton.rxbinding2.view.RxView
 import com.xwray.groupie.GroupAdapter
 import com.xwray.groupie.ViewHolder
 import io.reactivex.Observable
-import io.reactivex.android.schedulers.AndroidSchedulers.mainThread
+import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.subjects.PublishSubject
 import kotterknife.bindView
 import org.simple.clinic.R
 import org.simple.clinic.activity.TheActivity
-import org.simple.clinic.recentpatient.RecentPatientsScreenKey
 import org.simple.clinic.router.screen.ScreenRouter
 import org.simple.clinic.summary.OpenIntention
 import org.simple.clinic.summary.PatientSummaryScreenKey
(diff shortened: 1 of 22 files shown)


app/src/main/res/layout/see_all_item_view.xml
```diff
@@ -1,10 +1,18 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Button style="@style/Widget.AppCompat.Button.Borderless.Colored"
-  xmlns:android="http://schemas.android.com/apk/res/android"
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
vinaysshenoy: There aren't any attributes set in this `FrameLayout`. Is there a reason you have wrapped the button in the frame?

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
```diff
@@ -71,21 +76,33 @@ class RecentPatientsView(context: Context, attrs: AttributeSet) : FrameLayout(co
         .ofType<RecentPatientItemClicked>()
```
vinaysshenoy: Can we not handle clicks like this in the screen separately? This means that the events never get reported to Heap since this is done in the controller. The `adapterEvents` should be fed into the stream that goes to the controller and these events should be handled in the controller.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
```diff
@@ -71,21 +76,33 @@ class RecentPatientsView(context: Context, attrs: AttributeSet) : FrameLayout(co
...
+        .takeUntil(RxView.detaches(this))
+        .subscribe { openRecentPatientsScreen() }
   }
 
   private fun screenCreates() = Observable.just(ScreenCreated())
 
   fun updateRecentPatients(recentPatients: List<RecentPatientItem>) {
-    recentPatients.forEach { it.uiEvents = adapterEvents }
-    groupAdapter.update(recentPatients)
+    val allItemTypes = recentPatients + SeeAllItem
```
vinaysshenoy: Instead of doing this here, the full set must be constructed in the `RecentPatientsViewController` so that it can be tested.

In addition, we might want to disable the `See All` button for now, and then push the code to `master` until the feature is complete. Adding this item type in the screen does not allow us to do that in a nice way.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
```diff
@@ -17,6 +17,8 @@ import io.reactivex.subjects.PublishSubject
 import kotterknife.bindView
 import org.simple.clinic.R
 import org.simple.clinic.activity.TheActivity
+import org.simple.clinic.recentpatient.RecentPatientsScreenKey
```
vinaysshenoy: The `patient` package is meant for placing classes related to the business logic/data only. Can you move the View related classes into a separate package for now?

For what it's worth, I feel that the way we organise our packages could be more obvious. Let's discuss this at some point.

app/src/main/java/org/simple/clinic/recentpatient/RecentPatientsScreen.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    Observable.mergeArray(screenCreates(), screenDestroys)
+        .compose(controller)
+        .takeUntil(screenDestroys)
+        .observeOn(AndroidSchedulers.mainThread())
+        .subscribe { uiChange -> uiChange(this) }
+
+    adapterEvents
+        .ofType<RecentPatientItemClicked>()
+        .takeUntil(RxView.detaches(this))
```
vinaysshenoy: Same as an earlier comment about feeding this stream into the controller instead of handling it here.

app/src/main/java/org/simple/clinic/recentpatient/RecentPatientsScreen.kt
```diff
@@ -0,0 +1,98 @@
...
+import org.simple.clinic.summary.PatientSummaryScreenKey
+import org.simple.clinic.util.UtcClock
+import org.simple.clinic.widgets.ScreenCreated
+import org.simple.clinic.widgets.ScreenDestroyed
+import org.simple.clinic.widgets.UiEvent
+import org.threeten.bp.Instant
+import java.util.UUID
+import javax.inject.Inject
+
+class RecentPatientsScreen(context: Context, attrs: AttributeSet) : LinearLayout(context, attrs) {
```
vinaysshenoy: If we are going to make this screen accessible once the PR is merged, we also need to provide tests for the controller. Can you either:
- Add tests for the controller, or
- Don't show the `See All` button yet and add it along with the PR that adds the tests?

app/src/main/res/layout/see_all_item_view.xml
rakshakhegde: the button takes up limited width and is centered inside the FrameLayout

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
rakshakhegde: done

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
rakshakhegde: Right, I will include this once the feature is complete.

app/src/main/res/layout/see_all_item_view.xml
vinaysshenoy: Cool. I was wondering if that can also be achieved by setting the width of the button to `wrap_content` without wrapping it in a container.

app/src/main/java/org/simple/clinic/patient/recent/RecentPatientsView.kt
rakshakhegde: done

app/src/main/java/org/simple/clinic/recentpatient/RecentPatientsScreen.kt
rakshakhegde: done

app/src/main/java/org/simple/clinic/recentpatient/RecentPatientsScreen.kt
rakshakhegde: I'll be adding code to show "See all" button in a following PR.

app/src/main/res/layout/see_all_item_view.xml
vinaysshenoy: Nvm, my brain gave out for a moment.

PR #628, opened 2019-04-29
Fix: BP Passports do not sync when added to an existing patient

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/165669043

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
@@ -369,16 +369,12 @@ class PatientRepository @Inject constructor(
           val updatedPatientAddress = patientAddress.copy(updatedAt = Instant.now(utcClock))
           database.addressDao().save(updatedPatientAddress)
         }
-        .andThen(Completable.fromAction {
-          database.patientDao().updateSyncStatus(listOf(patientUuid), PENDING)
-        })
+        .andThen(setSyncStatus(listOf(patientUuid), PENDING))
   }
 
   fun updatePhoneNumberForPatient(patientUuid: UUID, phoneNumber: PatientPhoneNumber): Completable {
     return savePhoneNumber(phoneNumber.copy(updatedAt = Instant.now(utcClock)))
-        .andThen(Completable.fromAction {
-          database.patientDao().updateSyncStatus(listOf(patientUuid), PENDING)
-        })
+        .andThen(setSyncStatus(listOf(patientUuid), PENDING))
   }
 
   fun createPhoneNumberForPatient(patientUuid: UUID, number: String, phoneNumberType: PatientPhoneNumberType, active: Boolean): Completable {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -493,9 +487,26 @@ class PatientRepository @Inject constructor(
...
+  }
+
+  private fun touchPatientRecord(patientUuid: UUID): Completable {
+    return Completable
+        .fromAction {
+          database
+              .patientDao()
+              .touch(
+                  patientUuid = patientUuid,
+                  newUpdatedAt = Instant.now(utcClock),
```
sanchita: We do not need to change the `updatedAt` of `Patient` model.
We can just update the `syncStatus` using `database.patientDao().updateSyncStatus`.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -493,9 +487,26 @@ class PatientRepository @Inject constructor(
...
 
     return businessIdStream
-        .flatMap { businessId ->
-          Completable.fromAction { database.businessIdDao().save(listOf(businessId)) }
-              .toSingleDefault(businessId)
+        .flatMap { businessId -> saveBusinessId(businessId).toSingleDefault(businessId) }
+        .flatMap { businessId -> touchPatientRecord(patientUuid).toSingleDefault(businessId) }
+  }
+
+  private fun touchPatientRecord(patientUuid: UUID): Completable {
```
sanchita: Can you remove this method entirely and call `setSyncStatus` from `addIdentifierToPatient`. That will internally calls `database.patientDao().updateSyncStatus`?

app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -217,6 +217,13 @@ data class Patient(
         scheduled: Appointment.Status = Appointment.Status.SCHEDULED
     ): Flowable<Boolean>
 
+    @Query("""
+      UPDATE Patient
+      SET updatedAt = :newUpdatedAt, syncStatus = :newSyncStatus
+      WHERE uuid = :patientUuid
+    """)
+    abstract fun touch(patientUuid: UUID, newUpdatedAt: Instant, newSyncStatus: SyncStatus)
```
sanchita: We do not need to query since we only need to update the `syncStatus`.
There is already a method for that called `updateSyncStatus`.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: Right you are. I've fixed it!

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/patient/Patient.kt
vinaysshenoy: Done.

PR #630, opened 2019-04-29
Recent patients segregate by day

rakshakhegde:
Pending task:
- Format date

### DIFF
--- a/app/src/main/java/org/simple/clinic/recentpatient/RecentPatientScreenItemTypes.kt
+++ b/app/src/main/java/org/simple/clinic/recentpatient/RecentPatientScreenItemTypes.kt
@@ -30,7 +30,7 @@ data class RecentPatientItem(
     val age: Int,
     val lastBp: LastBp?,
     val gender: Gender,
-    val latestUpdatedAt: Instant
+    val updatedAt: Instant
 ) : RecentPatientScreenItemTypes<RecentPatientItemViewHolder>(uuid.hashCode().toLong()) {
 
   override fun getLayout(): Int = R.layout.recent_patient_item_view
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/simple/clinic/recentpatient/RecentPatientScreenItemTypes.kt
```diff
@@ -26,7 +29,8 @@ data class RecentPatientItem(
     val name: String,
     val age: Int,
     val lastBp: LastBp?,
-    val gender: Gender
+    val gender: Gender,
+    val latestUpdatedAt: Instant
```
vinaysshenoy: The fact that this `Instant` is the latest updated at timestamp selected from multiple timestamps is an implementation detail that is known to the `RecentPatient` model, but need not be known here. Rename this property to just `updatedAt`?

rakshakhegde: done

app/src/main/java/org/simple/clinic/util/TimeFunctions.kt
```diff
@@ -17,3 +18,8 @@ fun estimateCurrentAge(recordedAge: Int, ageRecordedAtTimestamp: Instant, clock:
 fun estimateCurrentAge(recordedDateOfBirth: LocalDate, clock: Clock): Int {
   return Period.between(recordedDateOfBirth, LocalDate.now(clock)).years
 }
+
+fun Instant.toLocalDateAtZone(zone: ZoneId): LocalDate =
```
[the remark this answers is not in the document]
rakshakhegde: done

PR #631, opened 2019-04-30
Refactor screen setup boilerplate to a separate function

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/165674246


vinaysshenoy:
@rakshakhegde This seems like a large PR, but it's mostly the same thing being done in all our screens where we have the binding between views and its controller.

PR #633, opened 2019-05-02
Set special occasion illustrations based on date

rakshakhegde:


app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
```diff
@@ -102,6 +105,19 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
         .observeOn(mainThread())
         .takeUntil(screenDestroys)
         .subscribe { uiChange -> uiChange(this) }
+
+    illustrationImageView.setImageResource(illustrationResourceId())
+  }
+
+  private fun illustrationResourceId(): Int {
+    val today = LocalDate.now()
+    val may_11_2019 = LocalDate.of(2019, Month.MAY, 11)
```
vinaysshenoy: World hypertension day repeats on the same day every year, so we should not be checking for the year as well.

Otherwise, this piece of code will need to be continuously updated every year.

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
```diff
@@ -102,6 +105,19 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
         .observeOn(mainThread())
         .takeUntil(screenDestroys)
         .subscribe { uiChange -> uiChange(this) }
+
+    illustrationImageView.setImageResource(illustrationResourceId())
+  }
+
+  private fun illustrationResourceId(): Int {
```
vinaysshenoy: Since we have only one resource to show, this looks fine for now. If we start adding more checks to show different images based on the date, we'll need to make this more generic.

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
```diff
@@ -102,6 +105,19 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
...
+    val today = LocalDate.now()
```
vinaysshenoy: Can you provide the `userClock` here instead of using the `.now()` method? This will compute the dates in the utc clock instead, which is not what we want.

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
vinaysshenoy: One more piece of feedback I have is that the variable name `may_11_2019` is pointless since I have no idea why this particular day is important. I can suggest changing the code that make it more obvious like so:

```kotlin
val worldHypertensionDay = today.withMonth(Month.MAY.value).withDayOfMonth(17)
val dateToShowBannerFrom = worldHypertensionDay.minusDays(6L)

return when(today) {
  in dateToShowBannerFrom..worldHypertensionDay -> R.drawable.ic_homescreen_world_hypertension_day
  else -> R.drawable.illustrations_homescreen
}
```

vinaysshenoy: This also handles the case where we need to show the banner the next year as well.

rakshakhegde: done

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
rakshakhegde: done

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
rakshakhegde: done

PR #636, opened 2019-05-06
Change Communication enums to use safe enum parsing method

vinaysshenoy:
This also adds support for the new enum value, `missed_visit_sms_reminder`, of the `communication_type` property.


vinaysshenoy:
Closing this because we decided to remove the `Communication` model since it's a dead resource.

PR #638, opened 2019-05-06
Add support for screen change animations 

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/165674246


vinaysshenoy:
@rakshakhegde @sanchita I have tentatively implemented a way to delay the UI change events by the screen change animation duration here.

It's a little cumbersome since it requires us to manually inject the `Duration` into every Screen and then forward that to the `bindUiToController` method. I've been thinking of ways to avoid this, but the only solution I can think of is to declare the animation duration as a global constant and then refer that. What do you feel?

vinaysshenoy:
@rakshakhegde @sanchita I am planning to go ahead an implement the version of screen change animations I have tried out in my latest commit. Please review.

app/src/main/java/org/simple/clinic/ViewControllerBinding.kt
```diff
@@ -7,19 +7,32 @@ import io.reactivex.android.schedulers.AndroidSchedulers.mainThread
...
+      .compose(delayEmissionsBy(uiChangeDelay))
       .observeOn(mainThread())
       .takeUntil(screenDestroys)
       .subscribe { uiChange -> uiChange(ui) }
 }
+
+private fun <T> delayEmissionsBy(duration: Duration): ObservableTransformer<(T) -> Unit, (T) -> Unit> {
+  return ObservableTransformer { upstream ->
+    when (duration) {
+      Duration.ZERO -> upstream
```
sanchita: This statement basically means that we do not want to delay emissions at all.
I was thinking instead of using one method to bind Ui, we could separate this out into two.

Something like this ->

```
fun <T> bindUiToController(
    ui: T,
    events: Observable<UiEvent>,
    controller: ObservableTransformer<UiEvent, (T) -> Unit>,
    screenDestroys: Observable<ScreenDestroyed>
) {
  events
      .mergeWith(screenDestroys)
      .observeOn(io())
      .compose(controller)
      .observeOn(mainThread())
      .takeUntil(screenDestroys)
      .subscribe { uiChange -> uiChange(ui) }
}

fun <T> bindUiToControllerWithDelay(
    ui: T,
    events: Observable<UiEvent>,
    controller: ObservableTransformer<UiEvent, (T) -> Unit>,
    screenDestroys: Observable<ScreenDestroyed>,
    uiChangeDelay: Duration = SCREEN_CHANGE_ANIMATION_DURATION
) {
  events
      .mergeWith(screenDestroys)
      .observeOn(io())
      .compose(controller)
      .delaySubscription(uiChangeDelay.toMillis(), TimeUnit.MILLISECONDS)
      .observeOn(mainThread())
      .takeUntil(screenDestroys)
      .subscribe { uiChange -> uiChange(ui) }
}
```

sanchita: Using the above way we won't have to pass the delay value as `ZERO` explicitly in cases where we do not want to add delay.
Wdyt?

vinaysshenoy: Good idea, I'll do this except with one change.

For most of our views, we'll want to add the delay so I'll leave the method named as `bindUiToController` and the other method I'll name `bindUiToControllerWithoutDelay`.

vinaysshenoy: Done.

sanchita:
@vinaysshenoy `PatientEditScreenUiTest` test is failing.
Can you please fix it?

vinaysshenoy:
> @vinaysshenoy PatientEditScreenUiTest test is failing.
> Can you please fix it?

What's happening is that because the delay to screen change actions was added, the text field does not get pre-filled by the time we run the assertion. The only way I can think to fix it at this point is to add a `Thread.sleep(SCREEN_CHANGE_ANIMATION_DURATION)` in the test before running the assertions, which is ugly.

The only thing this test does currently is:
- Has a base template for writing a UI test with espresso, and
- Verifies whether the `PatientEditScreen` is setting the date of birth in a specific format which is injected via DI

I feel that this test is not very useful anymore since we'll be having a different UI test suite written by the QA team and this is easily verifiable by manual QA. I think we can remove this test from the codebase. What do you think? @sanchita

vinaysshenoy:
> > @vinaysshenoy PatientEditScreenUiTest test is failing.
> > Can you please fix it?
> 
> What's happening is that because the delay to screen change actions was added, the text field does not get pre-filled by the time we run the assertion. The only way I can think to fix it at this point is to add a `Thread.sleep(SCREEN_CHANGE_ANIMATION_DURATION)` in the test before running the assertions, which is ugly.
> 
> The only thing this test does currently is:
> 
> * Has a base template for writing a UI test with espresso, and
> * Verifies whether the `PatientEditScreen` is setting the date of birth in a specific format which is injected via DI
> 
> I feel that this test is not very useful anymore since we'll be having a different UI test suite written by the QA team and this is easily verifiable by manual QA. I think we can remove this test from the codebase. What do you think? @sanchita

Discussed offline: We are going to remove this UI test for now since:
- It doesn't add much value at the moment since the only thing it verifies is whether the patient edit screen is using the injected date formatter.
- When the QA team starts writing a UI test suite for the app, we'll end up having to make some changes to the app to support the UI testing needs anyway.

PR #643, opened 2019-05-08
Add method to mark the current logged in user as unauthorized

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/165876930


sanchita:
We should update the `user-states` doc to include this new state.

vinaysshenoy:
> We should update the `user-states` doc to include this new state.

Good catch, I'll do this.

vinaysshenoy:
@sanchita Done.

PR #654, opened 2019-05-16
Report pending records to analytics when logout happens

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/165961128

### DIFF
--- a/app/src/test/java/org/simple/clinic/user/UserSessionTest.kt
+++ b/app/src/test/java/org/simple/clinic/user/UserSessionTest.kt
@@ -134,7 +134,8 @@ class UserSessionTest {
       appointmentSyncPullToken = appointmentPullToken,
       communicationSyncPullToken = communicationPullToken,
       medicalHistorySyncPullToken = medicalHistoryPullToken,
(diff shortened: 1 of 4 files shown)


app/src/test/java/org/simple/clinic/user/UserSessionTest.kt
```diff
@@ -868,4 +882,24 @@ class UserSessionTest {
   private fun `params for logout clear database failures`(): List<Any> {
     return listOf(NullPointerException(), RuntimeException())
   }
+
+  @Test
+  @Parameters(method = "params for logout report pending sync records failures")
+  fun `when reporting pending records fails, the logout must fail`(cause: Throwable) {
```
sanchita: I think that reporting analytics should not affect logout per se.
Reporting should be a side effect and not part of the flow.
We should make this a fire and forget call.

vinaysshenoy: Good point. I'll make this change.

vinaysshenoy: Done.

PR #657, opened 2019-05-17
Provide logged in user to PatientRepository methods

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/166083420

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
@@ -7,12 +7,14 @@ import io.reactivex.Single
 import io.reactivex.rxkotlin.Observables
 import org.simple.clinic.AppDatabase
 import org.simple.clinic.analytics.OperationTimingTracker
+import org.simple.clinic.bp.PatientToFacilityId
 import org.simple.clinic.di.AppScope
-import org.simple.clinic.facility.FacilityRepository
+import org.simple.clinic.facility.Facility
 import org.simple.clinic.patient.SyncStatus.DONE
 import org.simple.clinic.patient.SyncStatus.PENDING
 import org.simple.clinic.patient.businessid.BusinessId
-import org.simple.clinic.patient.businessid.BusinessIdMetaData
+import org.simple.clinic.patient.businessid.BusinessId.MetaDataVersion
+import org.simple.clinic.patient.businessid.BusinessIdMetaData.BpPassportMetaDataV1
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -92,19 +91,17 @@ class PatientRepository @Inject constructor(
           }
         }
         .doOnNext { timingTracker.start(sortByVisitedFacilityAnalytics) }
-        .compose(sortByCurrentFacility())
+        .compose(sortByCurrentFacility(loggedInUser))
```
sanchita: I can see that we need `loggedInUser` here to only get the current facility eventually.
Why can't we pass the current facility instead then? 
That would also add more sense to the method called `sortByCurrentFacility` which would obviously need the current facility.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -471,8 +468,12 @@ class PatientRepository @Inject constructor(
         .toObservable()
   }
 
-  fun addIdentifierToPatient(patientUuid: UUID, identifier: Identifier): Single<BusinessId> {
-    val businessIdStream = createBusinessIdMetaDataForIdentifier(identifier.type)
+  fun addIdentifierToPatient(
+      patientUuid: UUID,
+      identifier: Identifier,
+      loggedInUser: User
```
sanchita: Can we rename `loggedInUser` to be more meaningful in the context of creating business ids?
It could be renamed to `assigningUser`.

app/src/test/java/org/simple/clinic/searchresultsview/PatientSearchViewControllerTest.kt
```diff
@@ -104,115 +104,115 @@ class PatientSearchViewControllerTest {
...
-  fun `when register new patient clicked then RegisterNewPatient event should be emitted`() {
-    whenever(patientRepository.search(any())).thenReturn(Observable.empty())
-
-    uiEvents.onNext(SearchResultsViewCreated)
-    uiEvents.onNext(SearchResultPatientName(patientName = "name"))
-    uiEvents.onNext(RegisterNewPatientClicked)
-
-    verify(screen).registerNewPatient(RegisterNewPatient("name"))
-  }
+    @Test
```
sanchita: Please fix formatting in this class.

PR #658, opened 2019-05-20
Add class to report pending records to Analytics

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/165961128

### DIFF
--- a/app/src/main/java/org/simple/clinic/user/ReportPendingRecordsToAnalytics.kt
+++ b/app/src/main/java/org/simple/clinic/user/ReportPendingRecordsToAnalytics.kt
@@ -1,0 +1,78 @@
+package org.simple.clinic.user
+
+import io.reactivex.Completable
+import io.reactivex.rxkotlin.Singles
+import org.simple.clinic.analytics.Analytics
+import org.simple.clinic.bp.BloodPressureMeasurement
+import org.simple.clinic.bp.BloodPressureRepository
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/user/ReportPendingRecordsToAnalytics.kt
```diff
@@ -0,0 +1,78 @@
...
+    private val patientRepository: PatientRepository,
+    private val bloodPressureRepository: BloodPressureRepository,
+    private val appointmentRepository: AppointmentRepository,
+    private val prescriptionRepository: PrescriptionRepository,
+    private val medicalHistoryRepository: MedicalHistoryRepository
+) {
+  fun report(): Completable {
+    return Singles
+        .zip(
+            patientRepository.recordsWithSyncStatus(PENDING),
```
rakshakhegde: As discussed personally with @vinaysshenoy, we can use `SynceableRepository#pendingSyncRecordsCount`.

vinaysshenoy: I just realised that we can't do this. As part of pushing the records, we also push the oldest `updatedAt` timestamp to Analytics which will be used to figure out since when records have been lost. We'll need to load all the records anyway to compare the timestamps.

PR #660, opened 2019-05-21
Update patient recordedAt based on oldest BP

sanchita:
Pivotol story -> https://www.pivotaltracker.com/story/show/166119194

One part of this story is remaining where we need to consider deleted BPs and update `recordedAt` accordingly.

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -1735,4 +1735,46 @@ class PatientRepositoryAndroidTest {
     assertThat(latestBpPassport).isEqualTo(currentBpPassport)
   }
 
+  @Test
+  fun when_updating_patient_recordedAt_then_it_should_compare_and_update_the_date() {
+
+    fun createPatientProfile(patientUuid: UUID, recordedAt: Instant): PatientProfile {
+      return testData.patientProfile(patientUuid = patientUuid)
+          .run {
+            copy(patient = patient.copy(
+                recordedAt = recordedAt,
+                updatedAt = Instant.now(testClock),
+                syncStatus = DONE)
+            )
+          }
+    }
+
+    val patientUuid1 = UUID.randomUUID()
+    val recordedAtForPatient1 = Instant.now(testClock)
+    val patientProfile1 = createPatientProfile(patientUuid1, recordedAtForPatient1)
+    testClock.advanceBy(Duration.ofMinutes(1))
+    val instantToCompare1 = Instant.now(testClock)
+
+    val patientUuid2 = UUID.randomUUID()
+    val instantToCompare2 = Instant.now(testClock)
+    testClock.advanceBy(Duration.ofMinutes(1))
+    val recordedAtForPatient2 = Instant.now(testClock)
+    val patientProfile2 = createPatientProfile(patientUuid2, recordedAtForPatient2)
+
+    patientRepository.save(listOf(patientProfile1, patientProfile2)).blockingAwait()
+
+    patientRepository.compareAndUpdateRecordedAt(patientUuid1, instantToCompare1).blockingAwait()
+    patientRepository.compareAndUpdateRecordedAt(patientUuid2, instantToCompare2).blockingAwait()
+
+    val patient1 = patientRepository.patient(patientUuid1).blockingFirst().toNullable()!!
+    val patient2 = patientRepository.patient(patientUuid2).blockingFirst().toNullable()!!
+
+    assertThat(patient1.recordedAt).isEqualTo(recordedAtForPatient1)
+    assertThat(patient1.syncStatus).isEqualTo(DONE)
+
+    assertThat(patient2.recordedAt).isEqualTo(instantToCompare2)
+    assertThat(patient2.syncStatus).isEqualTo(PENDING)
+    assertThat(patient2.updatedAt).isEqualTo(Instant.now(testClock))
+  }
 }
+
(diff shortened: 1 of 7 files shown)


app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -1735,4 +1735,46 @@ class PatientRepositoryAndroidTest {
...
+
+    val patientUuid2 = UUID.randomUUID()
+    val bpRecordedAtForPatient2 = Instant.now(testClock)
+    testClock.advanceBy(Duration.ofMinutes(1))
+    val recordedAtForPatient2 = Instant.now(testClock)
+    val patientProfile2 = createPatientProfile(patientUuid2, recordedAtForPatient2)
+
+    patientRepository.save(listOf(patientProfile1, patientProfile2)).blockingAwait()
+
+    patientRepository.updatePatientRecordedAt(patientUuid1, bpRecordedAtForPatient1).blockingGet()
```
vinaysshenoy: Using `blockingGet()` makes me think that this method is returning a `Single<>` when it's actually a `Completable` which will return a `Throwable|null`. Should we change this to `blockingAwait()` instead?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -493,6 +495,7 @@ class BloodPressureEntrySheetController @Inject constructor(
               .flatMap { bpSaved ->
                 appointmentRepository
                     .markAppointmentsCreatedBeforeTodayAsVisited(newBp.patientUuid)
+                    .andThen(patientRepository.updatePatientRecordedAt(newBp.patientUuid, toUtcInstant(newBp.date)))
```
vinaysshenoy: Should we move this `toUtcInstant` lambda to `TimeFunctions.kt`? We have a lot of these useful utility methods related to Time that we end up rewriting a lot.

vinaysshenoy: Instead of computing the instant again, why don't we add the saved `BloodPressureMeasurement` instance as a property of `BloodPressureSaved` and use the `recordedAt` from there?

app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -126,6 +126,17 @@ data class Patient(
         newUpdatedAt: Instant
     )
 
+    @Query("""
+      UPDATE Patient
+      SET recordedAt = :bpRecordedAt, syncStatus = :pendingStatus, updatedAt = :updatedAt
+      WHERE uuid = :patientUuid AND recordedAt > :bpRecordedAt
+    """)
+    abstract fun updateRecordedAt(
```
vinaysshenoy: There is nothing blood pressure related about this query; we are just setting the `recordedAt` of the patient to the minimum of the patient's `recordedAt` and the BP's `recordedAt`. Should we rename the method and the `bpRecordedAt` parameter to better reflect this?

vinaysshenoy: This goes for the method and parameter naming in the `PatientRepository` as well since the `Instant` could be from anywhere. It's only the `BloodPressureEntrySheetController` that decides that the `recordedAt` of the BP could potentially be the new `recordedAt` of the patient.

sanchita: Good point. I will rename the parameter.
why do I need to rename the method though?

vinaysshenoy: This is just my opinion, but when I read the name `updateRecordedAt`, I feel it will always update the `recordedAt` of the patient. But this method conditionally updates it.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
sanchita: I would not want to do that since `BloodPressureSaved` is not for that purpose.
What we are doing with patient `recordedAt` is like a side effect as it does not affect the outcome of BP save in any way.

app/src/main/java/org/simple/clinic/patient/Patient.kt
sanchita: Done.

app/src/main/java/org/simple/clinic/util/TimeFunctions.kt
```diff
@@ -23,3 +24,10 @@ fun Instant.toLocalDateAtZone(zone: ZoneId): LocalDate =
     atZone(UTC)
         .withZoneSameInstant(zone)
         .toLocalDate()
+
+fun LocalDate.toUtcInstant(clock: Clock): Instant {
```
vinaysshenoy: This should take in the `UserClock` instead of a generic `Clock`, right? This is basically meant to convert the local time in the user's timezone to the equivalent UTC `Instant`.

vinaysshenoy: The test can change to create two different `TestUserClock` instances with different zone offsets.

app/src/test/java/org/simple/clinic/util/TimeFunctionsTest.kt
```diff
@@ -178,4 +178,19 @@ class TimeFunctionsTest {
...
       )
 
+  @Test
+  @Parameters(method = "params for testing conversion of local date to instant")
+  fun `different clocks should provide correct instant`(userClock: UserClock, localDate: LocalDate, expectedInstant: Instant) {
+    assertThat(localDate.toUtcInstant(userClock)).isEqualTo(expectedInstant)
+  }
+
+  fun `params for testing conversion of local date to instant`(): List<List<Any>> {
+    val userClock = TestUserClock()
```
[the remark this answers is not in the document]
sanchita: Oh! You're right!

PR #661, opened 2019-05-22
Show see all in recent patients only when the list is too long

rakshakhegde:
Show it only when longer than recent patient limit.

https://www.pivotaltracker.com/story/show/165962163

### DIFF
--- a/app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
+++ b/app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
@@ -5,6 +5,7 @@ import io.reactivex.ObservableSource
 import io.reactivex.ObservableTransformer
 import io.reactivex.rxkotlin.Observables
 import io.reactivex.rxkotlin.ofType
+import io.reactivex.rxkotlin.withLatestFrom
 import org.simple.clinic.ReplayUntilScreenIsDestroyed
 import org.simple.clinic.ReportAnalyticsEvents
 import org.simple.clinic.facility.FacilityRepository
@@ -60,7 +61,14 @@ class RecentPatientsViewController @Inject constructor(
 
     val updateRecentPatients = recentPatientsStream
         .map { it.map(::recentPatientItem) }
-        .map { it + SeeAllItem }
+        .withLatestFrom(patientConfig)
+        .map { (recentPatients, config) ->
+          if (recentPatients.size > config.recentPatientLimit) {
+            recentPatients + SeeAllItem
+          } else {
+            recentPatients
+          }
+        }
         .map { { ui: Ui -> ui.updateRecentPatients(it) } }
 
     val toggleEmptyState = recentPatientsStream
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
```diff
@@ -60,7 +61,10 @@ class RecentPatientsViewController @Inject constructor(
 
     val updateRecentPatients = recentPatientsStream
         .map { it.map(::recentPatientItem) }
-        .map { it + SeeAllItem }
+        .withLatestFrom(patientConfig)
+        .map { (recentPatients, config) ->
+          maybeAddSeeAllButton(recentPatients, config)
```
sanchita: We can actually inline this code since this method is not being reused.
Also the scope of `maybeAddSeeAllButton` method is limited to `showRecentPatients()`.

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
```diff
@@ -72,6 +76,10 @@ class RecentPatientsViewController @Inject constructor(
     return Observable.merge(updateRecentPatients, toggleEmptyState)
   }
 
+  private fun maybeAddSeeAllButton(recentPatients: List<RecentPatientItem>, config: PatientConfig) =
+      if (recentPatients.size > config.recentPatientLimit) recentPatients + SeeAllItem
+      else recentPatients
```
sanchita: We should add brackets to the `if-else` statements even if they are one lines only.

sanchita: If there are multiple lines in a method then we should avoid using `=` in a method and enclose the statements in `{ }`

app/src/test/java/org/simple/clinic/recentpatientsview/RecentPatientsViewControllerTest.kt
```diff
@@ -139,6 +140,90 @@ class RecentPatientsViewControllerTest {
...
+            age = Age(37, Instant.now(), LocalDate.MIN),
+            gender = TRANSGENDER,
+            lastBp = LastBp(systolic = 139, diastolic = 92, recordedAt = Instant.now().minus(3, ChronoUnit.DAYS))
+        )
+    )))
+
+    uiEvents.onNext(ScreenCreated())
+
+    verify(screen).updateRecentPatients(listOf(
+        RecentPatientItem(
```
sanchita: Can you extract creation of `RecentPatientItem` to variables and use that in both returning from `patientRepository.recentPatients(facility.uuid, recentPatientLimit)` and verifying it here?

sanchita: I realize that this change is more involved and this works for now.
Maybe we can think of doing this in the future.

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
rakshakhegde: Inlined

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
rakshakhegde: Inlined

PR #664, opened 2019-05-24
Switch to RxDogTag for improving RxJava stacktraces

vinaysshenoy:
- Traceur is an abandoned project and is also not compatible with
newer versions of RxJava
- We faced a problem where Traceur was causing a crash when an
exception was thrown in a blocking call (root cause not yet determined)

[RxDogTag](https://github.com/uber/RxDogTag) does not have any of these problems, and the stacktraces look
nicer and easier to parse as well.

Screen Shot 2019-05-24 at 11 59 35 AM


vinaysshenoy:
Closing this because `RxDogTag` doesn't give as much information as we'd like. Investigation on `Traceur` has shown that we'd probably have to disable it on production anyway.

PR #665, opened 2019-05-27
Reduce coupling between BloodPressureRepository and UserSession

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/166263263


app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -457,10 +464,32 @@ class BloodPressureEntrySheetController @Inject constructor(
         .ofType<BloodPressureDateToPrefillCalculated>()
         .map { it.date }
 
-    val newBpDataStream = Observables.combineLatest(dateValidations, bpValidations, patientUuidStream)
-        .map { (dateResult, bpResult, patientUuid) ->
+    val loggedInUserStream = userSession
+        .loggedInUser()
```
sanchita: Why don't we use `requireLoggedInUser()` here?
That already takes care of `Optional<User>`

vinaysshenoy: The difference is that `requireLoggedInUser` will throw an error if there is no user present locally, whereas this current chain will just not emit anything.

But you are right that we can use `requireLoggedInUser` here because it is not possible for the user to reach this code block if they are not logged in. I'll change this.

PR #666, opened 2019-05-27
Reduce coupling between AppointmentRepository and UserSession

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/166266668


app/src/test/java/org/simple/clinic/home/overdue/OverdueScreenControllerTest.kt
```diff
@@ -108,64 +108,42 @@ class OverdueScreenControllerTest {
   }
```
vinaysshenoy: The diff for this commit doesn't view very nicely because I deleted two tests in the middle and added a new test at the end. It might be better for the reviewer to view this file as a whole instead of as a diff.

PR #668, opened 2019-05-27
Show see all in recent patients only when the list is too long

rakshakhegde:
https://www.pivotaltracker.com/story/show/165962163

### DIFF
--- a/app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
+++ b/app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
@@ -5,6 +5,7 @@ import io.reactivex.ObservableSource
 import io.reactivex.ObservableTransformer
 import io.reactivex.rxkotlin.Observables
 import io.reactivex.rxkotlin.ofType
+import io.reactivex.rxkotlin.withLatestFrom
 import org.simple.clinic.ReplayUntilScreenIsDestroyed
 import org.simple.clinic.ReportAnalyticsEvents
 import org.simple.clinic.facility.FacilityRepository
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
```diff
@@ -53,14 +55,17 @@ class RecentPatientsViewController @Inject constructor(
     val recentPatientsStream = Observables
         .combineLatest(currentFacilityStream, patientConfig)
         .switchMap { (facility, config) ->
-          patientRepository.recentPatients(facility.uuid, limit = config.recentPatientLimit)
+          patientRepository.recentPatients(facility.uuid, limit = config.recentPatientLimit + 1)
```
vinaysshenoy: Can you add a comment here about why you are doing this?

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
```diff
@@ -72,6 +76,17 @@ class RecentPatientsViewController @Inject constructor(
     return Observable.merge(updateRecentPatients, toggleEmptyState)
   }
 
+  private fun addSeeAllIfListTooLong(
+      recentPatients: List<RecentPatientItem>,
+      recentPatientLimit: Int
+  ) =
+      if (recentPatients.size > recentPatientLimit) {
```
vinaysshenoy: I just realised that there is a `.take` extension function on `Iterable`, so you can also make this code block a bit more terse:

```kotlin
if (recentPatients.size > recentPatientLimit) {
  recentPatients.take(recentPatientLimit) + SeeAllItem
} else {
  recentPatients
}
```

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
rakshakhegde: Added comment

PR #671, opened 2019-05-28
Use a global api version for all Retrofit interfaces

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/166278675


app/src/main/java/org/simple/clinic/di/NetworkModule.kt
```diff
@@ -104,8 +104,10 @@ open class NetworkModule {
   @AppScope
   fun retrofit(commonRetrofitBuilder: Retrofit.Builder): Retrofit {
     val baseUrl = BuildConfig.API_ENDPOINT
+    val currentApiVersion = "v3"
+
     return commonRetrofitBuilder
-        .baseUrl(baseUrl)
+        .baseUrl("$baseUrl$currentApiVersion/")
```
vinaysshenoy: We needed to add this "/" at the end here because Retrofit complains if the base url does not end with a forward slash.

PR #676, opened 2019-05-29
Fix crash when starting the Release build

vinaysshenoy:
Earlier, we were initializing them as soon as the clas was instantiated
but this caused a crash when we tried to access them in ClinicApp. This
changes the creation to a getter so that they are initialized after
onCreate() is called

### DIFF
--- a/app/src/main/java/org/simple/clinic/ReleaseClinicApp.kt
+++ b/app/src/main/java/org/simple/clinic/ReleaseClinicApp.kt
@@ -18,7 +18,9 @@ class ReleaseClinicApp : ClinicApp() {
   @Inject
   lateinit var syncIndicatorStatusCalculator: SyncIndicatorStatusCalculator
 
-  override val analyticsReporters = listOf(HeapAnalyticsReporter(this))
+  override val analyticsReporters by lazy {
+    listOf(HeapAnalyticsReporter(this))
+  }
 
   override fun onCreate() {
     super.onCreate()


app/src/main/java/org/simple/clinic/ReleaseClinicApp.kt
```diff
@@ -18,7 +18,8 @@ class ReleaseClinicApp : ClinicApp() {
   @Inject
   lateinit var syncIndicatorStatusCalculator: SyncIndicatorStatusCalculator
 
-  override val analyticsReporters = listOf(HeapAnalyticsReporter(this))
+  override val analyticsReporters: List<HeapAnalyticsReporter>
+    get() = listOf(HeapAnalyticsReporter(this))
```
rakshakhegde: Are you sure you want a getter and not a lazy function? Because this will return a new instance of `HeapAnalyticsReporter` every time `analyticsReporters` is used.

rakshakhegde: Also your commit name requirement only says this needs to be lazily created.

vinaysshenoy: By lazily, I meant instantiated when required instead of when the class is created. Since this is only called once when the application is instantiated, it should be fine.

My intention is to get SBX working again, at the moment. If you feel that making it an explicit lazy initialiser is needed, I can do that in another PR.

rakshakhegde: This is relatively small change that can go in this PR I feel. Can you please try out the `lazy` block way of doing it?

vinaysshenoy: Sure. I've left for the day though, and I won't be able to change it until later tonight. This is why I was going to do it in another PR because I don't want to ask for an urgent review late night.

If you'll be able to review it later, I'll make the change in this PR.

rakshakhegde: That's alright :)

vinaysshenoy: Thanks!

PR #681, opened 2019-05-30
Handle dates in recent patients correctly

rakshakhegde:
https://www.pivotaltracker.com/story/show/165962179

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/RelativeTimestampGenerator.kt
+++ b/app/src/main/java/org/simple/clinic/summary/RelativeTimestampGenerator.kt
@@ -1,30 +1,37 @@
 package org.simple.clinic.summary
 
 import android.content.Context
+import androidx.annotation.VisibleForTesting
 import org.simple.clinic.R
 import org.threeten.bp.Instant
 import org.threeten.bp.LocalDateTime
 import org.threeten.bp.ZoneOffset
 import org.threeten.bp.format.DateTimeFormatter
-import org.threeten.bp.temporal.ChronoUnit
+import org.threeten.bp.temporal.ChronoUnit.DAYS
 import java.util.Locale
 import javax.inject.Inject
 
 class RelativeTimestampGenerator @Inject constructor() {
 
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/summary/RelativeTimestampGenerator.kt
```diff
@@ -63,3 +68,13 @@ data class OlderThanSixMonths(private val time: Instant) : RelativeTimestamp() {
...
   }
 }
+
+data class Future(private val time: Instant) : RelativeTimestamp() {
+  override fun displayText(context: Context): String {
+    return timestampFormatter.format(time.atZone(ZoneOffset.UTC).toLocalDateTime())
+  }
+
+  companion object {
+    val timestampFormatter = DateTimeFormatter.ofPattern("d MMM, yyyy", Locale.ENGLISH)!!
```
vinaysshenoy: Can you inject this via the `DateFormatterModule` instead of having a static instance?

app/src/main/java/org/simple/clinic/summary/RelativeTimestampGenerator.kt
```diff
@@ -15,17 +15,22 @@ class RelativeTimestampGenerator @Inject constructor() {
   fun generate(today: LocalDateTime, time: Instant): RelativeTimestamp {
```
vinaysshenoy: Since this is not being used anywhere except internally and in tests, should we add a `@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)` annotation?

app/src/main/java/org/simple/clinic/summary/RelativeTimestampGenerator.kt
vinaysshenoy: Never mind, I see the rest of the classes in this file also use the same convention. We'll anyway have to unify the relative date presentation across the codebase later, so let's leave this as is.

app/src/main/java/org/simple/clinic/summary/RelativeTimestampGenerator.kt
vinaysshenoy: Added this myself.

PR #684, opened 2019-06-03
Text getting chopped off in phone masking bottom sheet

rakshakhegde:
https://www.pivotaltracker.com/story/show/166267416

### DIFF
--- a/app/src/main/res/layout/sheet_phone_mask.xml
+++ b/app/src/main/res/layout/sheet_phone_mask.xml
@@ -15,7 +15,6 @@
     android:layout_marginEnd="@dimen/spacing_16"
     android:layout_marginStart="@dimen/spacing_16"
     android:gravity="center"
-    android:maxLines="1"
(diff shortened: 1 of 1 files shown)


app/src/main/res/layout/sheet_phone_mask.xml
```diff
@@ -15,7 +15,6 @@
     android:layout_marginEnd="@dimen/spacing_16"
     android:layout_marginStart="@dimen/spacing_16"
     android:gravity="center"
-    android:maxLines="1"
```
vinaysshenoy: Image

The text looks weird when it wraps to multiple lines. What do you think about setting the gravity to `start` instead?

rakshakhegde: I'm not sure. It'd look even weird. Can you please try it out?

vinaysshenoy: Tried. `center` is better. Let's keep as is.

PR #687, opened 2019-06-06
Finish bottom sheets when user becomes unauthorized

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/166491353


vinaysshenoy:
This is a little involved to test manually. The best way is:

1. In `ClinicApp`, uncomment line 76 
```kotlin
unauthorizeUser.listen(Schedulers.io())
```
This takes care of updating the user's logged in status to `UNAUTHORIZED` when the server throws a `401 Unauthorized`, which is the trigger for all these events. We commented it out because we were waiting for the feature to complete to integrate it.

2. Keep two emulators open and install the app on both of them.
3. Login on one emulator and not on the other one
4. Perform the actions to bring up  the bottom sheet that is being tested
5. Login on the other emulator with the same account
6. On the first emulator, bring down the notification tray (while the bottom sheet is still open) and find the Debug notification and trigger a manual sync
7. The bottom sheet should automatically close.

PR #688, opened 2019-06-06
Fix: Overdue Appointments not loading

vinaysshenoy:
Bug Tracker Link: https://www.pivotaltracker.com/story/show/166494209

Post-mortem:
We refactored the AppointmentRepository to accept the current facility
externally instead of looking it up itself in a past commit. As part of
that change, we updated OverdueScreenController to load the facility
and query the method like so -

  val facilityStream = userSession
	.requireLoggedInUser()
	.switchMap { facilityRepository.currentFacility(it) }

  events.ofType<ScreenCreated>()
	.withLatestFrom(facilityStream)
	.flatMap { (_, facility) ->
		appointmentRepository.overdue(facility)
	}

The failure happened because the ScreenCreated event was getting
emitted *BEFORE* the facility was queried, so when the
wlf(facilityStream) is called, there is no facility yet as part of that
stream to combine, so the overdue appointments nver get called.

The test passed because the test emits the ScreenCreated event after
the facility gets emitted in the test.

Fix: This was fixed by changing the stream to start querying the
current facility *AFTER* the ScreenCreated event is emitted.

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/OverdueScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/OverdueScreenController.kt
@@ -18,6 +18,7 @@ import org.threeten.bp.LocalDateTime
 import org.threeten.bp.Period
 import org.threeten.bp.ZoneOffset.UTC
 import org.threeten.bp.temporal.ChronoUnit.DAYS
+import timber.log.Timber
 import javax.inject.Inject
 
 typealias Ui = OverdueScreen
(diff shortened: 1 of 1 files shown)


vinaysshenoy:
> Can a failing test be added for this? @vinaysshenoy

@rakshakhegde I thought of writing a failing test initially for this, but this will affect all the tests for the controller. This is because we set up default values statically using `Observable.just` in our mocks in the test setup functions as well as subscribe to the event stream in the setup function.

If I have to write a failing test for this, I'd have to introduce a `Subject` instead and then ensure that there aren't any default values being provided in the `setUp` function. This will also require us to change all the tests in the class to set up the default values themselves.

While we can do this, I also realised that almost all the tests in the app are build in this way, and instead of changing it now for this isolated case, I feel we should all take a decision on how we need to change the tests across the app in a way that makes them behave more similarly to reality.

What do you think?

rakshakhegde:
Alright, we can do this in another PR. I've documented this as a story here: https://www.pivotaltracker.com/story/show/166495328

vinaysshenoy:
> Alright, we can do this in another PR. I've documented this as a story here: https://www.pivotaltracker.com/story/show/166495328

Thanks! Let's discuss this in more detail tomorrow.

PR #690, opened 2019-06-10
Add "logged out on this device" dialog

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/166571175

### DIFF
--- a/app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreenController.kt
@@ -35,11 +35,13 @@ class RegistrationPhoneScreenController @Inject constructor(
         .compose(ReportAnalyticsEvents())
         .replay()
 
-    return Observable.merge(
+    return Observable.mergeArray(
         createEmptyOngoingEntryAndPreFill(replayedEvents),
         showValidationError(replayedEvents),
         hideValidationError(replayedEvents),
-        saveOngoingEntryAndProceed(replayedEvents))
+        saveOngoingEntryAndProceed(replayedEvents),
+        showLoggedOutOnThisDeviceDialog(replayedEvents)
+    )
   }
 
   private fun createEmptyOngoingEntryAndPreFill(events: Observable<UiEvent>): Observable<UiChange> {
@@ -156,4 +158,16 @@ class RegistrationPhoneScreenController @Inject constructor(
       createdAt = loggedInUserPayload.createdAt,
       updatedAt = loggedInUserPayload.updatedAt
   )
+
+  private fun showLoggedOutOnThisDeviceDialog(events: Observable<UiEvent>): Observable<UiChange> {
+    return events
+        .ofType<RegistrationPhoneScreenCreated>()
+        .flatMap {
+          userSession
+              .isUserUnauthorized()
+              .take(1)
+        }
+        .filter { isUserUnauthorized -> isUserUnauthorized }
+        .map { { ui: Ui -> ui.showLoggedOutOfDeviceDialog() } }
+  }
 }
--- a/app/src/main/java/org/simple/clinic/registration/phone/loggedout/LoggedOutOfDeviceDialog.kt
+++ b/app/src/main/java/org/simple/clinic/registration/phone/loggedout/LoggedOutOfDeviceDialog.kt
@@ -76,8 +76,6 @@ class LoggedOutOfDeviceDialog : AppCompatDialogFragment() {
   }
 
   private fun setupDialog() {
-    okayButton.isEnabled = false
-
     bindUiToControllerWithoutDelay(
         ui = this,
         events = screenCreates(),
@@ -100,5 +98,9 @@ class LoggedOutOfDeviceDialog : AppCompatDialogFragment() {
 
   fun enableOkayButton() {
     okayButton.isEnabled = true
+  }
+
+  fun disableOkayButton() {
+    okayButton.isEnabled = false
   }
 }
--- a/app/src/main/java/org/simple/clinic/registration/phone/loggedout/LoggedOutOfDeviceDialog.kt
+++ b/app/src/main/java/org/simple/clinic/registration/phone/loggedout/LoggedOutOfDeviceDialog.kt
@@ -1,0 +1,104 @@
+package org.simple.clinic.registration.phone.loggedout
+
+import android.annotation.SuppressLint
+import android.app.Dialog
+import android.content.DialogInterface
+import android.os.Bundle
+import android.widget.Button
+import androidx.appcompat.app.AlertDialog
+import androidx.appcompat.app.AppCompatDialogFragment
+import androidx.fragment.app.FragmentManager
+import io.reactivex.Observable
+import io.reactivex.subjects.PublishSubject
+import org.simple.clinic.R
+import org.simple.clinic.activity.TheActivity
+import org.simple.clinic.bindUiToControllerWithoutDelay
+import org.simple.clinic.util.unsafeLazy
+import org.simple.clinic.widgets.ScreenCreated
+import org.simple.clinic.widgets.ScreenDestroyed
+import org.simple.clinic.widgets.UiEvent
+import javax.inject.Inject
+
+class LoggedOutOfDeviceDialog : AppCompatDialogFragment() {
+
+  companion object {
+    private const val FRAGMENT_TAG = "LoggedOutOfDeviceDialog"
+
+    fun show(fragmentManager: FragmentManager) {
+      val existingFragment = fragmentManager.findFragmentByTag(FRAGMENT_TAG)
+
+      if (existingFragment != null) {
+        fragmentManager
+            .beginTransaction()
+            .remove(existingFragment)
+            .commit()
+      }
+
+      val fragment = LoggedOutOfDeviceDialog().apply {
+        isCancelable = false
+      }
+
+      fragmentManager
+          .beginTransaction()
+          .add(fragment, FRAGMENT_TAG)
+          .commit()
+    }
+  }
+
+  @Inject
+  lateinit var controller: LoggedOutOfDeviceDialogController
+
+  private val okayButton: Button by unsafeLazy {
+    (dialog as AlertDialog).getButton(DialogInterface.BUTTON_POSITIVE)
+  }
+
+  private val screenDestroys = PublishSubject.create<ScreenDestroyed>()
+  private val onStarts = PublishSubject.create<Any>()
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    TheActivity.component.inject(this)
+  }
+
+  @SuppressLint("CheckResult")
+  override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+    val dialog = AlertDialog.Builder(requireContext())
+        .setTitle(R.string.registration_loggedout_dialog_title)
+        .setMessage(R.string.registration_loggedout_dialog_message)
+        .setPositiveButton(R.string.registration_loggedout_dialog_confirm, null)
+        .create()
+
+    onStarts
+        .take(1)
+        .subscribe { setupDialog() }
+
+    return dialog
+  }
+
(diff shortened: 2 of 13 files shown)


app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreenController.kt
```diff
@@ -181,4 +181,13 @@ class RegistrationPhoneScreenController @Inject constructor(
...
         )
   }
+
+  private fun showLoggedOutOnThisDeviceDialog(events: Observable<UiEvent>): Observable<UiChange> {
+    return events
+        .ofType<RegistrationPhoneScreenCreated>()
+        .flatMapSingle { userSession.loggedInUser().firstOrError() }
+        .filterAndUnwrapJust()
+        .filter { loggedInUser -> loggedInUser.loggedInStatus == UNAUTHORIZED }
+        .flatMap { Observable.just { ui: Ui -> ui.showLoggedOutOnThisDeviceDialog() } }
```
sanchita: This could be simplified to just this: 
`map { { ui: Ui -> ui.showLoggedOutOnThisDeviceDialog() } }`

sanchita: Also, can we extract if user is `UNAUTHORIZED` to `UserSession` itself ?
The controller need not know what unauthorised means.
Wdyt?

app/src/test/java/org/simple/clinic/registration/phone/RegistrationPhoneScreenControllerTest.kt
```diff
@@ -280,4 +289,62 @@ class RegistrationPhoneScreenControllerTest {
...
+        testCase(
+            loggedInUser = PatientMocker.loggedInUser(loggedInStatus = RESETTING_PIN),
+            shouldShowLoggedOutDialog = false
+        ),
+        testCase(
+            loggedInUser = PatientMocker.loggedInUser(loggedInStatus = RESET_PIN_REQUESTED),
+            shouldShowLoggedOutDialog = false
+        ),
+        testCase(
+            loggedInUser = PatientMocker.loggedInUser(loggedInStatus = UNAUTHORIZED),
```
sanchita: Another reason for moving the check for whether a user is `UNAUTHORIZED` or not.
The `RegistrationPhoneScreenControllerTest` need not be aware of various `loggedInStatus`.
This logic should not be part of the Controller itself.

vinaysshenoy:
> Testing is pending.
> Will add more feedback after testing.

@sanchita A better option for manual testing would be to wait until #691 is ready since that will be the final PR that integrates all these features together. Right now, the only thing you can manually test is that these changes do not affect the current flow.

app/src/main/java/org/simple/clinic/registration/phone/loggedout/LoggedOutOnThisDeviceDialog.kt
```diff
@@ -0,0 +1,104 @@
...
+import org.simple.clinic.R
+import org.simple.clinic.activity.TheActivity
+import org.simple.clinic.bindUiToControllerWithoutDelay
+import org.simple.clinic.util.unsafeLazy
+import org.simple.clinic.widgets.ScreenCreated
+import org.simple.clinic.widgets.ScreenDestroyed
+import org.simple.clinic.widgets.UiEvent
+import javax.inject.Inject
+
+class LoggedOutOnThisDeviceDialog : AppCompatDialogFragment() {
```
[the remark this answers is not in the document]
vinaysshenoy: Yeah, it can be. We already have a `LoggedOutOnOtherDeviceDialog` which is shown when the user enters the OTP and logs in, so this for me seemed like a good name for its counterpart. I can rename it if you feel this sounds better, though.

app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreenController.kt
```diff
@@ -52,15 +55,24 @@ class RegistrationPhoneScreenController @Inject constructor(
...
+    // One exception where we don't clear the user is if the user's logged in status
+    // is UNAUTHORIZED, because this means that this screen has been opened to
+    // ask the user to sign in again, and this is handled elsewhere.
+    return events
         .ofType<RegistrationPhoneScreenCreated>()
+        .flatMapSingle { userSession.loggedInUser().firstOrError() }
+        .filterAndUnwrapJust()
+        .filter { loggedInUser -> loggedInUser.loggedInStatus != User.LoggedInStatus.UNAUTHORIZED }
         .flatMap {
           userSession.clearLoggedInUser().andThen(Observable.empty<UiChange>())
```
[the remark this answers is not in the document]
vinaysshenoy: I agree with you. However, refactoring is a little tricky because `Flow` currently requires us to provide a key when initialising it and this is the key that is provided in most cases. This also needs to be provided in a blocking fashion and is invoked on the main thread, so we need to avoid database calls on this, so this clearing of data was being done right here.

I have added a maintenance story to come back to this at some point. https://www.pivotaltracker.com/story/show/166607823

app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreenController.kt
vinaysshenoy: > This could be simplified to just this:
> `map { { ui: Ui -> ui.showLoggedOutOnThisDeviceDialog() } }`

Oh yeah, this is doable. I don't know why I missed this. 😅

vinaysshenoy: > Also, can we extract if user is `UNAUTHORIZED` to `UserSession` itself ?
> The controller need not know what unauthorised means.
> Wdyt?

This is a good idea, we have duplicated this code in quite a few places. I'll do this.

app/src/test/java/org/simple/clinic/registration/phone/RegistrationPhoneScreenControllerTest.kt
vinaysshenoy: Yup. Raised a story to move this out at some point: https://www.pivotaltracker.com/story/show/166607823

app/src/main/java/org/simple/clinic/registration/phone/loggedout/LoggedOutOnThisDeviceDialog.kt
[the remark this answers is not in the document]
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreenController.kt
vinaysshenoy: Done.

app/src/test/java/org/simple/clinic/registration/phone/RegistrationPhoneScreenControllerTest.kt
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreenController.kt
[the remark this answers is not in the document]
sanchita: I actually meant that we could think of clearing data based on the current state of the user and not based on the UI changes like `RegistrationPhoneScreenCreated`.
Let's discuss this in-person.

sanchita:
The app is crashing on `PinEntryCardController` when we try to login.
Its happening on **line 60.**
I guess the `LoggedInUser` is being cleared. I m saying this because `loggedInUser()` is returning `null` and hence `UserSession#requireLoggedInUser` is throwing `AssertionError`.

vinaysshenoy:
Temporality put into WIP until https://www.pivotaltracker.com/story/show/166632435 is done.

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -525,6 +525,19 @@ class UserSession @Inject constructor(
...
+    return loggedInUser()
+        .filterAndUnwrapJust()
+        .map { user -> user.loggedInStatus }
+        .map { loggedInStatus ->
+          when (loggedInStatus) {
+            UNAUTHORIZED -> true
+            else -> false
+          }
+        }
+        .distinctUntilChanged()
```
saket: Would you prefer this?

```kotlin
return loggedInUser()
  .filterAndUnwrapJust()
  .map { it.loggedInStatus == UNAUTHORIZED }
  .distinctUntilChanged()
```

vinaysshenoy:
> @vinaysshenoy The app is crashing on `PinEntryCardController` when we try to login.
> Its happening on **line 60.**
> I guess the `LoggedInUser` is being cleared. I m saying this because `loggedInUser()` is returning `null` and hence `UserSession#requireLoggedInUser` is throwing `AssertionError`.

Finally fixed with the major refactor. Opening for review again.

Note: The final piece of the auto-logout feature will happen in #691 where we will actually set the user status to `UNAUTHORIZED` and move the user to the `RegistrationPhoneScreen` to login again. This commit actually doesn't introduce any new behaviour.

app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreenController.kt
```diff
@@ -156,4 +158,13 @@ class RegistrationPhoneScreenController @Inject constructor(
       createdAt = loggedInUserPayload.createdAt,
       updatedAt = loggedInUserPayload.updatedAt
   )
+
+  private fun showLoggedOutOnThisDeviceDialog(events: Observable<UiEvent>): Observable<UiChange> {
+    return events
+        .ofType<RegistrationPhoneScreenCreated>()
+        .flatMap { userSession.isUserUnauthorized() }
+        .take(1)
```
sanchita: `take(1)` will complete the upstream. 
Can you change this to the following? 

`.flatMap { userSession.isUserUnauthorized().take(1) }`

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -399,6 +399,18 @@ class UserSession @Inject constructor(
   fun syncAndClearData(patientRepository: PatientRepository, syncRetryCount: Int = 0, timeoutSeconds: Long = 15L): Completable {
     Timber.i("Syncing and clearing all patient related data")
 
+    val updateUserStatusToResettingPin = requireLoggedInUser()
+        .firstOrError()
+        .flatMapCompletable { user -> updateLoggedInStatusForUser(user.uuid, RESETTING_PIN) }
```
sanchita: This is another hidden behavior inside a method that should only be responsible for clearing data. 
This should have been a separate method to be invoked from the caller itself. In this case `ForgotPinConfirmPinScreenController`. 
No need to refactor it in this PR. Just an FYI.

Every refactor in this class leads to more cracks opening :(

app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreenController.kt
vinaysshenoy: Whoops, I didn't know this. I'll fix this.

app/src/main/java/org/simple/clinic/user/UserSession.kt
vinaysshenoy: Good point. I decided to refactor this and improve the tests a little bit.

app/src/main/java/org/simple/clinic/registration/phone/loggedout/LoggedOutOfDeviceDialog.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    onStarts
+        .take(1)
+        .subscribe { setupDialog() }
+
+    return dialog
+  }
+
+  private fun setupDialog() {
+    okayButton.isEnabled = false
```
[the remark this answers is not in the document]
vinaysshenoy: Makes sense, I'll do this.

PR #693, opened 2019-06-12
Check app updates available

sanchita:
Pivotol story -> https://www.pivotaltracker.com/story/show/163619300

### DIFF
--- a/app/src/main/java/org/simple/clinic/appupdate/AppUpdateConfig.kt
+++ b/app/src/main/java/org/simple/clinic/appupdate/AppUpdateConfig.kt
@@ -1,0 +1,6 @@
+package org.simple.clinic.appupdate
+
+data class AppUpdateConfig(
+    val inAppUpdateEnabled: Boolean,
+    val differenceBetweenVersionsToNudge: Long
+)
--- a/app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
+++ b/app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
@@ -1,0 +1,62 @@
+package org.simple.clinic.appupdate
+
+import android.app.Application
+import android.os.Build
+import androidx.annotation.VisibleForTesting
+import com.google.android.play.core.appupdate.AppUpdateInfo
+import com.google.android.play.core.appupdate.AppUpdateManagerFactory
+import com.google.android.play.core.install.model.AppUpdateType.FLEXIBLE
+import com.google.android.play.core.install.model.UpdateAvailability
+import io.reactivex.Single
+import org.simple.clinic.util.None
+import org.simple.clinic.util.Optional
+import org.simple.clinic.util.toOptional
+import javax.inject.Inject
+
+class CheckAppUpdateAvailability @Inject constructor(
+    private val appContext: Application,
+    private val config: AppUpdateConfig,
+    private val versionUpdateCheck: (Int, Application, AppUpdateConfig) -> Boolean = isVersionApplicableForUpdate
+) {
+
+  fun listen(): Single<Optional<AppUpdateData>> {
+    val appUpdateManager = AppUpdateManagerFactory.create(appContext)
+    val appUpdateInfoTask = appUpdateManager.appUpdateInfo
+
+    return Single.create { emitter ->
+
+      appUpdateInfoTask.addOnSuccessListener { appUpdateInfo ->
+        if (shouldNudgeForAppUpdate(appUpdateInfo)) {
+          emitter.onSuccess(AppUpdateData(appUpdateInfo).toOptional())
+        } else {
+          emitter.onSuccess(None)
+        }
+      }
+
+      appUpdateInfoTask.addOnFailureListener { exception ->
+        emitter.onError(exception)
+      }
+    }
+  }
+
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  fun shouldNudgeForAppUpdate(appUpdateInfo: AppUpdateInfo): Boolean {
+    return config.inAppUpdateEnabled
+        && appUpdateInfo.updateAvailability() == UpdateAvailability.UPDATE_AVAILABLE
+        && appUpdateInfo.isUpdateTypeAllowed(FLEXIBLE)
+        && versionUpdateCheck(appUpdateInfo.availableVersionCode(), appContext, config)
+  }
+}
+
+private val isVersionApplicableForUpdate = { availableVersionCode: Int, appContext: Application, config: AppUpdateConfig ->
+  val packageInfo = appContext.packageManager.getPackageInfo(appContext.packageName, 0)
+
+  if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+    val longVersionCode = packageInfo.longVersionCode
+    val versionCode = longVersionCode.and(0xffffffff)
+    availableVersionCode.minus(versionCode) >= config.differenceBetweenVersionsToNudge
+
+  } else {
+    availableVersionCode.minus(packageInfo.versionCode) >= config.differenceBetweenVersionsToNudge
+  }
+}
--- a/app/src/test/java/org/simple/clinic/appupdate/CheckAppUpdateAvailabilityTest.kt
+++ b/app/src/test/java/org/simple/clinic/appupdate/CheckAppUpdateAvailabilityTest.kt
@@ -1,0 +1,123 @@
+package org.simple.clinic.appupdate
+
+import android.app.Application
+import android.app.PendingIntent
+import com.google.android.play.core.appupdate.AppUpdateInfo
+import com.google.android.play.core.install.model.InstallStatus
+import com.google.android.play.core.install.model.UpdateAvailability
+import com.google.common.truth.Truth.assertThat
+import com.nhaarman.mockito_kotlin.mock
+import com.nhaarman.mockito_kotlin.whenever
+import junitparams.JUnitParamsRunner
+import junitparams.Parameters
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(JUnitParamsRunner::class)
+class CheckAppUpdateAvailabilityTest {
+
+  private val config = mock<AppUpdateConfig>()
+
+  lateinit var checkUpdateAvailable: CheckAppUpdateAvailability
+
+  private val currentAppVersionCode = 1
+  private val differenceInVersionsToShowUpdate = 1
+
+  @Before
+  fun setup() {
+
+    val versionCodeCheck = { versionCode: Int, _: Application, _: AppUpdateConfig ->
+      versionCode.minus(currentAppVersionCode) >= differenceInVersionsToShowUpdate
+    }
+
+    checkUpdateAvailable = CheckAppUpdateAvailability(mock(), config, versionCodeCheck)
+  }
+
+  @Test
+  @Parameters(method = "params for checking app update")
+  fun `when app update is available and it is eligible for updates, then the user should be nudged for an app update`(
+      availableVersionCode: Int,
+      updateAvailabilityState: Int,
+      flexibleUpdateIntent: PendingIntent?,
+      isInAppUpdateEnabled: Boolean,
+      shouldShow: Boolean
+  ) {
+    val packageName = "org.simple.clinic"
+    val immediateUpdateIntent = mock<PendingIntent>()
+
+    val appUpdateInfo = AppUpdateInfo(
+        packageName,
+        availableVersionCode,
+        updateAvailabilityState,
+        InstallStatus.DOWNLOADED,
+        immediateUpdateIntent,
+        flexibleUpdateIntent
+    )
+
+    whenever(config.inAppUpdateEnabled).thenReturn(isInAppUpdateEnabled)
+    whenever(config.differenceBetweenVersionsToNudge).thenReturn(1)
+
+    val shouldShowDialog = checkUpdateAvailable.shouldNudgeForAppUpdate(appUpdateInfo)
+
+    assertThat(shouldShowDialog).isEqualTo(shouldShow)
+  }
+
+  fun `params for checking app update`(): List<List<Any?>> {
+
+    fun testCase(
+        versionCode: Int,
+        updateAvailabilityState: Int,
+        flexibleUpdateIntent: PendingIntent?,
+        isInAppUpdateEnabled: Boolean,
+        shouldShow: Boolean
+    ): List<Any?> {
+      return listOf(versionCode, updateAvailabilityState, flexibleUpdateIntent, isInAppUpdateEnabled, shouldShow)
+    }
+
+    return listOf(
+        testCase(
+            versionCode = 1,
+            updateAvailabilityState = UpdateAvailability.UPDATE_AVAILABLE,
+            flexibleUpdateIntent = mock(),
+            isInAppUpdateEnabled = true,
+            shouldShow = false
+        ),
+        testCase(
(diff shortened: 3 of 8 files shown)


build.gradle
```diff
@@ -58,6 +58,7 @@ buildscript {
       firebaseCore        : '16.0.8',
       firebaseConfig      : '16.4.1',
       googleServices      : '4.0.1',
+      playCore            : '1.5.0'
```
vinaysshenoy: The current latest version on [Google Maven](https://maven.google.com) is **1.6.1**. The developer docs don't usually point to the latest version. Use this version instead?

app/src/main/java/org/simple/clinic/appupdate/AppUpdateConfig.kt
```diff
@@ -0,0 +1,6 @@
+package org.simple.clinic.appupdate
+
+data class AppUpdateConfig(
+    val inAppUpdateEnabled: Boolean,
+    val numberOfVersionsToDiff: Long
```
vinaysshenoy: The number of versions to diff seems like it would be better off being remotely configurable. Are we planning to do this before releasing this feature?

vinaysshenoy: Rename to `differenceBetweenVersionsToNudge`?

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
```diff
@@ -0,0 +1,65 @@
...
+
+  if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+    val longVersionCode = packageManager.longVersionCode
+    val versionCode = longVersionCode.and(0xffffffff)
+    Timber.d("VersionCode: $versionCode")
+    availableVersionCode.minus(versionCode) >= config.numberOfVersionsToDiff
+
+  } else {
+    availableVersionCode.minus(packageManager.versionCode) >= config.numberOfVersionsToDiff
+
```
vinaysshenoy: Extra line?

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
```diff
@@ -0,0 +1,65 @@
...
+  @VisibleForTesting
+  fun shouldShowUpdateDialog(appUpdateInfo: AppUpdateInfo): Boolean {
+    return config.inAppUpdateEnabled
+        && appUpdateInfo.updateAvailability() == UpdateAvailability.UPDATE_AVAILABLE
+        && appUpdateInfo.isUpdateTypeAllowed(FLEXIBLE)
+        && versionUpdateCheck(appUpdateInfo.availableVersionCode(), appContext, config)
+  }
+}
+
+private val isVersionApplicableForUpdate = { availableVersionCode: Int, appContext: Application, config: AppUpdateConfig ->
```
vinaysshenoy: This can be a function too.

```kotlin
class CheckAppUpdateAvailability @Inject constructor(
    private val appContext: Application,
    private val config: AppUpdateConfig,
    private val versionUpdateCheck: (Int, Application, AppUpdateConfig) -> Boolean = ::isVersionApplicableForUpdate
) {
}

private fun isVersionApplicableForUpdate(availableVersionCode: Int, appContext: Application, config: AppUpdateConfig): Boolean {
  val packageManager = appContext.packageManager.getPackageInfo(appContext.packageName, 0)

  return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
    val longVersionCode = packageManager.longVersionCode
    val versionCode = longVersionCode.and(0xffffffff)
    availableVersionCode.minus(versionCode) >= config.numberOfVersionsToDiff

  } else {
    availableVersionCode.minus(packageManager.versionCode) >= config.numberOfVersionsToDiff
  }
}
```

I'm not asking for this to be changed, just adding that this is also an option instead of needing to declare a lambda.

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
```diff
@@ -0,0 +1,65 @@
...
+  val packageManager = appContext.packageManager.getPackageInfo(appContext.packageName, 0)
+
+  if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+    val longVersionCode = packageManager.longVersionCode
+    val versionCode = longVersionCode.and(0xffffffff)
+    Timber.d("VersionCode: $versionCode")
```
vinaysshenoy: Leftover log?

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
```diff
@@ -0,0 +1,65 @@
...
+  fun shouldShowUpdateDialog(appUpdateInfo: AppUpdateInfo): Boolean {
+    return config.inAppUpdateEnabled
+        && appUpdateInfo.updateAvailability() == UpdateAvailability.UPDATE_AVAILABLE
+        && appUpdateInfo.isUpdateTypeAllowed(FLEXIBLE)
+        && versionUpdateCheck(appUpdateInfo.availableVersionCode(), appContext, config)
+  }
+}
+
+private val isVersionApplicableForUpdate = { availableVersionCode: Int, appContext: Application, config: AppUpdateConfig ->
+  val packageManager = appContext.packageManager.getPackageInfo(appContext.packageName, 0)
```
vinaysshenoy: This should be `packageInfo`, right?

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
```diff
@@ -0,0 +1,65 @@
...
+        }
+      }
+
+      appUpdateInfoTask.addOnFailureListener { exception ->
+        emitter.onError(exception)
+      }
+    }
+  }
+
+  @VisibleForTesting
```
vinaysshenoy: Can you add a little more information by making this `@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)`?

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
```diff
@@ -0,0 +1,65 @@
...
+import io.reactivex.Single
+import org.simple.clinic.util.None
+import org.simple.clinic.util.Optional
+import org.simple.clinic.util.toOptional
+import timber.log.Timber
+import javax.inject.Inject
+
+class CheckAppUpdateAvailability @Inject constructor(
+    private val appContext: Application,
+    private val config: AppUpdateConfig,
```
vinaysshenoy: We've always followed the convention of injecting configuration as `Observable<Config>`. Can we do the same here?

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
```diff
@@ -0,0 +1,65 @@
...
+      }
+
+      appUpdateInfoTask.addOnFailureListener { exception ->
+        emitter.onError(exception)
+      }
+    }
+  }
+
+  @VisibleForTesting
+  fun shouldShowUpdateDialog(appUpdateInfo: AppUpdateInfo): Boolean {
```
vinaysshenoy: The name `shouldShowUpdateDialog` indicates that this class knows how this update check is going to be presented to the user (that is via a dialog). Rename this to something else, like `shouldNudgeAppUpdate`?

app/src/test/java/org/simple/clinic/appupdate/CheckAppUpdateAvailabilityTest.kt
```diff
@@ -0,0 +1,75 @@
...
+
+@RunWith(JUnitParamsRunner::class)
+class CheckAppUpdateAvailabilityTest {
+
+  private val config = mock<AppUpdateConfig>()
+
+  lateinit var checkUpdateAvailable: CheckAppUpdateAvailability
+
+  private val currentAppVersionCode = 1
+  private val diffToShowUpdate = 1
```
vinaysshenoy: Rename to `differenceInVersionsToShowUpdate`?

app/src/test/java/org/simple/clinic/appupdate/CheckAppUpdateAvailabilityTest.kt
```diff
@@ -0,0 +1,75 @@
...
+  ) {
+    val packageName = "org.simple.clinic"
+    val immediateUpdateIntent = mock<PendingIntent>()
+
+    val appUpdateInfo = AppUpdateInfo(
+        packageName,
+        availableVersionCode,
+        updateAvailabilityState,
+        InstallStatus.DOWNLOADED,
+        immediateUpdateIntent,
```
vinaysshenoy: Since we are not using `immediateUpdateIntent`, maybe inline the `mock<PendingIntent>`?

vinaysshenoy: This goes for the `packageName` as well.

app/src/test/java/org/simple/clinic/appupdate/CheckAppUpdateAvailabilityTest.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    checkUpdateAvailable = CheckAppUpdateAvailability(mock(), config, versionCodeCheck)
+  }
+
+  @Test
+  @Parameters(method = "params for checking app update")
+  fun `when app update is available and version is applicable, then shouldShowUpdateDialog should return true`(
+      availableVersionCode: Int,
+      updateAvailabilityState: Int,
+      flexibleUpdateIntent: PendingIntent?,
```
vinaysshenoy: Is there a need to pass this as a parameter since we aren't even verifying anything on it? We can just inline the `mock<PendingIntent>()` when constructing the `AppUpdateInfo` instance.

app/src/test/java/org/simple/clinic/appupdate/CheckAppUpdateAvailabilityTest.kt
```diff
@@ -0,0 +1,75 @@
...
+        packageName,
+        availableVersionCode,
+        updateAvailabilityState,
+        InstallStatus.DOWNLOADED,
+        immediateUpdateIntent,
+        flexibleUpdateIntent
+    )
+
+    whenever(config.inAppUpdateEnabled).thenReturn(isInAppUpdateEnabled)
+    whenever(config.numberOfVersionsToDiff).thenReturn(1)
```
vinaysshenoy: We should use parameterize this property and verify that this value is used as well when checking for nudging app updates.

app/src/test/java/org/simple/clinic/appupdate/CheckAppUpdateAvailabilityTest.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    val shouldShowDialog = checkUpdateAvailable.shouldShowUpdateDialog(appUpdateInfo)
+
+    assert(shouldShowDialog == shouldShow)
+  }
+
+  fun `params for checking app update`(): List<List<Any?>> {
+    return listOf(
```
vinaysshenoy: I generally prefer to create a local function named `testCase` so I can construct the instances with named parameters that are easier to understand. Do you feel that would be a good idea here? Otherwise, it's hard to tell what the different values mean.

app/src/main/java/org/simple/clinic/appupdate/AppUpdateConfig.kt
sanchita: Yes. This is hard-coded for the time being.
Remote config will be implemented in another PR.

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
sanchita: Since we are not using streams here to get this config, I decided to make as just `AppUpdateConfig`

app/src/test/java/org/simple/clinic/appupdate/CheckAppUpdateAvailabilityTest.kt
sanchita: I specifically did this to make it clear that what these values are.
Since I cannot add named arguments here, I felt this will make it easier to read.

app/src/test/java/org/simple/clinic/appupdate/CheckAppUpdateAvailabilityTest.kt
sanchita: We are actually verifying. If we pass a `null` here, it should fail.

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
vinaysshenoy: What I mean is that we should be using streams here to get this config since that's what we do everywhere.

app/src/test/java/org/simple/clinic/appupdate/CheckAppUpdateAvailabilityTest.kt
vinaysshenoy: Maybe then mock the `AppUpdateInfo` itself?

```kotlin
val appUpdateInfo = mock<AppUpdatedInfo>()
whenever(appUpdateInfo.availableVersionCode).thenReturn(availableVersionCode)
```

With this, you will be able to only set the values that you need and not care about the rest.

app/src/test/java/org/simple/clinic/appupdate/CheckAppUpdateAvailabilityTest.kt
vinaysshenoy: Oh right, there was one which was `null`, which I missed.

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
sanchita: Good point.

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
sanchita: Cool. I will make a note of this.

app/src/test/java/org/simple/clinic/appupdate/CheckAppUpdateAvailabilityTest.kt
sanchita: As discussed, there is no point in testing this out since we are mocking the function `isVersionApplicableForUpdate` which uses this value.

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
sanchita: Yes. Done.

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
sanchita: Oops!

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
vinaysshenoy: Will we make this `Observable<AppUpdateConfig>` in a future PR?

PR #696, opened 2019-06-14
Change PinEntryCardView to accept the PIN to be verified externally

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/166693699

### DIFF
--- a/app/src/test/java/org/simple/clinic/security/pin/PinEntryCardControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/security/pin/PinEntryCardControllerTest.kt
@@ -2,9 +2,11 @@ package org.simple.clinic.security.pin
 
 import com.nhaarman.mockito_kotlin.any
 import com.nhaarman.mockito_kotlin.eq
+import com.nhaarman.mockito_kotlin.inOrder
 import com.nhaarman.mockito_kotlin.mock
 import com.nhaarman.mockito_kotlin.times
 import com.nhaarman.mockito_kotlin.verify
+import com.nhaarman.mockito_kotlin.verifyZeroInteractions
 import com.nhaarman.mockito_kotlin.whenever
 import io.reactivex.Completable
 import io.reactivex.Observable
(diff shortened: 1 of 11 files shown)


app/src/test/java/org/simple/clinic/security/pin/PinEntryCardControllerTest.kt
```diff
@@ -68,75 +65,83 @@ class  PinEntryCardControllerTest {
...
   fun `when 4 digits are entered then the PIN should be submitted automatically`() {
+    val pinDigest = "1234"
     uiEvents.onNext(PinTextChanged("1"))
     uiEvents.onNext(PinTextChanged("12"))
     uiEvents.onNext(PinTextChanged("123"))
     uiEvents.onNext(PinTextChanged("1234"))
+    uiEvents.onNext(PinDigestToVerify(pinDigest))
 
-    verify(passwordHasher, times(1)).compare(loggedInUser.pinDigest, "1234")
+    verify(passwordHasher, times(1)).compare(pinDigest, "1234")
```
rakshakhegde: We can remove explicit `times(1)` here?

rakshakhegde: Also, why do we have the variable `pinDigest` and not using it here `.compare(pinDigest, pinDigest)`

app/src/test/java/org/simple/clinic/security/pin/PinEntryCardControllerTest.kt
```diff
@@ -68,75 +65,83 @@ class  PinEntryCardControllerTest {
...
   }
 
   @Test
   fun `when the PIN is submitted then it should be validated`() {
+    val pinDigest = "1234"
     uiEvents.onNext(PinTextChanged("1234"))
+    uiEvents.onNext(PinDigestToVerify(pinDigest))
 
-    verify(passwordHasher).compare(loggedInUser.pinDigest, "1234")
+    verify(passwordHasher).compare(pinDigest, "1234")
```
rakshakhegde: Similarly here, why do we have the variable `pinDigest`?

app/src/test/java/org/simple/clinic/security/pin/PinEntryCardControllerTest.kt
vinaysshenoy: What this test is actually testing is that "when four digits are entered, the value should be compared to the PIN digest that was set".

This might seem a little confusing because both the value of `pinDigest` and the PIN text changed value that is emitted are set to the same value "1234".

vinaysshenoy: > We can remove explicit `times(1)` here?

Done.

app/src/test/java/org/simple/clinic/security/pin/PinEntryCardControllerTest.kt
vinaysshenoy: Same as the above comment.

PR #702, opened 2019-06-24
Add All Patients In Facility View

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/166811630

### DIFF
--- a/app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListAdapter.kt
+++ b/app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListAdapter.kt
@@ -1,0 +1,23 @@
+package org.simple.clinic.allpatientsinfacility
+
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.recyclerview.widget.ListAdapter
+import org.simple.clinic.widgets.recyclerview.ViewHolderX
+
+class AllPatientsInFacilityListAdapter : ListAdapter<AllPatientsInFacilityListItem, ViewHolderX>(
+    AllPatientsInFacilityListItem.AllPatientsInFacilityListItemCallback()
+) {
+
+  override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolderX {
+    return ViewHolderX(LayoutInflater.from(parent.context).inflate(viewType, parent, false))
+  }
+
+  override fun onBindViewHolder(holder: ViewHolderX, position: Int) {
+    getItem(position).render(holder)
+  }
+
+  override fun getItemViewType(position: Int): Int {
+    return getItem(position).layoutResId
+  }
+}
--- a/app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListItem.kt
+++ b/app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListItem.kt
@@ -11,6 +11,17 @@ import org.simple.clinic.widgets.recyclerview.ViewHolderX
 
 sealed class AllPatientsInFacilityListItem {
 
+  companion object {
+    fun mapSearchResultsToListItems(
+        facility: Facility,
+        patientSearchResults: List<PatientSearchResult>
+    ): List<AllPatientsInFacilityListItem> {
+      return patientSearchResults
+          .map { patientSearchResult -> SearchResult(facility, patientSearchResult) }
+          .let { searchResultItems -> listOf(FacilityHeader(facility.name)) + searchResultItems }
+    }
+  }
+
   abstract val layoutResId: Int
 
   abstract fun render(holder: ViewHolderX)
@@ -21,7 +32,7 @@ sealed class AllPatientsInFacilityListItem {
 
     override fun render(holder: ViewHolderX) {
       val resources = holder.itemView.resources
-      holder.facilityLabel.text = resources.getString(R.string.allpatientsinfacility_no_patients_title, facilityName)
+      holder.facilityLabel.text = resources.getString(R.string.allpatientsinfacility_nopatients_title, facilityName)
     }
   }
 
--- a/app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListItem.kt
+++ b/app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListItem.kt
@@ -32,7 +32,7 @@ sealed class AllPatientsInFacilityListItem {
 
     override fun render(holder: ViewHolderX) {
       val resources = holder.itemView.resources
-      holder.facilityLabel.text = resources.getString(R.string.allpatientsinfacility_nopatients_title, facilityName)
+      holder.facilityLabel.text = resources.getString(R.string.allpatientsinfacility_foundpatients_header, facilityName)
     }
   }
 
--- a/app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListItem.kt
+++ b/app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListItem.kt
@@ -1,0 +1,55 @@
+package org.simple.clinic.allpatientsinfacility
+
+import android.annotation.SuppressLint
+import androidx.recyclerview.widget.DiffUtil
+import kotlinx.android.synthetic.main.list_allpatientsinfacility_facility_header.*
+import kotlinx.android.synthetic.main.list_patient_search.*
+import org.simple.clinic.R
+import org.simple.clinic.facility.Facility
+import org.simple.clinic.patient.PatientSearchResult
+import org.simple.clinic.widgets.recyclerview.ViewHolderX
+
+sealed class AllPatientsInFacilityListItem {
+
+  abstract val layoutResId: Int
+
+  abstract fun render(holder: ViewHolderX)
+
+  data class FacilityHeader(val facilityName: String) : AllPatientsInFacilityListItem() {
+
+    override val layoutResId = R.layout.list_allpatientsinfacility_facility_header
+
+    override fun render(holder: ViewHolderX) {
+      val resources = holder.itemView.resources
+      holder.facilityLabel.text = resources.getString(R.string.allpatientsinfacility_no_patients_title, facilityName)
+    }
+  }
+
+  data class SearchResult(
+      val facility: Facility,
+      val patientSearchResult: PatientSearchResult
+  ) : AllPatientsInFacilityListItem() {
+
+    override val layoutResId = R.layout.list_patient_search
+
+    override fun render(holder: ViewHolderX) {
+      holder.patientSearchResultView.render(patientSearchResult, facility)
+    }
+  }
+
+  class AllPatientsInFacilityListItemCallback : DiffUtil.ItemCallback<AllPatientsInFacilityListItem>() {
+
+    override fun areItemsTheSame(oldItem: AllPatientsInFacilityListItem, newItem: AllPatientsInFacilityListItem): Boolean {
+      return when {
+        oldItem is FacilityHeader && newItem is FacilityHeader -> true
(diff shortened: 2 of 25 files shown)


app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityUiChangeProducer.kt
```diff
@@ -3,21 +3,25 @@ package org.simple.clinic.allpatientsinfacility
...
-        .map { { view: AllPatientsInFacilityView -> view.showNoPatientsFound(it.facility!!.name) } }
+        .map { { view: AllPatientsInFacilityUi -> view.showNoPatientsFound(it.facility!!.name) } }
 
     val hasPatientsInFacilityUiChanges = queriedPatientsViewStates
         .filter { it.patients.isNotEmpty() }
-        .map { { view: AllPatientsInFacilityView -> view.showPatients(it.facility!!, it.patients) } }
+        .map { { ui: AllPatientsInFacilityUi -> ui.showPatients(it.facility!!, it.patients) } }
 
-    return Observable.merge(noPatientsInFacilityUiChanges, hasPatientsInFacilityUiChanges)
+    return Observable.merge(logUiStates, noPatientsInFacilityUiChanges, hasPatientsInFacilityUiChanges)
```
rakshakhegde: Can you move all these rx chain merges into their own functions like we do in controllers? This may just be my personal preference but it becomes easier to parse what each chunk does. For example, instead we could do:
```kotlin
return Observable.merge(
  logUiStates(uiStates),
  noPatientsInFacilityUiChanges(uiStates),
  hasPatientsInFacilityUiChanges(uiStates)
)
```

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityView.kt
```diff
@@ -1,9 +1,65 @@
...
+
+    patientsList.apply {
+      setHasFixedSize(true)
+      layoutManager = LinearLayoutManager(context, RecyclerView.VERTICAL, false)
+      adapter = searchResultsAdapter
+    }
+
+    bindUiToController(
+        ui = this,
+        events = Observable.just(ScreenCreated() as UiEvent),
```
rakshakhegde: We don't need this typecast to `UiEvent`, do we?

app/src/main/res/layout/view_allpatientsinfacility.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8"?>
+<org.simple.clinic.allpatientsinfacility.AllPatientsInFacilityView xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:tools="http://schemas.android.com/tools"
+  android:layout_width="match_parent"
+  android:layout_height="match_parent"
+  android:animateLayoutChanges="true">
```
rakshakhegde: I'm not sure if `animateLayoutChanges` is advisable here because of recyclerview view changes too. Need to confirm this.

app/src/main/res/layout/view_allpatientsinfacility.xml
```diff
@@ -0,0 +1,46 @@
...
+      android:layout_marginStart="@dimen/spacing_32"
+      android:layout_marginEnd="@dimen/spacing_32"/>
+
+  </LinearLayout>
+
+  <androidx.recyclerview.widget.RecyclerView
+    android:id="@+id/patientsList"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:visibility="gone"
```
rakshakhegde: Instead of setting visibility here, please ensure that the controller sets the initial state correctly. We should avoid setting visibility in the XML because that is state contained in XML now.

app/src/main/res/values/styles_v2.xml
```diff
@@ -184,10 +184,11 @@
   </style>
 
   <style name="Clinic.V2.TextAppearance.Body2Tag">
-    <item name="android:textSize">12sp</item>
+    <item name="android:textSize">14sp</item>
```
rakshakhegde: Were the designs updated for this? Please link to Zeplin.

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListItem.kt
```diff
@@ -0,0 +1,70 @@
...
+import kotlinx.android.synthetic.main.list_patient_search.*
+import org.simple.clinic.R
+import org.simple.clinic.facility.Facility
+import org.simple.clinic.patient.PatientSearchResult
+import org.simple.clinic.widgets.recyclerview.ViewHolderX
+
+sealed class AllPatientsInFacilityListItem {
+
+  companion object {
+    fun mapSearchResultsToListItems(
```
rakshakhegde: This is logic that should be part of `AllPatientsInFacilityUiChangeProducer` isn't it?

quality/lint.xml
```diff
@@ -11,6 +11,9 @@
     <ignore path="src/main/res/drawable/ic_patient_male.xml" />
     <ignore path="src/main/res/drawable/ic_patient_transgender.xml" />
 
+    <!-- TODO: Remove once the layout is used in the future PR -->
+    <ignore path="src/main/res/layout/view_allpatientsinfacility.xml" />
```
rakshakhegde: Let's clean up `lint.xml`? All the resources mentioned above this line seem to be used somewhere.

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListItem.kt
```diff
@@ -0,0 +1,70 @@
...
+      return when {
+        oldItem is FacilityHeader && newItem is FacilityHeader -> true
+        oldItem is SearchResult && newItem is SearchResult -> oldItem.patientSearchResult.uuid == newItem.patientSearchResult.uuid
+        else -> false
+      }
+    }
+
+    @SuppressLint("DiffUtilEquals")
+    override fun areContentsTheSame(oldItem: AllPatientsInFacilityListItem, newItem: AllPatientsInFacilityListItem): Boolean {
+      return when {
```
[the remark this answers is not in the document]
vinaysshenoy: It can! I'll do this change.

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListAdapter.kt
```diff
@@ -0,0 +1,29 @@
...
+      field = value
+      submitList(field)
+    }
+
+  override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolderX {
+    return ViewHolderX(LayoutInflater.from(parent.context).inflate(viewType, parent, false))
+  }
+
+  override fun onBindViewHolder(holder: ViewHolderX, position: Int) {
+    items[position].render(holder)
```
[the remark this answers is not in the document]
vinaysshenoy: I didn't know `ListAdapter` provided this api. I'll make this change. 👍

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityUiChangeProducer.kt
vinaysshenoy: Makes sense, I'll try this out.

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityView.kt
vinaysshenoy: We do, since we get a compilation error because it sees `Observable<ScreenCreated>` instead. This problem goes away if we merge multiple streams,  but if we have only one, we need to explicitly cast it.

app/src/main/res/layout/view_allpatientsinfacility.xml
vinaysshenoy: I am unsure if this problem can happen if `animateLayoutChanges` is set on the `RecyclerView` itself or if it can happen if the property is set on the parent itself.

app/src/main/res/layout/view_allpatientsinfacility.xml
vinaysshenoy: Makes sense, I'll do this.

app/src/main/res/values/styles_v2.xml
vinaysshenoy: Yup, they were updated. I can't link to a Zeplin component, but the screen is [HERE](https://zpl.io/agkOWA0). You should  be able to find the component from there.

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListItem.kt
vinaysshenoy: @ragunathjawahar and I had a discussion about whether this conversion of domain models to view models should happen inside the `UiChangeProducer` or the `View`. The outcome of this discussion was that since this is a presentation detail (that the domain models are being converted into list items which is specific to `RecyclerView`), they should not happen here and should be handled by the view itself.

That way, the presentation of these items can change without affecting the ui change producer. Does this make sense?

quality/lint.xml
vinaysshenoy: I *think* this will need to be removed in the next PR when the view actually gets embedded in the parent screens. I'll remove this and confirm.

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListItem.kt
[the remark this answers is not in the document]
vinaysshenoy: Although, we'll have to be careful whenever we add a new type that it's either a `data class` or an `object`. Since it's in the same file, it should be fine.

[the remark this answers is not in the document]
vinaysshenoy: Ignore the previous comment, my brain forgot for a moment how `==` works. 😅

quality/lint.xml
vinaysshenoy: Tested just now, can't remove it in this PR. Will remove in the final PR.

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityUiChangeProducer.kt
vinaysshenoy: Done.

vinaysshenoy:
@rakshakhegde PR feedback addressed.

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityView.kt
sanchita: You could do this instead:
`Observable.just<UiEvent>(ScreenCreated())`

vinaysshenoy: I'll try this out!

vinaysshenoy: This looks nicer. However, when this opened in the IDE, you might see the `<UiEvent>` part marked as unnecessary by the IDE and be tempted to remove it. This works fine as well.

However, when you try to compile or build via Gradle, this breaks with a type inference problem. The problem is that Kotlin 1.3.40 [ships with a new and improved type inference engine](https://blog.jetbrains.com/kotlin/2019/06/kotlin-1-3-40-released/) which is enabled by default **only** in the IDE and not in Gradle which causes this change in behaviour.

tl;dr Please disable the new type inference engine in the IDE if you've bumped your Kotlin version.

PR #705, opened 2019-06-25
Show App update dialog

sanchita:
Pivotol story -> https://www.pivotaltracker.com/n/projects/2184102/stories/163619300
There is one part of this story left. Creating Remote config for this feature.

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientsModule.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientsModule.kt
@@ -22,4 +22,10 @@ class PatientsModule {
   fun hasUserDismissedApprovedStatus(rxSharedPrefs: RxSharedPreferences): Preference<Boolean> {
     return rxSharedPrefs.getBoolean("approved_status_dismissed", false)
   }
+
+  @Provides
+  @Named("app_update_last_shown_at")
+  fun timestampForAppUpdateDialogShown(rxSharedPrefs: RxSharedPreferences): Preference<Instant> {
+    return rxSharedPrefs.getObject("app_update_last_shown_at", Instant.EPOCH, InstantRxPreferencesConverter())
+  }
 }
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
@@ -12,6 +12,8 @@ import io.reactivex.rxkotlin.withLatestFrom
 import io.reactivex.schedulers.Schedulers.io
 import org.simple.clinic.ReplayUntilScreenIsDestroyed
 import org.simple.clinic.ReportAnalyticsEvents
+import org.simple.clinic.appupdate.AppUpdateState.ShowAppUpdate
+import org.simple.clinic.appupdate.CheckAppUpdateAvailability
 import org.simple.clinic.patient.PatientConfig
 import org.simple.clinic.sync.DataSync
 import org.simple.clinic.user.User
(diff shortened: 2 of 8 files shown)


app/src/test/java/org/simple/clinic/home/patients/PatientsScreenControllerTest.kt
```diff
@@ -37,13 +37,15 @@ import org.simple.clinic.user.UserStatus.WaitingForApproval
...
+import org.simple.clinic.util.TestUtcClock
 import org.simple.clinic.util.toOptional
 import org.simple.clinic.widgets.ScreenCreated
 import org.simple.clinic.widgets.TheActivityLifecycle
 import org.simple.clinic.widgets.UiEvent
+import org.threeten.bp.Duration
 import org.threeten.bp.Instant
 import org.threeten.bp.temporal.ChronoUnit
-import java.lang.IllegalStateException
+import timber.log.Timber
```
vinaysshenoy: Leftover log?

app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
```diff
@@ -218,9 +222,13 @@ class PatientsScreenController @Inject constructor(
         .listen()
         .filter { it is ShowAppUpdate }
 
+    val now = LocalDate.now(utcClock)
+
     return Observables
         .combineLatest(screenCreated, availableUpdate)
         .map { (_, update) -> update }
+        .filter { appUpdateDialogShownAtPref.get().atZone(utcClock.zone).toLocalDate().isBefore(now) }
```
vinaysshenoy: Since the feature requires us to check calendar dates, we should be using the user clock to generate the `LocalDate` instead of the utc clock.

I tried to do this and clean up the chain a little bit.

```kotlin
return Observables
        .combineLatest(screenCreated, availableUpdate)
        .map { (_, update) -> update }
        .filter { hasADayPassedSinceAppUpdateWasLastShown() }
        .doOnNext { appUpdateDialogShownAtPref.set(Instant.now(utcClock)) }
        .map { Ui::showAppUpdateDialog }

private fun hasADayPassedSinceAppUpdateWasLastShown(): Boolean {
    val dateOnWhichAppUpdateWasLastShown = appUpdateDialogShownAtPref.get().toLocalDateAtZone(userClock.zone)

    return dateOnWhichAppUpdateWasLastShown.isBefore(today)
}
```

I also changed the saving of the app update shown at preference do a side effect block.

app/src/main/java/org/simple/clinic/home/patients/PatientsModule.kt
```diff
@@ -22,4 +22,10 @@ class PatientsModule {
   fun hasUserDismissedApprovedStatus(rxSharedPrefs: RxSharedPreferences): Preference<Boolean> {
     return rxSharedPrefs.getBoolean("approved_status_dismissed", false)
   }
+
+  @Provides
+  @Named("app_update_dialog_shown_at")
```
vinaysshenoy: This timestamp exposes a presentation detail that the nudging of app update is done via a dialog. I feel we can make this name a little more generic, something like `app_update_last_nudged_at`. What do you think?

sanchita: Done

app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
sanchita: Good point. I will change this.

app/src/test/java/org/simple/clinic/home/patients/PatientsScreenControllerTest.kt
sanchita: Removed

PR #707, opened 2019-06-27
Appointment status safeenumtypeadapter

rakshakhegde:
https://www.pivotaltracker.com/story/show/165815995


app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -2310,6 +2310,53 @@ class DatabaseMigrationAndroidTest {
...
+
+  @Test
+  fun migrate_userstatus_from_37_to_38() {
+    val db_v37 = helper.createDatabase(version = 37)
+
+    fun insertAppointment(uuid: String, oldStatus: String) {
+      db_v37.execSQL("""
+        INSERT OR REPLACE INTO `Appointment`
+          (`uuid`,`patientUuid`,`facilityUuid`,`scheduledDate`,`status`,`appointmentType`,`syncStatus`,`createdAt`,`updatedAt`)
+        VALUES (
```
sanchita: We don't use ` `` ` in SQL queries anymore.
Refer the official [SQLLite docs](https://www.sqlite.org/lang_keywords.html) for the legends.
`""` - for identifiers (like table and column names)
`''` - for literals (like values)

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -2310,6 +2310,53 @@ class DatabaseMigrationAndroidTest {
...
+    val uuid2 = UUID.randomUUID().toString()
+    insertAppointment(uuid = uuid2, oldStatus = "CANCELLED")
+
+    val uuid3 = UUID.randomUUID().toString()
+    insertAppointment(uuid = uuid3, oldStatus = "VISITED")
+
+    val db_v38 = helper.migrateTo(version = 38)
+
+    fun verifyNewAppointmentStatus(uuid: String, newStatus: String) {
+      val cursor = db_v38.query("""SELECT "status" FROM "Appointment" WHERE "uuid"='$uuid'""")
```
sanchita: You don't need to add `""` for the column and table names since they are already in `""" """`.

```
"""SELECT status FROM Appointment WHERE uuid='$uuid' """
```

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -2310,6 +2310,53 @@ class DatabaseMigrationAndroidTest {
...
   }
+
+  @Test
+  fun migrate_userstatus_from_37_to_38() {
+    val db_v37 = helper.createDatabase(version = 37)
+
+    fun insertAppointment(uuid: String, oldStatus: String) {
+      db_v37.execSQL("""
+        INSERT OR REPLACE INTO `Appointment`
+          (`uuid`,`patientUuid`,`facilityUuid`,`scheduledDate`,`status`,`appointmentType`,`syncStatus`,`createdAt`,`updatedAt`)
```
sanchita: We can also simplify this and add all the values.
We wont need to pass column names then.

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -30,9 +30,9 @@ import org.simple.clinic.medicalhistory.MedicalHistoryRepository
 import org.simple.clinic.medicalhistory.OngoingMedicalHistoryEntry
```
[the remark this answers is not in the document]
sanchita: Oh I get it now. Smart!
Although I am not sure if this an appropriate way to refactor.
Can you merge this commit with the next one though and refactor at once after you change enums to sealed class?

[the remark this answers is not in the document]
rakshakhegde: I renamed it to what it needs to be called when it becomes a sealed class so that it is extremely easy to refactor. So for example: `SCHEDULED` becomes `Scheduled` across the project and then I convert them to a sealed class.

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -2310,6 +2310,53 @@ class DatabaseMigrationAndroidTest {
     verifyNewStatus(uuid = uuid2, newStatus = "allowed")
     verifyNewStatus(uuid = uuid3, newStatus = "denied")
   }
+
+  @Test
+  fun migrate_userstatus_from_37_to_38() {
```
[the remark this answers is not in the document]
rakshakhegde: Whoops, that was a mistake. Done.

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
rakshakhegde: done

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
rakshakhegde: done

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
rakshakhegde: I'd like to keep these because then later when the model changes it will be easier to know what value corresponds to what key in an older migration.

sanchita: So for that we actually name the values same as the column name.
Although its ok for now.

PR #708, opened 2019-06-27
Hook up all patients in facility view to UI

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/166754208

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -1963,4 +1963,154 @@ class PatientRepositoryAndroidTest {
     savePatientWithBp(facilityUuid = facilityUuid, patientStatus = INACTIVE)
     verifyRecentPatientOrder(patient2, patient1, facilityUuid = facilityUuid)
   }
+
+  @Test
+  fun querying_all_patients_in_a_facility_should_return_only_patients_in_that_facility_ordered_by_name() {
+
+    fun recordPatientAtFacility(
+        patientName: String,
+        status: PatientStatus,
+        facility: Facility
+    ) {
+      val patientUuid = UUID.randomUUID()
+      val patientProfile = testData.patientProfile(patientUuid = patientUuid).let { patientProfile ->
+        patientProfile.copy(patient = patientProfile.patient.copy(status = status, fullName = patientName))
+      }
+      val bp = testData.bloodPressureMeasurement(patientUuid = patientUuid, facilityUuid = facility.uuid)
+
+      patientRepository.save(listOf(patientProfile)).blockingAwait()
+      bloodPressureRepository.save(listOf(bp)).blockingAwait()
+    }
+
+    // given
+    val (facilityA, facilityB, facilityC) = facilityRepository.facilities().blockingFirst()
+
+    recordPatientAtFacility(
+        "Chitra",
+        ACTIVE,
+        facilityA
+    )
+    recordPatientAtFacility(
+        "Anubhav",
+        ACTIVE,
+        facilityA
+    )
+    recordPatientAtFacility(
+        "Bhim",
+        DEAD,
+        facilityA
+    )
+    recordPatientAtFacility(
+        "Elvis",
+        DEAD,
+        facilityB
+    )
+    recordPatientAtFacility(
+        "Farhan",
+        DEAD,
+        facilityB
+    )
+    recordPatientAtFacility(
+        "Dhruv",
+        ACTIVE,
+        facilityB
+    )
+
+    // when
+    fun patientsInFacility(facility: Facility): List<PatientSearchResult> {
+      return patientRepository
+          .allPatientsInFacility(facility)
+          .blockingFirst()
+    }
+
+    val allPatientFullNamesInFacilityA = patientsInFacility(facilityA).map { it.fullName }
+    val allPatientFullNamesInFacilityB = patientsInFacility(facilityB).map { it.fullName }
+    val allPatientFullNamesInFacilityC = patientsInFacility(facilityC).map { it.fullName }
+
+    // then
+    assertThat(allPatientFullNamesInFacilityA)
+        .containsExactly("Anubhav", "Chitra")
+        .inOrder()
+
+    assertThat(allPatientFullNamesInFacilityB)
+        .containsExactly("Dhruv")
+        .inOrder()
+
+    assertThat(allPatientFullNamesInFacilityC).isEmpty()
+  }
+
+  @Test
+  fun querying_all_patients_in_a_facility_should_set_the_last_bp_to_the_last_recorded_one_within_any_facility() {
+
+    fun createPatient(patientUuid: UUID, patientName: String): PatientProfile {
+      val patientProfile = testData.patientProfile(patientUuid = patientUuid).let { patientProfile ->
+        patientProfile.copy(patient = patientProfile.patient.copy(status = ACTIVE, fullName = patientName))
+      }
+      patientRepository.save(listOf(patientProfile)).blockingAwait()
+
+      return patientProfile
+    }
+
+    fun recordBp(uuid: UUID, patientUuid: UUID, facilityUuid: UUID, recordedAt: Instant) {
+      val bp = testData.bloodPressureMeasurement(uuid = uuid, patientUuid = patientUuid, facilityUuid = facilityUuid, recordedAt = recordedAt)
+
+      bloodPressureRepository.save(listOf(bp)).blockingAwait()
+    }
+
+    // given
+    val (facilityA, facilityB) = facilityRepository.facilities().blockingFirst()
+    val now = Instant.parse("2019-06-26T00:00:00Z")
+    val oneSecondLater = now.plusSeconds(1)
+    val oneSecondEarlier = now.minusSeconds(1)
+    val fiveSecondsLater = now.plusSeconds(5)
+
+    val uuidOfPatientA = UUID.fromString("e2a49529-3b46-4c9e-a7bc-af3090f1ecbb")
+    val uuidOfBp1OfPatientA = UUID.fromString("588b0d8c-6dce-4664-863f-8c79c06c6c3c")
+    val uuidOfBp2OfPatientA = UUID.fromString("ba875b98-4ca0-4097-bbc7-56041b377cf2")
+    val uuidOfBp3OfPatientA = UUID.fromString("4b92601e-4a86-40de-b6b7-707f9122a81b")
+    createPatient(patientUuid = uuidOfPatientA, patientName = "Patient with latest BP in Facility A")
+    recordBp(uuid = uuidOfBp1OfPatientA, patientUuid = uuidOfPatientA, facilityUuid = facilityA.uuid, recordedAt = oneSecondLater)
+    recordBp(uuid = uuidOfBp2OfPatientA, patientUuid = uuidOfPatientA, facilityUuid = facilityB.uuid, recordedAt = oneSecondEarlier)
+    recordBp(uuid = uuidOfBp3OfPatientA, patientUuid = uuidOfPatientA, facilityUuid = facilityA.uuid, recordedAt = now)
+
+    val uuidOfPatientB = UUID.fromString("7925e13f-3b04-46b0-b685-7005ebb1b6fd")
+    val uuidOfBp1OfPatientB = UUID.fromString("00e39456-f5e4-4538-956d-7e973ec5da88")
+    val uuidOfBp2OfPatientB = UUID.fromString("b72e28cd-1b46-4c72-84cd-58530c2829e3")
+    val uuidOfBp3OfPatientB = UUID.fromString("c18b41d5-260e-4a26-8030-45c621ded98d")
+    createPatient(patientUuid = uuidOfPatientB, patientName = "Patient with latest BP in Facility B")
+    recordBp(uuid = uuidOfBp1OfPatientB, patientUuid = uuidOfPatientB, facilityUuid = facilityB.uuid, recordedAt = fiveSecondsLater)
+    recordBp(uuid = uuidOfBp2OfPatientB, patientUuid = uuidOfPatientB, facilityUuid = facilityA.uuid, recordedAt = oneSecondEarlier)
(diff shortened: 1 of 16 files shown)


vinaysshenoy:
The last part of the feature is pending. But it's a tiny piece because the only thing we need to do is embed this view inside the `AddIdToPatientSearchScreen` and toggle the search button and list on typing. This will be done in another PR.

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListAdapter.kt
```diff
@@ -2,19 +2,26 @@ package org.simple.clinic.allpatientsinfacility
...
+
+  private val eventSubject = PublishSubject.create<AllPatientsInFacilityListItem.Event>()
+
+  val listItemEvents: Observable<AllPatientsInFacilityListItem.Event> = eventSubject.hide()
 
   override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolderX {
     return ViewHolderX(LayoutInflater.from(parent.context).inflate(viewType, parent, false))
   }
 
   override fun onBindViewHolder(holder: ViewHolderX, position: Int) {
```
sanchita: Why is this named `ViewHolderX`?

vinaysshenoy: Just a convention for Kotlin Android Extensions. Package names for kotlin android extensions related code all start with `kotlinx`, and this class is meant to be used for viewholders which need access to synthetic accessors, so `VIewHolderX`.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -1932,5 +1932,169 @@ class PatientRepositoryAndroidTest {
...
+            facilityB
+        ),
+        recordPatientAtFacility(
+            UUID.fromString("51d7785c-dfe1-4018-8207-e74cd2377f5d"),
+            "Patient F",
+            PatientStatus.DEAD,
+            facilityB
+        ),
+        recordPatientAtFacility(
+            UUID.fromString("072df9ac-6a00-462f-b87e-1a9ca3b24c1b"),
```
sanchita: We are not using this generated UUID to verify the patient itself.
Why do we need to pass these to `recordPatientAtFacility` and not use the auto-generated UUID in the `TestData` function itself?

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -1932,5 +1932,169 @@ class PatientRepositoryAndroidTest {
...
+            "Patient D",
```
sanchita: We are using `fullName` to determine the order.
This is not very obvious reading just the test.
Can you use better names instead like say `OrderByNameA`?
Or it would be a better idea to test this separately and then name the test properly.
Whichever suits you.

sanchita: Also I see that the assertions are happening on the position on the patient in the list rather than the `fullName` itself.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -1932,5 +1932,169 @@ class PatientRepositoryAndroidTest {
...
+
+    assertThat(allPatientsInFacilityB)
+        .containsExactly(patientsInFacilityB[2].patient.uuid)
+        .inOrder()
+
+    assertThat(allPatientsInFacilityC).isEmpty()
+  }
+
+  @Test
+  fun querying_all_patients_in_a_facility_with_bps_in_different_facilities_should_set_the_last_bp_to_the_latest_one() {
```
sanchita: Rephrase this to `querying_all_patients_in_a_facility_should_set_the_last_bp_to_the_last_recorded_one_within_any_facility`

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -1932,5 +1932,169 @@ class PatientRepositoryAndroidTest {
...
+
+    fun recordPatientWithBps(
```
sanchita: This function is doing two separate operations.
Can you split this in two?
One function should create the profile whereas the other could record Bps.

app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
```diff
@@ -90,6 +90,18 @@ data class PatientSearchResult(
...
+    @Query("""
+      SELECT * FROM ($mainQuery) AllSearchResults
+      INNER JOIN (
+        SELECT DISTINCT P.uuid FROM Patient P
+          INNER JOIN BloodPressureMeasurement BP ON BP.patientUuid = P.uuid
+          WHERE BP.deletedAt IS NULL AND P.status = :status AND BP.facilityUuid = :facilityUuid
+      ) PatientsAtFacility ON AllSearchResults.uuid = PatientsAtFacility.uuid
+      GROUP BY AllSearchResults.uuid ORDER BY AllSearchResults.fullName
+    """)
+    fun searchInFacilitySortedByName(facilityUuid: UUID, status: PatientStatus): Flowable<List<PatientSearchResult>>
```
sanchita: Rename this to `searchInFacilityAndSortByName`

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -1932,5 +1932,169 @@ class PatientRepositoryAndroidTest {
...
+      }
+
+      patientRepository.save(listOf(patientProfile)).blockingAwait()
+      bloodPressureRepository.save(bps).blockingAwait()
+
+      return patientProfile
+    }
+
+    fun allPatientIdsAndLastBpRecordedAtFacility(facility: Facility): List<Pair<UUID, Instant>> {
+      return patientRepository.allPatientsInFacility(facility).blockingFirst().map { it.uuid to it.lastBp!!.takenOn }
```
sanchita: More than two Rx operators are being chained.
Please format each operator to a separate line.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -1932,5 +1932,169 @@ class PatientRepositoryAndroidTest {
...
+    assertThat(allPatientsInFacilityC).isEmpty()
+  }
+
+  @Test
+  fun querying_all_patients_in_a_facility_with_bps_in_different_facilities_should_set_the_last_bp_to_the_latest_one() {
+
+    fun recordPatientWithBps(
+        patientUuid: UUID,
+        patientName: String,
+        recordBpsAt: List<Triple<UUID, Facility, Instant>>
```
sanchita: Using a `Triple` here makes it difficult to understand that what needs to be passed in this.
Can you instead use three function arguments?
This will also become concise when you split the function.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -1932,5 +1932,169 @@ class PatientRepositoryAndroidTest {
...
+        "Patient B",
+        listOf(
+            Triple(bpUuid1ForPatientB, facilityB, now.plusSeconds(5)),
+            Triple(bpUuid2ForPatientB, facilityA, now.minusSeconds(1)),
+            Triple(bpUuid3ForPatientB, facilityB, now)
+        )
+    )
+
+    // when
+    val allPatientsInFacilityA = allPatientIdsAndLastBpRecordedAtFacility(facilityA)
```
sanchita: Why do we need to create a `Pair` here?
Again this makes it difficult to read and understand what we want to verify.
In the assertions, can't we directly verify ids and bp recordedAt for `allPatientsInFacility`

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -1932,5 +1932,169 @@ class PatientRepositoryAndroidTest {
...
+        listOf(
+            Triple(bpUuid1ForPatientA, facilityA, now.plusSeconds(1)),
+            Triple(bpUuid2ForPatientA, facilityB, now.minusSeconds(1)),
+            Triple(bpUuid3ForPatientA, facilityA, now)
+        )
+    )
+
+    val bpUuid1ForPatientB = UUID.fromString("00e39456-f5e4-4538-956d-7e973ec5da88")
+    val bpUuid2ForPatientB = UUID.fromString("b72e28cd-1b46-4c72-84cd-58530c2829e3")
+    val bpUuid3ForPatientB = UUID.fromString("c18b41d5-260e-4a26-8030-45c621ded98d")
```
sanchita: I dont see any of these generated Ids being used further in the method.
We can leave Ids to be auto-generated.
Same for `Patient A` as well.

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityListItem.kt
```diff
@@ -43,11 +50,19 @@ sealed class AllPatientsInFacilityListItem {
...
+        holder: ViewHolderX,
+        eventSubject: Subject<Event>
+    ) {
       holder.patientSearchResultView.render(patientSearchResult, facility)
+      holder.itemView.setOnClickListener { eventSubject.onNext(Event.SearchResultClicked(patientSearchResult)) }
     }
   }
 
+  sealed class Event {
+    data class SearchResultClicked(val patientSearchResult: PatientSearchResult): AllPatientsInFacilityListItem.Event()
```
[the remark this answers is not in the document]
vinaysshenoy: @ragunathjawahar and I talked about removing the `UiEvent` interface and having individual event types per feature. This is a little hard to do for other screens at the moment, but we could start doing this for events for new features.

app/src/main/java/org/simple/clinic/search/PatientSearchScreenController.kt
```diff
@@ -114,4 +100,34 @@ class PatientSearchScreenController @Inject constructor() : ObservableTransforme
         .map { it.patientUuid }
         .map { patientUuid -> { ui: Ui -> ui.openPatientSummary(patientUuid) } }
   }
+
+  private fun toggleAllPatientsVisibility(events: Observable<UiEvent>): ObservableSource<UiChange> {
+    return events
+        .ofType<SearchQueryNameChanged>()
+        .map { it.name.isBlank().not() }
```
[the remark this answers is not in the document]
vinaysshenoy: Good point, I'll do this.

app/src/main/java/org/simple/clinic/search/PatientSearchScreenController.kt
```diff
@@ -114,4 +100,34 @@ class PatientSearchScreenController @Inject constructor() : ObservableTransforme
...
+              ui.showAllPatientsInFacility()
+            }
+          }
+        }
+  }
+
+  private fun toggleSearchButtonVisibility(events: Observable<UiEvent>): ObservableSource<UiChange> {
+    return events
+        .ofType<SearchQueryNameChanged>()
+        .map { it.name.isBlank().not() }
```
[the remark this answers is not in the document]
vinaysshenoy: Good point, I'll do this.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
vinaysshenoy: In a nutshell, we are planning to reduce the amount of variability in our tests. Part of doing this is reducing the randomly generated data in our tests and statically generate constant data in the test class itself. We can explain this in more detail next week.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
vinaysshenoy: Good point, I'll look at making this test more understandable.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
vinaysshenoy: Sure, I'll do this.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
vinaysshenoy: Right. I'll do this when splitting the functions.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
vinaysshenoy: We could. I'll look at making this test more understandable.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
vinaysshenoy: See above comment about reducing the amount of random data in our tests. I think @ragunathjawahar will be able to explain this decision in a better way than I can.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
vinaysshenoy: I tried to make it more clear what is happening. Let's go over this again in a bit and see if it makes better sense now?

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
vinaysshenoy: Take a look now? I've rewritten the test to make it (hopefully) more understandable.

PR #709, opened 2019-06-28
Add remote config for showing app update

sanchita:
https://www.pivotaltracker.com/story/show/163619300

### DIFF
--- a/app/src/main/java/org/simple/clinic/appupdate/AppUpdateConfig.kt
+++ b/app/src/main/java/org/simple/clinic/appupdate/AppUpdateConfig.kt
@@ -1,5 +1,6 @@
 package org.simple.clinic.appupdate
 
+import io.reactivex.Observable
 import org.simple.clinic.remoteconfig.ConfigReader
 
 data class AppUpdateConfig(
@@ -8,10 +9,12 @@ data class AppUpdateConfig(
 ) {
 
   companion object {
-    fun read(reader: ConfigReader): AppUpdateConfig {
-      return AppUpdateConfig(
-          inAppUpdateEnabled = reader.boolean(name = "appupdate_enabled", default = false),
-          differenceBetweenVersionsToNudge = reader.long(name = "appupdate_version_diff_to_nudge", default = 1000))
+    fun read(reader: ConfigReader): Observable<AppUpdateConfig> {
+      return Observable.fromCallable {
+        AppUpdateConfig(
+            inAppUpdateEnabled = reader.boolean(name = "appupdate_enabled", default = false),
+            differenceBetweenVersionsToNudge = reader.long(name = "appupdate_version_diff_to_nudge", default = 1000))
+      }
     }
   }
 }
--- a/app/src/main/java/org/simple/clinic/appupdate/AppUpdateConfig.kt
+++ b/app/src/main/java/org/simple/clinic/appupdate/AppUpdateConfig.kt
@@ -1,6 +1,17 @@
 package org.simple.clinic.appupdate
 
+import org.simple.clinic.remoteconfig.ConfigReader
+
 data class AppUpdateConfig(
     val inAppUpdateEnabled: Boolean,
     val differenceBetweenVersionsToNudge: Long
-)
+) {
+
+  companion object {
+    fun read(reader: ConfigReader): AppUpdateConfig {
+      return AppUpdateConfig(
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/appupdate/AppUpdateConfig.kt
```diff
@@ -1,6 +1,17 @@
...
+import org.simple.clinic.remoteconfig.ConfigReader
+
 data class AppUpdateConfig(
     val inAppUpdateEnabled: Boolean,
     val differenceBetweenVersionsToNudge: Long
-)
+) {
+
+  companion object {
+    fun read(reader: ConfigReader): AppUpdateConfig {
```
vinaysshenoy: Reading from Firebase config involves disk I/O, so we should inject an `Observable<AppUpdatedConfig>` and use `Observable.fromCallable {}` to read the config instead.

See `PatientConfig#read(ConfigReader)`.

sanchita: Done!

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
```diff
@@ -11,35 +12,64 @@ import io.reactivex.Observable
...
+  @VisibleForTesting(otherwise = PRIVATE)
+  fun shouldNudgeForUpdate(appUpdateInfo: AppUpdateInfo): Observable<AppUpdateState> {
+    val checkUpdate = config
+        .map { checkForUpdate(appUpdateInfo, it) }
+
+    val shouldShow = checkUpdate
+        .filter { showUpdate -> showUpdate }
+        .map { ShowAppUpdate }
+
+    val doNotShow = config
```
vinaysshenoy: We can replace 
```kotlin
config
  .map { checkForUpdate(appUpdateInfo, it) }
```

with `checkUpdate`, right?

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
```diff
@@ -11,35 +12,64 @@ import io.reactivex.Observable
...
 import javax.inject.Inject
 
 class CheckAppUpdateAvailability @Inject constructor(
     private val appContext: Application,
-    private val config: AppUpdateConfig,
+    private val config: Observable<AppUpdateConfig>,
     private val versionUpdateCheck: (Int, Application, AppUpdateConfig) -> Boolean = isVersionApplicableForUpdate
 ) {
 
   fun listen(): Observable<AppUpdateState> {
```
vinaysshenoy: Since `appUpdateCallback()` is a private API, I feel wrapping the result in a sealed class to handle only an exception is  overkill. We can change `appUpdateCallback` to return an `Observable<AppUpdateInfo>` and then add an `onErrorReturn` to the `listen()` chain to handle this without changing the public API.

```kotlin
  fun listen(): Observable<AppUpdateState> {
    return appUpdateCallback()
        .flatMap(this::shouldNudgeForUpdate)
        .onErrorReturn(::AppUpdateStateError)
  }

  private fun appUpdateCallback(): Observable<AppUpdateInfo> {
    val appUpdateManager = AppUpdateManagerFactory.create(appContext)
    val appUpdateInfoTask = appUpdateManager.appUpdateInfo

    return Observable.create { emitter ->
      appUpdateInfoTask.addOnSuccessListener { appUpdateInfo ->
        emitter.onNext(appUpdateInfo)
      }

      appUpdateInfoTask.addOnFailureListener { exception ->
        emitter.onError(exception)
      }
    }
  }
```

What do you think?

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
sanchita: Missed this.

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
sanchita: I felt the same about using `sealed class` here but I didn't think of `onErrorReturn` as an option. Thanks!
Will make this change.

PR #713, opened 2019-07-03
Overdue bug fix

sanchita:
https://www.pivotaltracker.com/story/show/165275614

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 ### Internal
 - Change `LoggedInUser#status` to be a forward-compatible enum
 
+### Fixes
+- Fix overdue list to show last recorded BP, instead of last updated BP
+
 ## 2019-07-01-4522
 ### Feature
 - Show "Last Seen" time of patient in recent patients list instead of "Last BP"
--- a/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
@@ -3,6 +3,7 @@ package org.simple.clinic.overdue
 import androidx.test.runner.AndroidJUnit4
 import com.google.common.truth.Truth.assertThat
 import io.bloco.faker.Faker
+import io.reactivex.Single
 import org.junit.After
 import org.junit.Before
 import org.junit.Rule
@@ -50,6 +51,7 @@ import org.threeten.bp.Duration
 import org.threeten.bp.Instant
 import org.threeten.bp.LocalDate
 import org.threeten.bp.LocalDateTime
+import org.threeten.bp.temporal.ChronoUnit
 import java.util.UUID
 import javax.inject.Inject
 
@@ -805,7 +807,7 @@ class AppointmentRepositoryAndroidTest {
   }
 
   @Test
-  fun marking_appointment_older_than_current_date_should_work_correctly() {
+  fun marking_appointment_older_than_current_date_as_visited_should_work_correctly() {
     val patientId = UUID.randomUUID()
     val appointmentUuid1 = UUID.randomUUID()
     val appointmentUuid2 = UUID.randomUUID()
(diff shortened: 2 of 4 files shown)


app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -875,4 +877,77 @@ class AppointmentRepositoryAndroidTest {
...
+        recordedAt = now.plus(10, ChronoUnit.MINUTES),
+        updatedAt = now
+    )
+
+    bpRepository.save(listOf(bpOneForPatient1, bpTwoForPatient1, bpOneForPatient2, bpTwoForPatient2)).blockingAwait()
+
+    scheduleAppointment(patient1).blockingGet()
+    scheduleAppointment(patient2).blockingGet()
+
+    val appointments = appointmentRepository.overdueAppointments(facility).blockingFirst().associateBy { it.fullName }
```
rakshakhegde: We can simplify this test without the `Map` like this?
```kotlin
    val appointments = appointmentRepository.overdueAppointments(facility).blockingFirst()

    assertThat(appointments.size).isEqualTo(2)
    assertThat(appointments[0].bloodPressure).isEqualTo(bpTwoForPatient2)
    assertThat(appointments[1].bloodPressure).isEqualTo(bpOneForPatient1)
```

app/src/main/java/org/simple/clinic/home/overdue/OverdueScreenController.kt
```diff
@@ -72,7 +72,7 @@ class OverdueScreenController @Inject constructor(
                 phoneNumber = it.phoneNumber?.number,
                 bpSystolic = it.bloodPressure.systolic,
                 bpDiastolic = it.bloodPressure.diastolic,
-                bpDaysAgo = calculateDaysAgoFromInstant(it.bloodPressure.updatedAt),
+                bpDaysAgo = calculateDaysAgoFromInstant(it.bloodPressure.recordedAt),
```
rakshakhegde: Let's add one associated controller test for this change?

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
sanchita: We cannot determine the order of appointments received in `appointmentRepository.overdueAppointments(facility)` since there are multiple factors that affect this order.
Also, the reason for using `fullName` is so that the test case is more readable.

app/src/main/java/org/simple/clinic/home/overdue/OverdueScreenController.kt
sanchita: `OverdueListItem` is a UI related implementation detail which there is no point writing a unit test for. 
You can actually run a UI test on this instead.

rakshakhegde: Yes we could add a UI test. Can we also not add a unit test that ensures that the controller calculates the `calculateDaysAgoFromInstant` using the `recordedAt` value instead of `updatedAt`? By having a recordedAt value that pre-dates updatedAt for instance.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 ### Internal
 - Change `LoggedInUser#status` to be a forward-compatible enum
 
+### Fixes
+- Fix overdue list showing last updated BP, instead of last recorded BP
```
rakshakhegde: Can we name this something similar to how we do in the commits. How about:
`Fix overdue list to show last recorded BP instead of last updated BP`

rakshakhegde:
@sanchita Also, there's one failing android test on bitrise. I see a 500 thrown. It must get fixed when you push again.

app/src/main/java/org/simple/clinic/home/overdue/OverdueScreenController.kt
sanchita: This is still a presentation detail.
I don't find any value adding an unnecessary test.

CHANGELOG.md
sanchita: I will do this.

app/src/main/java/org/simple/clinic/home/overdue/OverdueScreenController.kt
sanchita: Let me know if this makes sense. We can get on a call if you want to discuss this further.

rakshakhegde: Can you please link me to the test that verifies that "`recordedAt` is used to fetch overdue patients and not `updatedAt`"?

sanchita: This is part of `AppointmentRepositoryAndroidTest`

PR #715, opened 2019-07-04
Fix possible crash while checking app update availability

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/167088502


rakshakhegde:
LGTM. I've only updated code to change to boolean `.not()` operator.

PR #725, opened 2019-07-15
Fix: Patient details sometimes does not get filled in phone mask sheet

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/167284393

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/phonemask/PhoneMaskBottomSheetController.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/phonemask/PhoneMaskBottomSheetController.kt
@@ -72,7 +72,7 @@ class PhoneMaskBottomSheetController @Inject constructor(
     return PatientDetails(
         phoneNumber = phoneNumber.number,
         name = patient.fullName,
-        genderLetterRes = patient.gender.displayLetterRes,
+        gender = patient.gender,
         age = ageValue(patient)
     )
   }
--- a/app/src/main/java/org/simple/clinic/home/overdue/phonemask/PhoneMaskBottomSheetController.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/phonemask/PhoneMaskBottomSheetController.kt
@@ -3,11 +3,13 @@ package org.simple.clinic.home.overdue.phonemask
 import io.reactivex.Observable
 import io.reactivex.ObservableSource
 import io.reactivex.ObservableTransformer
+import io.reactivex.rxkotlin.Observables
 import io.reactivex.rxkotlin.ofType
 import io.reactivex.rxkotlin.withLatestFrom
 import org.simple.clinic.ReplayUntilScreenIsDestroyed
 import org.simple.clinic.ReportAnalyticsEvents
 import org.simple.clinic.patient.Patient
+import org.simple.clinic.patient.PatientPhoneNumber
 import org.simple.clinic.patient.PatientRepository
 import org.simple.clinic.patient.displayLetterRes
 import org.simple.clinic.phone.Dialer.Automatic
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/home/overdue/phonemask/PhoneMaskBottomSheetController.kt
```diff
@@ -45,31 +47,43 @@ class PhoneMaskBottomSheetController @Inject constructor(
...
-
-  private fun ageValue(patient: Patient): Int =
-      if (patient.dateOfBirth == null) {
-        val age = patient.age!!
-        estimateCurrentAge(age.value, age.updatedAt, clock)
-      } else {
-        estimateCurrentAge(patient.dateOfBirth, clock)
-      }
+    val patientStream = patientUuidStream
+        .switchMap { patientUuid -> patientRepository.patient(patientUuid) }
```
sanchita: Do we need a `switchMap` here?
`PhoneMaskBottomSheetCreated` won't be called more than once here right?

app/src/main/java/org/simple/clinic/home/overdue/phonemask/PhoneMaskBottomSheetController.kt
```diff
@@ -45,31 +47,43 @@ class PhoneMaskBottomSheetController @Inject constructor(
...
+        .filterAndUnwrapJust()
+
+    val patientPhoneNumberStream = patientUuidStream
+        .switchMap { patientUuid -> patientRepository.phoneNumber(patientUuid) }
+        .filterAndUnwrapJust()
+
+    return Observables.combineLatest(patientStream, patientPhoneNumberStream)
+        .map { (patient, phoneNumber) -> patientDetails(patient, phoneNumber) }
+        .map { patient ->
```
sanchita: Rename this to `patientDetails` to avoid confusion.

app/src/main/java/org/simple/clinic/home/overdue/phonemask/PhoneMaskBottomSheetController.kt
```diff
@@ -3,11 +3,13 @@ package org.simple.clinic.home.overdue.phonemask
 import io.reactivex.Observable
```
sanchita: Either remove the following function in this class and inline it in `patientPhoneNumberStream()` as its not being reused anymore,

```
 private fun sheetCreated(events: Observable<UiEvent>) =
events.ofType<PhoneMaskBottomSheetCreated>()
```

or reuse it in `setupView()`.

app/src/test/java/org/simple/clinic/home/overdue/phonemask/PhoneMaskBottomSheetControllerTest.kt
```diff
@@ -198,4 +203,45 @@ class PhoneMaskBottomSheetControllerTest {
...
           listOf(SecureCallClicked, NEVER_ASK_AGAIN, Manual)
       )
+
+  @Test
+  fun `when the screen is created, the patient details must be populated`() {
+    // given
+    val patientSubject = PublishSubject.create<Optional<Patient>>()
+    val patientPhoneNumberSubject = PublishSubject.create<Optional<PatientPhoneNumber>>()
+    whenever(patientRepository.patient(patientUuid)).thenReturn(patientSubject)
+    whenever(patientRepository.phoneNumber(patientUuid)).thenReturn(patientPhoneNumberSubject)
```
sanchita: Both the methods `patientRepository.patient(patientUuid)` and `patientRepository.phoneNumber(patientUuid)` return an `Optional`, this means that either one of them could give out a `None`.
Can we add test to tackle that as well?

sanchita: Also what should be the expected behavior in that case?

app/src/main/java/org/simple/clinic/home/overdue/phonemask/PhoneMaskBottomSheetController.kt
vinaysshenoy: You're right, I'll change this.

app/src/test/java/org/simple/clinic/home/overdue/phonemask/PhoneMaskBottomSheetControllerTest.kt
vinaysshenoy: In this screen, it is not possible for this to happen during normal flows because overdue appointments will never load when the patient and phone number do not exist. I'll change this `setupView` function to use `unwrapJust` instead of `filterAndUnwrapJust` because this is an abnormal state and the user should not be allowed to continue.

PR #727, opened 2019-07-17
Fix: Camera does not get released properly when scanning QR codes

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/167217634


app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -13,26 +14,18 @@ import com.budiyev.android.codescanner.ScanMode
...
 import org.simple.clinic.activity.TheActivity
-import org.simple.clinic.bindUiToController
-import org.simple.clinic.widgets.ScreenCreated
 import org.simple.clinic.widgets.ScreenDestroyed
 import org.simple.clinic.widgets.TheActivityLifecycle
-import org.simple.clinic.widgets.UiEvent
 import timber.log.Timber
 import javax.inject.Inject
 
 class QrCodeScannerView(context: Context, attrs: AttributeSet) : FrameLayout(context, attrs) {
```
rakshakhegde: Can we extend `CodeScannerView` instead of using a FrameLayout and then adding this view?

vinaysshenoy: We could. I'll do this.

vinaysshenoy: Never mind, I tried doing it and the library has made `CodeScannerView` a `final` class, so we can't inherit from it.

PR #735, opened 2019-07-22
Add phone number search support to patient repository

vinaysshenoy:
Pivotal story link: https://www.pivotaltracker.com/story/show/167343626

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -57,6 +57,7 @@ import org.threeten.bp.Duration
 import org.threeten.bp.Instant
 import org.threeten.bp.LocalDate
 import org.threeten.bp.temporal.ChronoUnit
+import java.util.Locale
 import java.util.UUID
 import javax.inject.Inject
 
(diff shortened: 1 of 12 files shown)


app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -1936,4 +1937,145 @@ class PatientRepositoryAndroidTest {
...
+    createPatientWithPhoneNumber(patientName = "esther d' souza", phoneNumber = "1234567890")
+    createPatientWithPhoneNumber(patientName = "Dhruv Saxena", phoneNumber = "1234567890")
+    createPatientWithPhoneNumber(patientName = "Dhruv Acharya", phoneNumber = "1234567890")
+
+    // when
+    val searchResults = searchResults(phoneNumber = "1234567890")
+        // Converting to lowercase because of different SQLite implementations in
+        // platforms causes flaky tests because the ordering of small and capital
+        // strings which are identical is platform-dependent.
+        .map { it.toLowerCase(Locale.ENGLISH) }
```
sanchita: Just a doubt.
Does this mean that the search SQL query can return different results based on the platform?

app/src/main/java/org/simple/clinic/patient/PatientSearchCriteria.kt
```diff
@@ -0,0 +1,8 @@
+package org.simple.clinic.patient
+
+sealed class PatientSearchCriteria {
+
+  data class ByName(val patientName: String): PatientSearchCriteria()
```
sanchita: 'Criteria' already covers what this sealed class is.
I think we can rename this to just `Name` and `PhoneNumber`.
So it means that search criteria is name or phone number.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -56,6 +56,20 @@ class PatientRepository @Inject constructor(
 
   private var ongoingNewPatientEntry: OngoingNewPatientEntry = OngoingNewPatientEntry()
 
+  fun search(criteria: PatientSearchCriteria): Observable<List<PatientSearchResult>> {
+    return when(criteria) {
+      is PatientSearchCriteria.ByName -> search(criteria.patientName)
+      is PatientSearchCriteria.ByPhoneNumber -> search(criteria.phoneNumber)
```
sanchita: `PatientSearchCriteria` only communicates what type of search should be performed on a particular `search string`. That is also synonymous with the way our search looks. We decide criteria based on input.

Passing the actual `search string` to `PatientSearchCriteria` types doesn't encapsulate information about that sealed class type itself. The `search string` can exist outside the criteria so as to make it more generic.

`search(searchString, criteria)`

Another way to do this could be to enforce to implement a search mechanism for each `PatientSearchCriteria` using an `abstract` method.
Wdyt?

sanchita: I also realised that if we use the second technique we can eliminate the need to add `when` blocks everywhere.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -57,20 +59,13 @@ class PatientRepository @Inject constructor(
   private var ongoingNewPatientEntry: OngoingNewPatientEntry = OngoingNewPatientEntry()
 
   fun search(criteria: PatientSearchCriteria): Observable<List<PatientSearchResult>> {
```
sanchita: Do you think now would be a good time to move search entirely out of `PatientRepository` as this does not need to be `SynceableRepository` or a repository at all?

sanchita: This change does not need to be a part of this PR.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
vinaysshenoy: The default ordering of results in SQL is undefined. However, the part which I am talking about is if we have cases like this where we explicitly order by name and ignore the case and there are names which are exactly the same but in different cases (like "Abhinav" and "abhinav").

Different runs of the test will sometimes return ["Abhinav", "abhinav"] and sometimes ["abhinav", "Abhinav"] and causes flakiness. So I explicitly lowercased the results in the tests.

app/src/main/java/org/simple/clinic/patient/PatientSearchCriteria.kt
vinaysshenoy: I'll make this change.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: The problem with moving the search string out of the search criteria is that future types of search criteria which might need more information than just a string will no longer be supported. For example, if there's a need to search with a name and colony, then we can just change the `ByName` criteria to `ByNameAndColony` without affecting anything else. It is just an implementation detail at the moment that both the different search criteria take in only a string as a parameter, which is evident from the different names for the parameters.

You do have a great point about having different classes which implement different search mechanisms based on the criteria, though. I will do this in a different PR because I had to temporarily disable performance time tracking for the search by name flow which I need to fix first.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: Good point, I will move the search capability out of the `PatientRepository` in another PR as discussed earlier.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
sanchita: And we don't care about this order to be the same everywhere?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
sanchita: > For example, if there's a need to search with a name and colony, then we can just change the ByName criteria to ByNameAndColony without affecting anything else.

Makes sense.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
vinaysshenoy: We can't force it. I think it's ok for a user to see ["Abhinav", "abhinav"] in one run and ["Abhinav, "abhinav"] in another, because the solution to make sure it always remains the same order is a lot more effort than the benefit, IMO.

PR #737, opened 2019-07-23
Remove deprecated validate function from UserInputDateValidator

ragunathjawahar:
https://www.pivotaltracker.com/story/show/167453590

### DIFF
--- a/app/src/test/java/org/simple/clinic/editpatient/PatientEditScreenControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/editpatient/PatientEditScreenControllerTest.kt
@@ -54,9 +54,9 @@ import org.simple.clinic.widgets.ageanddateofbirth.DateOfBirthAndAgeVisibility.A
 import org.simple.clinic.widgets.ageanddateofbirth.DateOfBirthAndAgeVisibility.BOTH_VISIBLE
 import org.simple.clinic.widgets.ageanddateofbirth.DateOfBirthAndAgeVisibility.DATE_OF_BIRTH_VISIBLE
 import org.simple.clinic.widgets.ageanddateofbirth.UserInputDateValidator
-import org.simple.clinic.widgets.ageanddateofbirth.UserInputDateValidator.Result2.Invalid.DateIsInFuture
-import org.simple.clinic.widgets.ageanddateofbirth.UserInputDateValidator.Result2.Invalid.InvalidPattern
-import org.simple.clinic.widgets.ageanddateofbirth.UserInputDateValidator.Result2.Valid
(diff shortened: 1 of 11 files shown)


app/src/test/java/org/simple/clinic/editpatient/PatientEditScreenControllerTest.kt
```diff
@@ -415,6 +415,13 @@ class PatientEditScreenControllerTest {
     verify(screen).showValidationErrors(setOf(expectedError))
   }
 
+  fun `params for date of birth should be validated`(): List<List<Any>> {
+    return listOf(
+        listOf("01/01/2000", InvalidPattern, INVALID_DATE_OF_BIRTH),
```
vinaysshenoy: Something I like to do whenever I need to add a method source for a parameterized test is to add a local `testCase` function so I can construct the list of parameters in a typed fashion and expose the intention using named parameters.

```kotlin
fun `params for date of birth should be validated`(): List<List<Any>> {
    fun testCase(
        dateOfBirth: String,
        validationResult: UserInputDateValidator.Result,
        expectedError: PatientEditValidationError
    ): List<Any> {
      return listOf(dateOfBirth, validationResult, expectedError)
    }

    return listOf(
        testCase(dateOfBirth = "01/01/2000", validationResult = InvalidPattern, expectedError = INVALID_DATE_OF_BIRTH),
        testCase(dateOfBirth = "01/01/2000", validationResult = DateIsInFuture, expectedError = DATE_OF_BIRTH_IN_FUTURE)
    )
}
```

IMO, this makes the test more maintainable. I'm not asking you to change this, just adding a suggestion about what I prefer to do.

vinaysshenoy:
Left a comment. If you do not want to take any action on it, the PR is ready to merge.

app/src/test/java/org/simple/clinic/editpatient/PatientEditScreenControllerTest.kt
ragunathjawahar: Would using data classes make this better, because it would be more type-safe?

vinaysshenoy: The problem will data classes will be that you'll need to convert them into a `List<Any>` again before you return from the method source.

Or the test function will need to accept the data class itself as the parameter. I personally feel that it's overkill, but we could try it out and see what it looks like.

ragunathjawahar: Used a data class.

PR #738, opened 2019-07-23
Map build variants to appropriate Play Store URLs for app update

ragunathjawahar:


app/src/main/java/org/simple/clinic/appupdate/dialog/AppUpdateDialog.kt
```diff
@@ -37,10 +38,8 @@ class AppUpdateDialog : DialogFragment() {
   }
 
   private fun launchPlayStoreForUpdate() {
-    val appPackageName = "org.simple.clinic"
     val intent = Intent(ACTION_VIEW).apply {
-      data = Uri.parse(
-          "https://play.google.com/store/apps/details?id=$appPackageName")
+      data = Uri.parse("https://play.google.com/store/apps/details?id=${BuildConfig.APPLICATION_ID}")
```
ragunathjawahar: I may also have to see if we are checking against the appropriate variant in the play store.

PR #741, opened 2019-07-24
Fix: Clear OngoingRegistrationEntry after registration

sanchita:
https://www.pivotaltracker.com/story/show/167487181

### DIFF
--- a/app/src/main/java/org/simple/clinic/registration/register/RegistrationLoadingScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/registration/register/RegistrationLoadingScreenController.kt
@@ -28,15 +28,32 @@ class RegistrationLoadingScreenController @Inject constructor(
     val retryClicks = events.ofType<RegisterErrorRetryClicked>()
     val creates = events.ofType<ScreenCreated>()
 
-    return Observable.merge(creates, retryClicks)
-        .flatMapSingle { userSession.register() }
-        .doOnNext { if (it is Success) userSession.clearOngoingRegistrationEntry().subscribe() }
-        .map {
-          when (it) {
-            Success -> { ui: Ui -> ui.openHomeScreen() }
-            NetworkError -> { ui: Ui -> ui.showNetworkError() }
-            UnexpectedError -> { ui: Ui -> ui.showUnexpectedError() }
-          }
+    val register = Observable
+        .merge(creates, retryClicks)
+        .flatMap {
+          userSession
+              .register()
+              .toObservable()
         }
+        .replay()
+        .refCount()
+
+    val clearOngoingEntry = register
+        .filter { it is Success }
+        .flatMap {
+          userSession.clearOngoingRegistrationEntry()
+              .andThen(Observable.never<UiChange>())
+        }
+
+    val showScreenChanges = register.map {
+      { ui: Ui ->
+        when (it) {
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/simple/clinic/registration/register/RegistrationLoadingScreenController.kt
```diff
@@ -30,6 +30,7 @@ class RegistrationLoadingScreenController @Inject constructor(
 
     return Observable.merge(creates, retryClicks)
         .flatMapSingle { userSession.register() }
+        .doOnNext { if (it is Success) userSession.clearOngoingRegistrationEntry() }
```
vinaysshenoy: Since `UserSession#clearOngoingRegistrationEntry` returns a `Completable`, this call does not do anything because that returned `Completable` is never subscribed to.

@ragunathjawahar We have run into a few cases like this and it's dependent on the reviewer to catch these errors. We should discuss how we can add checks in place to catch problems like this without depending on humans.

app/src/main/java/org/simple/clinic/registration/register/RegistrationLoadingScreenController.kt
```diff
@@ -30,6 +30,7 @@ class RegistrationLoadingScreenController @Inject constructor(
 
     return Observable.merge(creates, retryClicks)
         .flatMapSingle { userSession.register() }
+        .doOnNext { if (it is Success) userSession.clearOngoingRegistrationEntry().subscribe() }
```
ragunathjawahar: This is a dangling `subscribe` call and is a code smell. What we may need here is a `flatMap` or fork the stream and process them separately.

app/src/test/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenControllerTest.kt
```diff
@@ -400,15 +400,14 @@ class RegistrationFacilitySelectionScreenControllerTest {
         createdAt = Instant.now())
     whenever(userSession.ongoingRegistrationEntry()).thenReturn(Single.just(ongoingEntry))
     whenever(userSession.saveOngoingRegistrationEntry(any())).thenReturn(Completable.complete())
-    whenever(userSession.loginFromOngoingRegistrationEntry()).thenReturn(Completable.complete())
+    whenever(userSession.saveOngoingRegistrationEntryAsUser()).thenReturn(Completable.complete())
 
     val facility1 = PatientMocker.facility(name = "Hoshiarpur", uuid = UUID.randomUUID())
     uiEvents.onNext(RegistrationFacilityClicked(facility1))
 
-    val inOrder = inOrder(userSession, screen)
```
ragunathjawahar: Nice that you got rid of `inOrder`.

app/src/main/java/org/simple/clinic/registration/register/RegistrationLoadingScreenController.kt
ragunathjawahar: I agree.

sanchita: I have fixed this code for now by subscribing to this `Completable` explicitly.

PR #742, opened 2019-07-26
Fix: Refresh user status  on Patient Screen

sanchita:
https://www.pivotaltracker.com/story/show/167533992

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
@@ -93,13 +93,14 @@ class PatientsScreenController @Inject constructor(
         // So duplicate triggers are dropped by restricting flatMap's
         // concurrency to 1.
         .toFlowable(BackpressureStrategy.LATEST)
-        .flatMapMaybe({ _ ->
+        .flatMapMaybe({
           userSession.loggedInUser()
               .firstOrError()
-              .filter { (user) -> user!!.status == WaitingForApproval }
-              .doOnSuccess {
+              .filter { (user) -> user != null }
+              .doOnSuccess { (user) ->
+                val userStatus = user?.status
                 // Resetting this flag here to show the approved status later
-                if (hasUserDismissedApprovedStatusPref.get()) {
+                if (userStatus != ApprovedForSyncing && hasUserDismissedApprovedStatusPref.get()) {
                   hasUserDismissedApprovedStatusPref.set(false)
                 }
 
--- a/app/src/test/java/org/simple/clinic/home/patients/PatientsScreenControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/home/patients/PatientsScreenControllerTest.kt
@@ -3,10 +3,10 @@ package org.simple.clinic.home.patients
 import com.f2prateek.rx.preferences2.Preference
 import com.nhaarman.mockito_kotlin.any
 import com.nhaarman.mockito_kotlin.atLeastOnce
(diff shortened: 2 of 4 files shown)


app/src/test/java/org/simple/clinic/home/patients/PatientsScreenControllerTest.kt
```diff
@@ -108,43 +107,26 @@ class PatientsScreenControllerTest {
...
     whenever(approvalStatusApprovedAt.get()).thenReturn(Instant.now())
     whenever(hasUserDismissedApprovedStatus.asObservable()).thenReturn(Observable.just(false))
     whenever(hasUserDismissedApprovedStatus.get()).thenReturn(false)
 
     uiEvents.onNext(ScreenCreated())
     uiEvents.onNext(TheActivityLifecycle.Resumed())
-    uiEvents.onNext(TheActivityLifecycle.Resumed())
 
-    verify(userSession, never()).refreshLoggedInUser()
+    verify(userSession, times(2)).refreshLoggedInUser()
```
vinaysshenoy: It's not obvious from this check that each of the two different `UiEvent`s (`ScreenCreated` and `Resumed`) can independently trigger a user refresh. Can we either:
1. Pass the event in via the test parameter as well so we can test each event separately, or
2. Make one verification call after the first event, and another after the second event so we know that each event triggers a refresh independently.

```kotlin
uiEvents.onNext(ScreenCreated())
verify(userSession).refreshLoggedInUser()

clearInvocations(userSession)

uiEvents.onNext(TheActivityLifecycle.Resumed())
verify(userSession).refreshLoggedInUser()
```

app/src/test/java/org/simple/clinic/home/patients/PatientsScreenControllerTest.kt
```diff
@@ -177,6 +159,8 @@ class PatientsScreenControllerTest {
     whenever(userSession.loggedInUser()).thenReturn(Observable.just(Just(user)))
     whenever(hasUserDismissedApprovedStatus.asObservable()).thenReturn(Observable.just(false))
     whenever(hasUserDismissedApprovedStatus.get()).thenReturn(false)
+    whenever(userSession.refreshLoggedInUser()).thenReturn(Completable.never())
```
vinaysshenoy: We could move this to the `setUp` fixture and only override it in the tests that depend on this?

app/src/test/java/org/simple/clinic/home/patients/PatientsScreenControllerTest.kt
sanchita: Good point.

app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
```diff
@@ -93,13 +93,14 @@ class PatientsScreenController @Inject constructor(
         // So duplicate triggers are dropped by restricting flatMap's
```
vinaysshenoy: If you look at the `PatientsScreenController#refreshUserStatus` method, we can see that a complete data sync happens whenever the user gets refreshed if they are logged in.

This was meant to handle the case where sync should happen as soon as a user is approved. Now with this change, a complete sync will happen every time the screen is opened.

In order to maintain the current functionality, we should do this sync only whenever the user's logged in status and user status changes to (`Allowed`, Logged In`) from any other state.

sanchita: I will raise a new PR for this fix.

PR #744, opened 2019-07-29
Add ADR for enum forward compatibility.

ragunathjawahar:
This PR is a draft, can you please provide feedback so that I can iterate on it before we merge this?

### DIFF
--- a/doc/arch/006-forward-compatibility-for-enums.md
+++ b/doc/arch/006-forward-compatibility-for-enums.md
@@ -1,0 +1,108 @@
+# ADR: Forward compatibility for persistable enums from Server responses
+
+
+## Status
+
+Accepted
+
+
+## Context
+
+RESTful APIs tend to return enumerated values to represent properties with a fixed number of well-defined values. These are usually represented as `enum` classes on the client. When new business use cases are introduced, there could be additions to these values. If these values are consumed on the fly and are not persisted on the client, they could simply be treated as a special case (unknown) or ignored all together.
+
+Simple is an "offline-first" application and some of these values have to be persisted to the database. Hence, these enumerated values cannot be represented using a Kotlin `enum` class. Offline-first means, there are some unique constraints that we have to deal with.
+
+Apps on an older app version should be able to handle new enum values added to a type even though they don't know how to interpret the new value. This could be solved by adding an UNKNOWN fallback value which the server adds as well. Except,
+
+* Records are shared between different users and thus, between different app versions. A newer app version that understands the new enum value could create a record with the new value and upload it, but an older app version that does not know the new value will overwrite it with UNKNOWN and this value will overwrite the actual value on the server as well when it syncs.
+
+* An app on an older version could have a degraded, fallback experience for enum values it does not understand, but as soon as it updates, the app should be able to interpret the new values as if it was synced afresh.
+
+
+## Decision
+
+To maintain forward-compatibility with enum values, we have to roll out our own infrastructure. To accommodate, we use sealed classes to represent enumerated values so that,
+
+- We could save these values to the database.
+- Use these values in newer versions of the app.
+- Send the same enum values back to the server during sync even though the current version of the app cannot interpret these values during runtime.
+
+In order or achieve this, we require 4 classes.
+
+1. A sealed class hierarchy that is used to represent the enumerated values. Which also includes an `Unknown(val actualValue: String)` data class which is used to capture and handle newly introduced values.
+2. An implementation of the `SafeEnumTypeAdapter` class.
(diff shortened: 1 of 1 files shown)


ragunathjawahar:
There is also a mix of why and what. But I think the balance towards why can be improved.

doc/arch/006-forward-compatibility-for-enums.md
```diff
@@ -0,0 +1,55 @@
...
+## Status
+
+Accepted
+
+
+## Context
+
+RESTful APIs tend to return enumerated values to represent properties with a fixed number of well-defined values. These are usually represented as `enum` classes on the client. When new business use cases are introduced, there could be additions to these values. If these values are consumed on the fly and are not persisted on the client, they could simply be treated as a special case (unknown) or ignored all together.
+
+In Simple, some of these values have to be persisted in the database. Hence, these enumerated values cannot be represented using a Kotlin `enum` class.
```
vinaysshenoy: We should probably also add a section that explains the additional "offline-first" constraint on the product that stops us from being able to clear the data and sync again when an app update happens. This means that:

- Apps on an older app version should be able to handle new `enum` values added to a type even though they don't know how to interpret the new value. This *could* be solved by adding an `UNKNOWN` fallback value which the server adds as well, except
- Records are shared between different users and thus, between different app versions. A newer app version that understands the new `enum` value could create a record with the new value and upload it, but an older app version that does not know the new value will overwrite it with `UNKNOWN` and this value will overwrite the actual value on the server as well when it syncs.
- An app on an older version could have a degraded, fallback experience for `enum` values it does not understand, but as soon as it updates, the app should be able to interpret the new values as if it was synced afresh.

doc/arch/006-forward-compatibility-for-enums.md
```diff
@@ -0,0 +1,55 @@
...
+    private val unknownEnumToStringConverter: (T) -> String,
+    private val unknownStringToEnumConverter: (String) -> T
+) {
+  // …
+}
+```
+
+The `knownMappings` field takes in a `Map` that can convert to and from enumerations that the current version of the application understands.
+
+The `unknownEnumToStringConverter` converts strings to their appropriate sealed class representations.
```
vinaysshenoy: Isn't this the other way around? This is to convert unknown enum values to their actual string representation (usually used when pushing data to the server).

doc/arch/006-forward-compatibility-for-enums.md
```diff
@@ -0,0 +1,55 @@
...
+
+The `unknownStringToEnumConverter` can wrap the newly introduced value into the `Unknown` sealed data class defined earlier.
+
+
+## Consequences
+
+
+Not all enumerated values has to represented and conform to the existing forward compatible APIs. Only enums that are served through APIs and are saved to the database should use this convention.
```
vinaysshenoy: Grammar: Not all enumerated values ~~has to~~ have to be represented

vinaysshenoy: The first sentence seems a little awkward. I'd suggest a rewording

> Not all enumerated values have to conform to this forward compatibility spec and be represented in this fashion.

vinaysshenoy: Maybe emphasize the "and"?
> Only enums that are served through APIs **and** are saved to the database should use this convention.

doc/arch/006-forward-compatibility-for-enums.md
```diff
@@ -0,0 +1,55 @@
...
+
+The only way for developers to gain awareness about this is read this ADR or through tribal knowledge. Ideally, we should have a lint check that throws an error if an `enum class` is involved both in a Retrofit and a Room interface.
```
vinaysshenoy: Reword suggestion:
> The only way for developers to gain awareness about this, is to read this ADR or through tribal knowledge.

doc/arch/006-forward-compatibility-for-enums.md
```diff
@@ -0,0 +1,55 @@
...
+  // …
+}
+```
+
+The `knownMappings` field takes in a `Map` that can convert to and from enumerations that the current version of the application understands.
+
+The `unknownEnumToStringConverter` converts strings to their appropriate sealed class representations.
+
+The `unknownStringToEnumConverter` can wrap the newly introduced value into the `Unknown` sealed data class defined earlier.
+
```
vinaysshenoy: We should probably add a minimal example that shows how to add an enum value in this fashion. A good candidate would be `AppointmentType`.

ragunathjawahar:
@vinaysshenoy PTAL?

PR #745, opened 2019-07-29
Add phone number input to PatientSearchScreen

vinaysshenoy:
Pivotal Story Link: https://www.pivotaltracker.com/story/show/167343636

### DIFF
--- a/app/src/main/java/org/simple/clinic/search/PatientSearchScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/search/PatientSearchScreenController.kt
@@ -49,23 +49,25 @@ class PatientSearchScreenController @Inject constructor() : ObservableTransforme
           .map { it.text.trim() }
 
       val validationErrors = events.ofType<SearchClicked>()
-          .withLatestFrom(textChanges)
-          .map { (_, text) ->
-            val errors = mutableListOf<PatientSearchValidationError>()
-
-            if (text.isBlank()) {
-              errors += INPUT_EMPTY
-            }
-            SearchQueryValidated(errors)
-          }
+          .withLatestFrom(textChanges) { _, text -> text }
+          .map(this::validateInput)
 
       events.mergeWith(validationErrors)
     }
   }
 
+  private fun validateInput(inputText: String): SearchQueryValidationResult {
+    return if (inputText.isBlank()) {
+      SearchQueryValidationResult.Invalid(listOf(INPUT_EMPTY))
+    } else {
+      SearchQueryValidationResult.Valid(inputText)
+    }
+  }
+
   private fun showValidationErrors(events: Observable<UiEvent>): Observable<UiChange> {
-    return events.ofType<SearchQueryValidated>()
-        .flatMapIterable { it.validationErrors }
+    return events
+        .ofType<SearchQueryValidationResult.Invalid>()
+        .flatMapIterable { it.errors }
         .doOnNext { Analytics.reportInputValidationError(it.analyticsName) }
         .map {
           { ui: Ui ->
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsScreen.kt
```diff
@@ -75,13 +77,13 @@ class PatientSearchResultsScreen(context: Context, attrs: AttributeSet) : Relati
...
   }
 
-  private fun extractPatientName(searchPatientInput: SearchPatientInput): String {
-    return when (searchPatientInput) {
-      is SearchPatientInput.Name -> searchPatientInput.searchText
-      is SearchPatientInput.PhoneNumber -> TODO("not yet implemented")
+  private fun extractPatientName(criteria: PatientSearchCriteria): String {
+    return when (criteria) {
+      is Name -> criteria.patientName
+      is PhoneNumber -> TODO("not yet implemented")
```
sanchita: Please implement this.

app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsScreen.kt
```diff
@@ -75,13 +77,13 @@ class PatientSearchResultsScreen(context: Context, attrs: AttributeSet) : Relati
...
         .ofType<RegisterNewPatient>()
-        .map { PatientSearchResultRegisterNewPatient(extractPatientName(it.searchPatientInput)) }
+        .map { PatientSearchResultRegisterNewPatient(extractPatientName(it.criteria)) }
   }
 
-  private fun extractPatientName(searchPatientInput: SearchPatientInput): String {
-    return when (searchPatientInput) {
-      is SearchPatientInput.Name -> searchPatientInput.searchText
-      is SearchPatientInput.PhoneNumber -> TODO("not yet implemented")
+  private fun extractPatientName(criteria: PatientSearchCriteria): String {
```
sanchita: Rename this appropriately.

app/src/main/java/org/simple/clinic/search/PatientSearchScreenController.kt
```diff
@@ -36,17 +36,17 @@ class PatientSearchScreenController @Inject constructor() : ObservableTransforme
...
+          .withLatestFrom(textChanges)
+          .map { (_, text) ->
             val errors = mutableListOf<PatientSearchValidationError>()
 
-            if (name.isBlank()) {
-              errors += FULL_NAME_EMPTY
+            if (text.isBlank()) {
+              errors += INPUT_EMPTY
             }
             SearchQueryValidated(errors)
```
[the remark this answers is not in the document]
vinaysshenoy: Yeah, earlier this had name + phone number + age which were contributing their own errors. When we removed them, we didn't remove the list which was accumulating the errors. I'll remove this.

app/src/main/java/org/simple/clinic/search/PatientSearchScreenController.kt
```diff
@@ -75,10 +75,6 @@ class PatientSearchScreenController @Inject constructor() : ObservableTransforme
   }
 
   private fun openSearchResults(events: Observable<UiEvent>): Observable<UiChange> {
-    val nameChanges = events
-        .ofType<SearchQueryTextChanged>()
-        .map { it.text.trim() }
-
     val validationErrors = events
         .ofType<SearchQueryValidated>()
```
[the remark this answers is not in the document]
vinaysshenoy: I don't think we should remove multiple errors from the `PatientSearchValidationError` at the moment. There is currently no validation done on the phone number at the moment, but we can't say there will never be one in the future.

However, I do think we can abstract the fact that the input is valid means that the validation errors are empty by converting that chain into an `Observable<Boolean>`. I'll try this out.

app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsScreen.kt
vinaysshenoy: Whoops, good catch! This will be fixed when https://www.pivotaltracker.com/story/show/167343744 is picked up. For now, I will return an empty string so that the app does not crash.

app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsScreen.kt
vinaysshenoy: This was a temporary "fix" because we needed to pass in a name to the `RegisterNewPatient` event. This entire function will be removed once https://www.pivotaltracker.com/story/show/167343744 is done. I'd rather not rename a function which will be removed in an upcoming PR.

I have added a `TODO` that explains the same.

app/src/main/java/org/simple/clinic/search/PatientSearchScreenController.kt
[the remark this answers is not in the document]
vinaysshenoy: Tried this, looks a lot cleaner now!

vinaysshenoy:
@sanchita I have also removed the `TextInput` type that was introduced internally in `PatientSearchScreenController` and changed it to generate the `PatientSearchCriteria` directly.

PR #750, opened 2019-08-01
Appointment date picker

rakshakhegde:
https://www.pivotaltracker.com/story/show/166924812


vinaysshenoy:
@rakshakhegde I can select the current date on the calendar widget and schedule an appointment for "0 Days". This needs to be handled.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheet.kt
```diff
@@ -17,7 +17,8 @@ import kotterknife.bindView
 import org.simple.clinic.R
```
vinaysshenoy: Reword this commit to something more descriptive? I don't understand what is being restricted and why.

rakshakhegde:
@vinaysshenoy These were the following issues I found.
- The app crashes on choosing today's date from Calendar.
- The months turn into days in some places on pressing plus button

vinaysshenoy:
> @vinaysshenoy These were the following issues I found.
> 
> * The app crashes on choosing today's date from Calendar.
> * The months turn into days in some places on pressing plus button

Checking.

vinaysshenoy:
@rakshakhegde Fixed the issue. The current date was being allowed for selection, which I fixed. I am not able to reproduce the other problem.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -174,4 +255,56 @@ class ScheduleAppointmentSheetController @Inject constructor(
...
+      val scheduledFor: LocalDate,
+      val timeToAppointment: TimeToAppointment
+  ) : Comparable<PotentialAppointmentDate> {
+    override fun compareTo(other: PotentialAppointmentDate): Int {
+      return this.scheduledFor.compareTo(other.scheduledFor)
+    }
+  }
+}
+
+private fun LocalDate.plus(timeToAppointment: TimeToAppointment): LocalDate {
```
rakshakhegde: This could be nicer if converted to a `operator` function!

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -174,4 +255,56 @@ class ScheduleAppointmentSheetController @Inject constructor(
...
+  return this.plus(
```
rakshakhegde: Also `this` is not required?

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -174,4 +255,56 @@ class ScheduleAppointmentSheetController @Inject constructor(
...
+
+    return exactMatchingTimeToAppointment ?: Days(ChronoUnit.DAYS.between(today, date).toInt())
+  }
+
+  private fun generatePotentialAppointmentDate(scheduleAppointmentIn: TimeToAppointment): PotentialAppointmentDate {
+    val today = LocalDate.now(clock)
+    return PotentialAppointmentDate(today.plus(scheduleAppointmentIn), scheduleAppointmentIn)
+  }
+
+  private fun generatePotentialAppointmentDates(scheduleAppointmentsIn: List<TimeToAppointment>): List<PotentialAppointmentDate> {
```
rakshakhegde: We can simply deal with `TimeToAppointment` and create LocalDates only when we need them right? As mentioned before, the API of `PotentialAppointmentDate` allows for errors in input parameters passed.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -27,145 +34,219 @@ typealias UiChange = (Ui) -> Unit
...
 
     val closeSheetWithoutSavingAppointment = isPatientDefaulterStream
         .filter { isPatientDefaulter -> isPatientDefaulter.not() }
-        .map { { ui: Ui -> ui.closeSheet() } }
+        .map { Ui::closeSheet }
 
     return Observable.merge(saveAppointmentAndCloseSheet, closeSheetWithoutSavingAppointment)
   }
 
   private fun scheduleCreates(events: Observable<UiEvent>): Observable<UiChange> {
```
rakshakhegde: This function is scheduling the appointment and no more handling creation event. Let's call it `scheduleAppointmentDone`?

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheet.kt
```diff
@@ -105,37 +77,40 @@ class ScheduleAppointmentSheet : BottomSheetActivity() {
...
-    finish()
+  fun updateScheduledAppointment(appointmentDate: LocalDate, timeToAppointment: TimeToAppointment) {
+    currentAppointmentDate.text = dateFormatter.format(appointmentDate)
+    currentDateTextView.text = displayTextForTimeToAppointment(timeToAppointment)
   }
 
-  fun updateDisplayedDate(newIndex: Int) {
-    currentIndex = newIndex
-    currentDateTextView.text = possibleDates[currentIndex].displayText
+  private fun displayTextForTimeToAppointment(timeToAppointment: TimeToAppointment): CharSequence {
```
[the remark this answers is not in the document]
vinaysshenoy: It can, I'll do this.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheet.kt
```diff
@@ -105,37 +77,40 @@ class ScheduleAppointmentSheet : BottomSheetActivity() {
...
     setResult(Activity.RESULT_OK)
     finish()
   }
 
-  fun closeSheet(appointmentDate: LocalDate) {
-    val intent = Intent()
-    intent.putExtra(EXTRAS_SAVED_APPOINTMENT_DATE, appointmentDate)
-    setResult(Activity.RESULT_OK, intent)
-    finish()
+  fun updateScheduledAppointment(appointmentDate: LocalDate, timeToAppointment: TimeToAppointment) {
```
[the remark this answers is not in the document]
vinaysshenoy: Good point, I'll make this change.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
vinaysshenoy: It could, I just didn't do it because I wanted to maintain the convention of the rest of `LocalDate.plus` functions that don't have operator overloads.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
vinaysshenoy: It it not, just wanted to make it more explicit.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
vinaysshenoy: We can't do this because selecting the right `TimeToAppointment` when a calendar date is selected is not possible. We need to convert it to a `LocalDate` right at the beginning to make it more convenient.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheet.kt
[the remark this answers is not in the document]
vinaysshenoy: They are not exactly the same thing. The `appointmentDate` is the exact date on which the appointment will be scheduled. The `TimeToAppointment` is what to display it as. For the same `appointmentDate` can be shown in multiple `TimeToAppointment` ways (for example, `Weeks(1)` and `Days(7)` will both point to the same `appointmentDate` but the presentation will differ because of the type.

We have tests that ensure that the same value passed in the config is passed down to the screen when displaying it. We will depend on manual testing to ensure that the configuration itself is declared correctly.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
vinaysshenoy: We ensure in tests that the `PotentialAppointment` is created with the same `TimeToAppointment` that the date is generated from. We will depend on manual testing to verify that the configuration itself is done correctly.

rakshakhegde: We can do this. If you take a look at the commit I made just before this refactor, this is what was done. We computed the LocalDates finally only when they were required. Or at least LocalDate computation can also be made a property in `PotentialAppointmentDate`.

It is ok if you feel like skipping this change. We can come back to this design change in the future too.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
vinaysshenoy: Since this chain is still doing the job of actually creating the appointment, I have renamed it to `scheduleAppointment()`.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
rakshakhegde: This is the commit in this branch you'll want to check out 42726e2233c802fc26c36ac9c55d32af668e12d4

vinaysshenoy: I intentionally did not do this because I felt the generation of dates everywhere from `ScheduleAppointment` just in time was becoming too complex and hard to follow.

rakshakhegde:
@vinaysshenoy You can go ahead and merge it

PR #753, opened 2019-08-05
Add overdue list animations back

vinaysshenoy:
https://www.pivotaltracker.com/story/show/167245969


app/src/test/java/org/simple/clinic/home/overdue/OverdueScreenControllerTest.kt
```diff
@@ -128,32 +117,21 @@ class OverdueScreenControllerTest {
...
   @Test
   fun `when screen is created then the overdue appointments must be displayed`() {
+    // given
     val overdueAppointments = listOf(
-        PatientMocker.overdueAppointment(riskLevel = OverdueAppointment.RiskLevel.HIGHEST),
-        PatientMocker.overdueAppointment(riskLevel = OverdueAppointment.RiskLevel.LOW),
-        PatientMocker.overdueAppointment(riskLevel = OverdueAppointment.RiskLevel.NONE)
+        PatientMocker.overdueAppointment(riskLevel = OverdueAppointment.RiskLevel.HIGHEST, phoneNumber = null),
+        PatientMocker.overdueAppointment(riskLevel = OverdueAppointment.RiskLevel.LOW, phoneNumber = null),
+        PatientMocker.overdueAppointment(riskLevel = OverdueAppointment.RiskLevel.NONE, phoneNumber = null)
```
sanchita: Passing `null` in `phoneNumber` doesn't seem to make any difference here.
Any reason for it?

app/src/main/java/org/simple/clinic/widgets/ItemAdapter.kt
```diff
@@ -0,0 +1,40 @@
...
+import io.reactivex.subjects.Subject
+import org.simple.clinic.widgets.recyclerview.ViewHolderX
+
+open class ItemAdapter<I : ItemAdapter.Item<E>, E>(
+    diffCallback: DiffUtil.ItemCallback<I>
+) : ListAdapter<I, ViewHolderX>(diffCallback) {
+
+  private val eventSubject: Subject<E> = PublishSubject.create<E>()
+
+  val listItemEvents: Observable<E> = eventSubject.hide()
```
sanchita: Rename this to `itemEvents`?
Since we are not using the term `list` anywhere in this class. Its _obvious_ :)

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
```diff
@@ -475,4 +475,12 @@ object PatientMocker {
         businessIds = emptyList()
     )
   }
+
+  fun age(
+      ageValue: Int,
+      updatedAt: Instant = Instant.now(),
+      computedDateOfBirth: LocalDate = LocalDate.now(UTC)
```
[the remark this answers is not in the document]
vinaysshenoy: Interestingly, I looked at the code right now and noticed that this property is no longer being used anywhere. All usages of this property have been replaced with `TimeFunctions.estimateAge` and this is basically a useless property that is just there because it needs a migration to remove.

I will mark this property as deprecated for now and this will get removed when we do https://www.pivotaltracker.com/story/show/161671487.

app/src/main/java/org/simple/clinic/home/overdue/OverdueListItem.kt
```diff
@@ -0,0 +1,130 @@
...
+import org.simple.clinic.widgets.ItemAdapter
+import org.simple.clinic.widgets.UiEvent
+import org.simple.clinic.widgets.locationRectOnScreen
+import org.simple.clinic.widgets.marginLayoutParams
+import org.simple.clinic.widgets.recyclerview.ViewHolderX
+import org.simple.clinic.widgets.setCompoundDrawableStart
+import org.simple.clinic.widgets.visibleOrGone
+import java.util.UUID
+
+sealed class OverdueListItem : ItemAdapter.Item<UiEvent> {
```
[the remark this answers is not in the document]
vinaysshenoy: The original code I refactored from had it implemented as a sealed class and I retained it. But you're right, we have only one item and we don't need it to be. I'll remove this.

app/src/main/java/org/simple/clinic/home/overdue/OverdueListItem.kt
```diff
@@ -0,0 +1,130 @@
...
+      val name: String,
+      val gender: Gender,
+      val age: Int,
+      val phoneNumber: String? = null,
+      val bpSystolic: Int,
+      val bpDiastolic: Int,
+      val bpDaysAgo: Int,
+      val overdueDays: Int,
+      val isAtHighRisk: Boolean
+  ) : OverdueListItem() {
```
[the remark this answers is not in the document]
vinaysshenoy: See, see the earlier comment I left about why it was a sealed class. I'll make this change.

app/src/test/java/org/simple/clinic/home/overdue/OverdueScreenControllerTest.kt
vinaysshenoy: I think it got added automatically by the refactoring when I added the parameter to the method. I'll remove it.

app/src/test/java/org/simple/clinic/home/overdue/OverdueScreenControllerTest.kt
```diff
@@ -128,32 +117,21 @@ class OverdueScreenControllerTest {
 
   @Test
   fun `when screen is created then the overdue appointments must be displayed`() {
+    // given
     val overdueAppointments = listOf(
         PatientMocker.overdueAppointment(riskLevel = OverdueAppointment.RiskLevel.HIGHEST),
-        PatientMocker.overdueAppointment(riskLevel = OverdueAppointment.RiskLevel.LOW),
+        PatientMocker.overdueAppointment(riskLevel = OverdueAppointment.RiskLevel.LOW, phoneNumber = null),
```
sanchita: Did you forget to remove `phoneNumber = null` from here?

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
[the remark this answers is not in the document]
sanchita: Then can we remove this method altogether and use `estimateCurrentAge` instead?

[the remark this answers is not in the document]
vinaysshenoy: Removed the method in favour of just constructing the `Age` directly.

app/src/test/java/org/simple/clinic/home/overdue/OverdueScreenControllerTest.kt
vinaysshenoy: I originally left it in there to ensure appointments with no phone numbers are mapped correctly. Then I realised that the problem is that the property is nullable since we filter these out at the query level itself. I removed this for now; We'll need to make this property mandatory in the future,

sanchita: Yea I agree. This should not have been a `Nullable`.

PR #754, opened 2019-08-06
Fix: Duplicate results when searching patients by phone number

vinaysshenoy:
https://www.pivotaltracker.com/story/show/167702689

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # CHANGELOG
 
 ## Next Release
+### Fixes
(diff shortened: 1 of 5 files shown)


vinaysshenoy:
> As we need the first row from the sorted list for each group, can you also try out the technique used in `org.simple.clinic.patient.RecentPatient.RoomDao.RECENT_PATIENT_QUERY` where we use `SELECT MAX()` along with `GROUP BY` to get the latest record from each group. In case there are nullable fields, we'll also have to wrap those fields with `IFNULL` just like in recent patient query.

This worked out well. Changing the fix, thanks!

PR #756, opened 2019-08-08
Create an IntelliJ Plugin to generate UUIDs

ragunathjawahar:


ragunathjawahar:
@vinaysshenoy added instructions for building the plugin and also replaced the live templates file.

PR #765, opened 2019-08-14
Add short code search

sanchita:
https://www.pivotaltracker.com/story/show/167310717

### DIFF
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
@@ -2,11 +2,13 @@ package org.simple.clinic.scanid
 
 import android.content.Context
 import android.util.AttributeSet
+import android.view.View
 import androidx.constraintlayout.widget.ConstraintLayout
 import com.jakewharton.rxbinding2.view.RxView
 import com.jakewharton.rxbinding2.widget.RxTextView
 import io.reactivex.Observable
 import kotlinx.android.synthetic.main.screen_scan_simple.view.*
+import org.simple.clinic.R
 import org.simple.clinic.activity.TheActivity
 import org.simple.clinic.addidtopatient.searchforpatient.AddIdToPatientSearchScreenKey
 import org.simple.clinic.bindUiToController
@@ -61,8 +63,8 @@ class ScanSimpleIdScreen(context: Context, attrs: AttributeSet) : ConstraintLayo
 
   private fun doneClicks(): Observable<UiEvent> {
     return RxTextView
-        .editorActionEvents(editText)
-        .map { ShortCodeSearched(ShortCodeInput(editText.text.toString())) }
+        .editorActionEvents(shortCodeText)
+        .map { ShortCodeSearched(ShortCodeInput(shortCodeText.text.toString())) }
   }
 
   fun openPatientSummary(patientUuid: UUID) {
@@ -81,11 +83,12 @@ class ScanSimpleIdScreen(context: Context, attrs: AttributeSet) : ConstraintLayo
   }
 
   fun showShortCodeValidationError() {
-    TODO("not implemented")
+    shortCodeErrorText.visibility = View.VISIBLE
+    shortCodeErrorText.text = resources.getString(R.string.scansimpleid_shortcode_error)
   }
 
   fun hideShortCodeValidationError() {
-    TODO("not implemented")
+    shortCodeErrorText.visibility = View.GONE
   }
 
   fun openPatientShortCodeSearch(validShortCode: String) {
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
@@ -95,4 +95,8 @@ class ScanSimpleIdScreen(context: Context, attrs: AttributeSet) : ConstraintLayo
   fun hideQrCodeScannerView() {
     TODO("not implemented")
   }
+
+  fun showQrCodeScannerView() {
+    TODO("not implemented")
+  }
 }
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
@@ -91,4 +91,8 @@ class ScanSimpleIdScreen(context: Context, attrs: AttributeSet) : ConstraintLayo
   fun openPatientShortCodeSearch(validShortCode: String) {
     TODO("not implemented")
   }
+
+  fun hideQrCodeScannerView() {
+    TODO("not implemented")
+  }
 }
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
@@ -7,6 +7,7 @@ import androidx.constraintlayout.widget.ConstraintLayout
 import com.jakewharton.rxbinding2.view.RxView
 import com.jakewharton.rxbinding2.widget.RxTextView
 import io.reactivex.Observable
+import io.reactivex.subjects.PublishSubject
 import kotlinx.android.synthetic.main.screen_scan_simple.view.*
 import org.simple.clinic.R
 import org.simple.clinic.activity.TheActivity
@@ -36,6 +37,8 @@ class ScanSimpleIdScreen(context: Context, attrs: AttributeSet) : ConstraintLayo
   @Inject
   lateinit var utcClock: UtcClock
 
+  private val keyboardEvents = PublishSubject.create<ScanSimpleIdScreenEvent>()
+
   override fun onFinishInflate() {
     super.onFinishInflate()
     if (isInEditMode) {
@@ -49,7 +52,7 @@ class ScanSimpleIdScreen(context: Context, attrs: AttributeSet) : ConstraintLayo
 
     bindUiToController(
         ui = this,
-        events = Observable.mergeArray(qrScans(), doneClicks()),
+        events = Observable.mergeArray(qrScans(), keyboardEvents, doneClicks()),
         controller = controller,
         screenDestroys = RxView.detaches(this).map { ScreenDestroyed() }
     )
@@ -96,10 +99,10 @@ class ScanSimpleIdScreen(context: Context, attrs: AttributeSet) : ConstraintLayo
   }
 
   fun hideQrCodeScannerView() {
-    TODO("not implemented")
+    qrCodeScannerView.hideQrCodeScanner()
   }
 
   fun showQrCodeScannerView() {
-    TODO("not implemented")
+    qrCodeScannerView.showQrCodeScanner()
   }
 }
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
@@ -62,7 +62,7 @@ class ScanSimpleIdScreen(context: Context, attrs: AttributeSet) : ConstraintLayo
   private fun doneClicks(): Observable<UiEvent> {
     return RxTextView
         .editorActionEvents(editText)
-        .map { ShortCodeSearched }
+        .map { ShortCodeSearched(ShortCodeInput(editText.text.toString())) }
   }
 
   fun openPatientSummary(patientUuid: UUID) {
@@ -78,5 +78,17 @@ class ScanSimpleIdScreen(context: Context, attrs: AttributeSet) : ConstraintLayo
 
   fun openAddIdToPatientScreen(identifier: Identifier) {
     screenRouter.popAndPush(AddIdToPatientSearchScreenKey(identifier), RouterDirection.FORWARD)
+  }
+
+  fun showShortCodeValidationError() {
+    TODO("not implemented")
+  }
+
+  fun hideShortCodeValidationError() {
+    TODO("not implemented")
+  }
+
+  fun openPatientShortCodeSearch(validShortCode: String) {
+    TODO("not implemented")
   }
 }
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
@@ -2,12 +2,11 @@ package org.simple.clinic.scanid
 
 import android.content.Context
 import android.util.AttributeSet
-import androidx.appcompat.widget.Toolbar
 import androidx.constraintlayout.widget.ConstraintLayout
 import com.jakewharton.rxbinding2.view.RxView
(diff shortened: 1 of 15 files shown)


app/src/main/res/layout/screen_scan_simple.xml
```diff
@@ -27,4 +30,27 @@
...
+    app:layout_constraintStart_toStartOf="parent">
+
+    <EditText
+      android:id="@+id/editText"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:layout_gravity="center"
+      android:ems="10"
+      android:hint="123 4567"
+      android:imeOptions="actionGo"
```
vinaysshenoy: Shouldn't this be `actionSearch`? We won't know until the search has completed whether the target exists or not.

app/src/main/res/layout/screen_scan_simple.xml
```diff
@@ -27,4 +30,27 @@
...
 
+  <androidx.cardview.widget.CardView
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:layout_marginBottom="8dp"
+    android:layout_marginEnd="8dp"
+    android:layout_marginStart="8dp"
+    android:background="@color/white100"
+    android:visibility="gone"
+    app:layout_constraintBottom_toBottomOf="@+id/qrCodeScannerView"
```
vinaysshenoy: This should just be `@id/qrCodeScannerView`, right?

app/src/main/res/layout/screen_scan_simple.xml
```diff
@@ -27,4 +30,27 @@
...
+    app:layout_constraintEnd_toEndOf="@+id/qrCodeScannerView"
+    app:layout_constraintStart_toStartOf="parent">
+
+    <EditText
+      android:id="@+id/editText"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:layout_gravity="center"
+      android:ems="10"
```
vinaysshenoy: Seems weird that we define the `ems` as 10 here, and then again specify the `maxEms` as 7. Was this a mistake?

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdScreenControllerTest.kt
```diff
@@ -86,11 +77,74 @@ class ScanSimpleIdScreenControllerTest {
...
+  @Test
+  fun `when the keyboard is up, then don't process valid QR code scan events`() {
+    // given
+    val bpPassportUUid = "69170f2f-c112-4907-b184-a99d1001e269"
+    val patientUuid = UUID.fromString("b123dc5c-a02d-49cb-939e-f4602436f214")
+    whenever(patientRepository.findPatientWithBusinessId(bpPassportUUid))
+        .thenReturn(Observable.just(PatientMocker.patient(uuid = patientUuid).toOptional()))
+
+    // when
+    with(uiEvents) {
```
vinaysshenoy: TIL about the existence of `with()`. 👏

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenController.kt
```diff
@@ -39,21 +43,40 @@ class ScanSimpleIdScreenController @Inject constructor(
...
           }
         }
 
     events.mergeWith(scannedBpPassportCodes)
   }
 
   private fun handleScannedBpPassportCodes(events: Observable<UiEvent>): Observable<UiChange> {
-    val scannedBpPassportCodeStream = events
-        .ofType<ScanSimpleIdScreenPassportCodeScanned.ValidPassportCode>()
+    val showKeyboardEvents = events
```
vinaysshenoy: I am not entirely sure of this because it seems like the controller is too deeply coupled with the implementation of the screen and the keyboard behaviour. I wonder if we should consider delegating the behaviour of filtering the scanned passport codes while the keyboard is up to the view layer and make the controller not have to care about this.

Using the [RxJavaExtensions#valve](https://github.com/akarnokd/RxJavaExtensions#flowabletransformersvalve) operator might actually be really helpful here as well.

app/src/main/java/org/simple/clinic/scanid/ShortCodeInput.kt
```diff
@@ -0,0 +1,7 @@
+package org.simple.clinic.scanid
+
+data class ShortCodeInput(val shortCodeText: String) {
+  fun isValid(): Boolean {
+    return shortCodeText.length == 7
```
vinaysshenoy: We now have this magic number for the length of a BP Passport Short Code duplicated in two places:
- `UuidShortCodeCreatorModule`
- `ShortCodeInput`

The only use case this will change is if we decide to add a new type of BP Passport itself which will require us to rethink how the BP Passport feature works. 

Until then, what do you think about declaring this magic number as a global constant in `Globals.kt`? This only thing defined there now is the screen change animation duration, but I feel this is a good use case for a global constant.

app/src/test/java/org/simple/clinic/scanid/ShortCodeInputTest.kt
```diff
@@ -0,0 +1,44 @@
...
+class ShortCodeInputTest {
+
+  @Parameters(value = [
+    "3",
+    "34",
+    "345",
+    "3456",
+    "34567",
+    "345678",
+    "34567890"
```
vinaysshenoy: The hint we have defined earlier for the short code was `123 4567`, which means that the user might enter the short code with a space in between. Are we going to handle this case elsewhere?

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenController.kt
```diff
@@ -95,4 +98,30 @@ class ScanSimpleIdScreenController @Inject constructor(
...
+    val shortCodes = events
+        .ofType<ShortCodeSearched>()
+        .map { it.shortCode }
+        .share()
+
+    val showValidationErrors = shortCodes
+        .filter { !it.isValid() }
+        .map { { ui: Ui -> ui.showShortCodeValidationError() } }
+
+    val openPatientSearchScreenChanges = shortCodes
```
vinaysshenoy: Rename this to `openPatientSearchScreen`?

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdScreenControllerTest.kt
```diff
@@ -166,6 +168,51 @@ class ScanSimpleIdScreenControllerTest {
...
+    verifyNoMoreInteractions(screen)
+  }
+
+  @Test
+  fun `when short code text changes, then hide validation error`() {
+    //given
+    val invalidShortCode = "3456"
+    val invalidShortCodeInput = ShortCodeInput(invalidShortCode)
+
+    uiEvents.onNext(ShortCodeSearched(invalidShortCodeInput))
```
vinaysshenoy: This is already being tested in the previous test. Is this necessary to add again here?

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenController.kt
```diff
@@ -64,6 +64,7 @@ class ScanSimpleIdScreenController @Inject constructor(
 
     val keyboardEvents = Observable
```
vinaysshenoy: What do you feel about extracting the shared keyboard event stream outside this method and providing it as a parameter to this function? That way, we can also handle the hiding of the QR Code scanner view in a separate method instead of in this method.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
```diff
@@ -81,11 +83,12 @@ class ScanSimpleIdScreen(context: Context, attrs: AttributeSet) : ConstraintLayo
   }
 
   fun showShortCodeValidationError() {
-    TODO("not implemented")
+    shortCodeErrorText.visibility = View.VISIBLE
+    shortCodeErrorText.text = resources.getString(R.string.scansimpleid_shortcode_error)
```
vinaysshenoy: Since the error text is static, we can set it as a resource in the layout and just hide/show it here.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
```diff
@@ -64,6 +59,12 @@ class ScanSimpleIdScreen(context: Context, attrs: AttributeSet) : ConstraintLayo
         .map(::ScanSimpleIdScreenQrCodeScanned)
   }
 
+  private fun doneClicks(): Observable<UiEvent> {
+    return RxTextView
+        .editorActionEvents(editText)
```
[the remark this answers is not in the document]
sanchita: We will fix these things in a different PR where we will focus on cleaning up the UI and hooking it to the business logic.

app/src/main/res/layout/screen_scan_simple.xml
sanchita: Good catch.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
sanchita: This will change once we implement another error message for empty short code.
Let's keep it this way for now.

app/src/main/res/layout/screen_scan_simple.xml
sanchita: It is safer to add `@+id`.

app/src/main/res/layout/screen_scan_simple.xml
ragunathjawahar: Yes, I'll remove the `maxEms` property. Will use the appropriate one in the next PR. This layout element will not be visible to the end user after the merge.

app/src/main/res/layout/screen_scan_simple.xml
ragunathjawahar: `@id/qrCodeScannerView` will give a compilation error if the element is missing from the layout whereas `@+id` wouldn't (I'm guessing). However, at least nowadays we use `@+id/` everywhere. @vinaysshenoy do you have any reasons why we should use it?

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenEvents.kt
```diff
@@ -17,3 +19,9 @@ sealed class ScanSimpleIdScreenPassportCodeScanned : UiEvent {
     override val analyticsName = "Scan Simple Card:Invalid BP passport code scanned"
   }
 }
+
+object ShortCodeSearched : ScanSimpleIdScreenEvent()
```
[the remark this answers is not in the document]
ragunathjawahar: Done

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenEvents.kt
```diff
@@ -3,11 +3,13 @@ package org.simple.clinic.scanid
 import org.simple.clinic.widgets.UiEvent
 import java.util.UUID
 
-data class ScanSimpleIdScreenQrCodeScanned(val text: String) : UiEvent {
+sealed class ScanSimpleIdScreenEvent : UiEvent
```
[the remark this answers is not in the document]
ragunathjawahar: Yes, that is correct

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenController.kt
ragunathjawahar: We did this because of the way how keyboard scanning is implemented in the controller now. It is just exposed as a stream of scan events. And the controller does not have access to a camera / QR code scanner interface.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenController.kt
ragunathjawahar: This is done.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenController.kt
```diff
@@ -39,21 +43,40 @@ class ScanSimpleIdScreenController @Inject constructor(
...
+    val showKeyboardEvents = events
+        .ofType<ShowKeyboardEvent>()
+
+    val hideKeyboardEvents = events
+        .ofType<HideKeyboardEvent>()
+
+    val keyboardEvents = Observable
+        .merge(showKeyboardEvents, hideKeyboardEvents)
+
+    val sharedScannedBpPassportCodeStream = events
```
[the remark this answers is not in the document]
ragunathjawahar: Yes, that's because of lines `94` and `97` (in the new screen). We have to share the stream because the first subscription will dispose itself as soon as the keyboard event stream emits. Line `97` will ensure that we utilize QR code scans only when the keyboard is not shown on the screen. This stream relies on a lot of state (previous and current) across streams and I am unable to see how to provide a cue about this stateful event stream.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenController.kt
ragunathjawahar: I prefer the `valve` operator, however that would require a refactor which we may not be able to pull off in this iteration.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenController.kt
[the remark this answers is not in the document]
vinaysshenoy: Right. Let's revisit this after we complete the feature.

app/src/main/res/layout/screen_scan_simple.xml
vinaysshenoy: Lint was flagging this as errors in my IDE, and they went away when I removed the extra `+` symbol.

Let's leave the `+` symbol in now and figure out what to do about this later.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenController.kt
vinaysshenoy: Sure, let's revisit this after the feature is complete.

app/src/test/java/org/simple/clinic/scanid/ShortCodeInputTest.kt
ragunathjawahar: Raised a story for this on Pivotal.
https://www.pivotaltracker.com/story/show/167992229

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenController.kt
ragunathjawahar: Changed on the upcoming PR branch.

app/src/main/java/org/simple/clinic/scanid/ShortCodeInput.kt
vinaysshenoy: Discussed offline: Will be done in the next PR.

PR #769, opened 2019-08-20
Integrate Mixpanel into the app

vinaysshenoy:
https://www.pivotaltracker.com/story/show/167965001

### DIFF
--- a/app/src/main/java/org/simple/clinic/ReleaseClinicApp.kt
+++ b/app/src/main/java/org/simple/clinic/ReleaseClinicApp.kt
@@ -2,6 +2,7 @@ package org.simple.clinic
 
 import android.annotation.SuppressLint
 import org.simple.clinic.analytics.HeapAnalyticsReporter
+import org.simple.clinic.analytics.MixpanelAnalyticsReporter
 import org.simple.clinic.di.AppComponent
 import org.simple.clinic.di.AppModule
 import org.simple.clinic.di.DaggerAppComponent
@@ -20,7 +21,10 @@ class ReleaseClinicApp : ClinicApp() {
   lateinit var syncIndicatorStatusCalculator: SyncIndicatorStatusCalculator
 
   override val analyticsReporters by unsafeLazy {
-    listOf(HeapAnalyticsReporter(this))
+    listOf(
+        HeapAnalyticsReporter(this),
+        MixpanelAnalyticsReporter(this)
+    )
   }
 
   override fun onCreate() {
--- a/app/src/main/java/org/simple/clinic/ReleaseClinicApp.kt
+++ b/app/src/main/java/org/simple/clinic/ReleaseClinicApp.kt
@@ -3,6 +3,7 @@ package org.simple.clinic
 import android.annotation.SuppressLint
 import org.simple.clinic.analytics.HeapAnalyticsReporter
 import org.simple.clinic.analytics.MixpanelAnalyticsReporter
+import org.simple.clinic.analytics.swallowErrors
 import org.simple.clinic.di.AppComponent
 import org.simple.clinic.di.AppModule
 import org.simple.clinic.di.DaggerAppComponent
@@ -22,8 +23,8 @@ class ReleaseClinicApp : ClinicApp() {
 
   override val analyticsReporters by unsafeLazy {
     listOf(
-        HeapAnalyticsReporter(this),
-        MixpanelAnalyticsReporter(this)
+        HeapAnalyticsReporter(this).swallowErrors(),
+        MixpanelAnalyticsReporter(this).swallowErrors()
     )
   }
 
--- a/app/src/main/java/org/simple/clinic/analytics/Analytics.kt
+++ b/app/src/main/java/org/simple/clinic/analytics/Analytics.kt
@@ -39,7 +39,7 @@ object Analytics {
   fun clearUser() {
     reporters.forEach {
       it.safely("Error clearing user ID!") {
-        resetUserIdentity()
+        resetUser()
       }
     }
   }
--- a/app/src/main/java/org/simple/clinic/analytics/Analytics.kt
+++ b/app/src/main/java/org/simple/clinic/analytics/Analytics.kt
@@ -8,6 +8,7 @@ import android.net.NetworkCapabilities.TRANSPORT_LOWPAN
 import android.net.NetworkCapabilities.TRANSPORT_VPN
 import android.net.NetworkCapabilities.TRANSPORT_WIFI
 import android.net.NetworkCapabilities.TRANSPORT_WIFI_AWARE
+import org.simple.clinic.user.User
 import org.threeten.bp.Duration
 import org.threeten.bp.Instant
 import java.util.UUID
@@ -27,17 +28,15 @@ object Analytics {
     reporters = reporters - reporter
   }
 
-  fun setUserId(uuid: UUID) {
+  fun setUser(user: User) {
     reporters.forEach {
-      it.safely("Error setting user ID!") {
-        val uuidString = uuid.toString()
-
-        setUserIdentity(uuidString)
+      it.safely("Error setting user!") {
+        setUserIdentity(user.uuid.toString())
       }
     }
   }
 
-  fun clearUserId() {
+  fun clearUser() {
     reporters.forEach {
       it.safely("Error clearing user ID!") {
         resetUserIdentity()
--- a/app/src/main/java/org/simple/clinic/analytics/Analytics.kt
+++ b/app/src/main/java/org/simple/clinic/analytics/Analytics.kt
@@ -36,6 +36,14 @@ object Analytics {
     }
   }
 
+  fun setLoggedInUser(user: User) {
+    reporters.forEach {
+      it.safely("Error setting logged in user!") {
+        setLoggedInUser(user)
+      }
+    }
+  }
+
   fun clearUser() {
     reporters.forEach {
       it.safely("Error clearing user ID!") {
--- a/app/src/main/java/org/simple/clinic/analytics/Analytics.kt
+++ b/app/src/main/java/org/simple/clinic/analytics/Analytics.kt
@@ -31,7 +31,15 @@ object Analytics {
   fun setLoggedInUser(user: User) {
     reporters.forEach {
       it.safely("Error setting logged in user!") {
-        setLoggedInUser(user)
+        setLoggedInUser(user, false)
+      }
+    }
+  }
+
+  fun setNewlyRegisteredUser(user: User) {
+    reporters.forEach {
+      it.safely("Error setting newly registered user!") {
+        setLoggedInUser(user, true)
       }
     }
   }
--- a/app/src/main/java/org/simple/clinic/analytics/Analytics.kt
+++ b/app/src/main/java/org/simple/clinic/analytics/Analytics.kt
@@ -29,73 +29,57 @@ object Analytics {
   }
 
   fun setLoggedInUser(user: User) {
-    reporters.forEach {
-      it.safely("Error setting logged in user!") {
-        setLoggedInUser(user, false)
-      }
-    }
+    reporters.forEach { it.setLoggedInUser(user, false) }
   }
 
   fun setNewlyRegisteredUser(user: User) {
-    reporters.forEach {
-      it.safely("Error setting newly registered user!") {
-        setLoggedInUser(user, true)
-      }
-    }
+    reporters.forEach { it.setLoggedInUser(user, true) }
   }
 
   fun clearUser() {
-    reporters.forEach {
-      it.safely("Error clearing user ID!") {
-        resetUser()
-      }
-    }
+    reporters.forEach(AnalyticsReporter::resetUser)
   }
 
   fun reportUserInteraction(name: String) {
-    reporters.forEach {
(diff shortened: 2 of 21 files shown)


app/src/main/java/org/simple/clinic/analytics/HeapAnalyticsReporter.kt
```diff
@@ -3,18 +3,20 @@ package org.simple.clinic.analytics
...
 
 class HeapAnalyticsReporter(context: Context, debug: Boolean = false) : AnalyticsReporter {
 
   init {
     Heap.init(context.applicationContext, BuildConfig.HEAP_ID, debug)
   }
 
-  override fun setUserIdentity(id: String) {
-    Heap.identify(id)
+  override fun setLoggedInUser(user: User, isANewRegistration: Boolean) {
```
ragunathjawahar: `isANewRegistration` seems unused?

vinaysshenoy: Weird, something seems to gone wrong when I rebased on `master`. Will check.

vinaysshenoy: Oh wait, nothing has gone wrong.

`isANewRegistration` is used only in `MixpanelAnalyticsReporter` because Heap doesn't need anything different to be done when a user signs up vs logs in. `Mixpanel` needs an additional call to be made for a sign up so this parameter is being used there.

vinaysshenoy: I did consider having two separate methods on the reporters themselves, but I decided not to since the `Analytics` does have two different methods for this use case, and no class will use the reporters directly. They will always call the `Analytics` methods themselves.

app/src/main/java/org/simple/clinic/analytics/Analytics.kt
```diff
@@ -27,20 +28,26 @@ object Analytics {
...
+      it.safely("Error setting logged in user!") {
+        setLoggedInUser(user, false)
+      }
+    }
+  }
 
-        setUserIdentity(uuidString)
+  fun setNewlyRegisteredUser(user: User) {
+    reporters.forEach {
+      it.safely("Error setting newly registered user!") {
```
ragunathjawahar: What does `safely` do?

app/src/main/java/org/simple/clinic/analytics/MixpanelAnalyticsReporter.kt
```diff
@@ -0,0 +1,39 @@
...
+
+  private val mixpanel: MixpanelAPI = MixpanelAPI.getInstance(app, BuildConfig.MIXPANEL_TOKEN)
+
+  override fun setLoggedInUser(user: User, isANewRegistration: Boolean) {
+    synchronized(mixpanel) {
+      val userId = user.uuid.toString()
+
+      if (isANewRegistration) {
+        mixpanel.alias(userId, null)
+        mixpanel.identify(userId)
```
ragunathjawahar: Can `mixpanel.identify(userId)` move out of the if block altogether?

app/src/main/java/org/simple/clinic/analytics/MixpanelAnalyticsReporter.kt
```diff
@@ -0,0 +1,39 @@
...
+      } else {
+        mixpanel.identify(userId)
+      }
+
+      mixpanel.people.identify(userId)
```
ragunathjawahar: Would you consider,
```
with(mixpanel.people) {
  identify(userId)
  set("name", user.fullName)
}
```

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -118,11 +121,20 @@ class UserSession @Inject constructor(
         .doOnSuccess { Timber.i("Login result: $it") }
   }
 
+  private fun reportUserLoggedInToAnalytics() {
+    loggedInUser()
+        .subscribeOn(schedulersProvider.io())
+        .take(1)
+        .filterAndUnwrapJust()
+        .map(Analytics::setLoggedInUser)
+        .subscribe()
```
ragunathjawahar: I do understand that `take(1)` should handle unsubscription, but a dangling `.subscribe()` looks suspicious.

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -215,6 +228,15 @@ class UserSession @Inject constructor(
         .doOnSuccess { Timber.i("Registration result: $it") }
   }
 
+  private fun reportUserRegisteredToAnalytics() {
+    loggedInUser()
+        .subscribeOn(schedulersProvider.io())
+        .take(1)
+        .filterAndUnwrapJust()
+        .map(Analytics::setNewlyRegisteredUser)
+        .subscribe()
```
ragunathjawahar: Same here. Dangling `subscribe` call.

app/src/test/java/org/simple/clinic/analytics/AnalyticsTest.kt
```diff
@@ -31,13 +35,18 @@ class AnalyticsTest {
   }
 
   @Test
-  fun `when setting the user id without any reporters, no error should be thrown`() {
-    Analytics.setUserId(UUID.randomUUID())
+  fun `when setting the logged in user without any reporters, no error should be thrown`() {
+    Analytics.setLoggedInUser(user)
```
ragunathjawahar: Can we do something like,

```
try {
  Analytics.setLoggedInUser(user)
} catch (exception: Exception) {
  fail("Should not throw an exception: $exception")
}
```

Just the function invocation looks a bit suspicious.

app/src/test/java/org/simple/clinic/analytics/AnalyticsTest.kt
```diff
@@ -31,13 +35,18 @@ class AnalyticsTest {
...
+  }
+
+  @Test
```
ragunathjawahar: Same for the rest of the test functions in the class.

app/src/test/java/org/simple/clinic/analytics/UpdateAnalyticsUserIdTest.kt
```diff
@@ -61,55 +60,18 @@ class UpdateAnalyticsUserIdTest {
...
     whenever(userSession.loggedInUser()).thenReturn(Observable.just(Just(user)))
 
-    updateAnalyticsUserId.listen(scheduler)
+    // when
+    updateAnalyticsUserId.listen()
 
+    // then
     if (shouldSetUserId) {
-      assertThat(reporter.userId).isEqualTo(user.uuid.toString())
+      assertThat(reporter.user).isEqualTo(user)
```
ragunathjawahar: Can we remove this conditional from the test?

app/src/test/java/org/simple/clinic/user/UserSessionTest.kt
```diff
@@ -219,6 +227,9 @@ class UserSessionTest {
       if (syncWillFail) Completable.error(RuntimeException()) else Completable.complete()
     }
     whenever(ongoingLoginEntryRepository.clearLoginEntry()).thenReturn(Completable.complete())
+    whenever(facilityRepository.associateUserWithFacilities(any(), any(), any())).thenReturn(Completable.complete())
```
ragunathjawahar: `any(), any(), any()` This is something we should get rid of eventually.

app/src/test/java/org/simple/clinic/user/UserSessionTest.kt
```diff
@@ -416,7 +438,7 @@ class UserSessionTest {
   ) {
     val currentUser = PatientMocker.loggedInUser(pinDigest = "old-digest", loggedInStatus = RESETTING_PIN)
     whenever(userDao.user()).thenReturn(Flowable.just(listOf(currentUser)))
-
+    whenever(facilityRepository.associateUserWithFacilities(any(), any(), any())).thenReturn(Completable.complete())
```
ragunathjawahar: Too much mocking to run a test is a sign of a not so good test. Should we add a maintenance task to review the design of the SUT? I guess you are already looking into refactoring this god object `UserSession`.

app/src/test/java/org/simple/clinic/user/UserSessionTest.kt
```diff
@@ -928,4 +954,127 @@ class UserSessionTest {
         )
     )
   }
+
+  @Test
+  fun `when the login happens, set the user information in analytics`() {
```
ragunathjawahar: Large test!

app/src/main/java/org/simple/clinic/analytics/Analytics.kt
vinaysshenoy: Analytics is something that is not crucial to using the app. In the unlikely case that reporting anything to analytics throws an exception for whatever reason, the app must proceed as if nothing has happened.

This block is designed to swallow that error and just report it to Sentry (or the crash reporting service).

I realise now that it can be done in a better way by having a decorator `AnalyticsReporter` that swallows the error and reports it to crash reporting instead. I'll make this change.

app/src/main/java/org/simple/clinic/analytics/MixpanelAnalyticsReporter.kt
vinaysshenoy: Yes, it can! I'll do this.

app/src/main/java/org/simple/clinic/user/UserSession.kt
vinaysshenoy: I agree. This is something that is a side-effect and should happen regardless if the source stream was disposed or not.

However, this could be overkill on my part, and I could just return a `Completable` instead and let it be part of the source stream instead of a side effect. What do you think?

app/src/main/java/org/simple/clinic/user/UserSession.kt
vinaysshenoy: Same answer as the previous comment. I'll do the same here.

app/src/test/java/org/simple/clinic/analytics/AnalyticsTest.kt
vinaysshenoy: Good point, I'll do this.

app/src/test/java/org/simple/clinic/analytics/UpdateAnalyticsUserIdTest.kt
vinaysshenoy: Yup, I'll try doing this.

app/src/test/java/org/simple/clinic/user/UserSessionTest.kt
vinaysshenoy: Yes, I'll fix this when I move login and registration out into their own components as discussed offline.

app/src/test/java/org/simple/clinic/user/UserSessionTest.kt
vinaysshenoy: Yup, when I refactor login and registration out of `UserSession`, I'll change the tests to be better.

app/src/test/java/org/simple/clinic/user/UserSessionTest.kt
vinaysshenoy: Yes, it's a consequence of how the login process is implemented. I'll try seeing if I can fix this. Otherwise, I'll improve it when breaking up `UserSession`.

vinaysshenoy: I'll do this when refactoring `UserSession`.

app/src/test/java/org/simple/clinic/analytics/UpdateAnalyticsUserIdTest.kt
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/ReleaseClinicApp.kt
```diff
@@ -20,7 +22,10 @@ class ReleaseClinicApp : ClinicApp() {
   lateinit var syncIndicatorStatusCalculator: SyncIndicatorStatusCalculator
 
   override val analyticsReporters by unsafeLazy {
-    listOf(HeapAnalyticsReporter(this))
+    listOf(
+        HeapAnalyticsReporter(this).swallowErrors(),
```
ragunathjawahar: Nice

app/src/main/java/org/simple/clinic/analytics/ErrorSwallowingReporter.kt
```diff
@@ -0,0 +1,48 @@
...
+
+import org.simple.clinic.user.User
+import timber.log.Timber
+
+/**
+ * A decorator [AnalyticsReporter] that swallows errors thrown by the
+ * wrapped [AnalyticsReporter] so that the app does not crash when
+ * there's an error reporting analytics events.
+ **/
+class ErrorSwallowingReporter(private val reporter: AnalyticsReporter) : AnalyticsReporter {
```
ragunathjawahar: Nice!

PR #770, opened 2019-08-20
Show search results loader

rakshakhegde:
https://www.pivotaltracker.com/story/show/167997753

### DIFF
--- a/app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchView.kt
+++ b/app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchView.kt
@@ -79,6 +79,7 @@ class PatientSearchView(context: Context, attrs: AttributeSet) : RelativeLayout(
       currentFacility: Facility
   ) {
     loader.visibleOrGone(isVisible = false)
+    newPatientContainer.visibleOrGone(isVisible = true)
     if (results.hasNoResults) {
       setEmptyStateVisible(true)
       adapter.update(emptyList())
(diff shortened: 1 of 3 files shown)


vinaysshenoy:
@rakshakhegde Are there designs for this story? Can we link them in the Pivotal story?

app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchView.kt
```diff
@@ -47,6 +48,7 @@ class PatientSearchView(context: Context, attrs: AttributeSet) : RelativeLayout(
   private val emptyStateView by bindView<View>(R.id.searchresults_empty_state)
   private val newPatientRationaleTextView by bindView<TextView>(R.id.searchresults_new_patient_rationale)
   private val newPatientButton by bindView<Button>(R.id.searchresults_new_patient)
+  private val loader by bindView<ProgressBar>(R.id.searchresults_loader)
```
vinaysshenoy: Switch to the Kotlin view extensions? We have been moving screens to use them as and when we need to change them.

app/src/main/res/layout/patient_search_view.xml
```diff
@@ -7,6 +7,12 @@
   tools:context="org.simple.clinic.searchresultsview.PatientSearchView"
   tools:parentTag="RelativeLayout">
 
+  <ProgressBar
```
vinaysshenoy: The "Register as a new patient" button is still visible while the search results are loading. If the results take more than a few seconds to load (as it will on production because of the amount of data), it is possible that a user could tap on the button and register a patient before the search results load completely.

We should hide the button as well until the results load.

rakshakhegde:
@vinaysshenoy I've discussed with Akshay, there are no designs on Zeplin for this. For now, a standard progressbar suffices.

app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchView.kt
rakshakhegde: Done.

PR #771, opened 2019-08-21
Update to Android Studio 3.5

vinaysshenoy:
This PR bumps the Gradle plugin to 3.5.0. In addition, it updates libraries which use KAPT (Dagger, Room) to the latest versions and enabled KAPT incremental annotation processing.

### DIFF
--- a/spellfix/.gitignore
+++ b/spellfix/.gitignore
@@ -1,2 +1,3 @@
 /build
 .externalNativeBuild
+.cxx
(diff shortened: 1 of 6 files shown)


spellfix/.gitignore
```diff
@@ -1,2 +1,3 @@
 /build
 .externalNativeBuild
+.cxx
```
sanchita: What is this for?

sanchita: Whats this for?

vinaysshenoy: The new gradle plugin generates a new kind of native build artifact that we should ignore.

vinaysshenoy: Just like `.externalNativeBuild`, it is autogenerated when the project is imported and should not be checked into VCS.

PR #776, opened 2019-08-26
Move camera scan pause/resume logic to view from controller

ragunathjawahar:


app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -85,16 +89,23 @@ class QrCodeScannerView(context: Context, attrs: AttributeSet) : FrameLayout(con
         codeScanner.errorCallback = null
       }
     }
+
+    return qrCodeScans
+        .withLatestFrom(qrCodeScansValve)
```
vinaysshenoy: Is there a reason we're doing this manually instead of using [valve](https://github.com/akarnokd/RxJavaExtensions#flowabletransformersvalve)?

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenController.kt
```diff
@@ -77,31 +77,11 @@ class ScanSimpleIdScreenController @Inject constructor(
   }
 
   private fun handleScannedBpPassportCodes(events: Observable<UiEvent>): Observable<UiChange> {
-    val showKeyboardEvents = events
```
vinaysshenoy: 🚀

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
ragunathjawahar: The valve, buffers events that are received when it is off and delivers it downstream when it is back on. What we wanted was to drop event when the valve is shut. There is a overloaded value with a buffer size, but it requires a non-zero value. Hence, this.

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -30,6 +32,8 @@ class QrCodeScannerView(context: Context, attrs: AttributeSet) : FrameLayout(con
   @Inject
   lateinit var lifecycle: Observable<TheActivityLifecycle>
 
+  private val qrCodeScansValve = PublishSubject.create<Boolean>()
```
vinaysshenoy: Since this is a `PublishSubject`, and we're using `withLatestFrom` to pause/resume the stream, the QR code scanning will never work unless we open and close the keyboard at least once. Should this be a `BehaviorSubject` with an initial value of `true`?

ragunathjawahar: It was a `BehaviorSubject` 😮 good catch BTW.

PR #779, opened 2019-08-27
Remove hardcoded timezone and locale in relative timestamp generation

vinaysshenoy:
https://www.pivotaltracker.com/story/show/167310592


app/src/main/java/org/simple/clinic/util/RelativeTimestampGenerator.kt
```diff
@@ -40,43 +45,28 @@ class RelativeTimestampGenerator @Inject constructor() {
...
-    return context.getString(R.string.timestamp_today)
+  companion object {
+    val timestampFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern("d MMM, yyyy", Locale.ENGLISH)
   }
-}
 
-object Yesterday : RelativeTimestamp() {
-  override fun displayText(context: Context): String {
-    return context.getString(R.string.timestamp_yesterday)
+  fun displayText(context: Context): String {
```
ragunathjawahar: Nice change!

app/src/androidTest/java/org/simple/clinic/util/RelativeTimestampGeneratorAndroidTest.kt
```diff
@@ -26,21 +29,22 @@ class RelativeTimestampGeneratorAndroidTest {
 
   @Test
   fun it_should_generate_timestamp_strings() {
+    val generator = RelativeTimestampGenerator()
     val todayDateTime = LocalDateTime.of(2018, 7, 29, 16, 58)
 
     val todayTime = LocalDateTime.of(2018, 7, 29, 2, 0, 0).atOffset(ZoneOffset.UTC).toInstant()
-    assertThat(generator.generate(todayDateTime, todayTime).displayText(context)).isNotEmpty()
+    assertThat(generator.generate(todayDateTime, todayTime).displayText(context, formatter)).isNotEmpty()
```
ragunathjawahar: `.isNotEmpty()` does not tell the whole story here. There is no means for the maintainer to know what the expectations are, and what is/was the behavior of these calls.

We need to assert them against actual string values, which I understand is coupled to the Android UI. Without those, these tests don't help a lot.

app/src/main/java/org/simple/clinic/summary/SummaryBloodPressureListItem.kt
```diff
@@ -25,11 +26,12 @@ import org.threeten.bp.format.DateTimeFormatter
 
 data class SummaryBloodPressureListItem(
     val measurement: BloodPressureMeasurement,
-    private val daysAgo: RelativeTimestamp,
     val showDivider: Boolean,
     val formattedTime: String?,
     val addTopPadding: Boolean,
-    val dateFormatter: DateTimeFormatter = RelativeTimestamp.timestampFormatter
+    private val daysAgo: RelativeTimestamp,
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```
ragunathjawahar: Curious about why we are doing this.

app/src/test/java/org/simple/clinic/summary/PatientSummaryScreenControllerTest.kt
```diff
@@ -183,7 +183,7 @@ class PatientSummaryScreenControllerTest {
 
     uiEvents.onNext(PatientSummaryScreenCreated(patientUuid, openIntention = intention, screenCreatedTimestamp = Instant.now(clock)))
 
-    verify(screen).populateList(eq(SummaryPrescribedDrugsItem(prescriptions)), any(), any(), any())
+    verify(screen).populateList(eq(SummaryPrescribedDrugsItem(prescriptions, dateFormatter)), any(), any(), any())
```
ragunathjawahar: Is this the reason?

app/src/androidTest/java/org/simple/clinic/util/RelativeTimestampGeneratorAndroidTest.kt
vinaysshenoy: That is true. The reason these tests were added initially was as a failing test to fix a bug which caused a crash when formatting an `Instant` (because an `Instant` by itself is not sufficient to format a local date, you also need a time zone). 

If we do check the string here, this test will be deeply coupled to the UI and will be brittle. I feel we can remove this test.

app/src/main/java/org/simple/clinic/util/RelativeTimestampGenerator.kt
```diff
@@ -27,11 +25,11 @@ class RelativeTimestampGenerator @Inject constructor() {
     val tomorrowAtMidnight = todayAtMidnight.plusDays(1)
 
     return when {
-      dateTime > tomorrowAtMidnight -> Future(time)
+      dateTime > tomorrowAtMidnight -> ExactDate(time)
```
ragunathjawahar: Nice!

app/src/main/java/org/simple/clinic/summary/SummaryBloodPressureListItem.kt
vinaysshenoy: Yes.

app/src/main/java/org/simple/clinic/util/RelativeTimestampGenerator.kt
```diff
@@ -40,6 +41,25 @@ class RelativeTimestampGenerator @Inject constructor() {
     val today = LocalDateTime.now(ZoneOffset.UTC)
     return generate(today, time)
   }
+
+  fun generate(instant: Instant, userClock: UserClock): RelativeTimestamp {
+    val then = instant.toLocalDateAtZone(userClock.zone)
```
ragunathjawahar: For me `day` has better readability than `then` in this entire function. Consider changing it?

vinaysshenoy:
@ragunathjawahar While I was doing this change, I realised that the `RelativeTimestamp` sealed class is mostly useless since it is not used for control flow anywhere. This can be replaced with a function that just returns a `String`, and we can test that function separately.

This is too big a change to introduce in this PR, so I added a pivotal story (https://www.pivotaltracker.com/story/show/168131168) to track this. We should probably do this while moving to the `Encounter` model since most of the places where this class is being used will be removed.

app/src/main/java/org/simple/clinic/util/RelativeTimestampGenerator.kt
vinaysshenoy: Sure.

app/src/test/java/org/simple/clinic/summary/RelativeTimestampGeneratorTest.kt
```diff
@@ -44,4 +51,87 @@ class RelativeTimestampGeneratorTest {
...
+    val timestampJustBeforeSixMonthPeriod = generator.generate(instantJustBeforeSixMonthPeriod, clock)
+    val timestampTwoYearsAway = generator.generate(instantTwoYearsAway, clock)
+
+    // then
+    assertThat(timestampJustBeforeSixMonthPeriod).isEqualTo(ExactDate(instantJustBeforeSixMonthPeriod))
+    assertThat(timestampTwoYearsAway).isEqualTo(ExactDate(instantTwoYearsAway))
+  }
+
+  @Test
+  fun `when the given timestamp is in the future, the relative timestamp generated must be ExactDate`() {
```
ragunathjawahar: These are nice looking tests

vinaysshenoy: 🙏

app/src/androidTest/java/org/simple/clinic/util/RelativeTimestampGeneratorAndroidTest.kt
ragunathjawahar: If this can be removed, then please remove it. Brittle tests are better than tests that don't add any value. I fail to see how these tests provide any value at the moment.

app/src/main/java/org/simple/clinic/util/RelativeTimestampGenerator.kt
vinaysshenoy: I renamed it to `date`.

PR #780, opened 2019-08-27
Show Shortcode Search results

sanchita:
Final PR for this story.
Pivotol story -> https://www.pivotaltracker.com/story/show/167310717

### DIFF
--- a/app/src/main/java/org/simple/clinic/ViewControllerBinding.kt
+++ b/app/src/main/java/org/simple/clinic/ViewControllerBinding.kt
@@ -92,6 +92,7 @@ class ViewControllerBinding<E, S, T>(
 
     disposable = uiEvents
         .compose(uiStateProducer)
+        .share()
         .doOnNext { state -> stateSubject.onNext(state) }
         .compose(uiChangeProducer)
         .subscribe { uiChange -> uiChange(v as T) }
--- a/app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchResults.kt
+++ b/app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchResults.kt
@@ -6,5 +6,9 @@ data class PatientSearchResults(
     val visitedCurrentFacility: List<PatientSearchResult>,
     val notVisitedCurrentFacility: List<PatientSearchResult>
 ) {
+  companion object {
+    fun emptyResults() = PatientSearchResults(emptyList(), emptyList())
+  }
+
   val hasNoResults = visitedCurrentFacility.isEmpty() && notVisitedCurrentFacility.isEmpty()
 }
--- a/app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchResults.kt
+++ b/app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchResults.kt
@@ -1,13 +1,15 @@
 package org.simple.clinic.searchresultsview
 
+import org.simple.clinic.facility.Facility
 import org.simple.clinic.patient.PatientSearchResult
 
 data class PatientSearchResults(
     val visitedCurrentFacility: List<PatientSearchResult>,
-    val notVisitedCurrentFacility: List<PatientSearchResult>
+    val notVisitedCurrentFacility: List<PatientSearchResult>,
+    val currentFacility: Facility?
 ) {
   companion object {
-    fun emptyResults() = PatientSearchResults(emptyList(), emptyList())
+    fun emptyResults() = PatientSearchResults(emptyList(), emptyList(), null)
   }
 
   val hasNoResults = visitedCurrentFacility.isEmpty() && notVisitedCurrentFacility.isEmpty()
--- a/app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchViewController.kt
+++ b/app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchViewController.kt
@@ -50,10 +50,9 @@ class PatientSearchViewController @Inject constructor(
 
     return searchResultsStream
         .compose(PartitionSearchResultsByVisitedFacility(bloodPressureDao, currentFacilityStream))
-        .withLatestFrom(currentFacilityStream)
-        .map { (results, currentFacility) ->
+        .map { results ->
           { ui: Ui ->
-            ui.updateSearchResults(results, currentFacility)
+            ui.updateSearchResults(results)
           }
         }
   }
--- a/app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
+++ b/app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
@@ -3,7 +3,40 @@ package org.simple.clinic.shortcodesearchresult
 import android.content.Context
 import android.util.AttributeSet
 import android.widget.RelativeLayout
+import kotlinx.android.synthetic.main.screen_shortcode_search_result.view.*
+import org.simple.clinic.R
+import org.simple.clinic.router.screen.ScreenRouter
+import org.simple.clinic.text.style.TextAppearanceWithLetterSpacingSpan
+import org.simple.clinic.util.Truss
+import org.simple.clinic.util.Unicode
+import javax.inject.Inject
 
 class ShortCodeSearchResultScreen(context: Context, attributes: AttributeSet) : RelativeLayout(context, attributes) {
 
+  @Inject
+  lateinit var screenRouter: ScreenRouter
+
+  override fun onFinishInflate() {
+    super.onFinishInflate()
+    setupToolBar()
+  }
+
+  private fun setupToolBar() {
+    val screenKey = screenRouter.key<ShortCodeSearchResultScreenKey>(this)
+    val shortCode = screenKey.shortCode
+
+    // This is guaranteed to be exactly 7 characters in length.
+    val prefix = shortCode.substring(0, 3)
+    val suffix = shortCode.substring(3)
+
+    val formattedShortCode = "$prefix${Unicode.nonBreakingSpace}$suffix"
+
+    val textSpacingSpan = TextAppearanceWithLetterSpacingSpan(context, R.style.Clinic_V2_TextAppearance_Body0Left_NumericBold_White100)
+
+    toolBar.title = Truss()
+        .pushSpan(textSpacingSpan)
+        .append(formattedShortCode)
+        .popSpan()
+        .build()
+  }
 }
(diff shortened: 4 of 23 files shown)


app/src/main/java/org/simple/clinic/ViewControllerBinding.kt
```diff
@@ -92,6 +92,7 @@ class ViewControllerBinding<E, S, T>(
 
     disposable = uiEvents
         .compose(uiStateProducer)
+        .share()
```
ragunathjawahar: @vinaysshenoy you may have to look into this. This fixes a problem where the upstream pipeline has multiple subscribers. In our case, a `doOnNext` call was invoked once for every subscription.

app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchViewController.kt
```diff
@@ -50,10 +50,9 @@ class PatientSearchViewController @Inject constructor(
 
     return searchResultsStream
         .compose(PartitionSearchResultsByVisitedFacility(bloodPressureDao, currentFacilityStream))
-        .withLatestFrom(currentFacilityStream)
-        .map { (results, currentFacility) ->
+        .map { results ->
           { ui: Ui ->
-            ui.updateSearchResults(results, currentFacility)
+            ui.updateSearchResults(results)
```
ragunathjawahar: @sanchita I'm trying to understand the decision to get rid of the `currentFacilityStream` can you please explain.

app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchResults.kt
```diff
@@ -1,10 +1,16 @@
...
     val visitedCurrentFacility: List<PatientSearchResult>,
-    val notVisitedCurrentFacility: List<PatientSearchResult>
+    val notVisitedCurrentFacility: List<PatientSearchResult>,
+    val currentFacility: Facility?
 ) {
+  companion object {
+    fun emptyResults() = PatientSearchResults(emptyList(), emptyList(), null)
+  }
+
   val hasNoResults = visitedCurrentFacility.isEmpty() && notVisitedCurrentFacility.isEmpty()
```
ragunathjawahar: This is a tricky line 🤔 It works because all the properties in the data class are immutable. If that contract changes, `hasNoResults` will have its property set with the values that were used to create this class initially. A better way to protect against subtle bugs would be to make this a computed property.
```
val hasNoResults: Boolean
    get() = visitedCurrentFacility.isEmpty() && notVisitedCurrentFacility.isEmpty()
```

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+    hideKeyboard()
+
+    TheActivity.component.inject(this)
+
+    val screenKey = screenRouter.key<ShortCodeSearchResultScreenKey>(this)
+    setupToolBar(screenKey)
+    setupScreen()
+
+    val uiStateProducer = ShortCodeSearchResultStateProducer(
+        shortCode = screenKey.shortCode,
```
ragunathjawahar: The only reason we do too many field injections in this call is due to this dynamic dependency. If we create a submodule that accepts this value as a parameter. we should be able to get away with,

```
@Inject
lateinit var uiStateProducer: ShortCodeSearchResultStateProducer

@Inject
lateinit var uiChangeProducer: ShortCodeSearchResultUiChangeProducer
```

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+        patientRepository = patientRepository,
+        userSession = userSession,
+        facilityRepository = facilityRepository,
+        bloodPressureDao = bloodPressureDao,
+        ui = this,
+        schedulersProvider = schedulersProvider
+    )
+    binding = ViewControllerBinding.bindToView(this, uiStateProducer, uiChangeProducer)
```
ragunathjawahar: Extract this and the previous statement into a function? `setupControllerBinding`?

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+    if (::disposable.isInitialized) disposable.dispose()
+    super.onDetachedFromWindow()
+  }
+
+  private fun setupToolBar(screenKey: ShortCodeSearchResultScreenKey) {
+    val shortCode = screenKey.shortCode
+
+    // This is guaranteed to be exactly 7 characters in length.
+    val prefix = shortCode.substring(0, 3)
+    val suffix = shortCode.substring(3)
```
ragunathjawahar: Use a `ShortCodeFormatter` perhaps?

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+
+    val formattedShortCode = "$prefix${Unicode.nonBreakingSpace}$suffix"
+
+    val textSpacingSpan = TextAppearanceWithLetterSpacingSpan(context, R.style.Clinic_V2_TextAppearance_Body0Left_NumericBold_White100)
+
+    toolBar.title = Truss()
+        .pushSpan(textSpacingSpan)
+        .append(formattedShortCode)
+        .popSpan()
+        .build()
```
ragunathjawahar: All of these could go into a formatter.

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+
+    toolBar.setNavigationOnClickListener {
```
ragunathjawahar: val popScreenAction = { screenRouter.pop() }

with(toolBar) {
  setNavigationOnClickListener(popScreenAction)
  setOnClickListener(popScreenAction)
}

Something of this sort?

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+  override fun hideLoading() {
+    loader.visibility = View.GONE
+  }
+
+  override fun showSearchResults(foundPatients: PatientSearchResults) {
+    val items = SearchResultsItemType
+        .from(foundPatients)
+    items.map {
+      it.uiEvents = downstreamEvents
+    }
```
ragunathjawahar: We could use this to achieve what you are trying to achieve.

```kotlin
val items = SearchResultsItemType
        .from(foundPatients)
        .onEach { it.uiEvents = downstreamEvents }
```

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultState.kt
```diff
@@ -1,28 +1,27 @@
...
     val fetchPatientsAsyncOp: AsyncOp,
-    val patients: List<PatientSearchResultUiState>
+    val patients: PatientSearchResults
 ) {
   companion object {
     fun fetchingPatients(
         bpPassportNumber: String
     ): ShortCodeSearchResultState =
-        ShortCodeSearchResultState(bpPassportNumber, IN_FLIGHT, emptyList())
+        ShortCodeSearchResultState(bpPassportNumber, IN_FLIGHT, PatientSearchResults.emptyResults())
```
ragunathjawahar: Can we make this a constant property instead of a function invocation?

`PatientSearchResults.EMPTY_RESULTS`

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultStateProducer.kt
```diff
@@ -26,21 +35,35 @@ class ShortCodeSearchResultStateProducer(
...
-              .withLatestFrom(states) { patientSearchResults, state ->
-                if (patientSearchResults.isNotEmpty()) {
-                  state.patientsFetched(patientSearchResults)
-                } else {
-                  state.noMatchingPatients()
-                }
-              }
+              .searchByShortCode(shortCode)
+              .subscribeOn(schedulersProvider.io())
+        }
```
ragunathjawahar: `.share()`, because the database will be queried twice (giving out the same result hopefully, potential race-condition here) since the stream is split into two.

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultUiChangeProducer.kt
```diff
@@ -34,7 +36,7 @@ class ShortCodeSearchResultUiChangeProducer(
       states: Observable<ShortCodeSearchResultState>
   ): Observable<(ShortCodeSearchResultUi) -> Unit> {
     return states
-        .filter { it.fetchPatientsAsyncOp == SUCCEEDED && it.patients.isNotEmpty() }
+        .filter { it.fetchPatientsAsyncOp == SUCCEEDED && it.patients.hasNoResults.not() }
```
ragunathjawahar: More reasons why `hasNoResults` should be a computed property.

ragunathjawahar:
@sanchita @vinaysshenoy I accidentally approved this PR instead of requesting changes. How do we deal with this?

vinaysshenoy:
I suggest we make a new PR fixing the biggest concerns with the merged PR
immediately, then put the not so important changes into the backlog.

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
sanchita: Yes. This is what I was looking for.

app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchViewController.kt
sanchita: I refactored the class `PatientSearchResults` to capture all the required data for generating the partitioned search results by facility. That was the whole purpose of that class in the first place. I added `currentFacility` in that class itself and removed it from `SearchResultsItemType.from(patientSearchResults)`. Hence this refactor.

I think it would also make sense to rename `PatientSearchResults` to `PatinetSearchResultsByFacility` or something like that.
Wdyt?

app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchResults.kt
sanchita: Good catch. I will make this change.

PR #782, opened 2019-08-27
Format short code in the same format as its hint

ragunathjawahar:
In the short code input field, the hint is formatted in the following way - `123 4567`. This PR adds the same kind of formatting for the text that is being entered inside the `EditText`.

### DIFF
--- a/app/src/main/java/org/simple/clinic/scanid/ui/ShortCodeSpanWatcher.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ui/ShortCodeSpanWatcher.kt
@@ -1,0 +1,23 @@
+package org.simple.clinic.scanid.ui
+
+import android.text.Editable
+import android.text.Spannable
+import org.simple.clinic.text.style.TrailingSpaceForCharacterSpan
+import org.simple.clinic.util.TextWatcherAdapter
+
+private const val CHARS_REQUIRED_TO_INSERT_SPACE = 4
+private const val INSERT_SPACE_AT_INDEX = 2
+
+class ShortCodeSpanWatcher : TextWatcherAdapter() {
+    private val trailingSpaceSpan = TrailingSpaceForCharacterSpan()
+
+    override fun afterTextChanged(editable: Editable) {
+        with(editable) {
+            removeSpan(trailingSpaceSpan)
+
+            if (length >= CHARS_REQUIRED_TO_INSERT_SPACE) {
+                setSpan(trailingSpaceSpan, INSERT_SPACE_AT_INDEX, INSERT_SPACE_AT_INDEX + 1, Spannable.SPAN_INCLUSIVE_EXCLUSIVE)
+            }
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/scanid/ui/TextWatcherAdapter.kt
```diff
@@ -0,0 +1,18 @@
+package org.simple.clinic.scanid.ui
+
+import android.text.Editable
+import android.text.TextWatcher
+
+abstract class TextWatcherAdapter : TextWatcher {
```
vinaysshenoy: Since this is not specifically related to the Scan BP Passport feature, should we move it to a util package?

app/src/main/java/org/simple/clinic/scanid/ui/ShortCodeSpanWatcher.kt
```diff
@@ -0,0 +1,56 @@
...
+import android.graphics.Paint
+import android.text.Editable
+import android.text.Spannable
+import android.text.style.ReplacementSpan
+import kotlin.math.roundToInt
+
+private const val CHARS_REQUIRED_TO_INSERT_SPACE = 4
+private const val INSERT_SPACE_AT_INDEX = 2
+
+class ShortCodeSpanWatcher : TextWatcherAdapter() {
```
vinaysshenoy: Nice!

app/src/main/java/org/simple/clinic/scanid/ui/ShortCodeSpanWatcher.kt
```diff
@@ -0,0 +1,56 @@
...
+                setSpan(activeSpan, INSERT_SPACE_AT_INDEX, INSERT_SPACE_AT_INDEX + 1, Spannable.SPAN_INCLUSIVE_EXCLUSIVE)
+            } else {
+                removeSpan(activeSpan)
+                activeSpan = null
+            }
+        }
+    }
+}
+
+private class AddTrailingSpaceSpan : ReplacementSpan() {
```
vinaysshenoy: This is something that can be used even when formatting the title for the Scan BP passport screens. Can we break this out into a separate class? We already have a package where we keep our custom spans. (`org.simple.clinic.text.style`).

app/src/main/java/org/simple/clinic/scanid/ui/ShortCodeSpanWatcher.kt
```diff
@@ -0,0 +1,56 @@
...
+    override fun getSize(
+        paint: Paint,
+        text: CharSequence?,
+        start: Int,
+        end: Int,
+        fm: Paint.FontMetricsInt?
+    ): Int {
+        text ?: return 0
+        return paint.measureText("${text.substring(start, end)} ").roundToInt()
```
vinaysshenoy: Can we make this a non-breaking space? Check out the file `org.simple.clinic.util.Unicode` where we already have the character defined.

ragunathjawahar: We don't need a non-breaking space here, because we are only measuring the 'width' of a space character. The space character itself doesn't get rendered on the screen. We use it to decide how much area on the canvas we need to render the character. Concerns?

vinaysshenoy: Makes sense.

PR #785, opened 2019-08-29
Remove classes related to fuzzy search v1

vinaysshenoy:
https://www.pivotaltracker.com/story/show/164658189


app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -17,17 +17,6 @@ import org.threeten.bp.Instant
 import org.threeten.bp.LocalDate
 import java.util.UUID
 
-/**
```
vinaysshenoy: This block was removed and inlined in a Migration, which was the only place where we needed to retain it.

app/src/main/java/org/simple/clinic/patient/sync/PatientPayload.kt
```diff
@@ -62,35 +59,24 @@ data class PatientPayload(
...
         dateOfBirth = dateOfBirth,
-        age = age?.let {
-          Age(age, ageUpdatedAt!!, computedDateOfBirth(age, ageUpdatedAt))
-        },
+        age = age?.let { Age(age, ageUpdatedAt!!) },
         status = status,
         createdAt = createdAt,
         updatedAt = updatedAt,
         deletedAt = deletedAt,
-        // recordedAt should be changed here when PatientPayload starts receiving this field from server
```
vinaysshenoy: The server now sends the `recordedAt` property in all patient payloads, so we no longer have to retain it as a nullable property and add a fallback.

PR #786, opened 2019-08-30
Improve sync triggers and setup

vinaysshenoy:
https://www.pivotaltracker.com/story/show/165992187

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -232,7 +232,6 @@ dependencies {
   implementation "androidx.room:room-runtime:$versions.room"
   kapt "androidx.room:room-compiler:$versions.room"
   implementation "androidx.room:room-rxjava2:$versions.room"
-  implementation "android.arch.work:work-runtime-ktx:$versions.workRuntimeKtx"
   implementation "com.google.android.gms:play-services-location:$versions.playServicesLocation"
   implementation "com.google.firebase:firebase-core:$versions.firebaseCore"
   implementation "com.google.firebase:firebase-config:$versions.firebaseConfig"
@@ -288,6 +287,9 @@ dependencies {
   implementation "com.simplecityapps:recyclerview-fastscroll:$versions.fastScroll"
   implementation "com.google.android:flexbox:$versions.flexbox"
   implementation "com.mixpanel.android:mixpanel-android:$versions.mixpanel"
+
+  implementation "androidx.work:work-runtime:$versions.workManager"
+  implementation "androidx.work:work-gcm:$versions.workManager"
 }
 
 // This must always be present at the bottom of this file, as per:
--- a/app/src/debug/java/org/simple/clinic/DebugNotification.kt
+++ b/app/src/debug/java/org/simple/clinic/DebugNotification.kt
@@ -55,7 +55,7 @@ class DebugNotificationActionReceiver : BroadcastReceiver() {
   override fun onReceive(context: Context?, intent: Intent?) {
     DebugClinicApp.appComponent().inject(this)
 
-    dataSync.sync(null)
+    dataSync.syncTheWorld()
         .subscribeOn(io())
         .subscribe()
   }
--- a/app/src/debug/java/org/simple/clinic/DebugNotification.kt
+++ b/app/src/debug/java/org/simple/clinic/DebugNotification.kt
@@ -10,7 +10,6 @@ import android.os.Build
 import androidx.core.app.NotificationCompat
 import io.reactivex.schedulers.Schedulers.io
 import org.simple.clinic.sync.DataSync
-import timber.log.Timber
 import javax.inject.Inject
 
 object DebugNotification {
@@ -56,7 +55,6 @@ class DebugNotificationActionReceiver : BroadcastReceiver() {
   override fun onReceive(context: Context?, intent: Intent?) {
     DebugClinicApp.appComponent().inject(this)
 
-    Timber.tag("SyncWork").i("Sync from DebugNotification")
     dataSync.sync(null)
         .subscribeOn(io())
         .subscribe()
--- a/app/src/debug/java/org/simple/clinic/DebugNotification.kt
+++ b/app/src/debug/java/org/simple/clinic/DebugNotification.kt
@@ -10,6 +10,7 @@ import android.os.Build
 import androidx.core.app.NotificationCompat
 import io.reactivex.schedulers.Schedulers.io
 import org.simple.clinic.sync.DataSync
+import timber.log.Timber
 import javax.inject.Inject
 
 object DebugNotification {
@@ -55,6 +56,7 @@ class DebugNotificationActionReceiver : BroadcastReceiver() {
   override fun onReceive(context: Context?, intent: Intent?) {
     DebugClinicApp.appComponent().inject(this)
 
+    Timber.tag("SyncWork").i("Sync from DebugNotification")
     dataSync.sync(null)
         .subscribeOn(io())
         .subscribe()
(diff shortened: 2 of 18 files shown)


app/build.gradle
```diff
@@ -288,6 +287,9 @@ dependencies {
   implementation "com.simplecityapps:recyclerview-fastscroll:$versions.fastScroll"
   implementation "com.google.android:flexbox:$versions.flexbox"
   implementation "com.mixpanel.android:mixpanel-android:$versions.mixpanel"
+
+  implementation "androidx.work:work-runtime:$versions.workManager"
+  implementation "androidx.work:work-gcm:$versions.workManager"
```
ragunathjawahar: Curious,

1. Why are we removing the KTX version.
2. Why a GCM artifact?

app/src/main/java/org/simple/clinic/sync/SyncScheduler.kt
```diff
@@ -22,34 +22,62 @@ class SyncScheduler @Inject constructor(
...
+  }
+
+  /*
+   * This is meant to cancel the old periodic work that
+   * was scheduled and persisted before we moved to the
+   * "unique" work system.
+   * TODO 2019-09-02: Remove once the unique work feature has been deployed to enough devices
+   **/
+  private fun cancelPreviouslyScheduledPeriodicWork() {
+    workManager.cancelAllWorkByTag("patient-sync")
```
ragunathjawahar: Hard coded tag?

app/build.gradle
vinaysshenoy: 1. [Work Runtime KTX](https://developer.android.com/kotlin/ktx#workmanager) only adds support for coroutines, which we do not use at the moment. I felt we could use the core library instead.

2. For the GCM artifact, please see the release notes linked here: https://developer.android.com/jetpack/androidx/releases/work#2.2.0

app/src/main/java/org/simple/clinic/sync/SyncWorker.kt
```diff
@@ -11,7 +11,6 @@ import javax.inject.Inject
 class SyncWorker(context: Context, private val workerParams: WorkerParameters) : Worker(context, workerParams) {
 
   companion object {
-    const val TAG = "patient-sync"
     private const val NO_GROUP = "no-group-id"
     private const val SYNC_GROUP = "sync_group_id"
```
ragunathjawahar: Underscores vs hyphens

app/src/main/java/org/simple/clinic/sync/SyncScheduler.kt
vinaysshenoy: Yes, this is temporary and will be removed once enough users migrate to the newer build.

app/src/main/java/org/simple/clinic/sync/SyncDataOnApproval.kt
```diff
@@ -11,9 +13,12 @@ class SyncDataOnApproval @Inject constructor(
 
   override fun sync() {
     val shouldSync = userSession
-        .canSyncData()
-        .distinctUntilChanged()
-        .filter { canSyncData -> canSyncData }
+        .loggedInUser()
+        .filterAndUnwrapJust()
+        .map { it.status }
+        .buffer(2, 1)
```
ragunathjawahar: Why `buffer(2, 1)`

app/src/main/java/org/simple/clinic/sync/SyncDataOnApproval.kt
```diff
@@ -11,9 +13,12 @@ class SyncDataOnApproval @Inject constructor(
...
+        .filter { (previousStatus, currentStatus) -> currentStatus == UserStatus.ApprovedForSyncing && previousStatus != UserStatus.ApprovedForSyncing }
+        .map { Any() }
```
ragunathjawahar: `Unit` instead?

app/src/main/java/org/simple/clinic/sync/SyncDataOnApproval.kt
vinaysshenoy: Because we specifically want a sliding window of the last two changes to the `loggedInStatus`.

app/src/main/java/org/simple/clinic/sync/SyncScheduler.kt
```diff
@@ -20,36 +20,60 @@ class SyncScheduler @Inject constructor(
...
+  }
+
+  private fun scheduleWorkRequests(workRequests: List<Pair<PeriodicWorkRequest, String>>): Completable {
+    return Completable.fromAction {
+
+      workRequests.forEach { (workRequest, workName) ->
```
ragunathjawahar: Drop the prefixes from the lambda parameters?

app/src/main/java/org/simple/clinic/sync/SyncScheduler.kt
vinaysshenoy: This is an implementation detail because the tag is persisted to WorkManager's internal work database, so we need to keep this string somewhere. However, we do not want to make this available to the rest of the codebase.

app/src/test/java/org/simple/clinic/sync/SyncDataOnApprovalTest.kt
```diff
@@ -28,18 +40,40 @@ class SyncDataOnApprovalTest {
...
-      verify(dataSync).sync(null)
-    } else {
-      verify(dataSync, never()).sync(null)
-    }
+    // then
+    userSubject.onNext(user.toOptional())
+    verifyZeroInteractions(dataSync)
+
+    userSubject.onNext(user.withStatus(ApprovedForSyncing, clock).toOptional())
+    verify(dataSync).sync(null)
```
ragunathjawahar: Curious, why does `sync` take in `null`?

app/src/main/java/org/simple/clinic/sync/SyncDataOnApproval.kt
vinaysshenoy: To make it more clear, we want to do this only at the specific moment where the user's logged in status changes from anything to `ApprovedForSyncing`. However, since this is observing the database for changes, whenever the user model gets saved, this will emit.

So by buffering 2 emissions and skipping 1 value instead of 2, we get a sliding window of size 2 of the changes in the logged in status and we can explicitly check for the case where a user becomes `ApprovedForSyncing` from some other state.

app/src/test/java/org/simple/clinic/sync/SyncDataOnApprovalTest.kt
vinaysshenoy: This is bad API design which we haven't gotten around to fixing yet. `null` here is a sentinel value which means "sync everything" instead of a specific sync group.

vinaysshenoy: I can take a stab at fixing this.

app/src/main/java/org/simple/clinic/sync/SyncDataOnApproval.kt
vinaysshenoy: Added a comment explaining this.

app/src/test/java/org/simple/clinic/sync/SyncDataOnApprovalTest.kt
vinaysshenoy: Done.

vinaysshenoy:
@ragunathjawahar Feedback addressed.

app/src/debug/java/org/simple/clinic/DebugNotification.kt
```diff
@@ -55,7 +55,7 @@ class DebugNotificationActionReceiver : BroadcastReceiver() {
   override fun onReceive(context: Context?, intent: Intent?) {
     DebugClinicApp.appComponent().inject(this)
 
-    dataSync.sync(null)
+    dataSync.syncTheWorld()
```
ragunathjawahar: 😆

vinaysshenoy: 🌏

PR #794, opened 2019-09-05
Fix usages of UTC clock when working with appointments

vinaysshenoy:
https://www.pivotaltracker.com/story/show/168165692


app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
```diff
@@ -84,7 +86,7 @@ class AppointmentRepository @Inject constructor(
     return Completable.fromAction {
       appointmentDao.markAsAgreedToVisit(
           appointmentUUID = appointmentUuid,
-          reminderDate = LocalDate.now(utcClock).plusDays(30),
+          reminderDate = LocalDate.now(userClock).plusDays(30),
```
sanchita: Can you get the `clock` from the caller itself rather than injecting in the repository?

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -419,15 +419,15 @@ class AppointmentRepositoryAndroidTest {
...
+    userClock.setDate(LocalDate.parse("2018-01-31"))
 
     // when
     appointmentRepository.markAsAgreedToVisit(appointmentUuid).blockingAwait()
 
     // then
     val appointmentUpdatedAtTimestamp = Instant.parse("2018-01-01T00:00:01Z")
     with(getAppointmentByUuid(appointmentUuid)) {
-      assertThat(remindOn).isEqualTo(LocalDate.parse("2018-02-04"))
+      assertThat(remindOn).isEqualTo(LocalDate.parse("2018-02-28"))
```
sanchita: Good choice sample data.

app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
vinaysshenoy: Yup. 

Following up on our conversation offline, it makes sense to move this changing of reminder date to the controller and let the repository handle only setting the record timestamps (`createdAt`, `updatedAt`, and `deletedAt`). I have raised a pivotal story for this [here](https://www.pivotaltracker.com/story/show/168449411) and I'll get to it at some point.

PR #795, opened 2019-09-05
Show Simple video on Patient screen

sanchita:
https://www.pivotaltracker.com/story/show/168190582


app/src/main/res/layout/screen_patients.xml
```diff
@@ -54,53 +91,31 @@
         app:drawableTint="@color/white100"
         tools:ignore="MissingPrefix,RtlSymmetry" />
 
-      <ImageView
-        android:id="@+id/patients_record_bp_illustration"
+      <org.simple.clinic.widgets.ViewFlipperWithDebugPreview
```
ragunathjawahar: Curious, what does `ViewFlipperWithDebugPreview` do?

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
```diff
@@ -93,10 +94,17 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
         controller = controller,
         screenDestroys = RxView.detaches(this).map { ScreenDestroyed() }
     )
-
     illustrationImageView.setImageResource(illustrationResourceId())
   }
 
+  private fun showIllustration(@IdRes viewId: Int) {
+    illustrationFlipper.apply {
+      val indexOfChildId = indexOfChildId(viewId)
```
ragunathjawahar: Can this be,
```kotlin
illustrationFlipper.apply { 
  displayedChild = indexOfChildId(viewId)
}
```

app/src/main/res/layout/screen_patients.xml
ragunathjawahar: Can we rename `ViewFlipperWithDebugPreview`, something like `SimpleViewFlipper` should do. WDYT?

app/src/main/res/values/strings.xml
```diff
@@ -436,6 +436,9 @@
 
   <!-- Short code search -->
   <string name="shortcodesearchresult_enter_patient_name_button">Enter Patient\'s name</string>
+
+  <!-- Simple video -->
   <string name="simple_video_watch_video">Watch Video</string>
   <string name="simple_video_how_to_use_the_simple_app">How to use the Simple app</string>
+  <string name="simple_video_duration">5:07 minutes</string>
```
ragunathjawahar: Should we split the duration from minutes? If not, this could make i18n complicated if the duration of the video changes?

app/src/test/java/org/simple/clinic/home/patients/PatientsScreenControllerTest.kt
```diff
@@ -434,6 +434,37 @@ class PatientsScreenControllerTest {
     }
   }
 
+  @Test
+  fun `when screen is created then display simple video if patient registered count is less than 10`() {
```
ragunathjawahar: Nice!

app/src/test/java/org/simple/clinic/home/patients/PatientsScreenControllerTest.kt
```diff
@@ -434,6 +434,37 @@ class PatientsScreenControllerTest {
...
+    //when
+    uiEvents.onNext(ScreenCreated())
+
+    //then
+    verify(screen).showSimpleVideo()
+    verify(screen, never()).showIllustration()
+  }
+
+  @Test
+  fun `when screen is created then display illustration if patient registered count is exceeds 10`() {
```
ragunathjawahar: Nice!

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
```diff
@@ -59,17 +57,8 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
   @Inject
   lateinit var userClock: UserClock
 
-  private val searchButton by bindView<Button>(R.id.patients_search_patients)
```
ragunathjawahar: Nice!

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
```diff
@@ -5,17 +5,16 @@ import android.content.Context
...
-import android.widget.ImageView
 import android.widget.RelativeLayout
-import android.widget.ViewFlipper
 import androidx.annotation.IdRes
 import androidx.interpolator.view.animation.FastOutSlowInInterpolator
 import com.jakewharton.rxbinding2.view.RxView
 import io.reactivex.Observable
 import io.reactivex.disposables.Disposables
 import io.reactivex.rxkotlin.ofType
-import kotterknife.bindView
```
ragunathjawahar: Nice!

app/src/main/res/layout/screen_patients.xml
vinaysshenoy: May I suggest `ViewFlipperWithLayoutEditorPreview` instead? This makes it explicit why the class exists, otherwise `SimpleViewFlipper` might indicate that the class has something to do with the domain itself.

vinaysshenoy: Also, the fact that this needs to be done us is so indicative of the general quality of the Android UI framework 😥

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
```diff
@@ -57,6 +63,9 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
   @Inject
   lateinit var userClock: UserClock
 
+  @field:[Inject Named("training_video_youtube_id")]
```
ragunathjawahar: Curious,

Wouldn't
```kotlin
@Inject
@Named("training_video_youtube_id")
lateinit var youTubeVideoId: String 
```

suffice?

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
```diff
@@ -213,10 +222,29 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
   }
 
   fun showSimpleVideo() {
+    //Hard-coding to show this simple video view exists because, as of now,
+    // we are not sure if we will have variations of this training video.
+    // We should improve this implementation later.
```
ragunathjawahar: Can you also mention how we can improve to convey the kind of improvement that we should make? So the maintainer could have some context about the problem that we are attempting to solve?

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
```diff
@@ -213,10 +222,29 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
...
+      openYouTubeLinkForSimpleVideo()
+    }
   }
 
   fun showIllustration() {
     showHomeScreenBackground(homeIllustration.id)
   }
+
+  private fun openYouTubeLinkForSimpleVideo() {
+    val appIntent = Intent(Intent.ACTION_VIEW, Uri.parse("vnd.youtube:$youTubeVideoId"))
```
ragunathjawahar: We should resolve these intents using the `PackageManager` one after the other, in case we are unable to resolve the first intent. It's also not safe to assume the user will have access to the browser. Because with multiple user accounts, it is possible that the browser is not available.

It's okay to now solve for this case, but if the browser is not available we should at least log to sentry so that we can decide if we should handle this case later.

app/src/main/java/org/simple/clinic/patient/SimpleVideoModule.kt
```diff
@@ -0,0 +1,33 @@
...
+import dagger.Module
+import dagger.Provides
+import java.util.Locale
+import javax.inject.Named
+
+// SimpleVideoModule class is used to show a training video on home screen.
+// The URL is hard-coded right now which means the video view on UI is fixed as well.
+// This includes the video title, thumbnail and duration.
+// At this point, we are not sure what this training video section would look like in the future
+// so we are keeping this implementation as simple as possible.
```
ragunathjawahar: 🤔 not sure if we need this comment.

app/src/main/res/layout/view_simple_video.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
-  android:id="@+id/simpleVideoLayout"
+<RelativeLayout android:id="@+id/simpleVideoLayout"
```
ragunathjawahar: Place `android:id` after the XML namespace attribute?

app/src/main/res/layout/view_simple_video.xml
```diff
@@ -15,7 +15,7 @@
     android:layout_marginStart="@dimen/spacing_24"
     android:layout_marginTop="@dimen/spacing_24"
     android:src="@drawable/thumbnail_simple_video"
-    tools:ignore="ContentDescription" />
+  tools:ignore="ContentDescription" />
```
ragunathjawahar: Formatting issue?

app/src/main/res/layout/screen_patients.xml
ragunathjawahar: @vinaysshenoy yes, the `Debug` gives a _sense_ that this component should not be part of the production code. Which it is not.

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
```diff
@@ -204,14 +202,49 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
...
+    syncIndicator.visibility = View.VISIBLE
   }
 
   fun showAppUpdateDialog() {
     AppUpdateDialog.show(activity.supportFragmentManager)
   }
+
+  private fun showHomeScreenBackground(@IdRes viewId: Int) {
+    illustrationLayout.apply {
+      val indexOfChildId = indexOfChildId(viewId)
```
ragunathjawahar: Can this just be,
```kotlin
illustrationLayout.apply {
  displayedChild = indexOfChildId(viewId) 
}
```

app/src/main/res/values/strings.xml
sanchita: Good point!

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
sanchita: I actually added this to avoid multiple calls to updated `displayedChild`.
But I realized that this will only happen once at the time of creating this view, so its fine.

Will make this change.

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
sanchita: This doesn't work in Kotlin.
For field injection, we need to specify `@field` annotation because a property in Kotlin is backed by `set()` and `get()` and Dagger expects a field and not a property.
Read [here](https://medium.com/@WindRider/correct-usage-of-dagger-2-named-annotation-in-kotlin-8ab17ced6928)

app/src/main/java/org/simple/clinic/patient/SimpleVideoModule.kt
sanchita: Maybe I should be more clear in this comment.
I wanted to explain that the video url and other things should be configurable as well.
But the reason they are not is because we want to keep this as easy at the moment as possible.
I will update this comment.

app/src/main/res/layout/view_simple_video.xml
sanchita: Auto format does this. Since in other xmls in the project, we are following the same formatting, I am going to keep it this way for now.
Maybe we can update our formatter.

app/src/main/java/org/simple/clinic/patient/SimpleVideoModule.kt
sanchita: Removing it.

sanchita:
> Can we also collect some metrics around how many times the video was shown to the user and the number of times they clicked on it?

Added reporting

PR #813, opened 2019-09-16
🚚 Add a parallel implementation of the AllPatientsInFacilityView using Mobius 🚚

ragunathjawahar:
The implementation still does not launch the new screen. It will when I add a remote configuration flag to toggle between implementations.


ragunathjawahar:
@vinaysshenoy this is a fairly large PR, we may have to pair on this.

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityEvent.kt
```diff
@@ -0,0 +1,12 @@
+package org.simple.clinic.allpatientsinfacility
+
+import org.simple.clinic.facility.Facility
+import org.simple.clinic.widgets.UiEvent
+
+sealed class AllPatientsInFacilityEvent : UiEvent
```
ragunathjawahar: This need not be inheriting `UiEvent` anymore, right?

app/src/test/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityModelProducerTest.kt
```diff
@@ -141,4 +129,19 @@ class AllPatientsInFacilityUiStateProducerTest {
       assertNotTerminated()
     }
   }
+
+  private fun dispatchScreenCreatedEvent() {
+    MobiusTestFixture<AllPatientsInFacilityModel, AllPatientsInFacilityEvent, AllPatientsInFacilityEffect>(
+        uiEventsSubject,
```
ragunathjawahar: Since `uiEvents` is not being used anywhere, we can inline this or make it `Observable.never()`.

app/src/test/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityUiRendererTest.kt
```diff
@@ -0,0 +1,71 @@
...
+    // given
+    val noPatientsFoundInFacilityState = defaultModel
+        .facilityFetched(facility)
+        .noPatients()
+
+    // when
+    uiRenderer.render(noPatientsFoundInFacilityState)
+
+    // then
+    verify(ui).showNoPatientsFound(noPatientsFoundInFacilityState.facilityUiState!!.name)
```
ragunathjawahar: This should be `verify(ui).showNoPatientsFound(facility.name)`.

app/src/main/java/org/simple/clinic/mobius/DeferredEventSource.kt
```diff
@@ -0,0 +1,55 @@
...
+    val run = AtomicBoolean(true)
+    val thread = Thread {
+      while (run.get()) {
+        try {
+          val event = events.take()
+          if (run.get()) {
+            eventConsumer.accept(event)
+          }
+        } catch (e: Throwable) {
+          // TODO(rj) 8/Jan/19 - Log this exception.
```
ragunathjawahar: Should we do something about this?

app/src/main/java/org/simple/clinic/mobius/MobiusDelegate.kt
```diff
@@ -0,0 +1,86 @@
...
+    private val modelUpdateListener: (M) -> Unit
+) : Connectable<M, E> {
+  private val modelKey = defaultModel::class.java.name
+  private val viewStateKey = "ViewState_$modelKey"
+
+  private lateinit var controller: MobiusLoop.Controller<M, E>
+
+  private var lastKnownModel: M? = null
+
+  private val loop by lazy(NONE) {
```
ragunathjawahar: This can be `unsafeLazy`.

app/src/main/java/org/simple/clinic/mobius/MobiusDelegate.kt
```diff
@@ -0,0 +1,86 @@
...
+    RxMobius
+        .loop(
+            { model: M, event: E -> updateFunction(model, event) },
```
ragunathjawahar: Can this be a method reference?

app/src/main/java/org/simple/clinic/mobius/MobiusDelegate.kt
```diff
@@ -0,0 +1,86 @@
...
+            { effects -> effects.compose(effectHandler) }
+        )
+        .init(initFunction)
+        .eventSource(eventSource)
+  }
+
+  val eventSource by lazy(NONE) {
```
ragunathjawahar: This can be an `unsafeLazy`.

app/src/main/java/org/simple/clinic/mobius/MobiusDelegate.kt
```diff
@@ -0,0 +1,86 @@
...
+    DeferredEventSource<E>()
+  }
+
+  fun prepare() {
```
ragunathjawahar: We should guarantee that this gets called after `onRestoreInstanceState`.

ragunathjawahar: Otherwise, `lastKnownModel` could be `null` and the saved state won't get restored.

app/src/main/java/org/simple/clinic/allpatientsinfacility/migration/AllPatientsInFacilityViewMigration.kt
```diff
@@ -0,0 +1,6 @@
+package org.simple.clinic.allpatientsinfacility.migration
```
ragunathjawahar: This commit needs to be reordered before the previous two commits.

app/src/main/java/org/simple/clinic/mobius/MobiusDelegate.kt
ragunathjawahar: @vinaysshenoy the enclosing lambda is a type-inferred `Update` class. Hence, this won't be possible.

app/src/main/java/org/simple/clinic/mobius/migration/Architecture.kt
```diff
@@ -0,0 +1,6 @@
+package org.simple.clinic.mobius.migration
+
+enum class Architecture(val analyticsName: String) {
+  ORIGINAL("original"),
```
vinaysshenoy: Nice!

app/src/main/java/org/simple/clinic/analytics/Analytics.kt
```diff
@@ -131,6 +132,21 @@ object Analytics {
     reporters.forEach { it.createEvent("DataCleared", props) }
   }
 
+  fun reportArchitectureMigration(
```
vinaysshenoy: We should a test for this in `AnalyticsTest`.

vinaysshenoy: We really should add a test, because the `hostedScreen` event is not being sent. This is because `.apply` returns the same instance, and not the new instance that was created within the `apply {}` block.

I think we can just create a mutable map instead and just use a simple `if(!hostScreen.isNullOrBlank())` check instead of the current approach.

vinaysshenoy: We have a `DebugAnalyticsReporter` that can be added in `DebugClinicApp` to log the analytics events to logcat for verification.

vinaysshenoy:
List scroll and patient search result click events are not being handled in the host screens. I click on search results and scroll the list, but nothing happens.

app/src/main/java/org/simple/clinic/mobius/MobiusDelegate.kt
```diff
@@ -0,0 +1,88 @@
...
+    RxMobius
+        .loop(
+            { model: M, event: E -> updateFunction(model, event) },
+            { effects -> effects.compose(effectHandler) }
+        )
+        .init(initFunction)
+        .eventSource(eventSource)
+  }
+
+  val eventSource by unsafeLazy {
```
vinaysshenoy: We should explicitly set the type to `EventSource<E>` and hide the concrete type.

app/src/main/java/org/simple/clinic/analytics/Analytics.kt
ragunathjawahar: Nice catch! Will fix

PR #815, opened 2019-09-17
Sync everything when the app starts

vinaysshenoy:
https://www.pivotaltracker.com/story/show/168530182


vinaysshenoy:
@ragunathjawahar, Once this is merged in, we will need to `cherry-pick -x` the first three commits in this PR to the current release.

PR #817, opened 2019-09-17
Deleted a test and a minor api addition to MobiusTestFixture

ragunathjawahar:


app/src/test/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityModelProducerTest.kt
```diff
@@ -98,37 +97,6 @@ class AllPatientsInFacilityModelProducerTest {
     }
   }
 
-  @Ignore("This is automatically handled by Mobius, delete this test after manual verification.")
-  @Test
-  fun `when the screen is restored, then the ui change producer should emit the last known state`() {
```
vinaysshenoy: 👏

app/src/test/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityLogicTest.kt
```diff
@@ -32,12 +33,28 @@ class AllPatientsInFacilityLogicTest {
...
+  private lateinit var fixture: MobiusTestFixture<AllPatientsInFacilityModel, AllPatientsInFacilityEvent, AllPatientsInFacilityEffect>
+
   @Before
   fun setUp() {
     whenever(userSession.requireLoggedInUser())
         .thenReturn(Observable.just(user))
     whenever(facilityRepository.currentFacility(user))
         .thenReturn(Observable.just(facility))
+
+    fixture = MobiusTestFixture(
```
vinaysshenoy: Would it be possible to extract this into a separate `TestRule` that will automatically dispose the fixture? We could also try this out after we migrate another feature so we know what it needs to look like.

ragunathjawahar: Let's do this after we migrate another feature.

vinaysshenoy: Discussed offline: adding to the list of things to do.

PR #823, opened 2019-09-18
Show illustration on home screen

rakshakhegde:
https://www.pivotaltracker.com/story/show/168311301

Remote config sync is pending.

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
@@ -1,8 +1,10 @@
 package org.simple.clinic.home.patients
 
 import io.reactivex.Observable
+import io.reactivex.rxkotlin.ofType
 import org.simple.clinic.home.patients.illustration.HomescreenIllustration
 import org.simple.clinic.storage.files.FileStorage
+import org.simple.clinic.storage.files.GetFileResult
 import javax.inject.Inject
 
 class PatientScreenRepository @Inject constructor(
@@ -10,8 +12,13 @@ class PatientScreenRepository @Inject constructor(
     private val fileStorage: FileStorage
 ) {
 
-  fun illustrations(): Observable<HomescreenIllustration> =
+  fun illustrations(): Observable<GetFileResult.Success> =
       illustrationDao.illustrations()
           .filter { it.isNotEmpty() }
-          .map { it.first() }
+          .map { pickIllustration(it) }
+          .map { fileStorage.getFile(it.eventId) }
+          .ofType()
+
+  private fun pickIllustration(illustrations: List<HomescreenIllustration>): HomescreenIllustration =
+      illustrations.first()
 }
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
@@ -1,0 +1,17 @@
+package org.simple.clinic.home.patients
+
+import io.reactivex.Observable
+import org.simple.clinic.home.patients.illustration.HomescreenIllustration
+import org.simple.clinic.storage.files.FileStorage
+import javax.inject.Inject
+
+class PatientScreenRepository @Inject constructor(
+    private val illustrationDao: HomescreenIllustration.RoomDao,
+    private val fileStorage: FileStorage
+) {
+
+  fun illustrations(): Observable<HomescreenIllustration> =
+      illustrationDao.illustrations()
+          .filter { it.isNotEmpty() }
+          .map { it.first() }
+}
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
@@ -2,23 +2,46 @@ package org.simple.clinic.home.patients
 
 import io.reactivex.Observable
 import io.reactivex.rxkotlin.ofType
+import org.simple.clinic.home.patients.illustration.DayOfMonth
 import org.simple.clinic.home.patients.illustration.HomescreenIllustration
 import org.simple.clinic.storage.files.FileStorage
 import org.simple.clinic.storage.files.GetFileResult
+import org.simple.clinic.util.Just
+import org.simple.clinic.util.None
+import org.simple.clinic.util.Optional
+import org.simple.clinic.util.UserClock
+import org.simple.clinic.util.toOptional
+import org.simple.clinic.util.toUtcInstant
+import org.threeten.bp.Instant
+import org.threeten.bp.LocalDate
 import javax.inject.Inject
 
 class PatientScreenRepository @Inject constructor(
     private val illustrationDao: HomescreenIllustration.RoomDao,
-    private val fileStorage: FileStorage
+    private val fileStorage: FileStorage,
+    private val userClock: UserClock
 ) {
 
   fun illustrations(): Observable<GetFileResult.Success> =
       illustrationDao.illustrations()
-          .filter { it.isNotEmpty() }
           .map { pickIllustration(it) }
+          .ofType<Just<HomescreenIllustration>>()
+          .map { it.value }
           .map { fileStorage.getFile(it.eventId) }
           .ofType()
 
-  private fun pickIllustration(illustrations: List<HomescreenIllustration>): HomescreenIllustration =
-      illustrations.first()
+  private fun pickIllustration(illustrations: List<HomescreenIllustration>): Optional<HomescreenIllustration> {
+    illustrations.forEach {
+      if (userClock.instant() in toInstant(it.from)..toInstant(it.to)) {
+        return it.toOptional()
+      }
+    }
+    return None
+  }
+
+  private fun toInstant(dayOfMonth: DayOfMonth): Instant =
+      LocalDate.now(userClock)
+          .withMonth(dayOfMonth.month.value)
+          .withDayOfMonth(dayOfMonth.day)
+          .toUtcInstant(userClock)
 }
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
@@ -15,7 +15,6 @@ import org.simple.clinic.ReportAnalyticsEvents
 import org.simple.clinic.activity.TheActivityLifecycle
 import org.simple.clinic.appupdate.AppUpdateState.ShowAppUpdate
 import org.simple.clinic.appupdate.CheckAppUpdateAvailability
-import org.simple.clinic.home.patients.illustration.HomescreenIllustration
 import org.simple.clinic.patient.PatientConfig
 import org.simple.clinic.user.User
 import org.simple.clinic.user.User.LoggedInStatus.LOGGED_IN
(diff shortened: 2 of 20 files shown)


app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
```diff
@@ -0,0 +1,17 @@
+package org.simple.clinic.home.patients
+
+import io.reactivex.Observable
+import org.simple.clinic.home.patients.illustration.HomescreenIllustration
+import org.simple.clinic.storage.files.FileStorage
+import javax.inject.Inject
+
+class PatientScreenRepository @Inject constructor(
```
vinaysshenoy: I feel that `PatientScreenRepository` may not be the right name for this. It seems to me that the only reason this is being called `PatientScreenRepository` is because it is being used in the `PatientsScreen`.

Currently, we follow the convention of naming a repository around the database model (i.e., `Appointment.RoomDao` is managed by `AppointmentRepository`). Can we follow this convention and name it `HomescreenIllustrationRepository`?

vinaysshenoy: In addition, I feel this should go into the `illustration` package.

app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
```diff
@@ -0,0 +1,17 @@
+package org.simple.clinic.home.patients
+
+import io.reactivex.Observable
+import org.simple.clinic.home.patients.illustration.HomescreenIllustration
```
vinaysshenoy: Is there a reason the `illustration` package is a sub-package of `patients`? We might be using it in the `PatientsScreen`, but from a domain modelling perspective, the two are unrelated.

If we don't have a reason for putting this here, we should move this to a top-level package under `org.simple.clinic`.

app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
```diff
@@ -0,0 +1,17 @@
+package org.simple.clinic.home.patients
```
vinaysshenoy: Reword the commit message to be a bit more descriptive? I can suggest "Display illustration in the home screen".

app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
```diff
@@ -72,10 +71,16 @@ class PatientsScreenController @Inject constructor(
         toggleVisibilityOfSyncIndicator(replayedEvents),
         showAppUpdateDialog(replayedEvents),
         showSimpleVideo(replayedEvents),
-        openSimpleVideo(replayedEvents)
+        openSimpleVideo(replayedEvents),
```
vinaysshenoy: What happens if there is **both** a celebration, **and** the Simple training video needs to be displayed? Which one has a higher priority?

app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
```diff
@@ -1,17 +1,24 @@
...
+import org.simple.clinic.storage.files.GetFileResult
 import javax.inject.Inject
 
 class PatientScreenRepository @Inject constructor(
     private val illustrationDao: HomescreenIllustration.RoomDao,
     private val fileStorage: FileStorage
 ) {
 
-  fun illustrations(): Observable<HomescreenIllustration> =
+  fun illustrations(): Observable<GetFileResult.Success> =
```
vinaysshenoy: The repository is now leaking the abstraction that it uses the `FileStorage` component to manage and store images. This will cause an indirect coupling between the `PatientsScreen` and the `FileStorage` class. 

I suggest we can improve this by having the method return an `Observable<Optional<File>>`  instead and map the result from `FileStorage` internally.

app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
```diff
@@ -14,7 +14,6 @@ class PatientScreenRepository @Inject constructor(
 
   fun illustrations(): Observable<GetFileResult.Success> =
```
vinaysshenoy: We should squash this commit with the next one since this commit can potentially cause a crash since `pickIllustration` returns the first item in the list and we no longer filter it.

app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
```diff
@@ -2,22 +2,46 @@ package org.simple.clinic.home.patients
...
       illustrationDao.illustrations()
           .map { pickIllustration(it) }
+          .ofType<Just<HomescreenIllustration>>()
+          .map { it.value }
           .map { fileStorage.getFile(it.eventId) }
           .ofType()
 
-  private fun pickIllustration(illustrations: List<HomescreenIllustration>): HomescreenIllustration =
-      illustrations.first()
+  private fun pickIllustration(illustrations: List<HomescreenIllustration>): Optional<HomescreenIllustration> {
```
vinaysshenoy: I have a couple of suggestions to make to improve these two functions:

- Using an `Instant` is unnecessary since all we need are the dates. We can use `LocalDate` to compare instead.
- We can use the `.firstOrNull {}` extension on `Collection` to have a more readable method to find the right `HomescreenIllustration.

```kotlin
private fun pickIllustration(illustrations: List<HomescreenIllustration>): Optional<HomescreenIllustration> {

    return illustrations
        .firstOrNull { illustration ->
          val showIllustrationFrom = illustration.from.toLocalDate(userClock)
          val showIllustrationTill = illustration.to.toLocalDate(userClock)

          today in showIllustrationFrom..showIllustrationTill
        }
        .toOptional()
  }

  private fun DayOfMonth.toLocalDate(userClock: UserClock): LocalDate {
    return LocalDate.now(userClock)
        .withMonth(this.month.value)
        .withDayOfMonth(this.day)
  }
```

app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
```diff
@@ -2,22 +2,46 @@ package org.simple.clinic.home.patients
...
+    illustrations.forEach {
+      if (userClock.instant() in toInstant(it.from)..toInstant(it.to)) {
+        return it.toOptional()
+      }
+    }
+    return None
+  }
+
+  private fun toInstant(dayOfMonth: DayOfMonth): Instant =
+      LocalDate.now(userClock)
```
vinaysshenoy: Nice!

app/src/test/java/org/simple/clinic/home/patients/PatientScreenRepositoryTest.kt
```diff
@@ -0,0 +1,65 @@
...
+
+class PatientScreenRepositoryTest {
+
+  @get:Rule
+  val rxErrorsRule = RxErrorsRule()
+
+  val userClock = TestUserClock()
+
+  @Test
+  fun `verify correct illustration is picked up`() {
```
vinaysshenoy: Nice test! Can we also add a few more test cases, like

- On the bounds of a date range (current date is on the first and last day of the illustration date range)
- When the current date is not part of a illustration date range, it should return a `None`
- When the current date is part of an illustration date, but the downloaded image file is not present, it should return a `None`.

app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
rakshakhegde: Done.

app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
rakshakhegde: done

app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
vinaysshenoy: Then could you please add this as a task in another PR to ensure this? This is not taken care of in the current implementation.

app/src/main/java/org/simple/clinic/home/patients/PatientScreenRepository.kt
vinaysshenoy: I realised something about the `AndroidFileStorage` component and the `getFile()` method. It was written with the intention of receiving a `File` handle that would create the file if it didn't exist. However, that is not what we want here because we do not want to show the illustration if the file does not exist. This will end up showing a blank home screen instead of the default image if the image for an occasion has not been downloaded.

This is ok for now since there aren't any occasions being fetched. But I suggest we do two things before we integrate the final API:
- Rename `FileStorage#getFile` to `getOrCreateFile` so the method is explicit about what it does
- Add a new method `FileStorage#file(path)` that returns an `Optional<File>` that can be used for cases like this.

app/src/main/java/org/simple/clinic/illustration/HomescreenIllustrationRepository.kt
```diff
@@ -20,13 +21,14 @@ class HomescreenIllustrationRepository @Inject constructor(
     private val userClock: UserClock
 ) {
 
-  fun illustrations(): Observable<GetFileResult.Success> =
+  fun illustrations(): Observable<File> =
```
vinaysshenoy: This name seems inconsistent with the return type since it indicates that it would return a `Collection` of some sort, but instead it returns a single `File`.

I will rename this to something more _obvious_, like `illustrationImageToShow`. What do you think?

app/src/main/java/org/simple/clinic/illustration/HomescreenIllustrationRepository.kt
```diff
@@ -31,17 +28,16 @@ class HomescreenIllustrationRepository @Inject constructor(
...
-    illustrations.forEach {
-      if (userClock.instant() in toInstant(it.from)..toInstant(it.to)) {
-        return it.toOptional()
-      }
-    }
-    return None
+    val today = LocalDate.now(userClock)
+    return illustrations
+        .firstOrNull { illustration ->
+          today in toLocalDate(illustration.from)..toLocalDate(illustration.to)
```
vinaysshenoy: I will extract these into separate variables for better readability.

app/src/test/java/org/simple/clinic/illustration/HomescreenIllustrationRepositoryTest.kt
```diff
@@ -88,4 +88,16 @@ class HomescreenIllustrationRepositoryTest {
         .assertNoValues()
         .assertNoErrors()
   }
+
+  @Test
+  fun `verify illustration is not set when there is event today but no illustration file`() {
+    whenever(fileStorage.getFile(eventId)).thenReturn(GetFileResult.Failure(FileNotFoundException()))
```
vinaysshenoy: See earlier comment about the behaviour of the `FileStorage` class. Let's talk about this next week.

app/src/main/java/org/simple/clinic/home/patients/PatientsScreenController.kt
rakshakhegde: The current implementation works as expected

vinaysshenoy: I believe it only works because of coincidence. If the query to fetch the illustrations takes longer than the query to show the training video, the illustration will be shown instead.

We should make this explicit.

PR #831, opened 2019-09-24
Don't query the repository while querying data + split high/low fidelity tests to seperate classes

ragunathjawahar:


app/src/test/java/org/simple/clinic/editpatient/PatientEditScreenControllerTest.kt
```diff
@@ -76,20 +66,18 @@ class PatientEditScreenControllerTest {
 
   private lateinit var screen: PatientEditScreen
   private lateinit var patientRepository: PatientRepository
-  private val numberValidator: PhoneNumberValidator = IndianPhoneNumberValidator()
   private lateinit var controller: PatientEditScreenController
 
-  private lateinit var errorConsumer: (Throwable) -> Unit
-  private lateinit var dobValidator: UserInputDateValidator
   private val dateOfBirthFormat = DateTimeFormatter.ofPattern("dd/MM/yyyy", Locale.ENGLISH)
+  private val dobValidator: UserInputDateValidator = UserInputDateValidator(TestUserClock().zone, dateOfBirthFormat)
```
vinaysshenoy: We can just use `ZoneOffset.UTC` here since that is the default zone of the `TestUserClock`.

app/src/main/java/org/simple/clinic/editpatient/PatientEditScreenController.kt
```diff
@@ -185,14 +184,18 @@ class PatientEditScreenController @Inject constructor(
   private fun savePatientDetails(events: Observable<UiEvent>): Observable<UiChange> {
     val saveClicks = events.ofType<PatientEditSaveClicked>()
 
-    val patientUuidStream = events.ofType<PatientEditScreenCreated>()
+    val screenCreatedStream = events
+        .ofType<PatientEditScreenCreated>()
+        .share()
```
ragunathjawahar: @vinaysshenoy this was the perp!

vinaysshenoy: Whoops! :swe

PR #838, opened 2019-09-26
Sync illustration config

rakshakhegde:
https://www.pivotaltracker.com/story/show/168721888


app/src/main/java/org/simple/clinic/illustration/HomescreenIllustrationModule.kt
```diff
@@ -1,12 +1,26 @@
...
+import org.simple.clinic.remoteconfig.ConfigReader
 
 @Module
 class HomescreenIllustrationModule {
 
   @Provides
   fun illustrationDao(appDatabase: AppDatabase) = appDatabase.illustrationDao()
+
+  @Provides
+  fun illustrations(
```
vinaysshenoy: This will load and parse the illustrations on the UI thread. We should do this in the `HomescreenIllustrationRepository`, but load them from the config instead of the database.

app/src/main/java/org/simple/clinic/illustration/HomescreenIllustrationRepository.kt
```diff
@@ -13,13 +13,13 @@ import java.io.File
 import javax.inject.Inject
 
 class HomescreenIllustrationRepository @Inject constructor(
-    private val illustrationDao: HomescreenIllustration.RoomDao,
+    private val illustrations: List<HomescreenIllustration>,
```
vinaysshenoy: As mentioned above, doing this will make the illustrations be read and parsed at the time of injection, which is on the UI thread. Can we inject the `ConfigReader` here instead and load it at the time of invoking the method?

app/src/main/java/org/simple/clinic/illustration/IllustrationSync.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import io.reactivex.Completable
+import io.reactivex.Single
+import org.simple.clinic.sync.ModelSync
+import org.simple.clinic.sync.SyncConfig
+import javax.inject.Inject
+import javax.inject.Named
+
+class IllustrationSync @Inject constructor(
+    private val illustrations: List<HomescreenIllustration>,
```
vinaysshenoy: Can we inject the repository here instead and read the illustrations from the method? The reasons are the same as described above.

app/src/test/java/org/simple/clinic/illustration/HomescreenIllustrationRepositoryTest.kt
```diff
@@ -24,35 +24,36 @@ class HomescreenIllustrationRepositoryTest {
...
-            to = DayOfMonth(20, Month.SEPTEMBER)
-        ),
-        HomescreenIllustration(
-            eventId = "event-2",
-            illustrationUrl = "some-illustration-url-3",
-            from = DayOfMonth(10, Month.OCTOBER),
-            to = DayOfMonth(20, Month.OCTOBER)
-        )
-    )))
+    whenever(illustrationDao.illustrations()).thenReturn(Observable.just(illustrations))
```
vinaysshenoy: Since the `illustrationDao` is not being used anymore, we can just remove it.

app/src/main/java/org/simple/clinic/illustration/IllustrationSync.kt
```diff
@@ -2,21 +2,52 @@ package org.simple.clinic.illustration
...
+import retrofit2.Response
+import java.io.File
 import javax.inject.Inject
 import javax.inject.Named
 
 class IllustrationSync @Inject constructor(
     private val illustrations: List<HomescreenIllustration>,
-    @Named("sync_config_daily") private val configProvider: Single<SyncConfig>
+    private val fileDownloadService: FileDownloadService,
+    private val fileOperations: FileOperations,
```
vinaysshenoy: `IllustrationSync` should not be directly responsible for deciding which folder the images should be stored in. If we decide to change this, we'll have to change this in multiple places (the repository where the images are being read from, as well as this place) as well as makes testing harder because this class is doing too much (fetching the images as well as saving them).

We should delegate this task to the `HomescreenIllustrationRepository`. Could you please take a look at `ReportsSync`? We have done something similar there.

app/src/main/java/org/simple/clinic/illustration/IllustrationSync.kt
```diff
@@ -2,21 +2,52 @@ package org.simple.clinic.illustration
...
+  private fun saveIllustration(illustration: HomescreenIllustration) {
+    fileDownloadService.downloadFile(illustration.illustrationUrl)
+        .enqueue(object : retrofit2.Callback<ResponseBody> {
+          override fun onFailure(call: Call<ResponseBody>, t: Throwable) {
+            // TODO Make analytics call
+          }
+
+          override fun onResponse(call: Call<ResponseBody>, response: Response<ResponseBody>) {
+            response.body()?.byteStream()?.use { inputStream ->
+              val illustrationsFile = File(illustrationsFolder, illustration.eventId)
```
vinaysshenoy: `FileOperations` is not meant to be used outside of `FileStorage`. Can you inject `FileStorage` instead and use it instead?

app/src/main/java/org/simple/clinic/illustration/IllustrationSync.kt
```diff
@@ -33,6 +33,9 @@ class IllustrationSync @Inject constructor(
   override fun syncConfig() = configProvider
 
   private fun saveIllustration(illustration: HomescreenIllustration) {
+    val illustrationsFile = File(illustrationsFolder, illustration.eventId)
+    if (illustrationsFile.exists()) return
```
vinaysshenoy: Even if the file exists, we should continue to save it because a new image might have been uploaded for the same event.

app/src/main/java/org/simple/clinic/illustration/HomescreenIllustrationModule.kt
rakshakhegde: I would like to personally discuss with you on how we can do this.

app/src/main/java/org/simple/clinic/illustration/IllustrationSync.kt
rakshakhegde: Done

app/src/test/java/org/simple/clinic/illustration/HomescreenIllustrationRepositoryTest.kt
rakshakhegde: Done

app/src/main/java/org/simple/clinic/illustration/IllustrationSync.kt
rakshakhegde: Done

app/src/main/java/org/simple/clinic/illustration/IllustrationSync.kt
rakshakhegde: Done

app/src/main/java/org/simple/clinic/illustration/IllustrationSync.kt
rakshakhegde: Done

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -3319,6 +3319,14 @@ class DatabaseMigrationAndroidTest {
         setOf("eventId", "illustrationUrl", "from_day", "from_month", "to_day", "to_month")
     )
   }
+
+  fun verify_migration_to_47_drops_table_HomescreenIllustration() {
```
vinaysshenoy: You missed the `@Test` annotation. I'll add it.

vinaysshenoy:
LGTM for the most part. There is still one thing we need to handle:

- When saving the network stream to the local file, it is possible that the stream breaks in between and the stream is partially written to the file. In this case, trying to display the image will show a broken image. We need to handle this case.

Added a pivotal task for this: https://www.pivotaltracker.com/story/show/168860054

PR #839, opened 2019-09-26
Make BPs editable for a configurable duration only

sanchita:
https://www.pivotaltracker.com/story/show/168558948

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryConfig.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryConfig.kt
@@ -13,12 +13,12 @@ data class PatientSummaryConfig(
   companion object {
 
     fun read(configReader: ConfigReader): Observable<PatientSummaryConfig> {
-      val bpEditableDurationConfig = configReader.long("bp_editable_duration_in_minutes", 60)
+      val bpEditableDurationConfig = configReader.long("bp_editable_duration_in_seconds", 3600)
 
       return Observable.just(PatientSummaryConfig(
           numberOfBpPlaceholders = 3,
           numberOfBpsToDisplay = 100,
-          bpEditableDuration = Duration.ofMinutes(bpEditableDurationConfig)
+          bpEditableDuration = Duration.ofSeconds(bpEditableDurationConfig)
       ))
     }
   }
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryConfig.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryConfig.kt
@@ -13,13 +13,14 @@ data class PatientSummaryConfig(
   companion object {
 
     fun read(configReader: ConfigReader): Observable<PatientSummaryConfig> {
-      val bpEditableDurationConfig = configReader.long("bp_editable_duration_in_seconds", 3600)
+      return Observable.fromCallable {
+        val bpEditableDurationConfig = configReader.long("bp_editable_duration_in_seconds", 3600)
 
-      return Observable.just(PatientSummaryConfig(
-          numberOfBpPlaceholders = 3,
-          numberOfBpsToDisplay = 100,
-          bpEditableDuration = Duration.ofSeconds(bpEditableDurationConfig)
-      ))
+        PatientSummaryConfig(
+            numberOfBpPlaceholders = 3,
+            numberOfBpsToDisplay = 100,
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/summary/PatientSummaryConfig.kt
```diff
@@ -1,6 +1,25 @@
...
-    val numberOfBpsToDisplay: Int
-)
+    val numberOfBpsToDisplay: Int,
+    val bpEditableDuration: Duration
+) {
+
+  companion object {
+
+    fun read(configReader: ConfigReader): Observable<PatientSummaryConfig> {
+      val bpEditableDurationConfig = configReader.long("bp_editable_duration_in_minutes", 60)
```
vinaysshenoy: This would mean that we won't be able to change the editable duration to less than a minute remotely. Should we make this represent seconds instead of minutes?

vinaysshenoy: This will read from the disk, so maybe we should use `Observable.fromCallable {}` instead and inline the reading from the `ConfigReader` in there?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -10,6 +10,7 @@ import io.reactivex.rxkotlin.zipWith
 import org.simple.clinic.ReplayUntilScreenIsDestroyed
```
vinaysshenoy: There is an extra "adds" in the commit message.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -541,4 +545,12 @@ class PatientSummaryScreenController @Inject constructor(
         .ofType<PatientSummaryLinkIdCompleted>()
         .map { Ui::hideLinkIdWithPatientView }
   }
+
+
+  private fun isBpEditable(bloodPressureMeasurement: BloodPressureMeasurement, bpEditableFor: Duration): Boolean {
+    val now = Instant.now(utcClock)
+    val editExpiresAt = bloodPressureMeasurement.createdAt.plus(bpEditableFor)
```
vinaysshenoy: Suggestion for a slightly better name: `cannotEditAfter`?

app/src/main/java/org/simple/clinic/summary/SummaryBloodPressureListItem.kt
```diff
@@ -50,68 +52,64 @@ data class SummaryBloodPressureListItem(
     val context = holder.itemView.context
     val resources = context.resources
 
+    holder.itemView.isClickable = isBpEditable
+    holder.itemView.isFocusable = isBpEditable
+    if (isBpEditable) holder.itemView.setOnClickListener { uiEvents.onNext(PatientSummaryBpClicked(measurement)) }
```
vinaysshenoy: Won't we have to set the click listener to `null` if the bp is not editable.

app/src/main/java/org/simple/clinic/summary/PatientSummaryConfig.kt
sanchita: That's a good point.
I will make this change.

app/src/main/java/org/simple/clinic/summary/SummaryBloodPressureListItem.kt
sanchita: Since we are setting `isClickable` and `isFocusable` to `false`, we don't need to do this.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
sanchita: Removed.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
sanchita: I have refactored this code for better use of `Duration`.
Please take a look.

vinaysshenoy: Looks good, I am going to rename `durationSince` to `durationSinceBpCreated` since the former doesn't really tell us anything.

PR #840, opened 2019-09-27
Finish adding form validation and save / update patient functionality.

ragunathjawahar:
This is a fairly large PR, I am improving the internal quality of the following classes. Will send that in as a separate PR once this gets merged in.

- EditPatientEffectHandler
- EditPatientModel
- EditPatientUpdate
- EditPatientViewRenderer

(EDIT) The above mentioned classes were refactored and added as part of this PR.

PR #844, opened 2019-09-30
Add schedulers to effects, remove copy constructors from effect handler, and add `PatientSavedEvent`.

ragunathjawahar:
This PR contains changes that were discussed in #840, except

* ~~Make scrolling to first field in effect handler a discrete effect~~ (Not doing this as discussed because it is an implementation detail).

### DIFF
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientEffectHandler.kt
@@ -131,22 +131,19 @@ object EditPatientEffectHandler {
       dateOfBirthFormatter: DateTimeFormatter,
       utcClock: UtcClock
   ): Patient {
-    val patientWithoutAgeOrDateOfBirth = patient.copy(
-        fullName = ongoingEntry.name,
-        gender = ongoingEntry.gender,
-        dateOfBirth = null,
-        age = null
-    )
+    val patientWithoutAgeOrDateOfBirth = patient
+        .withNameAndGender(ongoingEntry.name, ongoingEntry.gender)
+        .withoutAgeAndDateOfBirth()
 
     return when (ongoingEntry.ageOrDateOfBirth) {
       is EntryWithAge -> {
         val age = coerceAgeFrom(patient.age, ongoingEntry.ageOrDateOfBirth.age, utcClock)
-        patientWithoutAgeOrDateOfBirth.copy(age = age)
+        patientWithoutAgeOrDateOfBirth.withAge(age)
       }
 
       is EntryWithDateOfBirth -> {
         val dateOfBirth = LocalDate.parse(ongoingEntry.ageOrDateOfBirth.dateOfBirth, dateOfBirthFormatter)
-        patientWithoutAgeOrDateOfBirth.copy(dateOfBirth = dateOfBirth)
+        patientWithoutAgeOrDateOfBirth.withDateOfBirth(dateOfBirth)
       }
     }
   }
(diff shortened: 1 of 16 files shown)


app/src/main/java/org/simple/clinic/editpatient/EditPatientEffectHandler.kt
```diff
@@ -50,17 +50,19 @@ object EditPatientEffectHandler {
   ) {
     val (patient, address, phoneNumber) = prefillFormFieldsEffect
 
-    ui.setPatientName(patient.fullName)
-    ui.setGender(patient.gender)
-    ui.setState(address.state)
-    ui.setDistrict(address.district)
+    with(ui) {
```
vinaysshenoy: Nice!

app/src/main/java/org/simple/clinic/editpatient/EditPatientEvent.kt
```diff
@@ -45,6 +45,9 @@ object SaveClicked: EditPatientEvent() {
   override val analyticsName = "Edit Patient Entry:Save Clicked"
 }
 
+// This is NOT an UiEvent, hence doesn't require analytics
```
vinaysshenoy: I'm wondering if we should still push this for better observability of what is happening though. All events need not necessarily be UI events, right? It could just be a signal that something of significance happened.

app/src/main/java/org/simple/clinic/editpatient/EditPatientEffectHandler.kt
```diff
@@ -131,22 +131,18 @@ object EditPatientEffectHandler {
       dateOfBirthFormatter: DateTimeFormatter,
       utcClock: UtcClock
   ): Patient {
-    val patientWithoutAgeOrDateOfBirth = patient.copy(
```
vinaysshenoy: Nice! 👏

app/src/main/java/org/simple/clinic/editpatient/EditPatientEffectHandler.kt
```diff
@@ -131,22 +131,18 @@ object EditPatientEffectHandler {
...
-        fullName = ongoingEntry.name,
-        gender = ongoingEntry.gender,
-        dateOfBirth = null,
-        age = null
-    )
+    val patientWithoutAgeOrDateOfBirth = patient
+        .withoutAgeOrDateOfBirth(ongoingEntry.name, ongoingEntry.gender)
```
vinaysshenoy: I think this naming is a little tricky because the name of the function and the parameters we are passing do not make sense with each other. What do you think of adding two more functions to transform the `Patient` so that this call would look more like:

```kotlin
val patientWithoutAgeOrDateOfBirth = patient
  .withNameAndGender(ongoingEntry.name, ongoingEntry.gender)
  .withoutAgeAndDateOfBirth()
```

app/src/main/java/org/simple/clinic/editpatient/EditPatientEffectHandler.kt
```diff
@@ -202,7 +194,7 @@ object EditPatientEffectHandler {
           hasExistingPhoneNumber(existingPhoneNumber, enteredPhoneNumber)
         }
         .flatMapCompletable { (patientUuid, existingPhoneNumber, enteredPhoneNumber) ->
-          patientRepository.updatePhoneNumberForPatient(patientUuid, existingPhoneNumber!!.copy(number = enteredPhoneNumber))
+          patientRepository.updatePhoneNumberForPatient(patientUuid, existingPhoneNumber!!.withNumber(enteredPhoneNumber))
```
vinaysshenoy: Should we replace the `!!` with a `requireNotNull(existingPhoneNumber)` check?

app/src/main/java/org/simple/clinic/editpatient/EditPatientEffectHandler.kt
```diff
@@ -85,10 +87,12 @@ object EditPatientEffectHandler {
   private fun savePatientTransformer(
       patientRepository: PatientRepository,
       utcClock: UtcClock,
-      dateOfBirthFormatter: DateTimeFormatter
+      dateOfBirthFormatter: DateTimeFormatter,
+      schedulersProvider: SchedulersProvider
```
vinaysshenoy: Can't we pass the `scheduler` directly as a parameter instead of the entire `SchedulersProvider`?

PR #845, opened 2019-10-01
Replace help icon with a button

rakshakhegde:
https://www.pivotaltracker.com/story/show/168836179


app/src/main/java/org/simple/clinic/home/HomeScreen.kt
```diff
@@ -5,10 +5,10 @@ import android.util.AttributeSet
 import android.view.ViewGroup
 import android.widget.Button
 import android.widget.RelativeLayout
-import androidx.appcompat.widget.Toolbar
 import androidx.viewpager.widget.ViewPager
 import com.jakewharton.rxbinding2.view.RxView
 import io.reactivex.Observable
+import kotlinx.android.synthetic.main.screen_home.view.*
```
ragunathjawahar: It's nice to see that we are moving to KTX, can we get rid of `kotterknife` entirely from this screen?

rakshakhegde: Done

ragunathjawahar: Nice!

PR #846, opened 2019-10-01
Implement Settings logic to update the model

vinaysshenoy:
https://www.pivotaltracker.com/story/show/166437675

### DIFF
--- a/app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
@@ -1,0 +1,30 @@
+package org.simple.clinic.settings
+
+import com.spotify.mobius.rx2.RxMobius
+import io.reactivex.ObservableTransformer
+import io.reactivex.Scheduler
+import org.simple.clinic.user.UserSession
+import org.simple.clinic.util.filterAndUnwrapJust
+import org.simple.clinic.util.scheduler.SchedulersProvider
+
+object SettingsEffectHandler {
+
+  fun create(
+      userSession: UserSession,
+      schedulersProvider: SchedulersProvider
+  ): ObservableTransformer<SettingsEffect, SettingsEvent> {
+    return RxMobius
+        .subtypeEffectHandler<SettingsEffect, SettingsEvent>()
+        .addTransformer(LoadUserDetailsEffect::class.java, loadUserDetails(userSession, schedulersProvider.io()))
+        .build()
+  }
+
+  private fun loadUserDetails(userSession: UserSession, scheduler: Scheduler): ObservableTransformer<LoadUserDetailsEffect, SettingsEvent> {
+    return ObservableTransformer { upstream ->
+      upstream
+          .switchMap { userSession.loggedInUser().subscribeOn(scheduler) }
+          .filterAndUnwrapJust()
+          .map { user -> UserDetailsLoaded(user.fullName, user.phoneNumber) }
+    }
+  }
+}
--- a/app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
+++ b/app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
@@ -1,0 +1,74 @@
+package org.simple.clinic.settings
+
+import com.nhaarman.mockito_kotlin.mock
+import com.nhaarman.mockito_kotlin.whenever
+import io.reactivex.Observable
+import io.reactivex.observers.TestObserver
+import io.reactivex.subjects.PublishSubject
+import org.junit.After
+import org.junit.Test
+import org.simple.clinic.patient.PatientMocker
+import org.simple.clinic.user.UserSession
+import org.simple.clinic.util.None
+import org.simple.clinic.util.scheduler.TrampolineSchedulersProvider
+import org.simple.clinic.util.toOptional
+import java.util.UUID
+
+class SettingsEffectHandlerTest {
+
+  private val userSession = mock<UserSession>()
+  private val effectsSubject = PublishSubject.create<SettingsEffect>()
+
+  private val effectHandlerTest: TestObserver<SettingsEvent> = effectsSubject
+      .compose(SettingsEffectHandler.create(
+          userSession = userSession,
+          schedulersProvider = TrampolineSchedulersProvider()
+      ))
+      .test()
+
+  @After
+  fun tearDown() {
+    effectHandlerTest.dispose()
+  }
+
+  @Test
+  fun `user details must be fetched when the load user details event is received`() {
+    // given
+    val savedUser = PatientMocker.loggedInUser(
+        uuid = UUID.fromString("eadfd8ef-6c88-4bc9-ba96-e05bec2d6d8b"),
+        name = "Mahalakshmi Puri",
+        phone = "1234567890"
+    )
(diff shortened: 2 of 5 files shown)


app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
```diff
@@ -0,0 +1,76 @@
...
+  private val effectHandlerTest: TestObserver<SettingsEvent> = effectsSubject
+      .compose(SettingsEffectHandler.create(
+          userSession = userSession,
+          schedulersProvider = TrampolineSchedulersProvider()
+      ))
+      .test()
+
+  @After
+  fun tearDown() {
+    effectHandlerTest.dispose()
```
ragunathjawahar: Nice!

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  @Test
+  fun `user details must be fetched when the load user details event is received`() {
```
ragunathjawahar: Isn't this test for a single user, why are we testing the response for two different users?

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
```diff
@@ -0,0 +1,76 @@
...
+        .assertValues(
+            UserDetailsLoaded(name = firstUser.fullName, phoneNumber = firstUser.phoneNumber),
+            UserDetailsLoaded(name = secondUser.fullName, phoneNumber = secondUser.phoneNumber)
+        )
+        .assertNotComplete()
+        .assertNotTerminated()
+  }
+
+  @Test
+  fun `when there are no user details, no event must be triggered`() {
```
ragunathjawahar: If there are not user details, then isn't this failing silently? The prerequisite for this screen is that, there should be a logged in user. Otherwise we won't land in this screen. Am I missing something here?

app/src/test/java/org/simple/clinic/settings/SettingsUpdateTest.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.spotify.mobius.test.UpdateSpec
+import com.spotify.mobius.test.UpdateSpec.assertThatNext
+import org.junit.Test
+
+class SettingsUpdateTest {
+
+  private val defaultModel = SettingsModel.FETCHING_USER_DETAILS
+
+  @Test
+  fun `the user details loaded event must update the model`() {
```
ragunathjawahar: Nice.

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
vinaysshenoy: This test is meant to verify that if the user gets updated in the background, the updated details are published as a new event so the model can be updated. I'll rework the test to make that more clear.

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
vinaysshenoy: We have a logout feature that will clear the local logged in user if they get signed out in the background. This test is to ensure that the app will not crash if this screen is open when that happens,

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
vinaysshenoy: Done.

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
```diff
@@ -0,0 +1,30 @@
...
+  ): ObservableTransformer<SettingsEffect, SettingsEvent> {
+    return RxMobius
+        .subtypeEffectHandler<SettingsEffect, SettingsEvent>()
+        .addTransformer(LoadUserDetailsEffect::class.java, loadUserDetails(userSession, schedulersProvider.io()))
+        .build()
+  }
+
+  private fun loadUserDetails(userSession: UserSession, scheduler: Scheduler): ObservableTransformer<LoadUserDetailsEffect, SettingsEvent> {
+    return ObservableTransformer { upstream ->
+      upstream
```
ragunathjawahar: Rename `upstream` in the next PR?

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
```diff
@@ -0,0 +1,80 @@
...
+  fun `user details must be fetched when the load user details event is received`() {
+    // given
+    val savedUser = PatientMocker.loggedInUser(
+        uuid = UUID.fromString("eadfd8ef-6c88-4bc9-ba96-e05bec2d6d8b"),
+        name = "Mahalakshmi Puri",
+        phone = "1234567890"
+    )
+    val savedUserWithUpdatedName = savedUser.copy(fullName = "Mahalakshmi Sharma")
+    val savedUserWithUpdatedPhoneNumber = savedUser.copy(phoneNumber = "1234567891")
+    val savedUserUpdatedNameAndPhoneNumber = savedUser.copy(fullName = "Mahalakshmi T Sharma", phoneNumber = "0987654321")
```
ragunathjawahar: We are just illustrating that whenever the underlying stream emits, there will be an update. Why not just do 2 events. More events make the test noisier and developers looking at this example will also pick up the pattern.

vinaysshenoy: Makes sense, I'll change this.

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
```diff
@@ -0,0 +1,80 @@
...
+            UserDetailsLoaded(name = savedUserWithUpdatedName.fullName, phoneNumber = savedUserWithUpdatedName.phoneNumber),
+            UserDetailsLoaded(name = savedUserWithUpdatedPhoneNumber.fullName, phoneNumber = savedUserWithUpdatedPhoneNumber.phoneNumber),
+            UserDetailsLoaded(name = savedUserUpdatedNameAndPhoneNumber.fullName, phoneNumber = savedUserUpdatedNameAndPhoneNumber.phoneNumber)
+        )
+        .assertNotComplete()
+        .assertNotTerminated()
+  }
+
+  @Test
+  fun `when there are no user details, there must not be any error`() {
```
ragunathjawahar: Can we add something like, when the user logs out ... to add more context to this test?

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
vinaysshenoy: Done

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
vinaysshenoy: Done.

PR #847, opened 2019-10-01
Introduce new edit patient screen behind a feature flag

ragunathjawahar:


app/src/main/java/org/simple/clinic/editpatient/EditPatientScreenKey.kt
```diff
@@ -27,5 +27,5 @@ data class EditPatientScreenKey(
   @IgnoredOnParcel
   override val analyticsName = "Edit Patient"
```
vinaysshenoy: Should the analytics name for the new screen be different?

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -79,31 +70,6 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
   @Inject
   lateinit var crashReporter: CrashReporter
 
-  private val formScrollView by bindView<ScrollView>(R.id.patientedit_form_scrollview)
```
vinaysshenoy: 🚀

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -64,12 +73,68 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
...
 
   @Inject
   lateinit var crashReporter: CrashReporter
 
+  private val screenKey by unsafeLazy {
+    screenRouter.key<EditPatientScreenKey>(this)
+  }
+
+  private val viewRenderer by unsafeLazy {
+    EditPatientViewRenderer(this)
```
vinaysshenoy: Why is the `viewRenderer` a lazy delegate? Is there a problem if we directly instantiate it? As far as I can tell, the renderer just holds the reference to the view instance, so it should be safe to just instantiate it here.

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -64,12 +73,68 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
...
+        ageTextChanges(),
+        backClicks()
+    ).compose(ReportAnalyticsEvents())
+        .cast()
+
+  private lateinit var eventsDisposable: Disposable
+
+  private val delegate by unsafeLazy {
+    val (patient, address, phoneNumber) = screenKey
+    MobiusDelegate(
```
vinaysshenoy: We should look at directly injecting the delegate using `@AssistedInject` so that the key can be provided at some point. What do you think?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -162,7 +169,16 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
...
 
   private fun setupEditButtonClicks() {
-    editButton.setOnClickListener { screenRouter.push(createEditPatientScreenKey(patientSummaryProfile!!)) }
+    editButton.setOnClickListener {
+      val (key, architecture) = if (mobiusMigrationConfig.useEditPatientScreen) {
+        createEditPatientScreenKey(patientSummaryProfile!!) to Architecture.MOBIUS
+      } else {
+        createPatientEditScreenKey(patientSummaryProfile!!) to Architecture.ORIGINAL
+      }
+      Analytics.reportArchitectureMigration("EditPatientScreen", architecture, "PatientSummaryScreen")
```
vinaysshenoy: Won't the `hostScreen` parameter be `null` since the `EditPatientScreen` is a discrete screen and not a part of some other screen?

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -78,61 +143,60 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
...
-    return Observable.just(patientEditScreenCreated)
+  override fun onAttachedToWindow() {
+    super.onAttachedToWindow()
+    delegate.start()
+    eventsDisposable = events.subscribe { delegate.eventSource.notifyEvent(it) }
+  }
+
+  override fun onDetachedFromWindow() {
+    delegate.stop()
+    if (::eventsDisposable.isInitialized && eventsDisposable.isDisposed.not()) {
```
vinaysshenoy: I find the way these events are being managed is cumbersome. Can we do something to simplify them? As a suggestion,

- Provide the `Observable<E>` to the `MobiusDelegate` via the constructor
- Subscribe and unsubscribed when `start()` and `stop()` are called

This is a suggestion for a future PR.

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
ragunathjawahar: Done.

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreenKey.kt
ragunathjawahar: @vinaysshenoy is there a reason why we should make that distinction?

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
ragunathjawahar: I don't know what `@AssistedInject` is, can we discuss this in person?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
ragunathjawahar: Good catch, done.

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
ragunathjawahar: Yes, can do!

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreenKey.kt
vinaysshenoy: I thought we might want to since it's a new screen, but then I realised that from the POV of the user, the screen hasn't changed. We already have a separate event for the architecture migration, so this is unnecessary. Closing the comment.

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
vinaysshenoy: Sure. But the tl;dr is that [Assisted Injection](https://github.com/square/AssistedInject) will allow components like the screens to contribute dependencies to the dagger graph (like the screen key in this case) when injecting it.

Anyway, it's not important to do in this PR. Just adding that it might be something that will allow us to inject mobius delegates into screens easily.

PR #851, opened 2019-10-03
Mobius delegate improvements

ragunathjawahar:
- Manage event subscription on its own
- Remove temporal coupling from `onRestoreInstanceState`

### DIFF
--- a/app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityView.kt
+++ b/app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityView.kt
@@ -67,6 +67,7 @@ class AllPatientsInFacilityView(
         .createEffectHandler(userSession, facilityRepository, patientRepository, schedulersProvider)
 
     MobiusDelegate(
+        Observable.never(),
         AllPatientsInFacilityModel.FETCHING_PATIENTS,
         ::allPatientsInFacilityInit,
         ::allPatientsInFacilityUpdate,
--- a/app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityView.kt
+++ b/app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityView.kt
@@ -56,7 +56,7 @@ class AllPatientsInFacilityView(
   @Inject
   lateinit var crashReporter: CrashReporter
 
-  private val viewRenderer by unsafeLazy { AllPatientsInFacilityUiRenderer(this) }
+  private val viewRenderer = AllPatientsInFacilityUiRenderer(this)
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityView.kt
```diff
@@ -118,9 +119,8 @@ class AllPatientsInFacilityView(
   }
 
   override fun onRestoreInstanceState(state: Parcelable?) {
-    super.onRestoreInstanceState(delegate.onRestoreInstanceState(state as Bundle?)).also {
-      delegate.prepare()
-    }
+    val viewState = delegate.onRestoreInstanceState(state as Bundle?)
+    super.onRestoreInstanceState(viewState)
```
ragunathjawahar: So tempted to leave a multi-line comment 😅

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -148,14 +146,10 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
   override fun onAttachedToWindow() {
     super.onAttachedToWindow()
     delegate.start()
-    eventsDisposable = events.subscribe { delegate.eventSource.notifyEvent(it) }
```
vinaysshenoy: 👏

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -148,14 +146,10 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
...
   }
 
   override fun onDetachedFromWindow() {
     delegate.stop()
-    if (::eventsDisposable.isInitialized && eventsDisposable.isDisposed.not()) {
```
vinaysshenoy: 👏

app/src/main/java/org/simple/clinic/mobius/MobiusDelegate.kt
```diff
@@ -97,7 +97,8 @@ class MobiusDelegate<M : Parcelable, E, F>(
 
   fun onRestoreInstanceState(bundle: Bundle?): Parcelable? {
     lastKnownModel = bundle?.getParcelable(modelKey) ?: defaultModel
```
vinaysshenoy: Can we change this to accept a `Parcelable` and then cast to `Bundle` internally? The delegate can then change the return value of its `onSaveInstanceState` to a `Parcelable` as well. That way, the implementation detail that it saves the state as a `Bundle` can be hidden from the callers and we can avoid the cast at the call site.

PR #855, opened 2019-10-03
Strangulate logic from screen controller to Mobius

ragunathjawahar:
The following logic was moved from `PatientEntryScreenController` to Mobius

- Screen created
- Date of birth & age updates
- Toggle identifier section
- Scrolling to the bottom of the screen when gender is selected for the first time

### DIFF
--- a/app/src/main/java/org/simple/clinic/mobius/MobiusDelegate.kt
+++ b/app/src/main/java/org/simple/clinic/mobius/MobiusDelegate.kt
@@ -85,7 +85,9 @@ class MobiusDelegate<M : Parcelable, E, F>(
     if (::eventsDisposable.isInitialized && eventsDisposable.isDisposed.not()) {
       eventsDisposable.dispose()
     }
-    controller.stop()
+
+    startControllerIfNotAlreadyRunning()
+    stopAndDisconnectController()
   }
 
   fun onSaveInstanceState(androidViewState: Parcelable?): Parcelable {
@@ -116,4 +118,15 @@ class MobiusDelegate<M : Parcelable, E, F>(
 
   private fun identity(): Function<M, M> =
       Function { it }
+
+  private fun startControllerIfNotAlreadyRunning() {
+    if (controller.isRunning.not()) {
+      controller.start()
+    }
+  }
+
+  private fun stopAndDisconnectController() {
+    controller.stop()
+    controller.disconnect()
+  }
 }
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
@@ -22,6 +22,7 @@ object PatientEntryEffectHandler {
         .subtypeEffectHandler<PatientEntryEffect, PatientEntryEvent>()
         .addTransformer(FetchPatientEntry::class.java, fetchOngoingEntryEffectHandler(userSession, facilityRepository, patientRepository, schedulersProvider.io()))
         .addConsumer(PrefillFields::class.java, { ui.preFillFields(it.patientEntry) }, schedulersProvider.ui())
+        .addAction(ScrollFormToBottom::class.java, { ui.scrollFormToBottom() }, schedulersProvider.ui())
         .build()
   }
 
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
@@ -2,14 +2,50 @@ package org.simple.clinic.newentry
 
 import com.spotify.mobius.rx2.RxMobius
 import io.reactivex.ObservableTransformer
+import io.reactivex.Scheduler
+import io.reactivex.rxkotlin.Singles
+import org.simple.clinic.facility.FacilityRepository
+import org.simple.clinic.patient.OngoingNewPatientEntry
+import org.simple.clinic.patient.PatientRepository
+import org.simple.clinic.user.UserSession
 import org.simple.clinic.util.scheduler.SchedulersProvider
 
 object PatientEntryEffectHandler {
   fun createEffectHandler(
+      userSession: UserSession,
+      facilityRepository: FacilityRepository,
+      patientRepository: PatientRepository,
+      ui: PatientEntryUi,
       schedulersProvider: SchedulersProvider
   ): ObservableTransformer<PatientEntryEffect, PatientEntryEvent> {
     return RxMobius
         .subtypeEffectHandler<PatientEntryEffect, PatientEntryEvent>()
+        .addTransformer(FetchPatientEntry::class.java, fetchOngoingEntryEffectHandler(userSession, facilityRepository, patientRepository, schedulersProvider.io()))
+        .addConsumer(PrefillFields::class.java, { ui.preFillFields(it.patientEntry) }, schedulersProvider.ui())
         .build()
+  }
+
+  private fun fetchOngoingEntryEffectHandler(
+      userSession: UserSession,
+      facilityRepository: FacilityRepository,
+      patientRepository: PatientRepository,
+      scheduler: Scheduler
+  ): ObservableTransformer<FetchPatientEntry, PatientEntryEvent> {
+    return ObservableTransformer { fetchPatientEntries ->
+      val getPatientEntryAndFacility = Singles.zip(
+          patientRepository.ongoingEntry(),
+          facilityRepository.currentFacility(userSession).firstOrError()
+      )
+
+      fetchPatientEntries
+          .flatMapSingle { getPatientEntryAndFacility }
+          .subscribeOn(scheduler)
+          .map { (entry, facility) ->
+            // TODO(rj): 2019-10-03 Extract as function!
+            entry.takeIf { it.address != null }
+                ?: entry.copy(address = OngoingNewPatientEntry.Address(colonyOrVillage = "", district = facility.district, state = facility.state))
+          }
+          .map { OngoingEntryFetched(it) }
+    }
   }
 }
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
@@ -1,0 +1,15 @@
+package org.simple.clinic.newentry
+
+import com.spotify.mobius.rx2.RxMobius
+import io.reactivex.ObservableTransformer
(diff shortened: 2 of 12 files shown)


app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
```diff
@@ -0,0 +1,15 @@
+package org.simple.clinic.newentry
+
+import com.spotify.mobius.rx2.RxMobius
+import io.reactivex.ObservableTransformer
+import org.simple.clinic.util.scheduler.SchedulersProvider
+
+object PatientEntryEffectHandler {
+  fun createEffectHandler(
```
vinaysshenoy: Since we will be accessing this method through the `PatientEntryEffectHandler` object, I feel the `EffectHandler` suffix to the method name is redundant. What do you think about calling it just `create()` instead?

app/src/main/java/org/simple/clinic/newentry/PatientEntryLogic.kt
```diff
@@ -3,17 +3,24 @@ package org.simple.clinic.newentry
...
+import com.spotify.mobius.Next.next
 import com.spotify.mobius.Next.noChange
 
 fun patientEntryInit(
     model: PatientEntryModel
-): First<PatientEntryModel, PatientEntryEffect> {
-  return first(model)
-}
+): First<PatientEntryModel, PatientEntryEffect> =
+    first(model, setOf(FetchPatientEntry))
```
vinaysshenoy: We won't need to emit the `FetchPatientEntry` on a view restoration event, right?

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
```diff
@@ -118,7 +126,11 @@ class PatientEntryScreenControllerTest {
 
     screenCreated()
 
-    verify(ui).preFillFields(OngoingNewPatientEntry(
+    // FIXME
```
vinaysshenoy: 👍

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
```diff
@@ -87,6 +106,22 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
    **/
   private var alreadyFocusedOnEmptyTextField: Boolean = false
 
+  private val viewRenderer = PatientEntryViewRenderer(this)
+
+  private val events = PublishSubject.create<PatientEntryEvent>().hide().share()
```
vinaysshenoy: We don't need to `share()` this stream since it has only one consumer. Did you mean to also replace the event stream going to the controller with this shared stream?

vinaysshenoy: Discussed offline: This will break the feature because the stream is not shared between the old controller and the new mobius loop.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
```diff
@@ -92,7 +93,7 @@ class PatientEntryScreenController @Inject constructor(
                   district = facility.district,
                   state = facility.state))
         }
-        .map { { ui: Ui -> ui.preFillFields(it) } }
+        .flatMap { Observable.never<UiChange>() }
```
vinaysshenoy: Instead of doing this, why don't we mark this method as `@Deprecated` and just remove this from being part of the merged observable chain in the `apply()` method. That way, all the code paths in this will never get executed as well.

app/src/main/java/org/simple/clinic/newentry/PatientEntryViewRenderer.kt
```diff
@@ -4,6 +4,12 @@ import org.simple.clinic.mobius.ViewRenderer
 
 class PatientEntryViewRenderer(val ui: PatientEntryUi) : ViewRenderer<PatientEntryModel> {
   override fun render(model: PatientEntryModel) {
-    // **crickets**
```
vinaysshenoy: Did you finally find the bug-spray?

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
```diff
@@ -332,27 +331,4 @@ class PatientEntryScreenController @Inject constructor(
         .take(1)
         .map { { ui: Ui -> ui.scrollFormToBottom() } }
   }
-
-  private fun toggleVisibilityOfIdentifierSection(events: Observable<UiEvent>): Observable<UiChange> {
```
vinaysshenoy: 👏

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
```diff
@@ -555,18 +555,18 @@ class PatientEntryScreenControllerTest {
     verify(patientRepository).saveOngoingEntry(expectedSavedEntry)
   }
 
-  @Test // TODO: Migrate to Mobius
```
vinaysshenoy: I think we can add the `// TODO: Migrate to Mobius` line either above or below the `@Test` annotation instead of on the same line. This way, reading the diff becomes easier since it will only create a single-line change each time instead of two.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
vinaysshenoy: Discussed offline: May not be valid in all cases since these might be mutating state in between.

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
vinaysshenoy: This is for future PRs.

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
```diff
@@ -22,6 +22,7 @@ object PatientEntryEffectHandler {
         .subtypeEffectHandler<PatientEntryEffect, PatientEntryEvent>()
         .addTransformer(FetchPatientEntry::class.java, fetchOngoingEntryEffectHandler(userSession, facilityRepository, patientRepository, schedulersProvider.io()))
         .addConsumer(PrefillFields::class.java, { ui.preFillFields(it.patientEntry) }, schedulersProvider.ui())
+        .addAction(ScrollFormToBottom::class.java, { ui.scrollFormToBottom() }, schedulersProvider.ui())
```
vinaysshenoy: This can be a method reference.

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
ragunathjawahar: Noted, will make this change as part of the next PR because it is 10+ commits ahead.

app/src/main/java/org/simple/clinic/newentry/PatientEntryLogic.kt
ragunathjawahar: We will at the moment. This is the same behavior we have in the code right now. That explains why we are using the Activity paused event to save the ongoing into the database and then we are re-querying it in the `ScreenCreated()` event.

Since we do have a model, we can persist it in the `Bundle` and restore it and avoid re-queries. I would still keep this behavior and change this after the migration by writing a test.

I have noted this and will do it once we delete the controller class.

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
ragunathjawahar: This makes sense. Will do that for the following features.

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
ragunathjawahar: Will make this change as part of the Mobius refactor refactor PR.

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
```diff
@@ -475,10 +475,12 @@ class PatientEntryScreenControllerTest {
     verify(patientRegisteredCount).set(any())
   }
 
-  @Test // TODO: Migrate to Mobius
+  @Test
```
sanchita: Since the tests for Mobius will co-exist with older tests, what do you think about adding a comment to indicate/segregate both the tests? 
Or to have Mobius tests in a separate file?
Else it might be hard to just read the tests and figure out how much has been migrated or how to write future tests in Mobius.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
ragunathjawahar: Fixed this.

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
ragunathjawahar: @sanchita Noted, Mobius tests will live inside their own classes. We are touching (please take this with a grain of salt) any of the tests for now.

vinaysshenoy:
@ragunathjawahar marked this as WIP until the crash problem is fixed.

ragunathjawahar:
@vinaysshenoy fixed the crash and a leak. PTAL?

app/src/main/java/org/simple/clinic/mobius/MobiusDelegate.kt
```diff
@@ -85,7 +85,13 @@ class MobiusDelegate<M : Parcelable, E, F>(
...
       eventsDisposable.dispose()
     }
-    controller.stop()
+    with(controller) {
+      if (isRunning.not()) {
+        start()
+      }
+      stop()
+      disconnect()
+    }
```
vinaysshenoy: Since `MobiusDelegate` already has `start()` and `stop()` functions, it is not *obvious* that the `start()` and `stop()` functions here are actually methods on the `Controller`. In addition, this could potentially cause problems since removing the `with()` clause won't cause compilation errors and will end up in a recursive function call.

Can we extract these into separate methods? Suggestions: `startControllerIfNotAlreadyRunning` and `stopAndDisconnectController`.

ragunathjawahar: Done!

PR #857, opened 2019-10-04
Refactor remote config setup

vinaysshenoy:
https://www.pivotaltracker.com/story/show/168949449

### DIFF
--- a/app/src/main/java/org/simple/clinic/di/NetworkModule.kt
+++ b/app/src/main/java/org/simple/clinic/di/NetworkModule.kt
@@ -122,8 +122,8 @@ open class NetworkModule {
   }
 
   @Provides
-  fun remoteConfigReader(firebase: FirebaseRemoteConfig): ConfigReader {
-    return FirebaseConfigReader(firebase)
+  fun remoteConfigReader(firebase: FirebaseRemoteConfig, cacheExpiration: FirebaseRemoteConfigCacheExpiration): ConfigReader {
+    return FirebaseConfigReader(firebase, cacheExpiration)
   }
 
   @Provides
--- a/app/src/main/java/org/simple/clinic/di/NetworkModule.kt
+++ b/app/src/main/java/org/simple/clinic/di/NetworkModule.kt
@@ -19,9 +19,7 @@ import org.simple.clinic.patient.businessid.BusinessId
 import org.simple.clinic.patient.businessid.Identifier
 import org.simple.clinic.patient.sync.PatientPayload
 import org.simple.clinic.remoteconfig.ConfigReader
-import org.simple.clinic.remoteconfig.FirebaseConfigReader
 import org.simple.clinic.remoteconfig.FirebaseRemoteConfigCacheExpiration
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/simple/clinic/remoteconfig/ConfigReader.kt
```diff
@@ -1,8 +1,11 @@
 package org.simple.clinic.remoteconfig
 
+import io.reactivex.Completable
+
 interface ConfigReader {
   fun string(name: String, default: String): String
   fun boolean(name: String, default: Boolean): Boolean
   fun double(name: String, default: Double): Double
   fun long(name: String, default: Long): Long
+  fun update(): Completable
```
sanchita: Adding an `update` method in a reader, looks out of place since the `ConfigReader` is a utility class to just access config.
Updating it should be a separate `interface` like `ConfigUpdate` or `ConfigSync`.
Wdyt?

vinaysshenoy: I intentionally added this to this interface because if we need to connect to a different remote config service, we only need to update in one place instead of multiple. Maybe a different name for this instead of `ConfigReader`?

There's also another option I can explore:

```kotlin
interface RemoteConfigService {
  fun reader(): ConfigReader
  fun update(): Completable
}

class FirebaseRemoteConfigService {
  fun reader() = FirebaseConfigReader()
  fun update(): Completable {
    // update remote config here
  }
}
```

What do you think?

sanchita: I understand the reason for making this change. I was only looking to separate the `ConfigReader` from the `update` operation itself.
Also, we were exposing `update()` to classes that only need to access config values.

The implementation suggested in you comment looks good. So this will mean that classes that need to read config will only be inject the `ConfigReader` and not `RemoteConfigService`.

vinaysshenoy: They will receive the service and retrieve the reader through the service. So when we switch do a different service, we will be forced to implement both.

I will go down this route and see where it goes.

vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/di/NetworkModule.kt
```diff
@@ -122,8 +123,8 @@ open class NetworkModule {
   }
 
   @Provides
-  fun remoteConfigReader(firebase: FirebaseRemoteConfig, cacheExpiration: FirebaseRemoteConfigCacheExpiration): ConfigReader {
-    return FirebaseConfigReader(firebase, cacheExpiration)
+  fun remoteConfigReader(service: RemoteConfigService): ConfigReader {
+    return service.reader()
```
sanchita: This looks great.

sanchita:
LGTM. Ship it!

PR #860, opened 2019-10-07
Fix and extract spacing styles of edit button from patient summary

rakshakhegde:
Fix spacing designs as per https://zpl.io/VkGA8GM

### DIFF
--- a/app/src/main/res/layout/screen_home.xml
+++ b/app/src/main/res/layout/screen_home.xml
@@ -31,13 +31,9 @@
 
     <Button
       android:id="@+id/helpButton"
-      android:layout_width="wrap_content"
-      android:layout_height="wrap_content"
+      style="@style/Clinic.V2.ToolbarButton"
       android:layout_gravity="end"
-      android:background="@drawable/background_rounded_clearbutton"
-      android:foreground="?selectableItemBackground"
       android:text="@string/home_help"
-      android:textAllCaps="true"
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/screen_home.xml
```diff
@@ -31,13 +31,9 @@
...
     <Button
       android:id="@+id/helpButton"
-      android:layout_width="wrap_content"
-      android:layout_height="wrap_content"
+      style="@style/Clinic.V2.ToolbarButton"
       android:layout_gravity="end"
-      android:background="@drawable/background_rounded_clearbutton"
-      android:foreground="?selectableItemBackground"
       android:text="@string/home_help"
-      android:textAllCaps="true"
```
vinaysshenoy: This attribute got lost. Can we add this to the style too?

app/src/main/res/values/styles_v2.xml
```diff
@@ -299,6 +299,19 @@
     <item name="android:paddingBottom">18dp</item>
   </style>
 
+  <style name="Clinic.V2.ToolbarButton">
```
vinaysshenoy: 👏

app/src/main/res/layout/screen_home.xml
rakshakhegde: This is already a part of the text style. It was removed intentionally.

PR #862, opened 2019-10-08
Move all remaining logic from controller except validation

ragunathjawahar:


app/src/main/java/org/simple/clinic/newentry/PatientEntryLogic.kt
```diff
@@ -15,24 +15,24 @@ fun patientEntryUpdate(
...
-    val updatedModel = model.withGender(event.gender)
 
-    if (event.gender.isNotEmpty() && model.isSelectingGenderForTheFirstTime) {
-      return next(updatedModel.copy(isSelectingGenderForTheFirstTime = false), setOf(ScrollFormToBottom))
-    } else {
-      return next(updatedModel)
+    is GenderChanged -> {
+      val updatedModel = model.withGender(event.gender)
+
+      return if (event.gender.isNotEmpty() && model.isSelectingGenderForTheFirstTime) {
```
vinaysshenoy: This `return` is redundant since the `when` block is already being returned.

ragunathjawahar: I have moved all the logic out of the controller and have begun refactoring this section. Can we make it part of the next PR?

app/src/main/java/org/simple/clinic/newentry/PatientEntryModel.kt
```diff
@@ -58,4 +58,9 @@ data class PatientEntryModel(
 
   private fun getPersonalDetails(patientEntry: OngoingNewPatientEntry?): PersonalDetails =
       patientEntry?.personalDetails ?: PersonalDetails("", null, null, null)
+
+  fun withDistrict(district: String): PatientEntryModel {
+    val address = patientEntry?.address ?: Address("", "", "")
```
vinaysshenoy: Should we extract `patientEntry?.address ?: Address("", "", "")` into a separate function like we did for `PersonalDetails`? This is being used in three places.

ragunathjawahar: Yes, this I am refactoring this class. This will land in the next PR.

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
```diff
@@ -195,18 +195,15 @@ class PatientEntryScreenControllerTest {
...
       onNext(GenderChanged(Just(Transgender)))
       onNext(ColonyOrVillageChanged("colony"))
+      onNext(DistrictChanged("district"))
       reset(ui, patientRegisteredCount, patientRegisteredCount)
 
-      onNext(DistrictChanged("district"))
-      verify(ui).showEmptyDistrictError(false)
-      verify(ui).hideIdentifierSection()
-      verify(ui).setDateOfBirthAndAgeVisibility(DateOfBirthAndAgeVisibility.DATE_OF_BIRTH_VISIBLE)
+      onNext(StateChanged("state"))
```
vinaysshenoy: Shouldn't the commit message be "Capture behavior for StateChanged"

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
```diff
@@ -8,11 +8,8 @@ import com.nhaarman.mockito_kotlin.doReturn
 import com.nhaarman.mockito_kotlin.inOrder
```
vinaysshenoy: ⚰️

vinaysshenoy: 💐

vinaysshenoy: 🙏

app/src/main/java/org/simple/clinic/util/DistinctValueCallback.kt
```diff
@@ -0,0 +1,12 @@
+package org.simple.clinic.util
```
vinaysshenoy: This class is being used in the previous commit. Can you either:
- Reorder it to be before the previous commit?
- Squash it with the previous commit?

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
```diff
@@ -78,4 +84,27 @@ object PatientEntryEffectHandler {
...
+    return ObservableTransformer { savePatientEffects ->
+      savePatientEffects
+          .map { it.entry }
+          .subscribeOn(schedulersProvider.io())
+          .flatMapCompletable {
+            patientRepository
+                .saveOngoingEntry(it)
+                .observeOn(schedulersProvider.ui())
+                .andThen(Completable.fromAction {
+                  patientRegisteredCount.set(patientRegisteredCount.get().plus(1))
```
vinaysshenoy: This is getting set on the UI thread. Should we move this above into a `doOnNext` block instead?

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
```diff
@@ -181,27 +176,4 @@ class PatientEntryScreenController @Inject constructor(
 
     return showErrors.mergeWith(scrollToFirstError)
   }
-
-  private fun savePatient(events: Observable<UiEvent>): Observable<UiChange> {
-    val ongoingEntryChanges = events
-        .ofType<OngoingPatientEntryChanged>()
-        .map { it.entry }
-
-    val canPatientBeSaved = ongoingEntryChanges
```
vinaysshenoy: This behaviour of checking validation errors before saving was lost. Please take a look?

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
ragunathjawahar: I have split this into an event + effect combo in the upcoming PR. Also, doesn't `SharedPreferences` have its own threading model?

vinaysshenoy: It does, but only if you use `apply()` instead of `commit()`. I looked at the implementation of the `Preference` library, and it looks like it does. So I guess this is fine for now.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreenController.kt
ragunathjawahar: Fixed.

ragunathjawahar:
@vinaysshenoy squashed 5 commits to 1. The last commit is BIG. Sorry.

app/src/main/java/org/simple/clinic/mobius/Extensions.kt
```diff
@@ -0,0 +1,7 @@
+package org.simple.clinic.mobius
+
+import com.spotify.mobius.First
+import com.spotify.mobius.Init
+
+fun <M, F> ((M) -> First<M, F>).toInit(): Init<M, F> =
```
vinaysshenoy: Nice!

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
```diff
@@ -108,21 +118,27 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
...
+        schedulersProvider
+    )
+
     MobiusDelegate(
         events.ofType(),
         PatientEntryModel.DEFAULT,
-        ::patientEntryInit,
-        ::patientEntryUpdate,
-        PatientEntryEffectHandler.createEffectHandler(userSession, facilityRepository, patientRepository, this, schedulersProvider),
+        ::patientEntryInit.toInit(),
```
vinaysshenoy: On second thought, I don't like the asymmetry between the init() and update() function invocations here. We should just make `::patientEntryInit` a concrete class and directly instantiate it here. What do you think?

vinaysshenoy: Or make the `PatientEntryUpdate` a function and convert it into a class using an extension like above.

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
```diff
@@ -85,31 +79,26 @@ class PatientEntryScreenControllerTest {
     whenever(facilityRepository.currentFacility(userSession)).thenReturn(Observable.just(PatientMocker.facility()))
     whenever(patientRepository.ongoingEntry()).thenReturn(Single.never())
 
-    errorConsumer = { throw it }
-
     val sharedEvents = uiEvents.hide().share()
```
vinaysshenoy: We no longer need `sharedEvents` and can directly use the `uiEvents`.

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
```diff
@@ -535,12 +509,13 @@ class PatientEntryScreenControllerTest {
...
   }
 
-  @Test // TODO: Migrate to Mobius
+  @Test
   fun `when the ongoing entry has an identifier it must be retained when accepting input`() {
     val identifier = Identifier(value = "id", type = BpPassport)
     whenever(patientRepository.ongoingEntry()).thenReturn(Single.just(OngoingNewPatientEntry(identifier = identifier)))
     whenever(patientRepository.saveOngoingEntry(any())).thenReturn(Completable.complete())
     whenever(patientRegisteredCount.get()).thenReturn(0)
+    screenCreatedForMobius()
```
vinaysshenoy: This can just renamed to `screenCreated` now, and the older method can be removed.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
ragunathjawahar: 1. We could make `PatientEntryInit` a concrete class.
2. We can't make `PatientEntryUpdate` a class, because it needs access to `dobValidator` and `phoneNumberValidator`.

I'll go with 1 instead.

PR #864, opened 2019-10-09
Manually override the default locale in activities

vinaysshenoy:

This enables overriding of the language when opening an `Activity`. Currently, it will show the default language because the changing of the language is not yet implemented; it will be done in another PR.

To test this change, replace the `locale` being provided to the `LocaleOverrideContextWrapper` in any activity with `Locale.forLanguageTag("hi-IN")`.

### DIFF
--- a/app/src/main/java/org/simple/clinic/activity/TheActivity.kt
+++ b/app/src/main/java/org/simple/clinic/activity/TheActivity.kt
@@ -26,6 +26,7 @@ import org.simple.clinic.router.screen.KeyChangeAnimator
 import org.simple.clinic.router.screen.NestedKeyChanger
 import org.simple.clinic.router.screen.RouterDirection
 import org.simple.clinic.router.screen.ScreenRouter
+import org.simple.clinic.util.unsafeLazy
 import javax.inject.Inject
 
 class TheActivity : AppCompatActivity() {
@@ -43,7 +44,9 @@ class TheActivity : AppCompatActivity() {
   @Inject
   lateinit var fullScreenKeyChangeAnimator: KeyChangeAnimator<FullScreenKey>
 
-  lateinit var screenRouter: ScreenRouter
+  private val screenRouter: ScreenRouter by unsafeLazy {
+    ScreenRouter.create(this, NestedKeyChanger(), screenResults)
+  }
 
   private val screenResults: ScreenResultBus = ScreenResultBus()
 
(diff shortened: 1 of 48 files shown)


app/src/main/java/org/simple/clinic/activity/TheActivity.kt
```diff
@@ -64,19 +67,12 @@ class TheActivity : AppCompatActivity() {
   }
 
   override fun attachBaseContext(baseContext: Context) {
+    setupDiGraph()
```
vinaysshenoy: This was needed to inject the locale before the `ScreenRouter` gets setup so that the context could be overriden with the selected `Locale`.

app/src/main/java/org/simple/clinic/util/LocaleOverrideContextWrapper.kt
```diff
@@ -0,0 +1,26 @@
+package org.simple.clinic.util
+
+import android.content.Context
+import android.content.ContextWrapper
+import java.util.Locale
+
+class LocaleOverrideContextWrapper(base: Context) : ContextWrapper(base) {
```
rakshakhegde: This constructor is only being used in this class, let's make it `private constructor`?

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsView.kt
```diff
@@ -87,7 +87,7 @@ class RecentPatientsView(context: Context, attrs: AttributeSet) : FrameLayout(co
   }
 
   fun openPatientSummary(patientUuid: UUID) {
-    activity.screenRouter.push(
+    screenRouter.push(
```
rakshakhegde: The injected `activity` above is unused. Let's remove it.

app/src/main/java/org/simple/clinic/activity/TheActivity.kt
```diff
@@ -68,7 +73,8 @@ class TheActivity : AppCompatActivity() {
 
   override fun attachBaseContext(baseContext: Context) {
     setupDiGraph()
-    val contextWithRouter = wrapContextWithRouter(baseContext)
+    val contextWithOverridenLocale = LocaleOverrideContextWrapper.wrap(baseContext, locale)
```
rakshakhegde: Nitpick 😛 It is `overridden` with double d.

vinaysshenoy: Whoops! will fix!

PR #865, opened 2019-10-09
Finish migrating patient entry screen to Mobius

ragunathjawahar:
Non-blocking PR feedback from the previous feedback have been addressed.

I have raised a PR to preserve validation errors during screen restoration. This will be part of another PR. Here is the corresponding Pivotal story - https://www.pivotaltracker.com/story/show/169030605


app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
```diff
@@ -104,26 +106,27 @@ object PatientEntryEffectHandler {
...
 
+  private fun savePatientEntry(
+      entry: OngoingNewPatientEntry,
+      patientRepository: PatientRepository
+  ): Single<Boolean> {
+    return patientRepository
+        .saveOngoingEntry(entry)
+        .andThen(Single.just(true))
+  }
+
```
vinaysshenoy: Since we are not using the result of this anywhere, should we just change the return type to `Single<Any>` instead?

app/src/main/java/org/simple/clinic/newentry/PatientEntryViewRenderer.kt
```diff
@@ -8,7 +8,7 @@ import org.simple.clinic.widgets.ageanddateofbirth.DateOfBirthAndAgeVisibility.B
 import org.simple.clinic.widgets.ageanddateofbirth.DateOfBirthAndAgeVisibility.DATE_OF_BIRTH_VISIBLE
 
 class PatientEntryViewRenderer(val ui: PatientEntryUi) : ViewRenderer<PatientEntryModel> {
-  private val dateOfBirthAndAgeVisibilityCallback: DistinctValueCallback<DateOfBirthAndAgeVisibility> = DistinctValueCallback()
+  private val distinctDateOfBirthAndAgeVisibilityCallback: DistinctValueCallback<DateOfBirthAndAgeVisibility> = DistinctValueCallback()
```
vinaysshenoy: Suggestion for renaming: `CallbackIfValueChanged` might expose more details about the intention of this class than `DistinctValueCallback`.

app/src/main/java/org/simple/clinic/newentry/PatientEntryViewRenderer.kt
```diff
@@ -1,27 +1,36 @@
...
   override fun render(model: PatientEntryModel) {
     val patientEntry = model.patientEntry ?: return
+    showOrHideIdentifier(patientEntry.identifier)
 
-    if (patientEntry.identifier != null) {
+    val personalDetails = patientEntry.personalDetails ?: return
+    changeDateOfBirthAndAgeVisibility(personalDetails)
+  }
+
+  private fun showOrHideIdentifier(identifier: Identifier?) {
```
vinaysshenoy: `showOrHideIdentifier` leaves some ambiguity from the name. Should we just call it `renderIdentifier` instead? Whether to hide or do something else is an implementation detail.

app/src/main/java/org/simple/clinic/mobius/MobiusFoo.kt
```diff
@@ -2,6 +2,11 @@ package org.simple.clinic.mobius
 
 import com.spotify.mobius.First
 import com.spotify.mobius.Init
+import com.spotify.mobius.Next
+import com.spotify.mobius.Next.dispatch
 
 fun <M, F> ((M) -> First<M, F>).toInit(): Init<M, F> =
     Init { this.invoke(it) }
+
+fun <M, F> justEffect(effect: F): Next<M, F> =
```
vinaysshenoy: Should this take in a `vararg` parameter?

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
ragunathjawahar: We could do a `Single<Unit>`

app/src/main/java/org/simple/clinic/mobius/MobiusFoo.kt
```diff
@@ -10,3 +10,6 @@ fun <M, F> ((M) -> First<M, F>).toInit(): Init<M, F> =
 
 fun <M, F> justEffect(effect: F): Next<M, F> =
     dispatch<M, F>(setOf(effect))
+
+fun <M, F> next(model: M, effect: F): Next<M, F> =
```
vinaysshenoy: Should this take in a `vararg` parameter for effects?

app/src/main/java/org/simple/clinic/newentry/PatientEntryUpdate.kt
```diff
@@ -1,58 +1,71 @@
...
+    }
+  }
 
-      is DateOfBirthFocusChanged -> {
-        val hasDateOfBirth = model.patientEntry?.personalDetails?.dateOfBirth?.isNotBlank() == true
-        // TODO(rj): 2019-10-04 Extract the justEffect function and use it instead.
-        Next.dispatch<PatientEntryModel, PatientEntryEffect>(setOf(ShowDatePatternInDateOfBirthLabel(event.hasFocus || hasDateOfBirth)))
-      }
+  private fun onTextChanged(effect: PatientEntryEffect, reducer: () -> PatientEntryModel): PatientEntryNext =
+      next(reducer(), effect)
```
vinaysshenoy: The reducer doesn't actually do anything here. Why can't we just pass in the updated model directly?

```kotlin
onTextChanged(HideEmptyDateOfBirthAndAgeError, model.withAge(event.age))
```

app/src/main/java/org/simple/clinic/newentry/PatientEntryInit.kt
```diff
@@ -2,8 +2,9 @@ package org.simple.clinic.newentry
 
 import com.spotify.mobius.First
 import com.spotify.mobius.First.first
+import com.spotify.mobius.Init
 
-fun patientEntryInit(
-    model: PatientEntryModel
-): First<PatientEntryModel, PatientEntryEffect> =
-    first(model, setOf(FetchPatientEntry))
+object PatientEntryInit : Init<PatientEntryModel, PatientEntryEffect> {
```
vinaysshenoy: Why is this an `object` class instead of being a normal class?

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenControllerTest.kt
```diff
@@ -564,7 +559,6 @@ class PatientEntryScreenControllerTest {
   }
 
   private fun screenCreated() {
```
vinaysshenoy: This function can be inlined since the only thing it does is call `screenCreatedForMobius()`.

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
vinaysshenoy: Works.

app/src/main/java/org/simple/clinic/mobius/MobiusFoo.kt
ragunathjawahar: I guess we should do the following,

```
fun <M, F> effects(effect: F, vararg effects: F): Next<M, F>
```

and keep the current one as is, thoughts?

vinaysshenoy: Works.

vinaysshenoy: Maybe choose the name as `justEffects` to retain consistency?

ragunathjawahar: I am looking at the patterns in the following features,
* All Patients In Facility
* Edit Patient
* New Entry
* Onboarding

I have enough samples to extract and reuse one or more functions. Can we make this part of a separate PR (I'm working on it now)?

app/src/main/java/org/simple/clinic/mobius/MobiusFoo.kt
ragunathjawahar: Same as above.

app/src/main/java/org/simple/clinic/mobius/MobiusFoo.kt
vinaysshenoy: Works.

vinaysshenoy: While you're doing this, there's also an unused extension function `First.toInit()` that we can remove.

PR #867, opened 2019-10-09
Add encounter model and link to BPs

sanchita:
Pivotol story -> https://www.pivotaltracker.com/story/show/168448997

FYI: This PR has two database schema files hence the high LOC.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -307,7 +307,7 @@ dependencies {
     because "Traceur is incompatible with the version of RxJava this artifact pulls in transitively"
   }
   testImplementation "org.threeten:threetenbp:$versions.threeTenBp"
-  implementation "com.github.f4b6a3:uuid-creator:1.4.1"
+  implementation "com.fasterxml.uuid:java-uuid-generator:$versions.uuidGenerator"
 }
 
 // This must always be present at the bottom of this file, as per:
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -307,6 +307,7 @@ dependencies {
     because "Traceur is incompatible with the version of RxJava this artifact pulls in transitively"
   }
   testImplementation "org.threeten:threetenbp:$versions.threeTenBp"
+  implementation "com.github.f4b6a3:uuid-creator:1.4.1"
 }
 
 // This must always be present at the bottom of this file, as per:
--- a/app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
@@ -20,12 +20,12 @@ import org.simple.clinic.user.User
 import org.simple.clinic.util.Just
 import org.simple.clinic.util.None
 import org.simple.clinic.util.Optional
+import org.simple.clinic.util.TestUserClock
 import org.simple.clinic.util.TestUtcClock
-import org.simple.clinic.util.createUuid5
+import org.simple.clinic.util.generateEncounterUuid
 import org.simple.clinic.util.toLocalDateAtZone
 import org.threeten.bp.Instant
 import org.threeten.bp.LocalDate
-import org.threeten.bp.ZoneOffset
 import java.util.UUID
 import javax.inject.Inject
 import javax.inject.Named
@@ -3460,8 +3460,13 @@ class DatabaseMigrationAndroidTest {
     val db_50 = helper.migrateTo(version = 50)
 
     //then
-    val expectedEncounteredOn = Instant.parse(recordedAtForFirstBP).toLocalDateAtZone(ZoneOffset.UTC).toString()
-    val expectedEncounterId = createUuid5(facilityUuid + patientUuid + expectedEncounteredOn).toString()
+    val userClock = TestUserClock()
+    val expectedEncounteredOn = Instant.parse(recordedAtForFirstBP).toLocalDateAtZone(userClock.zone)
+    val expectedEncounterId = generateEncounterUuid(
+        facilityUuid = UUID.fromString(facilityUuid),
+        patientUuid = UUID.fromString(patientUuid),
+        encounteredDate = expectedEncounteredOn
+    ).toString()
 
     db_50.query("""SELECT * FROM '$bloodPressureMeasurementTable' """).use {
       assertThat(it.count).isEqualTo(2)
@@ -3479,7 +3484,7 @@ class DatabaseMigrationAndroidTest {
     val expectedEncounterValues = mapOf(
         "uuid" to expectedEncounterId,
         "patientUuid" to patientUuid,
-        "encounteredOn" to expectedEncounteredOn,
+        "encounteredOn" to expectedEncounteredOn.toString(),
         "createdAt" to createdAtForFirstBP,
         "updatedAt" to createdAtForFirstBP,
         "deletedAt" to "null"
--- a/app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
@@ -21,8 +21,11 @@ import org.simple.clinic.util.Just
 import org.simple.clinic.util.None
 import org.simple.clinic.util.Optional
 import org.simple.clinic.util.TestUtcClock
+import org.simple.clinic.util.createUuid5
+import org.simple.clinic.util.toLocalDateAtZone
 import org.threeten.bp.Instant
 import org.threeten.bp.LocalDate
+import org.threeten.bp.ZoneOffset
 import java.util.UUID
 import javax.inject.Inject
 import javax.inject.Named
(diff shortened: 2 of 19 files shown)


app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -3329,6 +3329,29 @@ class DatabaseMigrationAndroidTest {
     val dbV47 = helper.migrateTo(47)
     dbV47.assertTableDoesNotExist("HomescreenIllustration")
   }
+
+  @Test
+  fun migration_from_47_to_48_should_create_encounter_table() {
+    //given
+    val tableName = "Encounter"
```
ragunathjawahar: Call this variable `encounterTable` instead? Lowers cognitive overhead.

app/src/main/java/org/simple/clinic/storage/Migration_47_48.kt
```diff
@@ -0,0 +1,22 @@
...
+    CREATE TABLE IF NOT EXISTS "Encounter" (
+    "uuid" TEXT NOT NULL,
+    "patientUuid" TEXT NOT NULL,
+    "encounteredOn" TEXT NOT NULL,
+    "createdAt" TEXT NOT NULL,
+    "updatedAt" TEXT NOT NULL,
+    "deletedAt" TEXT,
+    PRIMARY KEY("uuid")
+    )
+    """)
```
ragunathjawahar: Please check formatting?

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -3352,6 +3355,85 @@ class DatabaseMigrationAndroidTest {
         "deletedAt")
     )
   }
+
+  @Test
+  fun migration_from_48_to_49_create_only_one_encounter_for_per_day_but_add_encounterUuid_to_all_BPs() {
+    //given
+    val tableName = "BloodPressureMeasurement"
```
ragunathjawahar: variable name `bloodPressureMeasurementTable`?

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -3352,6 +3355,85 @@ class DatabaseMigrationAndroidTest {
...
+
+    val uuidForFirstBP = "3eb96138-5270-419c-bcdf-86db95d0bc8b"
+    val uuidForSecondBP = "a2455b0c-cf14-46c9-9f73-6e18972982b0"
+    val facilityUuid = "0086ba86-314c-49d9-8822-2563e53bed92"
+    val patientUuid = "c4233170-feb2-3d34-8115-98fef5a68837"
+    val recordedAtForFirstBP = "2018-07-03T00:00:00Z"
+    val recordedAtForSecondBP = "2018-07-03T01:30:00Z"
+    val createdAtForFirstBP = "2018-07-03T00:00:00Z"
+    val createdAtForSecondBP = "2018-07-03T01:30:00Z"
```
ragunathjawahar: Can we do `val createdAtForFirstBp = recordedAtForFirstBP`? and `recordedAtForSecondBP` as well?

sanchita: We can do that but we don't expect both these values to be the same always.
But I understand what you mean. I will change the `Instant` for both `createdAt`.

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -3352,6 +3355,85 @@ class DatabaseMigrationAndroidTest {
...
+
+    val db_47 = helper.createDatabase(version = 48)
+    db_47.insert(tableName, mapOf(
+        "uuid" to uuidForFirstBP,
+        "systolic" to "120",
+        "diastolic" to "90",
+        "syncStatus" to "DONE",
+        "userUuid" to "d689285f-5c92-4735-80ef-c0e360ad3f1d",
```
ragunathjawahar: Extract this as a variable `"d689285f-5c92-4735-80ef-c0e360ad3f1d"` ?

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -3352,6 +3355,85 @@ class DatabaseMigrationAndroidTest {
...
+        "deletedAt" to "null",
+        "recordedAt" to recordedAtForSecondBP
+    ))
+
+    //when
+    val db_49 = helper.migrateTo(version = 49)
+
+    //then
+    val expectedEncounteredOn = Instant.parse(recordedAtForFirstBP).toLocalDateAtZone(ZoneOffset.UTC).toString()
+    val expectedEncounterId = createUuid5(facilityUuid + patientUuid + expectedEncounteredOn).toString()
```
ragunathjawahar: `createUuid5(facilityUuid + patientUuid + expectedEncounteredOn)` this method signature reveals too much information and does not reveal intent, can we change the signature to `createUuid5(facilityUuid, patientUuid, expectedEncounteredOn)`?

app/src/androidTest/java/org/simple/clinic/TestData.kt
```diff
@@ -676,7 +677,8 @@ class TestData @Inject constructor(
         createdAt = createdAt,
         updatedAt = updatedAt,
         deletedAt = deletedAt,
-        recordedAt = recordedAt
+        recordedAt = recordedAt,
+        encounterUuid = createUuid5(facilityUuid.toString() + patientUuid + recordedAt)
```
ragunathjawahar: Same as mentioned above - function signature.

app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
```diff
@@ -48,7 +49,8 @@ class BloodPressureRepository @Inject constructor(
                 createdAt = now,
                 updatedAt = now,
                 deletedAt = null,
-                recordedAt = recordedAt)
+                recordedAt = recordedAt,
+                encounterUuid = createUuid5(currentFacility.uuid.toString() + patientUuid + recordedAt))
```
ragunathjawahar: Here and other usage sites as well.

app/src/main/java/org/simple/clinic/storage/Migration_48_49.kt
```diff
@@ -0,0 +1,54 @@
...
+@Suppress("ClassName")
+class Migration_48_49 : Migration(48, 49) {
+
+  override fun migrate(database: SupportSQLiteDatabase) {
+    val tableName = "BloodPressureMeasurement"
+
+    database.inTransaction {
+      //Using a random UUID to give a default value to non-null encounterUuid temporarily.
+      //The encounterUuid will be replaced with the correct value when the query executes in next line.
+      database.execSQL(""" ALTER TABLE $tableName ADD COLUMN "encounterUuid" TEXT NOT NULL DEFAULT '05f9c798-1701-4379-b14a-b1b18d937a33'  """)
```
ragunathjawahar: I don't understand why we are doing this - `05f9c798-1701-4379-b14a-b1b18d937a33` can you please explain?

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
```diff
@@ -110,7 +111,9 @@ object PatientMocker {
         createdAt = createdAt,
         updatedAt = updatedAt,
         deletedAt = deletedAt,
-        recordedAt = recordedAt)
+        recordedAt = recordedAt,
+        encounterUuid = createUuid5(facilityUuid.toString() + patientUuid + recordedAt)
```
ragunathjawahar: +1 here too.

app/build.gradle
```diff
@@ -307,6 +307,7 @@ dependencies {
     because "Traceur is incompatible with the version of RxJava this artifact pulls in transitively"
   }
   testImplementation "org.threeten:threetenbp:$versions.threeTenBp"
+  implementation "com.github.f4b6a3:uuid-creator:1.4.1"
```
ragunathjawahar: Would you consider using a library from a more reliable source?

https://github.com/cowtowncoder/java-uuid-generator

This library is from the Jackson folks.

app/src/main/java/org/simple/clinic/storage/Migration_47_48.kt
sanchita: The SQL query string doesn't get auto-formatted.
We should maybe add a formatter for these as well.

app/build.gradle
sanchita: The library we are using right now is already tested with between app and server.
If you think we should still update, I will make this change.

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
sanchita: I have added a separate method which is responsible for generating UUIDs for `Encounter`.
PTAL. :)

PR #868, opened 2019-10-09
Patient consent api changes

rakshakhegde:
https://www.pivotaltracker.com/story/show/168949014

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/ReminderConsent.kt
+++ b/app/src/main/java/org/simple/clinic/patient/ReminderConsent.kt
@@ -1,0 +1,47 @@
+package org.simple.clinic.patient
+
+import android.os.Parcelable
+import androidx.room.TypeConverter
+import com.squareup.moshi.FromJson
+import com.squareup.moshi.ToJson
+import kotlinx.android.parcel.Parcelize
+import org.simple.clinic.util.room.SafeEnumTypeAdapter
+
+sealed class ReminderConsent : Parcelable {
+
+  @Parcelize
+  object Granted : ReminderConsent()
+
+  @Parcelize
+  object Denied : ReminderConsent()
+
+  @Parcelize
+  data class Unknown(val actualValue: String) : ReminderConsent()
+
+  object TypeAdapter : SafeEnumTypeAdapter<ReminderConsent>(
+      knownMappings = mapOf(
+          Granted to "granted",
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/simple/clinic/patient/sync/PatientPayload.kt
```diff
@@ -75,7 +80,9 @@ data class PatientPayload(
         updatedAt = updatedAt,
         deletedAt = deletedAt,
         recordedAt = recordedAt,
-        syncStatus = newStatus)
+        syncStatus = newStatus,
+        reminderConsent = reminderConsent ?: Granted
```
vinaysshenoy: Can you add a `TODO` here to make this non-nullable when the reminder consent property is pushed to production?

app/src/main/java/org/simple/clinic/patient/sync/PatientSync.kt
```diff
@@ -73,7 +74,9 @@ class PatientSync @Inject constructor(
                 address = address.toPayload(),
                 phoneNumbers = numberPayloads,
                 businessIds = businessIdPayloads,
-                recordedAt = recordedAt)
+                recordedAt = recordedAt,
+                reminderConsent = Granted
```
vinaysshenoy: Shouldn't this be reading the `reminderConsent` from the `Patient` instead of hardcoding it to `Granted`?

app/src/main/java/org/simple/clinic/patient/ReminderConsent.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    @ToJson
+    fun toJson(status: ReminderConsent?): String? = TypeAdapter.fromEnum(status)
+  }
+
+  companion object {
+
+    @VisibleForTesting
+    fun random() = TypeAdapter.knownMappings.keys.shuffled().first()
+  }
```
vinaysshenoy: The `random()` function isn't being used in tests since we have hard-coded it to `Granted`. I think we can remove this.

app/src/main/java/org/simple/clinic/patient/sync/PatientSync.kt
rakshakhegde: Right! Will fix.

PR #870, opened 2019-10-10
UI for change language screen

vinaysshenoy:

This is a large PR, but most of it is boilerplate around UI and a layout resource.


app/src/main/java/org/simple/clinic/settings/SettingsRepositoryImpl.kt
```diff
@@ -0,0 +1,19 @@
+package org.simple.clinic.settings
+
+import io.reactivex.Completable
+import io.reactivex.Single
+
+class SettingsRepositoryImpl : SettingsRepository {
```
vinaysshenoy: This class is fine for now because the screen is unreachable from the UI. The final integration will happen in the next PR.

PR #871, opened 2019-10-10
Mobius Foo utility functions to get rid of the use of `setOf` when dealing with effects

ragunathjawahar:


app/src/main/java/org/simple/clinic/mobius/MobiusFoo.kt
```diff
@@ -1,15 +1,13 @@
...
 
-fun <M, F> ((M) -> First<M, F>).toInit(): Init<M, F> =
-    Init { this.invoke(it) }
-
 fun <M, F> justEffect(effect: F): Next<M, F> =
     dispatch<M, F>(setOf(effect))
 
-fun <M, F> next(model: M, effect: F): Next<M, F> =
-    Next.next(model, setOf(effect))
+fun <M, F> next(model: M, vararg effects: F): Next<M, F> = if (effects.isEmpty()) {
```
vinaysshenoy: 🚀

PR #872, opened 2019-10-10
Stabilize tests by replacing mock BP validator with a real implementation

ragunathjawahar:


ragunathjawahar:
@sanchita removing you as a reviewer, because we haven't on-boarded to Mobius yet.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTest.kt
```diff
@@ -549,27 +516,6 @@ class BloodPressureEntrySheetControllerTest {
...
-      onNext(SystolicChanged("-"))
-      onNext(DiastolicChanged("-"))
-      onNext(BloodPressureDateClicked)
-    }
-
-    verify(ui, never()).showDateEntryScreen()
-  }
-
   @Suppress("unused")
   fun `params for OpenAs and bp validation errors`(): List<Any> {
```
ragunathjawahar: Did not move this supplier function, because there is another test in the class that relies on this. Hence, this function was copied and later refactored.

PR #875, opened 2019-10-11
Remove constructors that take init and update functions (both fixture and delegate)

ragunathjawahar:


mobius-migration/src/main/java/org/simple/mobius/migration/MobiusTestFixture.kt
```diff
@@ -31,24 +31,6 @@ class MobiusTestFixture<M: Any, E, F>(
     modelUpdateListener: ModelUpdateListener<M>,
     requiresLogging: Boolean = false
 ) {
-  constructor(
```
vinaysshenoy: 🚀

PR #878, opened 2019-10-14
Implement platform logic in SettingsRepository

vinaysshenoy:

### DIFF
--- a/app/src/main/java/org/simple/clinic/settings/SettingsModule.kt
+++ b/app/src/main/java/org/simple/clinic/settings/SettingsModule.kt
@@ -21,7 +21,12 @@ class SettingsModule {
   }
 
   @Provides
-  fun provideSettingsRepository(): SettingsRepository {
-    return SettingsRepositoryImpl()
+  fun provideSettingsRepository(
+      @Named("preference_user_selected_locale") userSelectedLocalePreference: Preference<Optional<Locale>>
+  ): SettingsRepository {
+    return SettingsRepositoryImpl(
+        userSelectedLocalePreference = userSelectedLocalePreference,
+        supportedLanguages = emptyList()
+    )
   }
 }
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/simple/clinic/settings/SettingsModule.kt
```diff
@@ -21,7 +21,12 @@ class SettingsModule {
...
 
   @Provides
-  fun provideSettingsRepository(): SettingsRepository {
-    return SettingsRepositoryImpl()
+  fun provideSettingsRepository(
+      @Named("preference_user_selected_locale") userSelectedLocalePreference: Preference<Optional<Locale>>
+  ): SettingsRepository {
+    return SettingsRepositoryImpl(
+        userSelectedLocalePreference = userSelectedLocalePreference,
+        supportedLanguages = emptyList()
```
ragunathjawahar: Curious, why is supported languages an empty list?

vinaysshenoy: It's empty for now because it's not used in production. The final commit in this PR adds the actual languages we currently support.

app/src/main/java/org/simple/clinic/settings/SettingsRepositoryImpl.kt
```diff
@@ -1,12 +1,31 @@
...
+import com.f2prateek.rx.preferences2.Preference
 import io.reactivex.Completable
+import io.reactivex.Maybe
 import io.reactivex.Single
+import org.simple.clinic.util.Optional
+import org.simple.clinic.util.filterAndUnwrapJust
+import java.util.Locale
 
-class SettingsRepositoryImpl : SettingsRepository {
+class SettingsRepositoryImpl(
```
ragunathjawahar: Can we use a different name instead? `PreferencesSettingsRepository`?

vinaysshenoy: Done.

PR #880, opened 2019-10-14
Clean up old AllPatientsInFacilityView code

ragunathjawahar:


app/src/main/java/org/simple/clinic/addidtopatient/searchforpatient/AddIdToPatientSearchScreen.kt
```diff
@@ -72,15 +60,7 @@ class AddIdToPatientSearchScreen(context: Context, attrs: AttributeSet) : Relati
...
-    val (layoutResId, architecture) = if (mobiusMigrationConfig.useAllPatientsInFacilityView) {
-      R.layout.view_allpatientsinfacility to MOBIUS
-    } else {
-      R.layout.view_allpatientsinfacility_old to ORIGINAL
-    }
-
-    Analytics.reportArchitectureMigration("AllPatientsInFacilityView", architecture, "AddIdToPatientSearchScreen")
-
-    allPatientsViewStub.layoutResource = layoutResId
+    allPatientsViewStub.layoutResource = R.layout.view_allpatientsinfacility
```
vinaysshenoy: Do we still need to inflate this manually instead of adding it to the layout statically?

app/src/main/java/org/simple/clinic/addidtopatient/searchforpatient/AddIdToPatientSearchScreen.kt
```diff
@@ -72,15 +60,7 @@ class AddIdToPatientSearchScreen(context: Context, attrs: AttributeSet) : Relati
...
     allPatientsViewStub.inflate() as ExposesUiEvents
```
vinaysshenoy: We no longer need the `ExposedUiEvents` interface since that was meant to handle the fact that there are two different views, right?

PR #881, opened 2019-10-14
Restart the activity after the language is changed

vinaysshenoy:

### DIFF
--- a/app/src/main/java/org/simple/clinic/settings/changelanguage/ChangeLanguageScreen.kt
+++ b/app/src/main/java/org/simple/clinic/settings/changelanguage/ChangeLanguageScreen.kt
@@ -1,5 +1,6 @@
 package org.simple.clinic.settings.changelanguage
 
+import android.app.Activity
 import android.content.Context
 import android.os.Parcelable
 import android.util.AttributeSet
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/simple/clinic/settings/changelanguage/ChangeLanguageScreen.kt
```diff
@@ -138,9 +142,16 @@ class ChangeLanguageScreen(
   }
 }
 
-private class ChangeLanguageScreenUiActions(private val screenRouter: ScreenRouter) : UiActions {
+private class ChangeLanguageScreenUiActions(
+    private val screenRouter: ScreenRouter,
+    private val activity: Activity
+) : UiActions {
```
ragunathjawahar: Does this have to be a class? Can we just make the screen implement this interface instead?

ragunathjawahar: This can go in the next PR.

vinaysshenoy: I'll make this change.

PR #882, opened 2019-10-14
Open the change language screen from the settings screen

vinaysshenoy:


app/src/main/java/org/simple/clinic/settings/SettingsUpdate.kt
```diff
@@ -10,6 +11,7 @@ class SettingsUpdate : Update<SettingsModel, SettingsEvent, SettingsEffect> {
     return when (event) {
       is UserDetailsLoaded -> next(model.userDetailsFetched(name = event.name, phoneNumber = event.phoneNumber))
       is CurrentLanguageLoaded -> next(model.currentLanguageFetched(event.language))
+      is ChangeLanguage -> justEffect(OpenLanguageSelectionScreenEffect)
```
vinaysshenoy: It is fine for us to emit this effect now even though there is no handler because the event is not getting triggered from the UI. The handler will be added in a later commit.

app/src/main/java/org/simple/clinic/settings/changelanguage/ChangeLanguageListItem.kt
```diff
@@ -11,7 +11,7 @@ import org.simple.clinic.widgets.recyclerview.ViewHolderX
 
 data class ChangeLanguageListItem(
     private val language: ProvidedLanguage,
-    private val isSelected: Boolean
```
vinaysshenoy: This was renamed because it was actually picking up the `View#isSelected` property instead of the one from the list item.

app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
```diff
@@ -11,20 +11,36 @@ object SettingsEffectHandler {
...
   }
 
-  private fun loadUserDetails(userSession: UserSession, scheduler: Scheduler): ObservableTransformer<LoadUserDetailsEffect, SettingsEvent> {
-    return ObservableTransformer { upstream ->
-      upstream
+  private fun loadUserDetails(
+      userSession: UserSession,
+      scheduler: Scheduler
+  ): ObservableTransformer<LoadUserDetailsEffect, SettingsEvent> {
+    return ObservableTransformer { effectStream ->
```
ragunathjawahar: `loadUserDetailsEffects` instead of `effectStream`?

app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
```diff
@@ -11,20 +11,36 @@ object SettingsEffectHandler {
...
           .filterAndUnwrapJust()
           .map { user -> UserDetailsLoaded(user.fullName, user.phoneNumber) }
     }
   }
+
+  private fun loadCurrentSelectedLanguage(
+      settingsRepository: SettingsRepository,
+      scheduler: Scheduler
+  ): ObservableTransformer<LoadCurrentLanguageEffect, SettingsEvent> {
+    return ObservableTransformer { effectStream ->
```
ragunathjawahar: `loadCurrentLanguageEffects`?

app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
vinaysshenoy: I intentionally called it `effectStream` because I don't want to think about what to name it. The type of effect is anyway *obvious* from the function signature. If there is an effect whose properties I need to consume, I can name it at the site of usage as the lambda parameter.

Does this make sense?

app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
vinaysshenoy: Same as above.

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
```diff
@@ -40,7 +46,7 @@ class SettingsEffectHandlerTest {
         phone = "1234567890"
     )
     val updatedUser = savedUser.copy(fullName = "Mahalakshmi T Sharma", phoneNumber = "0987654321")
```
ragunathjawahar: Can we not expose the copy constructor and use a reducer function instead?
`savedUser.copy(fullName = "Mahalakshmi T Sharma", phoneNumber = "0987654321")`

vinaysshenoy: I didn't want to add this because that function will only be used in tests. However, I could make it an extension function that's private to this test. I'm not sure I'd want to do this since it would only be used in one place, however.

app/src/main/java/org/simple/clinic/settings/SettingsModel.kt
```diff
@@ -5,16 +5,18 @@ import kotlinx.android.parcel.Parcelize
...
 data class SettingsModel(
-    val userDetailsQueried: Boolean,
     val name: String?,
     val phoneNumber: String?
-): Parcelable {
+) : Parcelable {
 
   companion object {
-    val FETCHING_USER_DETAILS = SettingsModel(userDetailsQueried = false, name = null, phoneNumber = null)
+    val FETCHING_USER_DETAILS = SettingsModel(name = null, phoneNumber = null)
```
ragunathjawahar: Nice that you got rid of the boolean.

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
vinaysshenoy: Never mind, I'll add the reducer functions since there are places we can use it in production. We can change the production usages later.

app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
```diff
@@ -12,12 +12,14 @@ object SettingsEffectHandler {
   fun create(
       userSession: UserSession,
       settingsRepository: SettingsRepository,
-      schedulersProvider: SchedulersProvider
+      schedulersProvider: SchedulersProvider,
+      uiActions: UiActions
```
ragunathjawahar: Make `schedulersProvider` the last parameter?

app/src/main/res/layout/screen_change_language.xml
```diff
@@ -2,6 +2,7 @@
 <org.simple.clinic.settings.changelanguage.ChangeLanguageScreen xmlns:android="http://schemas.android.com/apk/res/android"
   xmlns:app="http://schemas.android.com/apk/res-auto"
   xmlns:tools="http://schemas.android.com/tools"
+  android:id="@+id/changeLanguageScreen"
```
ragunathjawahar: Nice! 🚀

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
ragunathjawahar: Copy constructors will cause flaky tests in case the implementation details change later. We should write a lint rule that flags any copy constructor invocations outside the data class.

app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
vinaysshenoy: Done.

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
vinaysshenoy: Done.

PR #885, opened 2019-10-15
Move screen created rendering logic to Mobius + Basic Mobius setup

ragunathjawahar:


app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -456,6 +456,8 @@ class BloodPressureEntrySheetController @Inject constructor(
 
     val currentFacilityStream = loggedInUserStream
         .flatMap(facilityRepository::currentFacility)
+        .replay()
```
vinaysshenoy: Why was this added?

ragunathjawahar: Since the stream is being consumed at two different places, `facilityRepository::currentFacility` was being invoked twice. I added this after the characteristic test showed two invocations.

Since this is a behavior we won't be able to replicate in Mobius, I added `replay()` ran the tests, ran the app and ensured the behavior remained the same.

vinaysshenoy: Will a `share()` work instead of a `replay().refCount()`?

ragunathjawahar: `share()` had problems in one of the older migration. Hence avoiding it.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -72,29 +79,44 @@ class BloodPressureEntrySheet : BottomSheetActivity(), BloodPressureEntryUi {
...
-    setContentView(R.layout.sheet_blood_pressure_entry)
-    TheActivity.component.inject(this)
+  private val viewRenderer = BloodPressureEntryViewRenderer(this)
 
-    bindUiToControllerWithoutDelay(
-        ui = this,
-        events = Observable.mergeArray(
+  private val delegate by unsafeLazy {
+    val defaultModel = when (val openAs = intent.extras!!.getParcelable(KEY_OPEN_AS) as OpenAs) {
+      is New -> BloodPressureEntryModel.newBloodPressureEntry(openAs.patientUuid)
```
vinaysshenoy: I don't understand the point of doing this since `BloodPressureEntryModel` just internally creates a new instance of the same `OpenAs` type anyway? We might as well just pass the `openAs` directly to the constructor of `BloodPressureEntryModel`.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -108,12 +130,47 @@ class BloodPressureEntrySheet : BottomSheetActivity(), BloodPressureEntryUi {
             dayTextChanges(),
             monthTextChanges(),
             yearTextChanges()
-        ),
+        )
+        .compose(ReportAnalyticsEvents())
```
vinaysshenoy: I just realised that doing this is a little dangerous since `BloodPressureEntrySheetController` also forwards events to analytics. This means that until the controller is completely migrated and deleted, the events will be pushed twice to analytics. This is fine if we are sure that we'll finish the migration in time for the next release, but it's still something to watch out for.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -108,12 +130,47 @@ class BloodPressureEntrySheet : BottomSheetActivity(), BloodPressureEntryUi {
...
+  }
+
+  override fun onStop() {
+    delegate.stop()
+    super.onStop()
+  }
+
+  override fun onSaveInstanceState(outState: Bundle?) {
+    delegate.onSaveInstanceState(outState)
+    super.onSaveInstanceState(outState)
```
vinaysshenoy: Should this be the state that is returned from `MobiusDelegate#onSaveInstanceState`? Maybe we should add a `@CheckReturnValue` annotation so that it isn't missed.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryViewRenderer.kt
```diff
@@ -1,11 +1,21 @@
...
 
+import org.simple.clinic.bp.entry.OpenAs.New
 import org.simple.clinic.mobius.ViewRenderer
 
 class BloodPressureEntryViewRenderer(
     val ui: BloodPressureEntryUi
 ) : ViewRenderer<BloodPressureEntryModel> {
   override fun render(model: BloodPressureEntryModel) {
-    // Do nothing!
+    renderOpenAs(model.openAs)
```
vinaysshenoy: Rename method to `setupUi`?

vinaysshenoy: I can't think of a good name, this essentially the intention of `set up the ui based on what it was opened for`.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
ragunathjawahar: Makes sense, will fix this.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
ragunathjawahar: Good catch, I missed removing `ReportAnalyticsEvents()` from the controller. Which I can do now. Therefore we could continue shipping the feature even if we are in the middle of refactoring.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryViewRenderer.kt
ragunathjawahar: Will fix.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
ragunathjawahar: Fixed

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
ragunathjawahar: I just realized that `onSaveInstanceState(Bundle?)` does not return a value while used inside an `Activity`. We make have to introduce new functions to deal with this case and later split them into two different classes. Will take a stab at this.

ragunathjawahar: I investigated this and it is evident that we need a different delegate for the `Activity`

ragunathjawahar: @vinaysshenoy addressed this feedback. Can you plese review #888 first. So that I can re-request review on this PR. Thanks.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -99,8 +99,8 @@ class BloodPressureEntrySheet : BottomSheetActivity(), BloodPressureEntryUi {
 
   private val delegate by unsafeLazy {
     val defaultModel = when (val openAs = intent.extras!!.getParcelable(KEY_OPEN_AS) as OpenAs) {
-      is New -> BloodPressureEntryModel.newBloodPressureEntry(openAs.patientUuid)
-      is Update -> BloodPressureEntryModel.updateBloodPressureEntry(openAs.bpUuid)
+      is New -> BloodPressureEntryModel.newBloodPressureEntry(New(openAs.patientUuid))
```
vinaysshenoy: What I meant was that we could just do

```kotlin
val openAs = intent.extras!!.getParcelable(KEY_OPEN_AS) as OpenAs
val defaultModel = BloodPressureEntryModel(openAs)
```

vinaysshenoy: The work of deciding the right `OpenAs` is already decided by the previous screen. This screen shouldn't have to worry about it.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
vinaysshenoy: Should we look at capturing the recorded analytics events as well in the characteristics tests, so we know that the events we record have not changed in any way?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
ragunathjawahar: Since I am migrating 1 feature at a time, I prefer to break into one execution path at once. We can unify this in the final PR.

vinaysshenoy: Alright.

PR #886, opened 2019-10-15
Add syncStatus to Encounter

sanchita:

### DIFF
--- a/app/src/main/java/org/simple/clinic/encounter/sync/EncounterPullResponse.kt
+++ b/app/src/main/java/org/simple/clinic/encounter/sync/EncounterPullResponse.kt
@@ -1,0 +1,16 @@
+package org.simple.clinic.encounter.sync
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+import org.simple.clinic.sync.DataPullResponse
+
+@JsonClass(generateAdapter = true)
+data class EncounterPullResponse(
+
+    @Json(name = "encounters")
+    override val payloads: List<EncounterPayload>,
+
+    @Json(name = "process_token")
+    override val processToken: String
+
+) : DataPullResponse<EncounterPayload>
--- a/app/src/main/java/org/simple/clinic/storage/Migration_50_51.kt
+++ b/app/src/main/java/org/simple/clinic/storage/Migration_50_51.kt
@@ -1,0 +1,13 @@
+package org.simple.clinic.storage
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+@Suppress("ClassName")
+class Migration_50_51 : Migration(50, 51) {
+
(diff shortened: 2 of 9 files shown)


app/src/main/java/org/simple/clinic/encounter/sync/EncounterPullResponse.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.simple.clinic.sync.DataPullResponse
+
+@JsonClass(generateAdapter = true)
+data class EncounterPullResponse(
+
+    @Json(name = "encounters")
+    override val payloads: List<EncounterPayload>,
+
+    @Json(name = "process_token")
+    override val processToken: String
```
ragunathjawahar: Curious, what is a process token?

app/src/main/java/org/simple/clinic/storage/Migration_50_51.kt
```diff
@@ -0,0 +1,13 @@
+package org.simple.clinic.storage
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+@Suppress("ClassName")
+class Migration_50_51 : Migration(50, 51) {
+
+  override fun migrate(database: SupportSQLiteDatabase) {
+    database.execSQL("""ALTER TABLE "Encounter" ADD COLUMN "syncStatus" TEXT NOT NULL DEFAULT "DONE" """)
```
ragunathjawahar: Curious, why are we setting this column value to `DONE`. How do we know that all the encounters have been synced? Context please.

app/src/main/java/org/simple/clinic/encounter/sync/EncounterPullResponse.kt
sanchita: Process token is an encoded string that has all the information related to batch syncing like last synced timestamp. This token is set by the server and we save it and send it across the next time we pull records from server. The server then send us records that were updated post the last sync.
There is more information in that token which is useful to the server mostly.

app/src/main/java/org/simple/clinic/storage/Migration_50_51.kt
sanchita: We do not have `Encounter` syncing implemented yet. Hence when this migration runs, the app won't have any `Encounter` records. So it's safe to assume this status.

The reason I am not adding `PENDING` as a default status is because that should only be added when you want records to be synced.
The other status is `INVALID` which will not be appropriate either because it is an error status.

app/src/main/java/org/simple/clinic/encounter/sync/EncounterPullResponse.kt
sanchita: FYI, `process_token` is added to all the APIs that involve syncing.

PR #888, opened 2019-10-16
Extract Mobius elements as a submodule from the project

ragunathjawahar:
I did this so that we can start modularising the app. I missed the `CrashReporter` in the Mobius package. Hence the lack of atomic commits. This commit also include a `MobiusActivityDelegate` that can support `Activity` classes.

There is a temporary branch `rj/16oct/counter` which showcases the `MobiusActivityDelegate` in action. This branch will be deleted once this PR gets merged into master.


app/src/main/java/org/simple/clinic/mobius/MobiusDelegate.kt
```diff
@@ -21,7 +22,7 @@ import org.simple.clinic.util.unsafeLazy
 class MobiusDelegate<M : Parcelable, E, F>(
     private val events: Observable<E>,
     private val defaultModel: M,
-    private val init: Init<M, F>,
+    private val init: Init<M, F>?,
```
vinaysshenoy: Instead of making this nullable, what do you think of making it an optional parameter and setting `Init { first(defaultModel) }` as a default value?

ragunathjawahar: I don't see any significant gain in doing that.

app/src/main/java/org/simple/clinic/mobius/migration/Architecture.kt
```diff
@@ -1,5 +1,8 @@
 package org.simple.clinic.mobius.migration
 
+@Deprecated("""
```
vinaysshenoy: 🚀

app/src/main/java/org/simple/clinic/mobius/MobiusActivityDelegate.kt
```diff
@@ -0,0 +1,110 @@
...
+import com.spotify.mobius.functions.Consumer
+import com.spotify.mobius.functions.Function
+import com.spotify.mobius.rx2.RxMobius
+import io.reactivex.Observable
+import io.reactivex.ObservableTransformer
+import io.reactivex.disposables.Disposable
+import org.simple.clinic.crash.CrashReporter
+import org.simple.clinic.util.unsafeLazy
+
+class MobiusActivityDelegate<M : Parcelable, E, F>(
```
vinaysshenoy: I am assuming that this is the same as `MobiusDelegate`, but with only the changes to the `onSaveInstanceState` and `onRestoreInstanceState` functions to accommodate the differences for an `Activity`.

vinaysshenoy: Do we plan to unify the two classes later once we migrate one activity to Mobius?

app/src/main/java/org/simple/clinic/mobius/MobiusActivityDelegate.kt
```diff
@@ -0,0 +1,110 @@
...
+    if (::eventsDisposable.isInitialized && eventsDisposable.isDisposed.not()) {
+      eventsDisposable.dispose()
+    }
+
+    startControllerIfNotAlreadyRunning()
+    stopAndDisconnectController()
+  }
+
+  fun onSaveInstanceState(outState: Parcelable?) {
+    (outState as? Bundle)?.putParcelable(modelKey, controller.model)
```
vinaysshenoy: Since we are using the same bundle passed in, should we add a defensive check and fail early if `modelKey` is already present in the bundle?

app/src/main/java/org/simple/clinic/mobius/MobiusActivityDelegate.kt
ragunathjawahar: There are some changes made to the start function as well and prepare is no longer exposed. This is due to subtle differences between the lifecycle of the `View` and the `Activity`.

app/src/main/java/org/simple/clinic/mobius/MobiusActivityDelegate.kt
ragunathjawahar: It's okay to overwrite the value because this is platform behavior and us choosing to add a defensive check can either prevent an overwrite / choose which version of the model is right. Both aren't much helpful in this case.

app/src/main/java/org/simple/clinic/mobius/MobiusActivityDelegate.kt
vinaysshenoy: Alright, let's discuss unifying them later.

simple-platform/src/main/java/org/simple/clinic/platform/crash/Breadcrumb.kt
```diff
@@ -1,4 +1,4 @@
-package org.simple.clinic.crash
+package org.simple.clinic.platform.crash
```
vinaysshenoy: Nice! I was planning to do this, but looks like you already have.

There is a difference between what you have done, and what I was planning though. My plan was to make this a pure java module. We can do this by removing the `init` function from the interface and moving the initialisation to the constructor of `SentryCrashReporter`.

We can do this in a later PR, though.

PR #889, opened 2019-10-16
Move prefill date for new BP entry logic to Mobius

ragunathjawahar:


ragunathjawahar:
Opening another PR with a few more events migrated to Mobius

PR #890, opened 2019-10-16
Move BP entry events to Mobius

ragunathjawahar:
These events were moved to Mobius
- Prefill date for new entry
- Hide BP errors during systolic changes
- Change focus to diastolic after systolic is entered
- Hide BP errors during diastolic changes
- Handle backspace presses in diastolic


app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
```diff
@@ -1,12 +1,32 @@
...
+      inputDatePaddingCharacter: UserInputDatePaddingCharacter,
+      schedulersProvider: SchedulersProvider
+  ): ObservableTransformer<BloodPressureEntryEffect, BloodPressureEntryEvent> {
+    return RxMobius
+        .subtypeEffectHandler<BloodPressureEntryEffect, BloodPressureEntryEvent>()
+        .addConsumer(PrefillDate::class.java, { prefillDate(ui, it.date, inputDatePaddingCharacter) }, schedulersProvider.ui())
+        .build()
+  }
+
+  fun prefillDate(
```
vinaysshenoy: Ideally, the date to be displayed should be part of the model. I am assuming that we will change this once the feature is migrated?

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandlerTest.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.junit.After
+import org.junit.Test
+import org.simple.clinic.mobius.EffectHandlerTestCase
+import org.simple.clinic.util.UserInputDatePaddingCharacter
+import org.simple.clinic.util.scheduler.TrampolineSchedulersProvider
+import org.threeten.bp.LocalDate
+
+class BloodPressureEntryEffectHandlerTest {
+  private val ui = mock<BloodPressureEntryUi>()
+  private val effectHandler = BloodPressureEntryEffectHandler
```
vinaysshenoy: This could be inlined since it's not used anywhere else.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
```diff
@@ -2,27 +2,30 @@ package org.simple.clinic.bp.entry
...
   }
 
-  fun prefillDate(
+  private fun prefillDateForNewEntry(
       ui: BloodPressureEntryUi,
-      date: LocalDate,
+      userClock: UserClock,
       inputDatePaddingCharacter: UserInputDatePaddingCharacter
   ) {
+    val date = userClock.instant().toLocalDateAtZone(userClock.zone)
```
vinaysshenoy: We generally use `Instant.now(clock)` instead of `clock.instant()`. If we don't have a reason to use the latter, could we stick to the former convention, please?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -162,7 +162,7 @@ class BloodPressureEntrySheetController @Inject constructor(
 
     val dateForNewBp = openAsStream
         .ofType<New>()
-        .map { LocalDate.now(userClock) }
+        .map { LocalDate.now() }
```
vinaysshenoy: Why did we remove the `userClock` parameter when constructing this instance?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -178,9 +178,16 @@ class BloodPressureEntrySheetController @Inject constructor(
...
-    return events
+    val localDates = events
         .ofType<DateToPrefillCalculated>()
         .map { it.date }
+
+    val updates = events
+        .ofType<SheetCreated>()
+        .filter { it.openAs is Update }
+
+    return Observables
```
vinaysshenoy: Should we use `updates.withLatestFrom(localDates)` instead?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
vinaysshenoy: Can't we just remove the `dateForNewBp` stream instead?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -88,16 +88,6 @@ class BloodPressureEntrySheetController @Inject constructor(
   }
 
   private fun automaticFocusChanges(events: Observable<UiEvent>): Observable<UiChange> {
-    val isSystolicValueComplete = { systolicText: String ->
```
vinaysshenoy: 👀

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
```diff
@@ -4,15 +4,26 @@ import com.spotify.mobius.Next
...
   override fun update(
       model: BloodPressureEntryModel,
       event: BloodPressureEntryEvent
   ): Next<BloodPressureEntryModel, BloodPressureEntryEffect> {
     return when (event) {
-      is SystolicChanged -> justEffect(HideBpErrorMessage)
+      is SystolicChanged -> if (isSystolicValueComplete(event.systolic)) {
+        justEffects(HideBpErrorMessage, ChangeFocusToDiastolic)
+      } else {
+        justEffect<BloodPressureEntryModel, BloodPressureEntryEffect>(HideBpErrorMessage)
```
vinaysshenoy: The signature of `justEffects` also supports sending a single effect. Should we use that instead of this single parameter method?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
```diff
@@ -7,3 +7,7 @@ object PrefillDateForNewEntry : BloodPressureEntryEffect()
 object HideBpErrorMessage : BloodPressureEntryEffect()
 
 object ChangeFocusToDiastolic : BloodPressureEntryEffect()
+
+object ChangeFocusToSystolic : BloodPressureEntryEffect()
+
+data class SetSystolic(val systolic: String) : BloodPressureEntryEffect()
```
vinaysshenoy: I am assuming that the `systolic` property will eventually become a part of the model.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
ragunathjawahar: Yes, that is correct. We will do this once we starting moving the update patient entry flow to Mobius.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandlerTest.kt
ragunathjawahar: Tried doing this, but it looks too crowded once inlined. Also, the effect handler will take in more parameters as we move more logic to Mobius.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
ragunathjawahar: Fixed this by putting back the user clock.

 We can't remove the `dateForNewBp` stream now because save entry is dependent on the state of this stream. We can remove this only after we migrate the save entry functionality.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
ragunathjawahar: That was my initial approach, but one of the streams were terminated and `withLatestFrom` was never executed. Hence the `combineLatest` with a `take(1)`. Tried your code again just to be sure but had the same effect.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
ragunathjawahar: I prefer to have two different functions to reveal more information. I am open to having a single function as long as it doesn't sound ambiguous. However I have changed `justEffect` to internally call `justEffects`.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
ragunathjawahar: Done.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
ragunathjawahar: Or we could use a function named `dispatch` that can do 1 or more effects. WDYT?

vinaysshenoy: Looks good to me, let's see in the other PR.

PR #896, opened 2019-10-18
Migrate more functionality to Mobius

ragunathjawahar:
- Prefill date when updating a BP entry
- Move BP validation when date button / save button is clicked
- Remove auto-dismiss BP entry sheet when synced BP was removed
- Go to date entry screen

### DIFF
--- a/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
+++ b/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
@@ -41,3 +41,5 @@ object Dismiss : BloodPressureEntryEffect()
 object HideDateErrorMessage : BloodPressureEntryEffect()
 
 data class ShowBpValidationError(val result: Validation) : BloodPressureEntryEffect()
+
+object ShowDateEntryScreen : BloodPressureEntryEffect()
--- a/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
+++ b/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
@@ -1,5 +1,6 @@
 package org.simple.clinic.bp.entry
 
+import org.simple.clinic.bp.entry.BpValidator.Validation
 import org.threeten.bp.Instant
 import java.util.UUID
 
@@ -38,3 +39,5 @@ data class ShowConfirmRemoveBloodPressureDialog(
 object Dismiss : BloodPressureEntryEffect()
 
 object HideDateErrorMessage : BloodPressureEntryEffect()
+
+data class ShowBpValidationError(val result: Validation) : BloodPressureEntryEffect()
--- a/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
+++ b/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
@@ -1,10 +1,21 @@
 package org.simple.clinic.bp.entry
 
+import org.threeten.bp.Instant
 import java.util.UUID
 
 sealed class BloodPressureEntryEffect
 
-object PrefillDateForNewEntry : BloodPressureEntryEffect()
+data class PrefillDate(val date: Instant? = null) : BloodPressureEntryEffect() {
+  companion object {
+    fun forNewEntry(): PrefillDate {
+      return PrefillDate()
+    }
+
+    fun forUpdateEntry(date: Instant): PrefillDate {
+      return PrefillDate(date)
+    }
+  }
+}
 
 object HideBpErrorMessage : BloodPressureEntryEffect()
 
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
```diff
@@ -1,10 +1,21 @@
 package org.simple.clinic.bp.entry
 
+import org.threeten.bp.Instant
 import java.util.UUID
 
 sealed class BloodPressureEntryEffect
 
-object PrefillDateForNewEntry : BloodPressureEntryEffect()
+data class PrefillDate(val date: Instant? = null) : BloodPressureEntryEffect() {
```
vinaysshenoy: I feel this should be a sealed class, since the date being `null` is just a signal that we should pre-fill the current date instead of a specific date. What do you think?

vinaysshenoy:
```kotlin
sealed class PrefillDate {
  object PrefillCurrentDate : PrefillDate()
  
  data class PrefillSpecificDate(val instant: Instant): PrefillDate()
}: BloodPressureEntryEffect()
```

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
```diff
@@ -58,4 +67,20 @@ object BloodPressureEntryEffectHandler {
...
+    when (validationError.result) {
+      is ErrorSystolicLessThanDiastolic -> ui.showSystolicLessThanDiastolicError()
+      is ErrorSystolicTooHigh -> ui.showSystolicHighError()
+      is ErrorSystolicTooLow -> ui.showSystolicLowError()
+      is ErrorDiastolicTooHigh -> ui.showDiastolicHighError()
+      is ErrorDiastolicTooLow -> ui.showDiastolicLowError()
+      is ErrorSystolicEmpty -> ui.showSystolicEmptyError()
+      is ErrorDiastolicEmpty -> ui.showDiastolicEmptyError()
+      is Success -> { /* Nothing to do here. */ }
+    }
```
vinaysshenoy: Should we add a `.exhaustive()` call here?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
```diff
@@ -63,6 +66,15 @@ class BloodPressureEntryUpdate : Update<BloodPressureEntryModel, BloodPressureEn
 
       is YearChanged -> dispatch(HideDateErrorMessage)
 
+      is BloodPressureDateClicked -> {
+        val result = bpValidator.validate(model.systolic, model.diastolic)
```
vinaysshenoy: Suggestion: Convert the if statement into a `when` block and inline the `result` into the `when` subject.

```kotlin
is BloodPressureDateClicked -> when (val result = bpValidator.validate(model.systolic, model.diastolic)) {
  is Success -> dispatch<BloodPressureEntryModel, BloodPressureEntryEffect>(ShowDateEntryScreen)
  else -> dispatch<BloodPressureEntryModel, BloodPressureEntryEffect>(ShowBpValidationError(result))
}

```

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -124,22 +123,6 @@ class BloodPressureEntrySheetController @Inject constructor(
         }
   }
 
-  private fun proceedToDateEntryWhenBpEntryIsDone(events: Observable<UiEvent>): Observable<UiChange> {
```
vinaysshenoy: 🚀

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
```diff
@@ -75,6 +75,15 @@ class BloodPressureEntryUpdate(
         }
       }
 
+      is SaveClicked -> {
```
vinaysshenoy: Same as earlier comment to collapse `result` into the subject of a `when` block.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
ragunathjawahar: I understand, however when we are migrating. I prefer focusing on moving to the newer architecture before refactoring the design of the migration itself. These are two separate problems. In most cases refactoring the Mobius classes may not yield optimum results unless the migration is complete.

However, I am adding this to my list of TODOs and will revisit while refactoring these classes once the migration is done (last PR for this feature).

ragunathjawahar: Please keep bringing these points so that I can refer to them when I refactor Mobius classes.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
ragunathjawahar: Same as above, when it comes to Mobius classes.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
ragunathjawahar: Same as above, when it comes to Mobius classes.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
ragunathjawahar: Fixed.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
vinaysshenoy: Sure, my intention was to add it as a suggestion for the final clean up PR after migration anyway.

PR #898, opened 2019-10-18
Add Encounter repository

sanchita:

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/TestData.kt
+++ b/app/src/androidTest/java/org/simple/clinic/TestData.kt
@@ -6,6 +6,9 @@ import org.simple.clinic.bp.sync.BloodPressureMeasurementPayload
 import org.simple.clinic.di.AppScope
 import org.simple.clinic.drugs.PrescribedDrug
 import org.simple.clinic.drugs.sync.PrescribedDrugPayload
+import org.simple.clinic.encounter.Encounter
+import org.simple.clinic.encounter.sync.EncounterObservationsPayload
+import org.simple.clinic.encounter.sync.EncounterPayload
 import org.simple.clinic.facility.Facility
 import org.simple.clinic.facility.FacilityPayload
 import org.simple.clinic.facility.FacilityRepository
@@ -775,6 +778,47 @@ class TestData @Inject constructor(
     return Identifier(
         value = value,
         type = type
+    )
+  }
+
+  fun encounter(
+      uuid: UUID = UUID.randomUUID(),
+      patientUuid: UUID = UUID.randomUUID(),
+      facilityUuid: UUID? = null,
+      encounteredOn: LocalDate = LocalDate.now(),
+      syncStatus: SyncStatus = SyncStatus.DONE,
+      createdAt: Instant = Instant.now(),
+      updatedAt: Instant = Instant.now(),
+      deletedAt: Instant? = null
+  ): Encounter {
+    return Encounter(
+        uuid = if (facilityUuid == null) uuid else generateEncounterUuid(facilityUuid, patientUuid, encounteredOn),
+        patientUuid = patientUuid,
+        encounteredOn = encounteredOn,
+        syncStatus = syncStatus,
+        createdAt = createdAt,
+        updatedAt = updatedAt,
+        deletedAt = deletedAt
+    )
+  }
+
+  fun encounterPayload(
+      uuid: UUID = UUID.randomUUID(),
+      patientUuid: UUID = UUID.randomUUID(),
+      encounteredOn: LocalDate = LocalDate.now(),
+      createdAt: Instant = Instant.now(),
+      updatedAt: Instant = Instant.now(),
+      deletedAt: Instant? = null,
+      bpPayloads: List<BloodPressureMeasurementPayload> = listOf(bpPayload(patientUuid = patientUuid))
+  ): EncounterPayload {
+    return EncounterPayload(
+        uuid = uuid,
+        patientUuid = patientUuid,
+        encounteredOn = encounteredOn,
+        createdAt = createdAt,
+        updatedAt = updatedAt,
+        deletedAt = deletedAt,
+        observations = EncounterObservationsPayload(bloodPressureMeasurements = bpPayloads)
     )
   }
 }
--- a/app/src/androidTest/java/org/simple/clinic/TestData.kt
+++ b/app/src/androidTest/java/org/simple/clinic/TestData.kt
@@ -782,8 +782,8 @@ class TestData @Inject constructor(
   }
 
   fun encounter(
-      uuid: UUID = UUID.randomUUID(),
-      patientUuid: UUID = UUID.randomUUID(),
+      uuid: UUID,
+      patientUuid: UUID,
       facilityUuid: UUID? = null,
       encounteredOn: LocalDate = LocalDate.now(),
       syncStatus: SyncStatus = SyncStatus.DONE,
@@ -803,8 +803,8 @@ class TestData @Inject constructor(
   }
 
   fun encounterPayload(
-      uuid: UUID = UUID.randomUUID(),
-      patientUuid: UUID = UUID.randomUUID(),
+      uuid: UUID,
+      patientUuid: UUID,
       encounteredOn: LocalDate = LocalDate.now(),
       createdAt: Instant = Instant.now(),
       updatedAt: Instant = Instant.now(),
--- a/app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
@@ -5,6 +5,7 @@ import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
 import org.junit.rules.RuleChain
+import org.simple.clinic.AppDatabase
 import org.simple.clinic.TestClinicApp
 import org.simple.clinic.TestData
 import org.simple.clinic.patient.SyncStatus
@@ -17,7 +18,7 @@ import javax.inject.Inject
 class EncounterRepositoryAndroidTest {
 
   @Inject
-  lateinit var appDatabase: org.simple.clinic.AppDatabase
+  lateinit var appDatabase: AppDatabase
 
   @Inject
   lateinit var repository: EncounterRepository
--- a/app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
@@ -11,6 +11,7 @@ import org.simple.clinic.TestData
 import org.simple.clinic.patient.SyncStatus
 import org.simple.clinic.rules.LocalAuthenticationRule
 import org.simple.clinic.util.RxErrorsRule
+import org.threeten.bp.Instant
 import java.util.UUID
 import javax.inject.Inject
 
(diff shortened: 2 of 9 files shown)


app/src/main/java/org/simple/clinic/encounter/Encounter.kt
```diff
@@ -33,6 +36,24 @@ data class Encounter(
   interface RoomDao {
 
     @Insert(onConflict = REPLACE)
-    fun save(encounters: List<Encounter>)
+    fun save(encounters: List<Encounter>): Completable
+
+    @Query("UPDATE Encounter SET syncStatus = :newStatus WHERE syncStatus = :oldStatus")
+    fun updateSyncStatus(oldStatus: SyncStatus, newStatus: SyncStatus): Completable
```
ragunathjawahar: Curious, what is the rationale behind changing the `SyncStatus` for all entries. The criteria being a specific `SyncStatus`?

app/src/main/java/org/simple/clinic/encounter/Encounter.kt
```diff
@@ -33,6 +36,24 @@ data class Encounter(
...
+
+    @Query("UPDATE Encounter SET syncStatus = :newStatus WHERE uuid IN (:uuids)")
+    fun updateSyncStatus(uuids: List<UUID>, newStatus: SyncStatus): Completable
+
+    @Query("SELECT * FROM Encounter WHERE syncStatus = :syncStatus")
+    fun recordsWithSyncStatus(syncStatus: SyncStatus): Observable<List<Encounter>>
+
+    @Query("SELECT COUNT(uuid) FROM Encounter")
+    fun recordCount(): Observable<Int>
```
ragunathjawahar: Curious, what are we doing with this information?

app/src/main/java/org/simple/clinic/encounter/Encounter.kt
```diff
@@ -33,6 +36,24 @@ data class Encounter(
...
+
+    @Query("SELECT COUNT(uuid) FROM Encounter WHERE syncStatus = :syncStatus")
+    fun recordCount(syncStatus: SyncStatus): Observable<Int>
+
+    @Query("SELECT * FROM Encounter WHERE uuid = :uuid")
+    fun getOne(uuid: UUID): Encounter?
```
ragunathjawahar: Should this return type be a `Maybe<Encounter>` or `Single<Encounter>` (depending upon what is supported by room)?

app/src/main/java/org/simple/clinic/encounter/Encounter.kt
ragunathjawahar: If this information is used to display static data, should we consider using a `Single` instead?

app/src/main/java/org/simple/clinic/encounter/Encounter.kt
ragunathjawahar: NVM, this is enforced by the repository contract.

app/src/main/java/org/simple/clinic/encounter/EncounterRepository.kt
```diff
@@ -0,0 +1,47 @@
...
+  override fun recordCount(): Observable<Int> {
+    return database.encountersDao().recordCount()
+  }
+
+  override fun pendingSyncRecordCount(): Observable<Int> {
+    return database.encountersDao().recordCount(syncStatus = PENDING)
+  }
+}
+
+
```
ragunathjawahar: Nitpick, extra line 😛 can go in the next PR.

app/src/androidTest/java/org/simple/clinic/TestData.kt
```diff
@@ -777,4 +780,45 @@ class TestData @Inject constructor(
         type = type
     )
   }
+
+  fun encounter(
+      uuid: UUID = UUID.randomUUID(),
+      patientUuid: UUID = UUID.randomUUID(),
```
ragunathjawahar: Can we provide explicit `UUID` instead of generating them randomly. It becomes really hard to debug when the test fail. Considering that this is not an integration test. Can we remove the default arguments for `uuid` and `patientUuid`?

app/src/androidTest/java/org/simple/clinic/TestData.kt
```diff
@@ -777,4 +780,45 @@ class TestData @Inject constructor(
...
+        syncStatus = syncStatus,
+        createdAt = createdAt,
+        updatedAt = updatedAt,
+        deletedAt = deletedAt
+    )
+  }
+
+  fun encounterPayload(
+      uuid: UUID = UUID.randomUUID(),
+      patientUuid: UUID = UUID.randomUUID(),
```
ragunathjawahar: Same as above for `uuid` and `patientUuid`

app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.simple.clinic.rules.LocalAuthenticationRule
+import org.simple.clinic.util.RxErrorsRule
+import java.util.UUID
+import javax.inject.Inject
+
+
+class EncounterRepositoryAndroidTest {
+
+  @Inject
+  lateinit var appDatabase: org.simple.clinic.AppDatabase
```
ragunathjawahar: FQCN `org.simple.clinic.AppDatabase`. Replace with `AppDatabase`?

app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
```diff
@@ -0,0 +1,73 @@
...
+    assertThat(count).isEqualTo(1)
+  }
+
+  @Test
+  fun encounter_payload_should_merge_correctly_with_encounter_database_model() {
+    //given
+    val patientUuid = UUID.fromString("c2273c6c-036b-41f0-a128-84d5551be3ce")
+    val encounterUuid = UUID.fromString("0b144e4a-ce13-431d-93f1-1d10d0639c09")
+
+    val encountersPayload = testData.encounterPayload(uuid = encounterUuid, patientUuid = patientUuid)
```
ragunathjawahar: This is nice 👏, using actual UUIDs.

app/src/main/java/org/simple/clinic/bp/sync/BloodPressureMeasurementPayload.kt
```diff
@@ -59,4 +59,21 @@ data class BloodPressureMeasurementPayload(
         //FIXME: [Encounter] This code is throw-away. Once the Encounter API is implemented, this should be fixed.
         encounterUuid = generateEncounterUuid(facilityUuid, patientUuid, recordedAt.toLocalDateAtZone(ZoneOffset.UTC)))
   }
+
+  fun toDatabaseModel(syncStatus: SyncStatus, encounterUuid: UUID): BloodPressureMeasurement {
```
ragunathjawahar: Make this function slightly more descriptive. So that maintainers don't have to look at the function's return type to figure out what it is doing?

app/src/main/java/org/simple/clinic/encounter/EncounterAndObservations.kt
```diff
@@ -0,0 +1,10 @@
+package org.simple.clinic.encounter
+
+import org.simple.clinic.bp.BloodPressureMeasurement
+
+data class EncounterAndObservations (
```
ragunathjawahar: Suggestion, should we just call this `Observations`?

app/src/main/java/org/simple/clinic/encounter/EncounterRepository.kt
```diff
@@ -32,7 +34,32 @@ class EncounterRepository @Inject constructor(
...
 
   override fun mergeWithLocalData(payloads: List<EncounterPayload>): Completable {
-    TODO("not implemented")
+    return Single.fromCallable {
+      payloads
+          .filter { payload ->
+            database.encountersDao()
+                .getOne(payload.uuid)
+                ?.syncStatus.canBeOverriddenByServerCopy()
+          }
```
ragunathjawahar: I am not really sure about using a `filter` to run async operations. Can we use a `flatMap` to perform these queries and use a filter to just specify if the local copy be overridden by the server copy?

app/src/main/java/org/simple/clinic/encounter/EncounterRepository.kt
```diff
@@ -32,7 +34,32 @@ class EncounterRepository @Inject constructor(
...
+    return EncounterAndObservations(encounter = payload.toDatabaseModel(DONE), bloodPressures = bloodPressures)
+  }
+
+  private fun saveMergedEncounters(records: List<EncounterAndObservations>): Completable {
+    return Completable.fromAction {
+      database
+          .bloodPressureDao()
+          .save(records.flatMap { it.bloodPressures })
+
+      database.encountersDao().save(records.map { it.encounter })
```
ragunathjawahar: Suggestion,
```kotlin
  private fun saveMergedEncounters(records: List<EncounterAndObservations>): Completable {
      val bloodPressures = records.flatMap { it.bloodPressures }
      val encounters = records.map { it.encounter }

      with(database) {
        bloodPressureDao().save(bloodPressures)
        encountersDao().save(encounters)
      }
    }
  }
```
Also, I feel uncomfortable saving the blood pressures and encounters outside a database transaction.

app/src/androidTest/java/org/simple/clinic/TestData.kt
sanchita: These are just default values. The whole idea of creating these utility methods is to provide an easy way to create multiple encounters in tests. The caller of this method can provide explicit UUIDs. This will be used when we actually start using `Encounter` in our app.

app/src/main/java/org/simple/clinic/bp/sync/BloodPressureMeasurementPayload.kt
sanchita: This method is kind of a standard throughout the app for all our `Payload` classes.
So if I change the name or signature here, I will have to change it in other places as well.

Its `Payload.toDatabaseModel` which I feel is explicit enough. Can you suggest what should I change this to?

app/src/main/java/org/simple/clinic/encounter/EncounterAndObservations.kt
sanchita: We cannot call it that since it would be incorrect.
`Encounter` and `Observations` have a 1:many relationship. This class represents a `Encounter` and all of its `Observations`. I can call it `EncounterWithObservations` or `ObservationsForEncounter`. Wdyt?

app/src/androidTest/java/org/simple/clinic/TestData.kt
ragunathjawahar: Yes, and that is causing problems because it is hard to track when tests fail. That is why we need to move away from this pattern. There are some places where we could make exceptions. For instance, writing integration tests. Otherwise, we should stop using this pattern.

app/src/main/java/org/simple/clinic/bp/sync/BloodPressureMeasurementPayload.kt
ragunathjawahar: If this is convention, I am probably not used to this. It is fine then.

app/src/main/java/org/simple/clinic/encounter/EncounterAndObservations.kt
ragunathjawahar: `ObservationsForEncounter` seems more accurate.

app/src/main/java/org/simple/clinic/encounter/Encounter.kt
sanchita: Actually this method is not needed by most `SynceableRepository` including this one.
But in order to satisfy the contract, we need to implement this.
We can actually consider removing this altogether method from `Repository` altogether.
Also, we could change the contract to accept `Single` instead of `Observable`.
I will add this as a separate refactoring.

app/src/main/java/org/simple/clinic/encounter/Encounter.kt
sanchita: The usage of this method does not require this to be reactive.
Also, we only expect one value here since the `WHERE` clause is on `uuid` which is a `PRIMARY KEY`.
I don't think it is necessary to use `Maybe` or `Single` here.

app/src/main/java/org/simple/clinic/encounter/EncounterRepository.kt
sanchita: I have changed to the following:
```
      database.openHelper.writableDatabase.inTransaction {
        with(database) {
          bloodPressureDao().save(records.flatMap { it.bloodPressures })
          encountersDao().save(records.map { it.encounter })
        }
      }

```
PTAL.

app/src/main/java/org/simple/clinic/encounter/EncounterRepository.kt
sanchita: I don't understand the issue in `filter` running in async ops. Can you explain this a little? 
 I will make the change anyway.

app/src/androidTest/java/org/simple/clinic/TestData.kt
sanchita: Ok. I will make this change.

app/src/androidTest/java/org/simple/clinic/TestData.kt
sanchita: Even though I have made this change, but in the next PR I will have to revert this.
We need the uuid params to have default values because this test data is used in instrumentation tests.

app/src/main/java/org/simple/clinic/encounter/Encounter.kt
ragunathjawahar: Works!

app/src/androidTest/java/org/simple/clinic/TestData.kt
ragunathjawahar: Can we not supply `UUID.random()` in the instrumentation tests alone? Otherwise, it is easy for us to still follow this pattern where failing tests make our lives miserable.

1. All UUIDs should be explicit by default.
2. UUIDs used for integration tests should explicitly use `UUID.random()`.

Does this seem reasonable?

app/src/main/java/org/simple/clinic/encounter/EncounterRepository.kt
ragunathjawahar: Will this work?

```kotlin
  override fun mergeWithLocalData(payloads: List<EncounterPayload>): Completable {
    val encounterPayloads = Observable.fromIterable(payloads)
    val encounters = encounterPayloads.flatMap(::getEncounter)

    return Observables.zip(encounterPayloads, encounters)
        .filter { (_, encounter) -> encounter.isNotEmpty() }
        .map { (payload, encounter) -> payload to encounter.toNullable()!! }
        .filter { (_, encounter) -> encounter.syncStatus.canBeOverriddenByServerCopy() }
        .map { (payload, _) -> payloadToEncounters(payload) }
        .toList()
        .flatMapCompletable { saveMergedEncounters(it) }
  }

  private fun getEncounter(payload: EncounterPayload): Observable<Optional<Encounter>> =
      Observable.fromCallable { database.encountersDao().getOne(payload.uuid).toOptional() }
```

app/src/main/java/org/simple/clinic/encounter/EncounterRepository.kt
ragunathjawahar:
```kotlin
with(database) {
  openHelper.writableDatabase.inTransaction {
    bloodPressureDao().save(bloodPressures)
    encountersDao().save(encounters)
  }
}
```

app/src/androidTest/java/org/simple/clinic/TestData.kt
sanchita: This makes sense. Will take care of this in the upcoming PR.

PR #901, opened 2019-10-22
Fix soft keyboard not showing up automatically during registration

vinaysshenoy:
https://www.pivotaltracker.com/story/show/169280383

### DIFF
--- a/app/src/main/java/org/simple/clinic/registration/name/RegistrationFullNameScreen.kt
+++ b/app/src/main/java/org/simple/clinic/registration/name/RegistrationFullNameScreen.kt
@@ -4,16 +4,12 @@ import android.animation.LayoutTransition
 import android.content.Context
 import android.util.AttributeSet
 import android.view.View
-import android.view.ViewGroup
 import android.view.inputmethod.EditorInfo
-import android.widget.EditText
-import android.widget.ImageButton
 import android.widget.RelativeLayout
-import android.widget.TextView
 import com.jakewharton.rxbinding2.view.RxView
 import com.jakewharton.rxbinding2.widget.RxTextView
 import io.reactivex.Observable
-import kotterknife.bindView
+import kotlinx.android.synthetic.main.screen_registration_name.view.*
 import org.simple.clinic.R
 import org.simple.clinic.main.TheActivity
 import org.simple.clinic.bindUiToController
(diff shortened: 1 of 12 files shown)


sanchita:
Do we know why is this issue happening?

vinaysshenoy:
> Do we know the root cause of this issue?
> `requestFocus` should automatically pop the keyboard.
> Why is this happening now?

I do not know the root cause of this issue. The reason I think it is happening now is because the way screens are getting managed has changed. Earlier, we would remove the older view from the screen **after** the incoming view gets added *and* measured, i.e, we remove it later. We changed it to both add the new view and remove the older view synchronously.

Since we already asked for the keyboard to be shown in the first screen in the registration flow, I think some combination of removing the older view synchronously *and* requesting focus manually ends up putting it in this state where it cannot be brought back until you background the app and resume it again. Removing the manual `<requestFocus />` was the only thing I could find that actually fixed the problem.

app/src/main/java/org/simple/clinic/registration/name/RegistrationFullNameScreen.kt
```diff
@@ -4,16 +4,12 @@ import android.animation.LayoutTransition
 import android.content.Context
```
sanchita: Nice of you to do this :)

app/src/main/res/layout/screen_registration_name.xml
```diff
@@ -50,10 +50,7 @@
         android:gravity="center"
         android:imeOptions="actionDone"
         android:inputType="textPersonName|textCapWords"
-        android:textAppearance="@style/Clinic.LoginInputTextAppearance">
-
-        <requestFocus />
```
sanchita: I have also seen this happen on `EnterOtpScreen` for entering OTP.
Although the issue gets resolved if you close and open the screen again.
We should check for the usages of `requestFocus` throughout the app and fix this issue.

vinaysshenoy: Good point, I'll go through the codebase and see.

I think this issue only happens in cases where the previous screen in the flow already had a keyboard open, though. If a screen is being opened without the keyboard being open, there isn't any problem.

vinaysshenoy: I checked the codebase, and we have four more usages.

1. Enter OTP screen. I ran it and verified, the problem does not happen.
2. Blood Pressure Entry sheet. I ran and verified, the problem does not happen.
3. Appointment Reminder Sheet. For some reason, this is on a `TextView`. No idea why it's there, but there is no editable field on that sheet.
4. Schedule Appointment Sheet. Same as above

I can't see any other places where this needs to be fixed at the moment.

sanchita: I can see this problem on `Enter OTP Screen` in QA atleast.
I wonder why?

sanchita: I think we can make this change on `Enter OTP screen` only.
Rest of the screens look fine.

vinaysshenoy: Done.

PR #903, opened 2019-10-22
Provide alternate injection method for screens

vinaysshenoy:
https://www.pivotaltracker.com/story/show/169253555

The current injection method for screens does not allow us to use alternate injectors for a given Screen and couples it tightly to a specific activity. This PR is meant to give us a way to abstract the injector from the screen so we can move a Screen to a different activity in stages.


app/src/main/java/org/simple/clinic/activity/TheActivityComponent.kt
```diff
@@ -73,15 +73,14 @@ import org.threeten.bp.Instant
...
-interface TheActivityComponent {
+interface TheActivityComponent: OnboardingScreenInjector {
 
   fun inject(target: TheActivity)
   fun inject(target: HomeScreen)
   fun inject(target: PatientsScreen)
   fun inject(target: LoginPinScreen)
   fun inject(target: AppLockScreen)
   fun inject(target: OverdueScreen)
-  fun inject(target: OnboardingScreen)
```
ragunathjawahar: Nice!

PR #904, opened 2019-10-23
Move save and update BP entry logic to Mobius

ragunathjawahar:
@vinaysshenoy apologies for this monstrous PR. The create / update functionality and `BloodPressureSaved` analytics event were slightly tricky to migrate. Please let me know if you'd like to pair on this.


app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
```diff
@@ -101,4 +118,14 @@ class BloodPressureEntryUpdate(
     return (systolicText.length == 3 && systolicText.matches("^[123].*$".toRegex()))
         || (systolicText.length == 2 && systolicText.matches("^[789].*$".toRegex()))
   }
+
+  private fun formatToPaddedDate(day: String, month: String, twoDigitYear: String, fourDigitYear: String): String {
```
vinaysshenoy: Ideally, this should be done in the UI and passed to the mobius loop as a date changed event with the padded string as the property, right?

ragunathjawahar: Yes, that is correct. I have noted this down and will work on this as a separate PR.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -60,7 +59,6 @@ class BloodPressureEntrySheetController @Inject constructor(
         .replay()
 
     return Observable.mergeArray(
-        showBpEntry(replayedEvents),
```
vinaysshenoy: It's kind of funny that the commit message says "Move showing of data validation errors when back is clicked...", and the method that gets removed is `showBpEntry`. 😅

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
```diff
@@ -63,6 +64,14 @@ class BloodPressureEntryUpdate(
 
       is BackPressed -> if (model.activeScreen == BP_ENTRY) {
         dispatch(Dismiss as BloodPressureEntryEffect)
+      } else if (model.activeScreen == DATE_ENTRY) {
+        val dateText = formatToPaddedDate(model.day, model.month, model.twoDigitYear, model.year)
```
vinaysshenoy: Just a suggestion: Since the `BloodPressureEntryUpdate` already has access to the current calendar date via `dateInUserTimeZone`, we can remove the `year` property from the model and read it from `dateInUserTimeZone`.

I checked the property and it is only used in formatting the padded date.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTest.kt
```diff
@@ -307,9 +307,7 @@ class BloodPressureEntrySheetControllerTest {
     verify(ui).showConfirmRemoveBloodPressureDialog(bloodPressure.uuid)
   }
 
-  // TODO Migrate logic to Mobius
   @Test
-  @Suppress("IMPLICIT_CAST_TO_ANY")
```
vinaysshenoy: I'm curious, why is this no longer required? Was the problem fixed with either a Kotlin version update, or the Android SDK update?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
```diff
@@ -110,4 +147,91 @@ object BloodPressureEntryEffectHandler {
...
+      facilityRepository: FacilityRepository,
+      patientRepository: PatientRepository,
+      bloodPressureRepository: BloodPressureRepository,
+      appointmentsRepository: AppointmentRepository,
+      userClock: UserClock,
+      scheduler: Scheduler
+  ): ObservableTransformer<CreateNewBpEntry, BloodPressureEntryEvent> {
+    return ObservableTransformer { createNewBpEntries ->
+      createNewBpEntries
+          .subscribeOn(scheduler)
```
vinaysshenoy: Subscribing to this on a different scheduler won't have any effect because we `flatMap` on a `Room` `Observable` query which switches the thread internally to Room's own scheduler. This is also seen in the `updateBpEntryTransformer` call since it doesn't accept a scheduler.

This is hidden behaviour, but since we are not unlikely to swap Room out for anything else, do you think we should remove the scheduler parameter from this method and just depend on Room to manage the threading for us at the moment?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
```diff
@@ -146,139 +129,7 @@ class BloodPressureEntrySheetController @Inject constructor(
     events.mergeWith(validations)
   }
 
-  private fun saveBpWhenDateEntryIsDone() = ObservableTransformer<UiEvent, UiEvent> { events ->
```
vinaysshenoy: 🚀

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTest.kt
```diff
@@ -404,14 +395,6 @@ class BloodPressureEntrySheetControllerTest {
       onNext(SaveClicked)
     }
 
-    val newUserUuid = UUID.fromString("a726d885-b12d-40b7-9fe1-5391f3fc0f88")
```
vinaysshenoy: This was intentionally done to verify that if a different user than the one who records a BP edits it, the BP measurement is updated with the new user and facility ID.

Is this being tested elsewhere?

vinaysshenoy: We might want to add a different test for it so that this behaviour is tested explicitly rather than implicitly.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTest.kt
```diff
@@ -884,14 +865,6 @@ class BloodPressureEntrySheetControllerTest {
     whenever(bloodPressureRepository.updateMeasurement(any())).doReturn(Completable.complete())
     whenever(patientRepository.compareAndUpdateRecordedAt(any(), any())).doReturn(Completable.complete())
 
-    val newUserUuid = UUID.fromString("a726d885-b12d-40b7-9fe1-5391f3fc0f88")
```
vinaysshenoy: Same as earlier comment.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetController.kt
ragunathjawahar: 😂 This is because validating the BP happens in multiple scenarios. One of them was when we click the show BP entry button. This was the last code path that I removed which was triggering the validation logic.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
ragunathjawahar: I want to rethink how validation is done. I am not comfortable passing in the validator to the Update function itself (which means we should guarantee that the validators are pure and it is easy for other developers to pick up this pattern without knowing the intention). Now that we have a different view of the solution. I think, this can be a part of the `formatToPaddedDate` refactoring update.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTest.kt
ragunathjawahar: It was removed because we replaced the `List<Any>` with a `List<TestParams>`.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTest.kt
ragunathjawahar: I had an understanding of it when I removed the test. The test was incorrect because, it was not simulating the correct scenario. What we wanted was - what happens a different user updates the BP. What the test was doing - the user of the device has changed, now let's see if the BP entry has the updated user ID and the facility ID. Will add a test to cover this case.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTest.kt
ragunathjawahar: As explained above.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
ragunathjawahar: I could remove the `.subscribeOn` call.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
```diff
@@ -8,16 +8,20 @@ import java.util.UUID
 
 sealed class BloodPressureEntryEffect
 
-data class PrefillDate(val date: Instant? = null) : BloodPressureEntryEffect() {
+sealed class PrefillDate : BloodPressureEntryEffect() {
```
vinaysshenoy: 👌

app/src/main/java/org/simple/clinic/bp/entry/ConfirmRemoveBloodPressureDialog.kt
```diff
@@ -22,6 +23,9 @@ import java.util.UUID
 import javax.inject.Inject
 
 class ConfirmRemoveBloodPressureDialog : AppCompatDialogFragment() {
+  interface RemoveBloodPressureListener {
```
vinaysshenoy: Our convention is generally to keep declared interfaces in the bottom of the class. Could we follow this convention please?

app/src/main/java/org/simple/clinic/bp/entry/ConfirmRemoveBloodPressureDialog.kt
```diff
@@ -54,6 +58,8 @@ class ConfirmRemoveBloodPressureDialog : AppCompatDialogFragment() {
   @Inject
   lateinit var controller: ConfirmRemoveBloodPressureDialogController
 
+  private var removeBloodPressureListener: RemoveBloodPressureListener? = null
```
vinaysshenoy: Can this ever be `null`? We throw an error when the activity has not implemented the interface, so this could be a `lateinit var`, right?

app/src/main/java/org/simple/clinic/bp/entry/ConfirmRemoveBloodPressureDialog.kt
```diff
@@ -107,4 +115,12 @@ class ConfirmRemoveBloodPressureDialog : AppCompatDialogFragment() {
     super.onDestroyView()
     screenDestroys.onNext(ScreenDestroyed())
   }
+
+  override fun onAttach(context: Context?) {
+    super.onAttach(context)
+    removeBloodPressureListener = context as? RemoveBloodPressureListener
```
vinaysshenoy: I am not comfortable doing this because this ties the dialog to be used in an `Activity` instead of a `Screen`. If we ever need to open this from a screen, then the activity the screen lives in will have to implement this even though the dialog itself is opened in the screen and then it will have to pass the event to the screen is some fashion, which will be ugly.

We do have a `ScreenRouter` which exposes a bus for passing events between screens in this fashion. The dialog can push an event to the bus, and regardless of whether it is being invoked in an activity or a screen, the bus can be subscribed to for this event. 

What do you think about using this method instead? We can do this in another PR.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -58,21 +53,6 @@ class BloodPressureEntrySheet : BottomSheetActivity(), BloodPressureEntryUi, Rem
...
-  private val backImageButton by bindView<ImageButton>(R.id.bloodpressureentry_back_button)
-  private val enterBloodPressureTitleTextView by bindView<TextView>(R.id.bloodpressureentry_enter_blood_pressure)
-  private val editBloodPressureTitleTextView by bindView<TextView>(R.id.bloodpressureentry_edit_blood_pressure)
-  private val removeBloodPressureButton by bindView<Button>(R.id.bloodpressureentry_remove)
-  private val dayEditText by bindView<EditText>(R.id.bloodpressureentry_day)
-  private val monthEditText by bindView<EditText>(R.id.bloodpressureentry_month)
-  private val yearEditText by bindView<EditText>(R.id.bloodpressureentry_year)
-  private val viewFlipper by bindView<ViewFlipperWithLayoutEditorPreview>(R.id.bloodpressureentry_view_flipper)
-  private val dateErrorTextView by bindView<TextView>(R.id.bloodpressureentry_date_error)
-
```
vinaysshenoy: ✂️

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
```diff
@@ -8,16 +8,20 @@ import java.util.UUID
...
 
     fun forUpdateEntry(date: Instant): PrefillDate {
-      return PrefillDate(date)
+      return UpdateEntryPrefill(date)
     }
   }
+
+  object NewEntryPrefill : PrefillDate()
+
+  data class UpdateEntryPrefill(val date: Instant) : PrefillDate()
```
vinaysshenoy: Suggestions for alternate names:

`NewEntryPrefill` -> `PrefillCurrentDate`
`UpdateEntryPrefill` -> `PrefillSpecificDate`

Thoughts?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
```diff
@@ -60,14 +60,22 @@ data class CreateNewBpEntry(
     val patientUuid: UUID,
     val systolic: Int,
     val diastolic: Int,
-    val date: LocalDate
-) : BloodPressureEntryEffect()
+    val parsedDateFromForm: LocalDate,
+    val prefilledDate: LocalDate
+) : BloodPressureEntryEffect() {
+  val wasDateChanged: Boolean
+    get() = parsedDateFromForm != prefilledDate
```
vinaysshenoy: `parsedDateFromForm` was a little hard to read since both `from` and `form` are anagrams and right next to each other. Can we rename it to something else like `userEnteredDate` or something?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffect.kt
```diff
@@ -60,14 +60,22 @@ data class CreateNewBpEntry(
...
     val bpUuid: UUID,
     val systolic: Int,
     val diastolic: Int,
-    val date: LocalDate
-) : BloodPressureEntryEffect()
+    val parsedDateFromForm: LocalDate,
+    val prefilledDate: LocalDate
+) : BloodPressureEntryEffect() {
+  val wasDateChanged: Boolean
+    get() = parsedDateFromForm != prefilledDate
```
vinaysshenoy: Same as above.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -201,17 +197,6 @@ class BloodPressureEntrySheet : BottomSheetActivity(), BloodPressureEntryUi, Rem
     }
   }
 
-  private fun sheetCreates(): Observable<UiEvent> {
```
vinaysshenoy: ✂️

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
```diff
@@ -37,110 +38,84 @@ import org.simple.clinic.widgets.ageanddateofbirth.UserInputDateValidator.Result
...
-
-    val updateBpEntryTransformer = updateBpEntryTransformer(
-        userSession,
-        facilityRepository,
-        patientRepository,
-        bloodPressureRepository,
-        userClock
-    )
-
+  fun create(): ObservableTransformer<BloodPressureEntryEffect, BloodPressureEntryEvent> {
```
vinaysshenoy: Discussed offline: Make this a factory function instead of an instance method.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
```diff
@@ -1,7 +1,7 @@
 package org.simple.clinic.bp.entry
```
vinaysshenoy: This commit is hard to read because mixes many things together:

- Moving the dependencies to the constructor
- Extracting methods
- Renaming variables

I think it would have been much easier to understand if each of these had been three separate commits.

app/src/main/java/org/simple/clinic/bp/entry/ConfirmRemoveBloodPressureDialog.kt
ragunathjawahar: This can become `null` during a configuration change. In such cases, we have to replace the null check with a `::property.isInitialized` check instead.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
ragunathjawahar: Noted, will be more conscious about these in the next PR.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTest.kt
ragunathjawahar: Added a test for this.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntrySheetControllerTest.kt
ragunathjawahar: And added a test for this as well.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
ragunathjawahar: This is done.

app/src/main/java/org/simple/clinic/bp/entry/ConfirmRemoveBloodPressureDialog.kt
ragunathjawahar: There are few things to note here.

1. This is a very specific dialog, used in a very specific feature. Which is a confirm blood pressure dialog, used in a BP entry screen.
2. The underlying UI itself lies inside an `Activity`. Since all our bottom sheets are designed as Activities, we won't be moving away from these unless we are moving away from Flow.
3. The `ScreenRouter` roughly compensates for a coordinator, which we will bring in at some point in time. However, when we start facilitating communication between components that has a hierarchy, our strategy would be to facilitate downstream events via a stream and upstream communication via a callback. Hence, using the `ScreenResultBus` may not be the best solution at the moment. Also, we are losing type-safety by passing across objects of type `Any`.

app/src/main/java/org/simple/clinic/bp/entry/ConfirmRemoveBloodPressureDialog.kt
ragunathjawahar: Done.

app/src/main/java/org/simple/clinic/bp/entry/ConfirmRemoveBloodPressureDialog.kt
vinaysshenoy: Got it, let's revisit this later.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
```diff
@@ -53,7 +53,32 @@ class BloodPressureEntryEffectHandler(
 ) {
   private val reportAnalyticsEvents = ReportAnalyticsEvents()
 
-  fun create(): ObservableTransformer<BloodPressureEntryEffect, BloodPressureEntryEvent> {
+  companion object {
```
vinaysshenoy: We should switch to this pattern for all effect handlers at some point for consistency.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
```diff
@@ -889,6 +890,133 @@ class BloodPressureEntrySheetLogicTest {
...
+  fun `when a different user clicks on save while updating a BP, then the updated BP measurement should be saved with the user's ID and the corresponding facility ID`() {
+    // given
+    reset(userSession, facilityRepository)
+    val userFromDifferentFacility = PatientMocker.loggedInUser(uuid = UUID.fromString("4844b826-a162-49fe-b92c-962da172e86c"))
+    val differentFacility = PatientMocker.facility(uuid = UUID.fromString("f895b54f-ee32-4471-bc0c-a91b80368778"))
+
+    whenever(userSession.requireLoggedInUser()).doReturn(Observable.just(userFromDifferentFacility))
+    whenever(facilityRepository.currentFacility(userFromDifferentFacility)).doReturn(Observable.just(differentFacility))
+
+    val oldCreatedAt = LocalDate.of(1990, 1, 13).toUtcInstant(testUserClock)
```
vinaysshenoy: We have a live template, `instant`, in the current tooling plugin which will generate something like `Instant.parse("2018-01-01T00:00:00Z")`.

We can use this instead of going this round about way.

vinaysshenoy: I can make this change and push if you like.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
```diff
@@ -889,6 +890,133 @@ class BloodPressureEntrySheetLogicTest {
...
+        userUuid = user.uuid,
+        facilityUuid = facility.uuid,
+        systolic = 9000,
+        diastolic = 8999,
+        createdAt = oldCreatedAt,
+        updatedAt = oldCreatedAt,
+        recordedAt = oldCreatedAt
+    )
+
+    val newInputDate = LocalDate.of(1991, 2, 14)
```
vinaysshenoy: We also have another live template, `date`, which will generate `LocalDate.parse("2018-01-01")` which we can then edit for our purposes.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
```diff
@@ -889,6 +890,133 @@ class BloodPressureEntrySheetLogicTest {
     verifyNoMoreInteractions(ui)
   }
 
+  @Test
+  fun `when a different user clicks on save while updating a BP, then the updated BP measurement should be saved with the user's ID and the corresponding facility ID`() {
```
vinaysshenoy: 🚀

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
vinaysshenoy: The property is not used until later (line 937). We should move it closer to where it's required.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
vinaysshenoy: Did you do this intentionally to make the behaviour closer to production (converting an entered `LocalDate` to an `Instant`)?

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
```diff
@@ -889,6 +890,133 @@ class BloodPressureEntrySheetLogicTest {
...
+    verify(bloodPressureRepository).updateMeasurement(updatedBp)
+    verify(patientRepository).compareAndUpdateRecordedAt(updatedBp.patientUuid, updatedBp.recordedAt)
+
+    verify(bloodPressureRepository, never()).saveMeasurement(any(), any(), any(), any(), any(), any())
+    verify(appointmentRepository, never()).markAppointmentsCreatedBeforeTodayAsVisited(any())
+    verify(ui).setBpSavedResultAndFinish()
+  }
+
+  @Test
+  fun `when done button is clicked by a user from a different facility in update BP entry, then save BP with entered date immediately`() {
```
vinaysshenoy: Isn't this testing the same thing as the earlier test? I can't tell the difference.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
ragunathjawahar: Will do!

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
ragunathjawahar: One is clicking the **Save** button on the BP entry sheet. The other one is using the IME's *Done* button.

vinaysshenoy: But these events both get mapped to `SaveClicked` at the UI level, right?

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
ragunathjawahar: I picked up examples from the earlier tests to do this. Hence, this pattern.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
vinaysshenoy: I also looked at the screen. There is no separate button on that sheet. The only way to save the BP is by the IME done button.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
vinaysshenoy: Yeah, the older tests were written before the tooling plugin. Not a problem though, we can always start using the live templates for newer tests.

ragunathjawahar: Fixed.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
ragunathjawahar: Fixed.

ragunathjawahar: Fixed.

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
ragunathjawahar: I guess, then the test checks saving the BP from two different screens.

vinaysshenoy: Hmm, that's not the case as well. It's the `BP_ENTRY` screen on both. 😅

vinaysshenoy: Never mind, I just relooked at the tests and saw that the earlier test also changes the screen. Makes sense now.

PR #906, opened 2019-10-23
Do not show onboarding screen if the user has already completed onboarding in Setup Activity

vinaysshenoy:
https://www.pivotaltracker.com/n/projects/2358313/stories/169253555

The final PR in this set of PRs will change `SetupActivity` to be the launcher activity.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -40,6 +40,17 @@
       android:value="true" />
 
     <activity
+      android:name=".setup.SetupActivity"
+      android:enabled="false"
+      android:launchMode="singleTask"
+      android:screenOrientation="portrait">
+      <intent-filter>
+        <action android:name="android.intent.action.MAIN" />
+        <category android:name="android.intent.category.LAUNCHER" />
+      </intent-filter>
+    </activity>
+
+    <activity
       android:name=".main.TheActivity"
       android:launchMode="singleTask"
       android:screenOrientation="portrait"
@@ -83,10 +94,6 @@
       android:name=".home.overdue.phonemask.PhoneMaskBottomSheet"
       android:theme="@style/Clinic.V2.Theme.BottomSheetActivity"
       android:windowSoftInputMode="adjustResize" />
-
-    <activity
-      android:name=".setup.SetupActivity"
-      android:enabled="false"/>
 
     <receiver
       android:name=".login.OtpSmsReceiver"
--- a/app/src/main/java/org/simple/clinic/activity/PlaceholderScreenKey.kt
+++ b/app/src/main/java/org/simple/clinic/activity/PlaceholderScreenKey.kt
@@ -1,0 +1,29 @@
+package org.simple.clinic.activity
+
+import android.app.Activity
+import kotlinx.android.parcel.IgnoredOnParcel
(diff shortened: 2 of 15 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -39,6 +39,17 @@
       android:name="com.mixpanel.android.MPConfig.DisableViewCrawler"
       android:value="true" />
 
+    <activity
+      android:name=".setup.SetupActivity"
+      android:enabled="false"
```
ragunathjawahar: I assume that since the `SetupActivity` is disabled, we will still continue launching `TheActivity`. Is this a safe assumption to make?

app/src/main/java/org/simple/clinic/activity/PlaceholderScreenKey.kt
```diff
@@ -0,0 +1,29 @@
...
+ * The only reason this exists is for cases where the
+ * screen that must be displayed when creating an [Activity] is because
+ *
+ * - [ScreenRouter] requires us to pass in a key when constructing it
+ * - The actual key that we want to load is conditional and might require some database I/O to figure out
+ *
+ * // TODO(vs): 2019-10-23 Either remove the requirement from ScreenRouter or remove Flow
+ **/
+@Parcelize
+class PlaceholderScreenKey : FullScreenKey {
```
ragunathjawahar: Off topic: Is this where I am supposed to add the loading screen?

app/src/test/java/org/simple/clinic/setup/SetupActivityInitTest.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.spotify.mobius.test.FirstMatchers.hasEffects
+import com.spotify.mobius.test.FirstMatchers.hasModel
+import com.spotify.mobius.test.InitSpec
+import com.spotify.mobius.test.InitSpec.assertThatFirst
+import org.junit.Test
+
+class SetupActivityInitTest {
+
+  @Test
+  fun `when the screen is created, the user details must be loaded`() {
```
ragunathjawahar: What if the details are already loaded, in case of a restoration?

ragunathjawahar: It doesn't really matter in this case, because it is okay to fetch user details again.

app/src/main/AndroidManifest.xml
vinaysshenoy: Yup, the final PR will change `SetupActivity` to be the only launcher.

app/src/main/java/org/simple/clinic/activity/PlaceholderScreenKey.kt
vinaysshenoy: Yes.

app/src/test/java/org/simple/clinic/setup/SetupActivityInitTest.kt
vinaysshenoy: Currently, this only fetches whether the user has done the onboarding or not. Later on, this will get more information like whether a user exists, or whether they have fetched a country. I will update this then.

PR #907, opened 2019-10-23
Add Encounter API and sync

sanchita:


app/src/main/java/org/simple/clinic/encounter/EncounterModule.kt
```diff
@@ -0,0 +1,28 @@
...
+
+  @Provides
+  fun syncApi(retrofit: Retrofit): EncounterSyncApi {
+    return retrofit.create(EncounterSyncApi::class.java)
+  }
+
+  @Provides
+  @Named("last_encounter_pull_token")
+  fun lastPullToken(rxSharedPreferences: RxSharedPreferences): Preference<Optional<String>> {
+    return rxSharedPreferences.getObject("last_encounter_pull_token_v3", None, OptionalRxPreferencesConverter(StringPreferenceConverter()))
```
vinaysshenoy: Why is there a `_v3` suffix for the pull token key? We needed this earlier because the client would need to change the token when needing to sync from the beginning, but that is unnecessary now with the server controlled resync feature.

app/src/androidTest/java/org/simple/clinic/encounter/sync/EncounterSyncAndroidTest.kt
```diff
@@ -0,0 +1,78 @@
...
+  lateinit var repository: EncounterRepository
+
+  @Inject
+  lateinit var testData: TestData
+
+  @Inject
+  lateinit var encounterSyncApi: EncounterSyncApi
+
+  @Inject
+  @field:Named("last_encounter_pull_token")
```
vinaysshenoy: We should combine the annotations.

```kotlin
@field:[Inject Named("last_encounter_pull_token")]
lateinit var lastPullToken: Preference<Optional<String>>
```

app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
```diff
@@ -63,22 +62,38 @@ class EncounterRepositoryAndroidTest {
...
-    assertThat(encounter.first().uuid).isEqualTo(encounterUuid2)
-
-    with(encounterPending.first()) {
+    assertThat(observationsForEncounters.size).isEqualTo(1)
+    with(observationsForEncounters.first()) {
+      assertThat(encounter.uuid).isEqualTo(encounterUuid2)
+      assertThat(bloodPressures.size).isEqualTo(1)
+      assertThat(bloodPressures.first().uuid).isEqualTo(bpUuid2)
+    }
+    with(pendingObservationsForEncounters.first().encounter) {
```
vinaysshenoy: Should we also assert that the size of this collection is also 1?

app/src/main/java/org/simple/clinic/encounter/EncounterRepository.kt
```diff
@@ -17,13 +17,13 @@ import javax.inject.Inject
 
 class EncounterRepository @Inject constructor(
     private val database: AppDatabase
-) : SynceableRepository<Encounter, EncounterPayload> {
+) : SynceableRepository<ObservationsForEncounter, EncounterPayload> {
 
-  override fun save(records: List<Encounter>): Completable {
-    return Completable.fromAction { database.encountersDao().save(encounters = records) }
+  override fun save(records: List<ObservationsForEncounter>): Completable {
+    return saveMergedEncounters(records)
```
vinaysshenoy: I looked at the `saveMergedEncounters` method implementation, and I saw the following code block:

```kotlin
with(database) {
  openHelper.writableDatabase.inTransaction {
    bloodPressureDao().save(bloodPressures)
    encountersDao().save(encounters)
  }
}
```

The `SupportSQLiteDatabase#inTransaction` extension method was written to be used in migrations and the migration tests, since we do not have access to the `AppDatabase` in those classes.

If you have access to the `AppDatabase`, Room already provides a `runInTransaction` method which is more convenient instead of accessing the internal open helper.

```kotlin
with(database) {
  runInTransaction {
    bloodPressureDao().save(bloodPressures)
    encountersDao().save(encounters)
  }
}
```

Would you like to switch to using this method instead?

vinaysshenoy: If we do, we should probably also add a comment on the `SupportSQLiteDatabase#inTransaction` method that clarifies where it is meant to be used and pointing users to the Room method as an alternative.

app/src/main/java/org/simple/clinic/encounter/ObservationsForEncounter.kt
```diff
@@ -1,10 +1,13 @@
...
 import org.simple.clinic.bp.BloodPressureMeasurement
 
-data class ObservationsForEncounter (
-
+data class ObservationsForEncounter(
+    @Embedded
     val encounter: Encounter,
 
+    @Relation(parentColumn = "uuid", entityColumn = "encounterUuid")
     val bloodPressures: List<BloodPressureMeasurement>
```
vinaysshenoy: Nice! 🚀

app/src/main/java/org/simple/clinic/encounter/sync/EncounterSync.kt
```diff
@@ -41,9 +41,21 @@ class EncounterSync @Inject constructor(
       pushNetworkCall = { api.push(toRequest(it)) }
   )
 
-  private fun toRequest(encounters: List<ObservationsForEncounter>): EncounterPushRequest {
-    //TODO : Replace with an actual request
-    return EncounterPushRequest(listOf())
+  private fun toRequest(observationsForEncounter: List<ObservationsForEncounter>): EncounterPushRequest {
+    val encounterPayloads = observationsForEncounter.map { (encounter, bps) ->
+      val observationsPayload = EncounterObservationsPayload(bloodPressureMeasurements = bps.map { it.toPayload() })
```
vinaysshenoy: Correct me if I am wrong, but doesn't this replace the separate `BloodPressureSync`? At what point do we stop syncing blood pressure measurements separately?

We can discuss this offline.

app/src/main/java/org/simple/clinic/encounter/EncounterModule.kt
sanchita: I kept this for consistency. I can remove it if you think it is not needed.

app/src/main/java/org/simple/clinic/encounter/sync/EncounterSync.kt
sanchita: We are going to do this in the next PR.
Since this PR still doesn't create `Encounter` for new BPs, we have not removed `BloodPressureSync`. For the same reason we have not added `EncounterSync` to `DataSync` yet for frequently syncing.

app/src/main/java/org/simple/clinic/encounter/EncounterModule.kt
vinaysshenoy: I don't think it's necessary anymore, we should remove it.

app/src/androidTest/java/org/simple/clinic/encounter/sync/EncounterSyncAndroidTest.kt
sanchita: Both these annotations are same.
Hence not making this change.

app/src/main/java/org/simple/clinic/storage/Databases.kt
```diff
@@ -2,6 +2,10 @@ package org.simple.clinic.storage
 
 import androidx.sqlite.db.SupportSQLiteDatabase
 
+/**
+ * This extension function is to be used in Room Migrations only.
```
vinaysshenoy: 👌

PR #908, opened 2019-10-23
Hide the Edit button in patient summary until the profile is populated

ragunathjawahar:
This fixes these crashes (i.e. Kotlin NPE when **Edit** button is clicked before the profile is populated).

https://sentry.io/organizations/resolve-to-save-lives/issues/searches/2487760/?project=1212614&statsPeriod=14d

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
@@ -127,7 +127,14 @@ class PatientSummaryScreenController @Inject constructor(
 
     return Observables
         .combineLatest(sharedPatients, addresses, phoneNumbers, bpPassport, ::PatientSummaryProfile)
-        .map { { ui: Ui -> ui.populatePatientProfile(it) } }
+        .map { patientSummaryProfile -> { ui: Ui -> showPatientSummaryProfile(ui, patientSummaryProfile) } }
+  }
+
+  private fun showPatientSummaryProfile(ui: Ui, patientSummaryProfile: PatientSummaryProfile) {
+    with(ui) {
+      populatePatientProfile(patientSummaryProfile)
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -127,7 +127,7 @@ class PatientSummaryScreenController @Inject constructor(
 
     return Observables
         .combineLatest(sharedPatients, addresses, phoneNumbers, bpPassport, ::PatientSummaryProfile)
-        .map { { ui: Ui -> ui.populatePatientProfile(it) } }
+        .map { { ui: Ui -> with(ui) { populatePatientProfile(it); showEditButton() } } }
```
vinaysshenoy: Why are these two statements on the same line?

app/src/main/res/layout/screen_patient_summary.xml
```diff
@@ -35,6 +35,7 @@
       android:layout_marginEnd="@dimen/spacing_8"
       android:text="@string/patientsummary_edit"
       android:textAppearance="@style/Clinic.V2.TextAppearance.Button2.White100"
+      android:visibility="invisible"
```
vinaysshenoy: We should override the visibility in the editor so that it's visible in the preview?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
vinaysshenoy: Can we make this a multi-line lambda, please?

app/src/main/res/layout/screen_patient_summary.xml
ragunathjawahar: Nice catch.

PR #911, opened 2019-10-24
A utility class to publish various events at specific points in time

ragunathjawahar:
This class enables us to show different messages / UI on the screen when the user is waiting for an asynchronous operation to complete.


app/src/debug/java/org/simple/clinic/playground/AwaitActivity.kt
```diff
@@ -0,0 +1,44 @@
...
+import io.reactivex.android.schedulers.AndroidSchedulers
+import io.reactivex.disposables.Disposable
+import kotlinx.android.synthetic.debug.await_activity.*
+import org.simple.clinic.R
+import org.simple.clinic.await.Await
+import org.simple.clinic.await.Checkpoint
+import org.simple.clinic.util.unsafeLazy
+
+class AwaitActivity : AppCompatActivity() {
+  private lateinit var awaitDisposable: Disposable
```
rakshakhegde: Instead of just `lateinit var Disposable` let's use `val CompositeDisposable`? That way we do not have to check for `isInitialized` in `onStop`.

app/src/debug/java/org/simple/clinic/playground/AwaitActivity.kt
```diff
@@ -0,0 +1,44 @@
...
+  override fun onStart() {
+    super.onStart()
+    awaitDisposable = await.items()
+        .observeOn(AndroidSchedulers.mainThread())
+        .subscribe { message -> messageTextView.text = message }
+  }
+
+  override fun onStop() {
+    if (::awaitDisposable.isInitialized && awaitDisposable.isDisposed.not()) {
+      awaitDisposable.dispose()
```
rakshakhegde: We can simply call `dispose()` once we make the CompositeDisposable change. We need not check `isDisposed.not()`. It is internally checked inside dispose function of CompositeDisposable.

ragunathjawahar: I've changed this to a nullable field. Also, this is not a `CompositeDisposable`

app/src/debug/java/org/simple/clinic/playground/AwaitActivity.kt
ragunathjawahar: Done

app/src/debug/java/org/simple/clinic/playground/AwaitActivity.kt
ragunathjawahar: Done

PR #916, opened 2019-10-29
Initial setup for the Select Country screen

vinaysshenoy:

### DIFF
--- a/app/src/main/java/org/simple/clinic/selectcountry/Country.kt
+++ b/app/src/main/java/org/simple/clinic/selectcountry/Country.kt
@@ -1,0 +1,10 @@
+package org.simple.clinic.selectcountry
+
+import java.net.URI
+
+data class Country(
+    val code: String,
+    val endpoint: URI,
+    val displayName: String,
+    val isdCode: String
+)
--- a/app/src/main/java/org/simple/clinic/selectcountry/SelectCountryModel.kt
+++ b/app/src/main/java/org/simple/clinic/selectcountry/SelectCountryModel.kt
@@ -1,3 +1,16 @@
 package org.simple.clinic.selectcountry
 
-object SelectCountryModel
+data class SelectCountryModel(
+    val supportedCountries: List<Country>?
+) {
+
+  fun hasFetchedCountries() = supportedCountries != null
+
+  fun withSupportedCountries(countries: List<Country>): SelectCountryModel {
+    return copy(supportedCountries = countries)
+  }
+
+  companion object {
+    val FETCHING = SelectCountryModel(null)
+  }
+}
--- a/app/src/main/java/org/simple/clinic/selectcountry/SelectCountryModel.kt
+++ b/app/src/main/java/org/simple/clinic/selectcountry/SelectCountryModel.kt
@@ -1,0 +1,3 @@
+package org.simple.clinic.selectcountry
+
+object SelectCountryModel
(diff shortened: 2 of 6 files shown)


app/src/main/java/org/simple/clinic/selectcountry/Country.kt
```diff
@@ -0,0 +1,10 @@
+package org.simple.clinic.selectcountry
+
+import java.net.URL
```
ragunathjawahar: Not sure with the `java.net.URL`. Was this intentional?

ragunathjawahar: https://stackoverflow.com/a/3771123/421372

app/src/main/java/org/simple/clinic/selectcountry/SelectCountryModel.kt
```diff
@@ -1,3 +1,16 @@
 package org.simple.clinic.selectcountry
 
-object SelectCountryModel
+data class SelectCountryModel(
+    val supportedCountries: List<Country>?
```
ragunathjawahar: Use `List<Country>` with an `emptyList()` instead of a nullable property?

app/src/main/java/org/simple/clinic/selectcountry/Country.kt
vinaysshenoy: Using a type instead of a `String` was intentional. This is to expose the intention behind the property instead of using a primitive.

I didn't know about the problem with the `java.net.URL` type, will change to `java.net.URI` instead!

app/src/main/java/org/simple/clinic/selectcountry/SelectCountryModel.kt
vinaysshenoy: This was intentional. A nullable list indicates that the manifest has not yet been fetched. An empty list indicates that the manifest has been fetched, but there were no countries in the manifest.

Does this make sense?

vinaysshenoy: This might also change when we implement the `Update` since we will need to display network errors, so we could change it then.

app/src/test/java/org/simple/clinic/selectcountry/SelectCountryInitTest.kt
```diff
@@ -0,0 +1,44 @@
+package org.simple.clinic.selectcountry
+
+import com.spotify.mobius.test.FirstMatchers.hasEffects
+import com.spotify.mobius.test.FirstMatchers.hasModel
+import com.spotify.mobius.test.FirstMatchers.hasNoEffects
+import com.spotify.mobius.test.InitSpec
+import com.spotify.mobius.test.InitSpec.assertThatFirst
+import org.junit.Test
+import java.net.URI
+import java.net.URL
```
ragunathjawahar: Unused import?

vinaysshenoy: I'll fix it right not

vinaysshenoy: Done

app/src/main/java/org/simple/clinic/selectcountry/Country.kt
ragunathjawahar: Can we use our own type instead? Using a `URI` or `URL` means, it is possible to initiate a side-effect from `Init` or `Update`.

vinaysshenoy: This makes sense. @rakshakhegde and I will make this change in the next PR?

ragunathjawahar: Works

PR #917, opened 2019-10-29
Update encounter when BP is edited

sanchita:
Pivotol -> https://www.pivotaltracker.com/story/show/168947291

This PR completes the story above and implements creating or updating `Encounter` when BPs are saved or updated. This PR is the final PR to complete the feature to add support for `Encounter` model in the app, with the exception of the UI.


app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
```diff
@@ -62,51 +58,17 @@ class BloodPressureRepository @Inject constructor(
...
-  override fun save(records: List<BloodPressureMeasurement>): Completable {
-    return Completable.fromAction { dao.save(records) }
-  }
-
   fun updateMeasurement(measurement: BloodPressureMeasurement): Completable {
     return encounterRepository.updateBloodPressure(measurement)
   }
 
-  override fun recordsWithSyncStatus(syncStatus: SyncStatus): Single<List<BloodPressureMeasurement>> {
+  //TODO: Remove this method once its usage is replaced in ReportPendingRecordsToAnalytics
```
sanchita: This TODO is added because we are not sure what is the expected count to be sent for analytics once Encounter model UI is added. Whether we should send BP count or Encounter count, this is unclear.
We should handle this when we build the UI.

app/src/main/java/org/simple/clinic/encounter/EncounterRepository.kt
```diff
@@ -19,7 +19,7 @@ class EncounterRepository @Inject constructor(
 ) : SynceableRepository<ObservationsForEncounter, EncounterPayload> {
 
   override fun save(records: List<ObservationsForEncounter>): Completable {
-    return saveMergedEncounters(records)
+    return saveObservationsForEncounters(records)
```
ragunathjawahar: 👍

app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
```diff
@@ -99,4 +102,106 @@ class EncounterRepositoryAndroidTest {
...
+    //given
+    val patientUuid = UUID.fromString("6f725ffd-7008-4018-9e7c-33346964a0c1")
+    val facilityUuid = UUID.fromString("22bdeedb-061e-4a17-8739-e946a4206593")
+    val bpUuid = UUID.fromString("2edd4c06-e2de-4a18-a8e7-43e2e30c9aba")
+    val encounteredDate = LocalDate.parse("2018-01-01")
+    val bloodPressureMeasurement = testData.bloodPressureMeasurement(
+        uuid = bpUuid,
+        patientUuid = patientUuid,
+        facilityUuid = facilityUuid,
+        recordedAt = Instant.parse("2018-01-01T00:00:00Z"),
```
ragunathjawahar: Curious, will the `encounteredDate` be the same as `recordedAt`?

app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
```diff
@@ -99,4 +102,106 @@ class EncounterRepositoryAndroidTest {
...
+    //when
+    repository.saveBloodPressureMeasurement(bloodPressureMeasurement, encounteredDate).blockingAwait()
+
+    //then
+    val bps = appDatabase.bloodPressureDao().bloodPressure(bpUuid).blockingFirst()
+    val encounters = appDatabase.encountersDao().recordsWithSyncStatus(SyncStatus.PENDING).blockingFirst()
+
+    assertThat(encounters).isNotEmpty()
+
+    with(encounters.first()) {
```
ragunathjawahar: This enclosing `with` block makes it harder to comprehend what is going on. Can we split this into two separate assertion blocks?

app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
```diff
@@ -99,4 +102,106 @@ class EncounterRepositoryAndroidTest {
...
+
+      with(encounter) {
+        assertThat(encounteredOn).isEqualTo(encounteredDate)
+        assertThat(createdAt).isEqualTo(bps.createdAt)
+        assertThat(updatedAt).isEqualTo(bps.updatedAt)
+        assertThat(deletedAt).isEqualTo(bps.deletedAt)
+        assertThat(syncStatus).isEqualTo(bps.syncStatus)
+      }
```
ragunathjawahar: This can be one split.

app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
```diff
@@ -99,4 +102,106 @@ class EncounterRepositoryAndroidTest {
...
+
+      assertThat(bps).isEqualTo(bloodPressures.first())
+
+      with(bps) {
+        assertThat(this.patientUuid).isEqualTo(patientUuid)
+        assertThat(encounterUuid).isEqualTo(encounter.uuid)
+      }
```
ragunathjawahar: These lines can be another split.

ragunathjawahar: These lines can be another split.

app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
ragunathjawahar: This can be one split.

app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
```diff
@@ -184,24 +194,65 @@ class EncounterRepositoryAndroidTest {
...
+    val updatedBp = appDatabase.bloodPressureDao().bloodPressure(bpUuid).blockingFirst()
+    val encounters = appDatabase.encountersDao().recordsWithSyncStatus(PENDING).blockingFirst()
+
+    assertThat(encounters.size).isEqualTo(1)
+    with(encounters.first()) {
+      assertThat(bloodPressures.first()).isEqualTo(updatedMeasurement)
+
+      with(encounter) {
+        assertThat(encounteredOn).isEqualTo(encounteredDate)
+        assertThat(createdAt).isEqualTo(updatedBp.createdAt)
```
ragunathjawahar: Curious, why is the encounter's `createdAt` and `deletedAt` properties are being updated?

1. Shouldn't we not update the `createdAt` property?
2. If only all the BP entries are deleted, then we update the `deletedAt` property for an `Encounter`?

Am I missing something here?

app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
```diff
@@ -67,12 +67,15 @@ class BloodPressureRepository @Inject constructor(
...
-    return Completable.fromAction {
-      val updatedMeasurement = measurement.copy(
-          updatedAt = Instant.now(utcClock),
-          syncStatus = SyncStatus.PENDING
-      )
+    val updatedMeasurement = measurement.copy(
+        updatedAt = Instant.now(utcClock),
+        syncStatus = SyncStatus.PENDING
+    )
+    //TODO [Encounter] Uncomment to enable
```
ragunathjawahar: Please explain, Uncomment to enable what?

app/src/main/java/org/simple/clinic/encounter/EncounterRepository.kt
```diff
@@ -112,16 +113,39 @@ class EncounterRepository @Inject constructor(
     )))
   }
 
-  fun updateBloodPressure(updatedMeasurement: BloodPressureMeasurement): Completable {
-    val encounterUuid = updatedMeasurement.encounterUuid
-    val encounteredOn = updatedMeasurement.recordedAt.toLocalDateAtZone(userClock.zone)
+  fun updateBloodPressure(measurement: BloodPressureMeasurement): Completable {
```
ragunathjawahar: Can we run all these operations within a transaction?

app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
```diff
@@ -314,4 +314,35 @@ class EncounterRepositoryAndroidTest {
       assertThat(encounter.deletedAt).isNull()
     }
   }
+
+  @Test
+  fun deleting_an_encounter_should_check_for_any_associated_active_observations(){
```
ragunathjawahar: Curious, what do we do after checking for active observations?

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -76,6 +76,7 @@ class UserSession @Inject constructor(
     @Named("last_prescription_pull_token") private val prescriptionSyncPullToken: Preference<Optional<String>>,
     @Named("last_appointment_pull_token") private val appointmentSyncPullToken: Preference<Optional<String>>,
     @Named("last_medicalhistory_pull_token") private val medicalHistorySyncPullToken: Preference<Optional<String>>,
+    @Named("last_encounter_pull_token") private val encounterSyncPullToken: Preference<Optional<String>>,
```
ragunathjawahar: Curious, why do we have different pull tokens for different sync mechanisms?

app/src/androidTest/java/org/simple/clinic/bp/sync/BloodPressureSyncAndroidTest.kt
```diff
@@ -1,79 +0,0 @@
-package org.simple.clinic.bp.sync
```
ragunathjawahar: Nice!

app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
sanchita: Yes. `encounteredDate` is derived from `recordedAt`.
There is no way to set the `encounterDate` by the user. Also the `encounterUuid` itself is generated using the `encounterDate` itself. This is not enforced through the APIs itself but it is handled in the `EncounterRepository` itself. Refer `saveBloodPressureMeasurement` method in `EncounterRepository`.

app/src/main/java/org/simple/clinic/bp/BloodPressureRepository.kt
sanchita: This is throw-away code. It is removed in the future commits. 
This was added to not break existing functionality while developing.
Do you want me to rebase edit this commit still?

app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
sanchita: Good catch!
Your understanding is correct. I added these assertions to ensure that only `updatedAt` is changed.
I will change these assertions.

app/src/androidTest/java/org/simple/clinic/encounter/EncounterRepositoryAndroidTest.kt
sanchita: We basically want to verify that when deleting an encounter we should check that it only gets if there are no observations associated.
I will reword this test case.

app/src/main/java/org/simple/clinic/user/UserSession.kt
sanchita: We have one token for every sync API. We cannot guarantee that all APIs will get synced in one go.
Hence we maintain different tokens to indicate what was last synced timestamp for each one so that the server can send the response accordingly.

app/src/main/java/org/simple/clinic/encounter/EncounterRepository.kt
ragunathjawahar: Done.

PR #919, opened 2019-10-30
Select country logic

rakshakhegde:

### DIFF
--- a/app/src/main/java/org/simple/clinic/selectcountry/SelectCountryModel.kt
+++ b/app/src/main/java/org/simple/clinic/selectcountry/SelectCountryModel.kt
@@ -11,6 +11,6 @@ data class SelectCountryModel(
   }
 
   companion object {
-    val FETCHING = SelectCountryModel(null)
+    val FETCHING = SelectCountryModel(supportedCountries = null)
   }
 }
--- a/app/src/main/java/org/simple/clinic/selectcountry/SelectCountryModel.kt
+++ b/app/src/main/java/org/simple/clinic/selectcountry/SelectCountryModel.kt
@@ -7,7 +7,7 @@ data class SelectCountryModel(
 
   fun hasFetchedCountries(): Boolean = countries != null
 
-  fun withCountries(countries: List<Country>): SelectCountryModel {
+  fun manifestFetched(countries: List<Country>): SelectCountryModel {
     return copy(countries = countries)
   }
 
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/simple/clinic/selectcountry/SelectCountryModel.kt
```diff
@@ -1,16 +1,16 @@
 package org.simple.clinic.selectcountry
 
 data class SelectCountryModel(
-    val supportedCountries: List<Country>?
+    val countries: List<Country>?
 ) {
 
-  fun hasFetchedCountries() = supportedCountries != null
+  fun hasFetchedCountries() = countries != null
```
ragunathjawahar: Please specify return type explicitly.

app/src/test/java/org/simple/clinic/selectcountry/SelectCountryUpdateTest.kt
```diff
@@ -13,23 +13,25 @@ class SelectCountryUpdateTest {
...
 
+  private val bangladesh = Country(
+      code = "BD",
+      endpoint = URI("https://bd.simple.org"),
+      displayName = "Bangladesh",
+      isdCode = "880"
+  )
+
+  private val countries = listOf(
+      Country(
```
ragunathjawahar: Can we extract this as a country as well just like we did with Bangaladesh?

app/src/test/java/org/simple/clinic/selectcountry/SelectCountryUpdateTest.kt
```diff
@@ -97,4 +97,18 @@ class SelectCountryUpdateTest {
...
+  @Test
+  fun `when retry is clicked, then fetch manifest`() {
+    val model = defaultModel
+        .manifestFetchError(NetworkError)
+
+    spec
+        .given(model)
+        .whenEvent(RetryClicked)
+        .then(assertThatNext(
+            hasModel(SelectCountryModel.FETCHING),
```
ragunathjawahar: This has to be a transition. (i.e.,) `SelectCountryModel.fetching()`. If it isn't, then we'll lose any information that was gathered between transitions.

vinaysshenoy: Right, I'll make this change.

app/src/test/java/org/simple/clinic/selectcountry/SelectCountryUpdateTest.kt
vinaysshenoy: Done.

app/src/test/java/org/simple/clinic/selectcountry/SelectCountryUpdateTest.kt
vinaysshenoy: Done.

PR #920, opened 2019-10-31
Add more comprehensive types to the http error resolver

vinaysshenoy:

This is needed to show appropriate errors when the app manifest fails to be fetched.

### DIFF
--- a/app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
+++ b/app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
@@ -3,7 +3,9 @@ package org.simple.clinic.util
 import io.reactivex.exceptions.CompositeException
 import okhttp3.internal.http2.ConnectionShutdownException
 import okhttp3.internal.http2.StreamResetException
-import org.simple.clinic.util.ResolvedError.*
+import org.simple.clinic.util.ResolvedError.NetworkRelated
+import org.simple.clinic.util.ResolvedError.Unauthorized
+import org.simple.clinic.util.ResolvedError.Unexpected
 import retrofit2.HttpException
 import java.io.IOException
 import java.net.ConnectException
@@ -54,4 +56,6 @@ sealed class ResolvedError(val actualCause: Throwable) {
   class Unexpected(actualCause: Throwable) : ResolvedError(actualCause)
 
   class Unauthorized(actualCause: Throwable) : ResolvedError(actualCause)
+
+  class ServerError(actualCause: Throwable) : ResolvedError(actualCause)
 }
--- a/app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
+++ b/app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
@@ -5,7 +5,7 @@ import okhttp3.internal.http2.ConnectionShutdownException
 import okhttp3.internal.http2.StreamResetException
 import org.simple.clinic.util.ResolvedError.NetworkRelated
 import org.simple.clinic.util.ResolvedError.ServerError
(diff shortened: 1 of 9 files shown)


app/src/test/java/org/simple/clinic/util/ErrorResolverTest.kt
```diff
@@ -131,6 +131,22 @@ class ErrorResolverTest {
     )
   }
 
+  @Test
+  @Parameters(value = ["403", "404", "500", "502"])
```
ragunathjawahar: Should `403` be considered as `Unexpected`?

app/src/test/java/org/simple/clinic/util/ErrorResolverTest.kt
```diff
@@ -103,32 +105,18 @@ class ErrorResolverTest {
...
-      httpException: HttpException,
-      expectedResolvedError: ResolvedError
-  ) {
-    val resolvedError = ErrorResolver.resolve(httpException)
-    assertThat(resolvedError::class).isSameAs(expectedResolvedError::class)
-    assertThat(resolvedError.actualCause).isSameAs(httpException)
-  }
+  fun `http unauthorized errors must be identified correctly`() {
+    // given
+    val exception = httpException(401)
```
ragunathjawahar: `401` - Unauthenticated?
`403` - Unauthorized?

app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
```diff
@@ -47,6 +48,14 @@ object ErrorResolver {
     // like UndeliverableException, etc. in the future.
     return actualError
   }
+
+  private fun mapHttpExceptionToResolvedError(actualCause: HttpException): ResolvedError {
+    return when (actualCause.code()) {
+      401 -> Unauthorized(actualCause)
```
ragunathjawahar: `Unauthenticated`?

app/src/test/java/org/simple/clinic/util/ErrorResolverTest.kt
vinaysshenoy: `403` is `HTTP Forbidden`. We currently do not explicitly handle `Forbidden` in the app, and it will be treated as a generic error message. This is why I left it as an `Unexpected` for now since we add a new type when needed.

Does this make sense?

app/src/test/java/org/simple/clinic/util/ErrorResolverTest.kt
vinaysshenoy: That is incorrect. If you look at https://httpstatuses.com/, the official terms are `Unauthorized` and `Forbidden`. I decided to follow that naming scheme.

app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
vinaysshenoy: I followed standard conventions based on https://httpstatuses.com/.

app/src/test/java/org/simple/clinic/util/ErrorResolverTest.kt
vinaysshenoy: Discussed offline: Will rename `Unauthorized` -> `Unauthenticated`.

app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
vinaysshenoy: Discussed offline: Will rename `Unauthorized` -> `Unauthenticated`.

PR #925, opened 2019-11-04
Add instrumentation for tracking the size of saved states in TheActivity

vinaysshenoy:
See [THIS](https://sentry.io/share/issue/8dc985c628804f458ca452372cae8e97/) crash report for an example of what the breadcrumbs would look like.

### DIFF
--- a/router/src/main/java/flow/StateSizeLogger.kt
+++ b/router/src/main/java/flow/StateSizeLogger.kt
@@ -1,0 +1,23 @@
+package flow
+
+import java.lang.reflect.Modifier
+
+/**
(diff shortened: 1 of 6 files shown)


router/src/main/java/flow/StateSizeLogger.kt
```diff
@@ -0,0 +1,23 @@
+package flow
```
ragunathjawahar: This doesn't need to be in the same package. As long as you are accessing the private property after setting the `isAccessible` property on the `Field`. It doesn't really matter anyways.

ragunathjawahar: Makes sense.

PR #929, opened 2019-11-05
Implement the repository for app configuration

vinaysshenoy:

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/appconfig/SelectedCountryPersistenceAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/appconfig/SelectedCountryPersistenceAndroidTest.kt
@@ -1,0 +1,57 @@
+package org.simple.clinic.appconfig
+
+import android.content.SharedPreferences
+import androidx.test.runner.AndroidJUnit4
+import com.f2prateek.rx.preferences2.Preference
+import com.google.common.truth.Truth.assertThat
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.simple.clinic.TestClinicApp
+import org.simple.clinic.util.Just
+import org.simple.clinic.util.Optional
+import java.net.URI
+import javax.inject.Inject
+
+@RunWith(AndroidJUnit4::class)
+class SelectedCountryPersistenceAndroidTest {
+
+  @Inject
+  lateinit var preferences: SharedPreferences
+
+  @Inject
+  lateinit var selectedCountryPreference: Preference<Optional<Country>>
+
+  @Before
+  fun setUp() {
+    TestClinicApp.appComponent().inject(this)
+  }
+
+  @Test
+  fun persisted_selected_country_should_be_read() {
+    // If this test fails, it means that you have changed the structure of the Country class
+    // without migrating the shared preference and will break the app when an update happens.
+    val savedJson = """
+      {
+        "country_code": "IN",
+        "endpoint": "https://in.simple.org/",
+        "display_name": "India",
+        "isd_code": "91"
+      }
+    """
+
+    preferences
+        .edit()
+        .putString("preference_selected_country_v1", savedJson)
+        .commit()
+
+    val expectedSavedCountry = Country(
+        code = "IN",
+        endpoint = URI.create("https://in.simple.org/"),
+        displayName = "India",
+        isdCode = "91"
+    )
+    assertThat(selectedCountryPreference.get())
+        .isEqualTo(Just(expectedSavedCountry))
+  }
+}
--- a/app/src/androidTest/java/org/simple/clinic/appconfig/SelectedCountryPersistenceAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/appconfig/SelectedCountryPersistenceAndroidTest.kt
@@ -46,7 +46,7 @@ class SelectedCountryPersistenceAndroidTest {
         .commit()
 
     val expectedSavedCountry = Country(
-        code = "IN",
+        isoCountryCode = "IN",
         endpoint = URI.create("https://in.simple.org/"),
         displayName = "India",
         isdCode = "91"
--- a/app/src/main/java/org/simple/clinic/appconfig/Country.kt
+++ b/app/src/main/java/org/simple/clinic/appconfig/Country.kt
@@ -8,7 +8,7 @@ import java.net.URI
 data class Country(
 
     @Json(name = "country_code")
-    val code: String,
+    val isoCountryCode: String,
 
     @Json(name = "endpoint")
     val endpoint: URI,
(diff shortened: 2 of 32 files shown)


app/src/test/java/org/simple/clinic/appconfig/AppConfigRepositoryTest.kt
```diff
@@ -30,4 +36,83 @@ class AppConfigRepositoryTest {
         .assertValue(FetchSucceeded(countries))
         .assertNoErrors()
   }
+
+  @Test
+  fun `failed network calls to fetch the app manifest to return an appropriate failure result`() {
```
sanchita: Rename to `failed network calls to fetch the app manifest should return network related failure result`

app/src/test/java/org/simple/clinic/appconfig/AppConfigRepositoryTest.kt
```diff
@@ -30,4 +36,83 @@ class AppConfigRepositoryTest {
...
+    // given
+    val cause = ConnectException()
+    whenever(api.fetchManifest()).thenReturn(Single.error(cause))
```
sanchita: Are we using `thenReturn` or moving to `doReturn`?

app/src/test/java/org/simple/clinic/appconfig/AppConfigRepositoryTest.kt
```diff
@@ -30,4 +36,83 @@ class AppConfigRepositoryTest {
...
+    // then
+    repository
+        .fetchAppManifest()
+        .test()
+        .assertValue(FetchError(ResolvedError.NetworkRelated(cause)))
+        .assertNoErrors()
+  }
+
+  @Test
+  fun `server errors when fetching the app manifest should return an appropriate failure result`() {
```
sanchita: Rename to `server errors when fetching the app manifest should return failure result as server error`

app/src/test/java/org/simple/clinic/appconfig/AppConfigRepositoryTest.kt
```diff
@@ -30,4 +36,83 @@ class AppConfigRepositoryTest {
...
+    // then
+    repository
+        .fetchAppManifest()
+        .test()
+        .assertValue(FetchError(ResolvedError.ServerError(cause)))
+        .assertNoErrors()
+  }
+
+  @Test
+  fun `unauthenticated errors when fetching the app manifest should return an appropriate failure result`() {
```
sanchita: Rename to `401 errors when fetching the app manifest should return failure result as unauthenticated`

app/src/main/java/org/simple/clinic/util/preference/MoshiPreferenceConverter.kt
```diff
@@ -0,0 +1,20 @@
+package org.simple.clinic.util.preference
+
+import com.f2prateek.rx.preferences2.Preference
+import com.squareup.moshi.Moshi
+
+class MoshiPreferenceConverter<T>(
```
sanchita: Do you think we need to be more specific and call this `MoshiObjectPreferenceConverter`?
I don't feel strongly about this. This is up to you :)

app/src/androidTest/java/org/simple/clinic/appconfig/SelectedCountryPersistenceAndroidTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+  @Before
+  fun setUp() {
+    TestClinicApp.appComponent().inject(this)
+  }
+
+  @Test
+  fun persisted_selected_country_should_be_read() {
+    // If this test fails, it means that you have changed the structure of the Country class
+    // without migrating the shared preference and will break the app when an update happens.
```
sanchita: Awesome!

app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
```diff
@@ -60,11 +60,11 @@ object ErrorResolver {
 
 sealed class ResolvedError(val actualCause: Throwable) {
 
-  class NetworkRelated(actualCause: Throwable) : ResolvedError(actualCause)
+  data class NetworkRelated(private val _actualCause: Throwable) : ResolvedError(_actualCause)
```
sanchita: Just for my information, what is the reason to add `_` to `actualCause`?
Is this a kind of best practice?

app/src/main/java/org/simple/clinic/appconfig/Country.kt
```diff
@@ -1,10 +1,21 @@
...
 
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
 import java.net.URI
 
+@JsonClass(generateAdapter = true)
 data class Country(
+
+    @Json(name = "country_code")
     val code: String,
```
sanchita: We should rename `code` to `countryCode` or `isoCode`.  We may end up adding more codes in the future so it might be ideal to specify what code is this one since all the codes are of type `String`.

app/src/test/java/org/simple/clinic/appconfig/AppConfigRepositoryTest.kt
vinaysshenoy: I managed to do something to remove the IDE errors. I'll use this for a couple more days and see. If the errors have gone, I'll revert to `thenReturn` everywhere else.

app/src/main/java/org/simple/clinic/util/preference/MoshiPreferenceConverter.kt
vinaysshenoy: Yeah, makes sense. I'll change this.

app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
vinaysshenoy: The only reason is because `ResolvedError` also exposes a property and I cannot make these types data class unless I convert them to properties. This `_` was added to avoid the name clash.

The "right" solution, IMO, is maybe to not expose the property from `ResolvedError` and let callers get the cause from the individual error types but that is too invasive a change to do at the moment.

I could also rename them `causeInternal` to fit with our usual convention.

app/src/main/java/org/simple/clinic/appconfig/Country.kt
vinaysshenoy: Makes sense. My long term plan is to move these types to [Inline classes](https://kotlinlang.org/docs/reference/inline-classes.html) when they are stable, so I guess naming them differently would help.

app/src/main/java/org/simple/clinic/util/ErrorResolver.kt
vinaysshenoy: Discussed offline: We will retain the names as they are for now.

PR #931, opened 2019-11-05
Remove Heap analytics

vinaysshenoy:


app/src/main/java/org/simple/clinic/analytics/HeapAnalyticsReporter.kt
```diff
@@ -1,26 +0,0 @@
-package org.simple.clinic.analytics
```
ragunathjawahar: Burn after reading 🔥!

gradle.properties
```diff
@@ -29,12 +29,6 @@ stagingApiEndpoint=do_not_change_here
 sandboxApiEndpoint=do_not_change_here
 productionApiEndpoint=do_not_change_here
 
-# Heap project IDs are declared here so they can be configured on CI
-qaHeapId=do_not_change_here
```
ragunathjawahar: Have we removed these values from the CI server?

vinaysshenoy: We cannot remove them now because the next prod release still depends on Heap. I have a note to remove it once the next prod build is released.

app/src/main/java/org/simple/clinic/analytics/HeapAnalyticsReporter.kt
vinaysshenoy: 🔥

PR #934, opened 2019-11-06
Change the default appointment based on protocol

rakshakhegde:
https://www.pivotaltracker.com/story/show/168980979

### DIFF
--- a/app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
+++ b/app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
@@ -16,11 +16,14 @@ import org.simple.clinic.overdue.Appointment.AppointmentType.Manual
 import org.simple.clinic.overdue.AppointmentConfig
 import org.simple.clinic.overdue.AppointmentRepository
 import org.simple.clinic.patient.PatientRepository
+import org.simple.clinic.protocol.ProtocolRepository
 import org.simple.clinic.scheduleappointment.TimeToAppointment.Days
 import org.simple.clinic.scheduleappointment.TimeToAppointment.Months
 import org.simple.clinic.scheduleappointment.TimeToAppointment.Weeks
 import org.simple.clinic.user.UserSession
 import org.simple.clinic.util.UserClock
+import org.simple.clinic.util.filterAndUnwrapJust
+import org.simple.clinic.util.toOptional
 import org.simple.clinic.widgets.UiEvent
 import org.threeten.bp.LocalDate
 import org.threeten.bp.temporal.ChronoUnit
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -119,9 +123,17 @@ class ScheduleAppointmentSheetController @Inject constructor(
   }
 
   private fun scheduleDefaultAppointmentDateForSheetCreates(events: Observable<UiEvent>): Observable<PotentialAppointmentDate> {
-    return events
-        .ofType<ScheduleAppointmentSheetCreated>()
-        .withLatestFrom(configProvider) { _, config -> config.defaultTimeToAppointment }
+    val protocolStream = currentFacilityStream()
+        .map { Optional.toOptional(it.protocolUuid) }
```
vinaysshenoy: We also have an extension function `toOptional()` that can change this to look like:

```kotlin
.map { it.protocolUuid.toOptional() }
```

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -119,9 +123,17 @@ class ScheduleAppointmentSheetController @Inject constructor(
...
+        .filterAndUnwrapJust()
```
vinaysshenoy: The problem with this approach is that if we have users who update to this version from before the protocols were added to the `Facility` model, this event would never emit and the screen would not update with a date at all.

How does the screen behave in this case? Should we use the default value we were using earlier?

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -119,9 +123,17 @@ class ScheduleAppointmentSheetController @Inject constructor(
   }
 
   private fun scheduleDefaultAppointmentDateForSheetCreates(events: Observable<UiEvent>): Observable<PotentialAppointmentDate> {
-    return events
-        .ofType<ScheduleAppointmentSheetCreated>()
-        .withLatestFrom(configProvider) { _, config -> config.defaultTimeToAppointment }
```
vinaysshenoy: If the `defaultTimeToAppointment` property is no longer necessary, we could remove this. However, we might want to retain it as a fallback for cases where users have not yet synced facilities after updating to this version (see the other comment).

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
vinaysshenoy: Screenshot_1573100430

This is what the screen ends up looking like, and the `DONE` button doesn't do anything since there are no appointment dates that have been generated. We should handle this case, probably use the value from the config as a fallback.

rakshakhegde:
I've fixed all of these in a new PR here. #945 . I'm closing this PR.

PR #935, opened 2019-11-07
Wire the mobius loop with the select country screen

vinaysshenoy:

### DIFF
--- a/app/src/main/java/org/simple/clinic/di/InjectorProviderContextWrapper.kt
+++ b/app/src/main/java/org/simple/clinic/di/InjectorProviderContextWrapper.kt
@@ -3,23 +3,38 @@ package org.simple.clinic.di
 import android.content.Context
 import android.content.ContextWrapper
 
+const val INJECTOR_SERVICE_KEY = "org.simple.clinic.injector_provider"
+
 class InjectorProviderContextWrapper(
     base: Context,
-    private val injectors: Map<String, Any>
+    private val injectors: List<Any>
 ) : ContextWrapper(base) {
 
   companion object {
     fun wrap(
         base: Context,
-        injectors: Map<Class<*>, Any>
-    ): Context = InjectorProviderContextWrapper(base, injectors.mapKeys { it.key.name })
+        vararg injectors: Any
+    ): Context = InjectorProviderContextWrapper(base, injectors.toList())
+  }
+
+  fun find(clazz: Class<*>): Any {
+    val injector = injectors.find { clazz.isAssignableFrom(it.javaClass) }
+
+    requireNotNull(injector) {
+      "Could not find an appropriate injector for ${clazz.name}! Did you forget to implement it on the component or to provide it to the context wrapper?"
+    }
+
+    return injector
   }
 
   override fun getSystemService(name: String): Any? {
-    return if (name in injectors) injectors.getValue(name) else super.getSystemService(name)
+    return if (name == INJECTOR_SERVICE_KEY) this else super.getSystemService(name)
   }
 }
 
-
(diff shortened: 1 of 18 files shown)


app/src/main/java/org/simple/clinic/di/InjectorProviderContextWrapper.kt
```diff
@@ -3,23 +3,38 @@ package org.simple.clinic.di
...
-        injectors: Map<Class<*>, Any>
-    ): Context = InjectorProviderContextWrapper(base, injectors.mapKeys { it.key.name })
+        vararg injectors: Any
+    ): Context = InjectorProviderContextWrapper(base, injectors.toList())
+  }
+
+  fun find(clazz: Class<*>): Any {
+    val injector = injectors.find { clazz.isAssignableFrom(it.javaClass) }
+
+    requireNotNull(injector) {
```
ragunathjawahar: Nice!

app/src/main/java/org/simple/clinic/setup/SetupActivity.kt
```diff
@@ -106,7 +105,7 @@ class SetupActivity : AppCompatActivity(), UiActions {
     val wrappedContext = baseContext
```
ragunathjawahar: We are doing this everywhere, should we extract this into a separate function? Thoughts?

app/src/main/java/org/simple/clinic/selectcountry/SelectCountryScreen.kt
```diff
@@ -86,6 +101,25 @@ class SelectCountryScreen(
     }
   }
 
+  private fun countrySelectionChanges(): Observable<CountryChosen> {
+    return supportedCountriesAdapter
+        .itemEvents
+        .ofType<Event.CountryClicked>()
```
ragunathjawahar: `Event.CountryClicked`?

app/src/main/java/org/simple/clinic/selectcountry/adapter/SelectableCountryItem.kt
```diff
@@ -5,24 +5,31 @@ import io.reactivex.subjects.Subject
...
 import org.simple.clinic.selectcountry.adapter.Event.CountryClicked
 import org.simple.clinic.widgets.ItemAdapter
 import org.simple.clinic.widgets.recyclerview.ViewHolderX
 
 data class SelectableCountryItem(
     val country: Country,
-    val isSelected: Boolean,
-    val showDivider: Boolean
+    private val isCountryChosenByUser: Boolean,
+    private val showDivider: Boolean,
```
ragunathjawahar: Is this how we are doing it? Curious if we are not using `RecyclerView.ItemDecoration` for a reason?

app/src/main/java/org/simple/clinic/setup/SetupActivity.kt
vinaysshenoy: We should, I'll pick this up after finishing the feature.

app/src/main/java/org/simple/clinic/selectcountry/SelectCountryScreen.kt
vinaysshenoy: I don't see the difference?

app/src/main/java/org/simple/clinic/selectcountry/adapter/SelectableCountryItem.kt
vinaysshenoy: I've never used the item decoration feature before so I didn't think about it. I'll revisit this after the feature is complete.

PR #937, opened 2019-11-07
Add Select Country screen to the app navigation graph

vinaysshenoy:
https://www.pivotaltracker.com/story/show/169418103

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/TestData.kt
+++ b/app/src/androidTest/java/org/simple/clinic/TestData.kt
@@ -1,6 +1,7 @@
 package org.simple.clinic
 
 import io.bloco.faker.Faker
+import org.simple.clinic.appconfig.Country
 import org.simple.clinic.bp.BloodPressureMeasurement
 import org.simple.clinic.bp.sync.BloodPressureMeasurementPayload
 import org.simple.clinic.di.AppScope
@@ -58,6 +59,7 @@ import org.simple.clinic.util.toLocalDateAtZone
 import org.threeten.bp.Instant
 import org.threeten.bp.LocalDate
 import org.threeten.bp.ZoneOffset.UTC
+import java.net.URI
 import java.util.UUID
 import javax.inject.Inject
 import kotlin.reflect.KClass
@@ -839,6 +841,20 @@ class TestData @Inject constructor(
         updatedAt = updatedAt,
         deletedAt = deletedAt,
         observations = EncounterObservationsPayload(bloodPressureMeasurements = bpPayloads)
+    )
+  }
+
+  fun country(
+      isoCountryCode: String = "IN",
+      endpoint: String = "https://simple.org",
+      displayName: String = "India",
+      isdCode: String = "91"
+  ): Country {
+    return Country(
+        isoCountryCode = isoCountryCode,
+        endpoint = URI.create(endpoint),
+        displayName = displayName,
+        isdCode = isdCode
     )
   }
 }
--- a/app/src/androidTest/java/org/simple/clinic/user/UserSessionAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/user/UserSessionAndroidTest.kt
@@ -1,6 +1,7 @@
 package org.simple.clinic.user
 
 import androidx.test.runner.AndroidJUnit4
+import com.f2prateek.rx.preferences2.Preference
 import com.google.common.truth.Truth.assertThat
 import org.junit.Before
 import org.junit.Rule
@@ -8,13 +9,16 @@ import org.junit.Test
 import org.junit.rules.RuleChain
 import org.junit.runner.RunWith
 import org.simple.clinic.AppDatabase
-import org.simple.clinic.rules.LocalAuthenticationRule
 import org.simple.clinic.TestClinicApp
 import org.simple.clinic.TestData
+import org.simple.clinic.appconfig.Country
 import org.simple.clinic.facility.FacilityRepository
+import org.simple.clinic.rules.LocalAuthenticationRule
 import org.simple.clinic.user.User.LoggedInStatus.LOGGED_IN
 import org.simple.clinic.user.User.LoggedInStatus.UNAUTHORIZED
+import org.simple.clinic.util.Just
 import org.simple.clinic.util.None
+import org.simple.clinic.util.Optional
 import org.simple.clinic.util.RxErrorsRule
 import org.simple.clinic.util.toOptional
 import javax.inject.Inject
@@ -33,6 +37,9 @@ class UserSessionAndroidTest {
 
   @Inject
   lateinit var facilityRepository: FacilityRepository
+
+  @Inject
+  lateinit var selectedCountryPreference: Preference<Optional<Country>>
 
   @get:Rule
   val ruleChain = RuleChain
@@ -97,5 +104,15 @@ class UserSessionAndroidTest {
     userSession.unauthorize().blockingAwait()
 
     assertThat(userSession.loggedInUserImmediate()!!.loggedInStatus).isEqualTo(UNAUTHORIZED)
+  }
+
+  @Test
+  fun when_user_is_logged_out_then_the_selected_country_preference_must_not_be_removed() {
+    val country = testData.country()
+    selectedCountryPreference.set(country.toOptional())
+
+    userSession.logout().blockingGet()
+
+    assertThat(selectedCountryPreference.get()).isEqualTo(Just(country))
   }
 }
--- a/app/src/main/java/org/simple/clinic/selectcountry/SelectCountryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/selectcountry/SelectCountryScreen.kt
@@ -1,8 +1,10 @@
 package org.simple.clinic.selectcountry
 
 import android.content.Context
+import android.content.Intent
 import android.os.Parcelable
 import android.util.AttributeSet
+import androidx.appcompat.app.AppCompatActivity
 import androidx.constraintlayout.widget.ConstraintLayout
 import androidx.recyclerview.widget.LinearLayoutManager
 import com.jakewharton.rxbinding2.view.RxView
@@ -16,6 +18,7 @@ import org.simple.clinic.appconfig.AppConfigRepository
 import org.simple.clinic.appconfig.Country
 import org.simple.clinic.appconfig.displayname.CountryDisplayNameFetcher
 import org.simple.clinic.di.injector
+import org.simple.clinic.main.TheActivity
 import org.simple.clinic.mobius.MobiusDelegate
 import org.simple.clinic.platform.crash.CrashReporter
 import org.simple.clinic.selectcountry.adapter.Event
@@ -43,6 +46,9 @@ class SelectCountryScreen(
 
   @Inject
   lateinit var countryDisplayNameFetcher: CountryDisplayNameFetcher
+
+  @Inject
+  lateinit var activity: AppCompatActivity
 
   private val uiRenderer = SelectCountryUiRenderer(this)
 
@@ -171,6 +177,14 @@ class SelectCountryScreen(
   }
 
   override fun goToNextScreen() {
-    // TODO(vs): 2019-11-07 Open TheActivity
+    // This navigation should not be done here, we need a way to publish
+    // an event to the parent activity (maybe via the screen router's
+    // event bus?) and handle the navigation there.
+    // TODO(vs): 2019-11-07 Move this to an event that is subscribed in the parent activity
+    val intent = TheActivity.newIntent(activity).apply {
+      flags = Intent.FLAG_ACTIVITY_NO_ANIMATION
+    }
+    activity.startActivity(intent)
+    activity.overridePendingTransition(0, 0)
   }
 }
--- a/app/src/main/java/org/simple/clinic/setup/SetupActivityEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/setup/SetupActivityEffectHandler.kt
@@ -20,6 +20,7 @@ object SetupActivityEffectHandler {
       uiActions: UiActions,
       userDao: User.RoomDao,
       appConfigRepository: AppConfigRepository,
+      fallbackCountry: Country,
       schedulersProvider: SchedulersProvider
   ): ObservableTransformer<SetupActivityEffect, SetupActivityEvent> {
     return RxMobius
@@ -40,6 +41,7 @@ object SetupActivityEffectHandler {
         // effect so that the intention is clear.
         .addTransformer(InitializeDatabase::class.java, initializeDatabase(userDao, schedulersProvider.io()))
         .addAction(ShowCountrySelectionScreen::class.java, uiActions::showCountrySelectionScreen, schedulersProvider.ui())
+        .addTransformer(SetFallbackCountryAsCurrentCountry::class.java, setFallbackCountryAsSelected(appConfigRepository, fallbackCountry, schedulersProvider))
         .build()
   }
 
@@ -81,6 +83,21 @@ object SetupActivityEffectHandler {
       effectStream
           .flatMapSingle { userDao.userCount().subscribeOn(scheduler) }
           .map { DatabaseInitialized }
+    }
+  }
+
+  private fun setFallbackCountryAsSelected(
+      appConfigRepository: AppConfigRepository,
+      fallbackCountry: Country,
+      schedulersProvider: SchedulersProvider
+  ): ObservableTransformer<SetFallbackCountryAsCurrentCountry, SetupActivityEvent> {
+    return ObservableTransformer { effectStream ->
+      effectStream.flatMapSingle {
+        appConfigRepository
+            .saveCurrentCountry(fallbackCountry)
+            .subscribeOn(schedulersProvider.io())
+            .toSingleDefault(FallbackCountrySetAsSelected)
+      }
     }
   }
 }
--- a/app/src/main/java/org/simple/clinic/setup/SetupActivityEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/setup/SetupActivityEffectHandler.kt
@@ -5,8 +5,10 @@ import com.spotify.mobius.rx2.RxMobius
 import io.reactivex.ObservableTransformer
 import io.reactivex.Scheduler
 import io.reactivex.Single
+import io.reactivex.rxkotlin.zipWith
 import org.simple.clinic.user.User
 import org.simple.clinic.util.scheduler.SchedulersProvider
+import org.simple.clinic.util.toOptional
 
 object SetupActivityEffectHandler {
 
@@ -18,7 +20,7 @@ object SetupActivityEffectHandler {
   ): ObservableTransformer<SetupActivityEffect, SetupActivityEvent> {
     return RxMobius
         .subtypeEffectHandler<SetupActivityEffect, SetupActivityEvent>()
-        .addTransformer(FetchUserDetails::class.java, fetchUserDetails(onboardingCompletePreference, schedulersProvider.io()))
+        .addTransformer(FetchUserDetails::class.java, fetchUserDetails(onboardingCompletePreference, userDao, schedulersProvider.io()))
         .addAction(GoToMainActivity::class.java, uiActions::goToMainActivity, schedulersProvider.ui())
         .addAction(ShowOnboardingScreen::class.java, uiActions::showOnboardingScreen, schedulersProvider.ui())
         // We could technically also implicitly wait on the database to
@@ -38,13 +40,20 @@ object SetupActivityEffectHandler {
 
   private fun fetchUserDetails(
       onboardingCompletePreference: Preference<Boolean>,
+      userDao: User.RoomDao,
       scheduler: Scheduler
   ): ObservableTransformer<FetchUserDetails, SetupActivityEvent> {
     return ObservableTransformer { effectStream ->
       effectStream
-          .flatMapSingle { Single.just(onboardingCompletePreference.get()) }
-          .subscribeOn(scheduler)
-          .map(::UserDetailsFetched)
+          .flatMapSingle {
+            val hasUserCompletedOnboarding = Single.fromCallable { onboardingCompletePreference.get() }
+            val loggedInUser = Single.fromCallable { userDao.userImmediate().toOptional() }
+
+            hasUserCompletedOnboarding
+                .zipWith(loggedInUser)
+                .subscribeOn(scheduler)
+          }
+          .map { (hasUserCompletedOnboarding, loggedInUser) -> UserDetailsFetched(hasUserCompletedOnboarding, loggedInUser) }
     }
   }
 
--- a/app/src/main/java/org/simple/clinic/setup/SetupActivityEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/setup/SetupActivityEffectHandler.kt
@@ -5,8 +5,11 @@ import com.spotify.mobius.rx2.RxMobius
 import io.reactivex.ObservableTransformer
 import io.reactivex.Scheduler
 import io.reactivex.Single
-import io.reactivex.rxkotlin.zipWith
+import io.reactivex.rxkotlin.Singles
+import org.simple.clinic.appconfig.AppConfigRepository
+import org.simple.clinic.appconfig.Country
 import org.simple.clinic.user.User
+import org.simple.clinic.util.Optional
 import org.simple.clinic.util.scheduler.SchedulersProvider
 import org.simple.clinic.util.toOptional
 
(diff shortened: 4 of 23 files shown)


app/src/main/java/org/simple/clinic/selectcountry/SelectCountryScreen.kt
```diff
@@ -171,6 +177,14 @@ class SelectCountryScreen(
   }
 
   override fun goToNextScreen() {
-    // TODO(vs): 2019-11-07 Open TheActivity
+    // This navigation should not be done here, we need a way to publish
+    // an event to the parent activity (maybe via the screen router's
+    // event bus?) and handle the navigation there.
+    // TODO(vs): 2019-11-07 Move this to an event that is subscribed in the parent activity
```
ragunathjawahar: Consider this before we implement communication inside a hierarchy? https://github.com/uber/RIBs/wiki#communication-between-ribs

app/src/test/java/org/simple/clinic/setup/SetupActivityEffectHandlerTest.kt
```diff
@@ -34,15 +38,16 @@ class SetupActivityEffectHandlerTest {
...
-    whenever(onboardingCompletePreference.get()).doReturn(true)
+    whenever(onboardingCompletePreference.get()) doReturn true
+    whenever(userDao.userImmediate()).doReturn<User?>(null)
 
     // when
     testCase.dispatch(FetchUserDetails)
 
     // then
-    testCase.assertOutgoingEvents(UserDetailsFetched(hasUserCompletedOnboarding = true))
+    testCase.assertOutgoingEvents(UserDetailsFetched(hasUserCompletedOnboarding = true, loggedInUser = None))
```
ragunathjawahar: Consider using factory functions to set these properties, gives user more context?

app/src/test/java/org/simple/clinic/setup/SetupActivityEffectHandlerTest.kt
```diff
@@ -80,4 +85,20 @@ class SetupActivityEffectHandlerTest {
...
+    // given
+    whenever(onboardingCompletePreference.get()) doReturn false
+    val user = PatientMocker.loggedInUser(uuid = UUID.fromString("426d2eb9-ebf7-4a62-b157-1de221c7c3d0"))
+    whenever(userDao.userImmediate()).doReturn(user)
+
+    // when
+    testCase.dispatch(FetchUserDetails)
+
+    // then
+    val expected = UserDetailsFetched(hasUserCompletedOnboarding = false, loggedInUser = Just(user))
```
ragunathjawahar: Same as above.

app/src/test/java/org/simple/clinic/setup/SetupActivityEffectHandlerTest.kt
```diff
@@ -80,4 +85,20 @@ class SetupActivityEffectHandlerTest {
...
+    testCase.assertOutgoingEvents(expected)
```
ragunathjawahar: `expected`? Can we use a more descriptive name here?

app/src/test/java/org/simple/clinic/setup/SetupActivityUpdateTest.kt
```diff
@@ -26,7 +27,7 @@ class SetupActivityUpdateTest {
   fun `if the user has not completed onboarding, the onboarding screen must be shown`() {
     updateSpec
         .given(defaultModel)
-        .whenEvent(UserDetailsFetched(hasUserCompletedOnboarding = false))
+        .whenEvent(UserDetailsFetched(hasUserCompletedOnboarding = false, loggedInUser = None))
```
ragunathjawahar: Same as above. Factory functions to create these instances.

app/src/main/java/org/simple/clinic/setup/SetupActivityUpdate.kt
```diff
@@ -4,16 +4,30 @@ import com.spotify.mobius.Next
...
-      is UserDetailsFetched -> if (event.hasUserCompletedOnboarding) GoToMainActivity else ShowOnboardingScreen
+      is UserDetailsFetched -> initialScreenEffect(event.loggedInUser, event.hasUserCompletedOnboarding)
       is DatabaseInitialized -> FetchUserDetails
       else -> null
     }
 
     return if (effect == null) noChange() else dispatch(effect)
   }
+
+  private fun initialScreenEffect(
```
ragunathjawahar: Suggestion `goToNextScreenEffect`?

app/src/main/java/org/simple/clinic/setup/SetupActivityModel.kt
```diff
@@ -18,4 +22,12 @@ data class SetupActivityModel(
   companion object {
     val SETTING_UP = SetupActivityModel(isUserLoggedIn = null, hasUserSelectedACountry = null)
   }
+
+  fun withLoggedInUser(user: Optional<User>): SetupActivityModel {
+    return copy(isUserLoggedIn = user is Just)
```
ragunathjawahar: Perhaps use `user.isNotEmpty()`?

app/src/main/java/org/simple/clinic/setup/SetupActivityModel.kt
```diff
@@ -18,4 +22,12 @@ data class SetupActivityModel(
...
+  }
+
+  fun withSelectedCountry(country: Optional<Country>): SetupActivityModel {
+    return copy(hasUserSelectedACountry = country is Just)
```
ragunathjawahar: Same as above?

app/src/main/java/org/simple/clinic/setup/SetupActivityUpdate.kt
```diff
@@ -4,20 +4,27 @@ import com.spotify.mobius.Next
...
-    val effect = when (event) {
-      is UserDetailsFetched -> initialScreenEffect(event.loggedInUser, event.hasUserCompletedOnboarding)
-      is DatabaseInitialized -> FetchUserDetails
-      else -> null
-    }
+    return when (event) {
+      is UserDetailsFetched -> {
+        val updatedModel = model
+            .withLoggedInUser(event.loggedInUser)
+            .withSelectedCountry(None)
```
ragunathjawahar: Instead of chaining functions with different values, can we create a reducer function for each of these state transitions? Such as `userLoggedIn`, `userNotLoggedIn`, and so on.

app/src/main/java/org/simple/clinic/setup/SetupActivityEffectHandler.kt
```diff
@@ -42,22 +46,33 @@ object SetupActivityEffectHandler {
...
+  ): Single<Triple<Boolean, Optional<User>, Optional<Country>>> {
+    val hasUserCompletedOnboarding = Single.fromCallable { onboardingCompletePreference.get() }
+    val loggedInUser = Single.fromCallable { userDao.userImmediate().toOptional() }
+    val userSelectedCountry = Single.fromCallable { appConfigRepository.currentCountry() }
+
+    return Singles
+        .zip(hasUserCompletedOnboarding, loggedInUser, userSelectedCountry)
+        .subscribeOn(scheduler)
+  }
+
```
ragunathjawahar: As discussed offline,

```kotlin
private fun fetchUserDetails(scheduler: Scheduler): ObservableTransformer<FetchUserDetails, SetupActivityEvent> {
    return ObservableTransformer { effectStream ->
      effectStream
          .flatMapSingle { Single.fromCallable(::readUserDetailsFromStorage).subscribeOn(scheduler) }
          .map { (hasUserCompletedOnboarding, loggedInUser, userSelectedCountry) ->
            UserDetailsFetched(hasUserCompletedOnboarding, loggedInUser, userSelectedCountry)
          }
    }
  }

  private fun readUserDetailsFromStorage(): Triple<Boolean, Optional<User>, Optional<Country>> {
    val hasUserCompletedOnboarding = onboardingCompletePreference.get()
    val loggedInUser = userDao.userImmediate().toOptional()
    val userSelectedCountry = appConfigRepository.currentCountry()

    return Triple(hasUserCompletedOnboarding, loggedInUser, userSelectedCountry)
  }
```

app/src/main/java/org/simple/clinic/selectcountry/SelectCountryScreen.kt
vinaysshenoy: Discussed offline: our current screen navigation component does not full support the callback style upstream communication in some cases. We will need to rethink this, punting on it for later.

app/src/main/java/org/simple/clinic/setup/SetupActivityUpdate.kt
vinaysshenoy: Makes sense. I'll see what I can do.

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
```diff
@@ -488,4 +490,18 @@ object PatientMocker {
...
+
+  fun country(
+      isoCountryCode: String = "IN",
+      endpoint: String = "https://simple.org",
+      displayName: String = "India",
+      isdCode: String = "91"
+  ): Country {
+    return Country(
+        isoCountryCode = isoCountryCode,
+        endpoint = URI.create(endpoint),
```
ragunathjawahar: I am slightly concerned about this, because `country` is no longer a pure function, because `URI.create(String)` can throw exceptions 😕

app/src/test/java/org/simple/clinic/setup/SetupActivityEffectHandlerTest.kt
```diff
@@ -42,12 +45,13 @@ class SetupActivityEffectHandlerTest {
...
     whenever(onboardingCompletePreference.get()) doReturn true
     whenever(userDao.userImmediate()).doReturn<User?>(null)
+    whenever(appConfigRepository.currentCountry()) doReturn None
 
     // when
     testCase.dispatch(FetchUserDetails)
 
     // then
-    testCase.assertOutgoingEvents(UserDetailsFetched(hasUserCompletedOnboarding = true, loggedInUser = None))
+    testCase.assertOutgoingEvents(UserDetailsFetched(hasUserCompletedOnboarding = true, loggedInUser = None, userSelectedCountry = None))
```
ragunathjawahar: Constructors in data classes expose too much information. Hence we should always consider exposing them via factory instance creation functions. These set of parameters have a meaning, but as a maintainer, I have to try hard to understand what it means. Also, when we keep adding new parameters to these functions, the tests and production code can get out of sync pretty quickly.

app/src/test/java/org/simple/clinic/setup/SetupActivityUpdateTest.kt
```diff
@@ -24,7 +24,7 @@ class SetupActivityUpdateTest {
 
     updateSpec
         .given(defaultModel)
-        .whenEvent(UserDetailsFetched(hasUserCompletedOnboarding = true, loggedInUser = None))
+        .whenEvent(UserDetailsFetched(hasUserCompletedOnboarding = true, loggedInUser = None, userSelectedCountry = None))
```
ragunathjawahar: Same as above.

app/src/test/java/org/simple/clinic/setup/SetupActivityUpdateTest.kt
```diff
@@ -39,7 +39,7 @@ class SetupActivityUpdateTest {
 
     updateSpec
         .given(defaultModel)
-        .whenEvent(UserDetailsFetched(hasUserCompletedOnboarding = false, loggedInUser = None))
+        .whenEvent(UserDetailsFetched(hasUserCompletedOnboarding = false, loggedInUser = None, userSelectedCountry = None))
```
ragunathjawahar: Same as above.

app/src/test/java/org/simple/clinic/setup/SetupActivityUpdateTest.kt
```diff
@@ -68,7 +68,7 @@ class SetupActivityUpdateTest {
 
     updateSpec
         .given(defaultModel)
-        .whenEvent(UserDetailsFetched(hasUserCompletedOnboarding = true, loggedInUser = user.toOptional()))
+        .whenEvent(UserDetailsFetched(hasUserCompletedOnboarding = true, loggedInUser = user.toOptional(), userSelectedCountry = None))
```
ragunathjawahar: Same as above.

app/src/main/java/org/simple/clinic/setup/SetupActivityEffectHandler.kt
```diff
@@ -83,4 +85,19 @@ object SetupActivityEffectHandler {
           .map { DatabaseInitialized }
     }
   }
+
+  private fun setFallbackCountryAsSelected(
+      appConfigRepository: AppConfigRepository,
+      fallbackCountry: Country,
+      schedulersProvider: SchedulersProvider
```
ragunathjawahar: `Scheduler`?

app/src/main/java/org/simple/clinic/setup/SetupActivityModule.kt
```diff
@@ -3,11 +3,26 @@ package org.simple.clinic.setup
...
 
   @Provides
   fun provideUserDao(appDatabase: AppDatabase): User.RoomDao = appDatabase.userDao()
+
+  @Provides
+  @Named("fallback")
+  fun providesFallbackCountry(): Country {
+    return Country(
+        isoCountryCode = "IN",
+        endpoint = URI.create(BuildConfig.FALLBACK_ENDPOINT),
```
ragunathjawahar: `URI.create(String)` can throw!

app/src/main/java/org/simple/clinic/setup/SetupActivityUpdate.kt
```diff
@@ -17,22 +18,28 @@ class SetupActivityUpdate : Update<SetupActivityModel, SetupActivityEvent, Setup
...
     }
   }
 
   private fun initialScreenEffect(
       loggedInUser: Optional<User>,
-      hasUserCompletedOnboarding: Boolean
+      hasUserCompletedOnboarding: Boolean,
+      selectedCountry: Optional<Country>
   ): SetupActivityEffect {
+    val hasUserLoggedInCompletely = loggedInUser is Just && selectedCountry is Just
```
ragunathjawahar: Do you also feel `.isNotEmpty()` has better readability, than comparing it to the type?

app/src/main/java/org/simple/clinic/setup/SetupActivityUpdate.kt
```diff
@@ -17,22 +18,28 @@ class SetupActivityUpdate : Update<SetupActivityModel, SetupActivityEvent, Setup
...
+    val userPresentButCountryNotSelected = loggedInUser is Just && selectedCountry is None
```
ragunathjawahar: Same as above.

app/src/test/java/org/simple/clinic/setup/SetupActivityUpdateTest.kt
```diff
@@ -60,8 +60,31 @@ class SetupActivityUpdateTest {
...
+    val country = PatientMocker.country()
+
+    //then
+    val expectedModel = defaultModel
+        .withLoggedInUser(user.toOptional())
+        .withSelectedCountry(country.toOptional())
+
+    updateSpec
+        .given(defaultModel)
+        .whenEvent(UserDetailsFetched(hasUserCompletedOnboarding = true, loggedInUser = user.toOptional(), userSelectedCountry = country.toOptional()))
```
ragunathjawahar: Factory function?

app/src/main/java/org/simple/clinic/setup/SetupActivityEffectHandler.kt
```diff
@@ -13,19 +12,39 @@ import org.simple.clinic.util.Optional
...
+          schedulersProvider = schedulersProvider
+      ).build()
+    }
+  }
+
+  private fun build(): ObservableTransformer<SetupActivityEffect, SetupActivityEvent> {
     return RxMobius
         .subtypeEffectHandler<SetupActivityEffect, SetupActivityEvent>()
-        .addTransformer(FetchUserDetails::class.java, fetchUserDetails(onboardingCompletePreference, userDao, appConfigRepository, schedulersProvider.io()))
+        .addTransformer(FetchUserDetails::class.java, fetchUserDetails())
```
ragunathjawahar: Pass in the `Scheduler` probably?

app/src/main/java/org/simple/clinic/setup/SetupActivityEffectHandler.kt
```diff
@@ -39,58 +58,41 @@ object SetupActivityEffectHandler {
...
-  ): Single<Triple<Boolean, Optional<User>, Optional<Country>>> {
+  private fun readUserDetailsFromStorage(): Single<Triple<Boolean, Optional<User>, Optional<Country>>> {
     val hasUserCompletedOnboarding = Single.fromCallable { onboardingCompletePreference.get() }
     val loggedInUser = Single.fromCallable { userDao.userImmediate().toOptional() }
     val userSelectedCountry = Single.fromCallable { appConfigRepository.currentCountry() }
 
     return Singles
         .zip(hasUserCompletedOnboarding, loggedInUser, userSelectedCountry)
-        .subscribeOn(scheduler)
+        .subscribeOn(schedulersProvider.io())
```
ragunathjawahar: Same as above.

app/src/main/java/org/simple/clinic/setup/SetupActivityEffectHandler.kt
```diff
@@ -39,58 +58,41 @@ object SetupActivityEffectHandler {
...
 
-  private fun initializeDatabase(
-      userDao: User.RoomDao,
-      scheduler: Scheduler
-  ): ObservableTransformer<InitializeDatabase, SetupActivityEvent> {
+  private fun initializeDatabase(): ObservableTransformer<InitializeDatabase, SetupActivityEvent> {
     return ObservableTransformer { effectStream ->
       effectStream
-          .flatMapSingle { userDao.userCount().subscribeOn(scheduler) }
+          .flatMapSingle { userDao.userCount().subscribeOn(schedulersProvider.io()) }
```
ragunathjawahar: Same as above?

app/src/androidTest/java/org/simple/clinic/TestData.kt
```diff
@@ -841,4 +843,18 @@ class TestData @Inject constructor(
...
+
+  fun country(
+      isoCountryCode: String = "IN",
+      endpoint: String = "https://simple.org",
+      displayName: String = "India",
+      isdCode: String = "91"
+  ): Country {
+    return Country(
+        isoCountryCode = isoCountryCode,
+        endpoint = URI.create(endpoint),
```
ragunathjawahar: `URI.create(endpoint)` 🧨 😅

app/src/androidTest/java/org/simple/clinic/user/UserSessionAndroidTest.kt
```diff
@@ -98,4 +105,14 @@ class UserSessionAndroidTest {
 
     assertThat(userSession.loggedInUserImmediate()!!.loggedInStatus).isEqualTo(UNAUTHORIZED)
   }
+
+  @Test
+  fun when_user_is_logged_out_then_the_selected_country_preference_must_not_be_removed() {
```
ragunathjawahar: Nice!!

app/src/test/java/org/simple/clinic/user/UserSessionTest.kt
```diff
@@ -836,6 +854,9 @@ class UserSessionTest {
     val preferencesEditor = mock<SharedPreferences.Editor>()
     whenever(preferencesEditor.clear()).thenReturn(preferencesEditor)
     whenever(sharedPrefs.edit()).thenReturn(preferencesEditor)
+    whenever(preferencesEditor.putString(eq("key"), any())) doReturn preferencesEditor
+    whenever(selectedCountryPreference.key()) doReturn "key"
+    whenever(sharedPrefs.getString(eq("key"), any())) doReturn ""
```
ragunathjawahar: Ufff... these tests 😅. Cascading changes across tests. Guess, we already have a pivotal ticket for refactoring `UserSession`?

vinaysshenoy: Yes, a lot of things will be broken out of `UserSession` once the breaking up of the DI scope happens. We can revisit how much `UserSession` needs to be refactored out after.

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
vinaysshenoy: This is intentional. We do not want an invalid endpoint to be provided. In production, this is something we control so given an invalid endpoint should be a fatal error.

app/src/main/java/org/simple/clinic/setup/SetupActivityModule.kt
vinaysshenoy: Intentionally added.

vinaysshenoy: Since this is something we control completely, an invalid endpoint is not something the app should handle and should fail instantly.

app/src/main/java/org/simple/clinic/setup/SetupActivityUpdate.kt
vinaysshenoy: Yes, it does. I'll make this change.

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
ragunathjawahar: Should the app crash in case of a misconfiguration?

app/src/main/java/org/simple/clinic/setup/SetupActivityEffectHandler.kt
vinaysshenoy: What benefit does it provide? We needed to do it earlier because the effect handler was a factory function, but now everything is a part of the class properties. This is anyway a private function so it does not need to be exposed for testing.

ragunathjawahar: Looking at the subtype effect handler tells us what scheduler the effect handler is being executed?

vinaysshenoy: Is that really important? The functions are quite small and easy to parse. Additionally, this reasoning breaks if we need to pass two schedulers to the effect handler for different use cases because we will know that the effect handler uses these two, but not for what.

My opinion on this is that if someone is at a point where they need to know what scheduler an effect handler is running on, they will be anyway looking closely at the function body and it is quite obvious which scheduler is being used there and where.

Does this make sense?

vinaysshenoy: In addition, I feel that adding the scheduler as a method parameter is exposing an implementation detail.

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
vinaysshenoy: I am of the opinion that it should.

app/src/test/java/org/simple/clinic/setup/SetupActivityEffectHandlerTest.kt
vinaysshenoy: Discussed offline: all three of the properties will always get set together and not individually. The tests have been updated to reflect this, so there was no need to create separate factory functions.

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
vinaysshenoy: Discussed offline: we'll retain the `URI.create()` for now and see if you need to change this later since Retrofit will anyway throw if the url is malformed.

app/src/test/java/org/simple/clinic/setup/SetupActivityEffectHandlerTest.kt
vinaysshenoy: For this specific event, all three properties will always have to be set together since this is an event. The tests have been updated to reflect this.

app/src/test/java/org/simple/clinic/setup/SetupActivityUpdateTest.kt
vinaysshenoy: Added a bunch of factory functions in the test to better expose intent.

app/src/main/java/org/simple/clinic/setup/SetupActivityEffectHandler.kt
vinaysshenoy: Discussed offline: Decided to pass the scheduler as a parameter to the effect handler function to maintain consistency with how we'll use schedulers when adding effect handler consumers and actions.

app/src/test/java/org/simple/clinic/setup/SetupActivityEffectHandlerTest.kt
vinaysshenoy: This test has been removed after unifying the different conditions for fetching user details.

PR #938, opened 2019-11-08
Refactor to more idiomatic Mobius while adding support for the new address model.

ragunathjawahar:


app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
```diff
@@ -3956,6 +3956,24 @@ class DatabaseMigrationAndroidTest {
...
+  fun migration_from_52_to_53_should_add_house_address_and_zone_columns() {
+    val db_52 = helper.createDatabase(52)
+    val selectAllPatientAddressesQuery = "SELECT * FROM PatientAddress"
+    db_52.query(selectAllPatientAddressesQuery).use {
+      assertThat(it.columnCount)
+          .isEqualTo(8)
+    }
+
+    val db_53 = helper.migrateTo(53)
+    db_53.query(selectAllPatientAddressesQuery).use {
```
vinaysshenoy: This assertion does not verify that the other columns were not modified as part of this migration. We have an extension function `SupportSQLiteDatabase.assertColumns` which requires passing in all the columns that you expect the table to have. Use that instead?

app/src/androidTest/java/org/simple/clinic/TestData.kt
```diff
@@ -186,7 +186,9 @@ class TestData @Inject constructor(
       deletedAt: Instant? = null
   ) = PatientAddress(
       uuid = uuid,
-      colonyOrVillage = colonyOrVilage,
+      houseAddress = null,
```
vinaysshenoy: Instead of setting this to `null`, should we make default parameters and set them to specific values?

app/src/main/java/org/simple/clinic/patient/sync/PatientPayload.kt
```diff
@@ -117,7 +117,9 @@ data class PatientAddressPayload(
   fun toDatabaseModel(): PatientAddress {
     return PatientAddress(
         uuid = uuid,
+        houseAddress = null,
```
vinaysshenoy: Should we add a `TODO` here to read this from the api?

app/src/main/java/org/simple/clinic/storage/Migration_52_53.kt
```diff
@@ -0,0 +1,15 @@
+package org.simple.clinic.storage
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+class Migration_52_53 : Migration(52, 53) {
+  override fun migrate(database: SupportSQLiteDatabase) {
+    with(database) {
+      inTransaction {
+        execSQL("ALTER TABLE PatientAddress ADD COLUMN houseAddress TEXT DEFAULT NULL")
```
vinaysshenoy: We've been using the convention of setting these as string templates (triple-quoted), and quoting the identifiers in the SQL statement as per the SQL conventions. Can we follow the convention please?

```kotlin
execSQL("""ALTER TABLE "PatientAddress" ADD COLUMN "houseAddress" TEXT DEFAULT NULL""")
execSQL("""ALTER TABLE "PatientAddress" ADD COLUMN "zone" TEXT DEFAULT NULL""")
```

app/src/test/java/org/simple/clinic/patient/PatientMocker.kt
```diff
@@ -82,7 +82,9 @@ object PatientMocker {
   ): PatientAddress {
     return PatientAddress(
         uuid = uuid,
+        houseAddress = null,
```
vinaysshenoy: Same as earlier comment about reading the values for these from method parameters and setting default values for them.

app/src/main/java/org/simple/clinic/newentry/PatientEntryUpdate.kt
```diff
@@ -44,7 +44,7 @@ class PatientEntryUpdate(
   private fun onGenderChanged(model: PatientEntryModel, gender: Optional<Gender>): PatientEntryNext {
     val updatedModel = model.withGender(gender)
     return if (gender.isNotEmpty() && model.isSelectingGenderForTheFirstTime) {
-      next(updatedModel.selectedGender(), HideMissingGenderError, ScrollFormToBottom)
+      next(updatedModel, HideMissingGenderError, ScrollFormToBottom)
```
vinaysshenoy: Since the model has not changed here, we can just dispatch the effects, right?

vinaysshenoy: Never mind, I just saw that the model is being changed in an earlier statement.

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
```diff
@@ -76,13 +77,14 @@ class PatientEntryEffectHandler(
...
             DateOfBirth -> hideDateOfBirthErrors()
+            Gender -> ui.showMissingGenderError(false)
             ColonyOrVillage -> ui.showEmptyColonyOrVillageError(false)
             District -> ui.showEmptyDistrictError(false)
             State -> ui.showEmptyStateError(false)
           }
         }, schedulersProvider.ui())
         .addAction(HideEmptyDateOfBirthAndAgeError::class.java, { ui.showEmptyDateOfBirthAndAgeError(false) }, schedulersProvider.ui())
-        .addAction(HideMissingGenderError::class.java, { ui.showMissingGenderError(false) }, schedulersProvider.ui())
+        // .addAction(HideMissingGenderError::class.java, { ui.showMissingGenderError(false) }, schedulersProvider.ui())
```
vinaysshenoy: Forgot to remove this action?

app/src/main/java/org/simple/clinic/patient/PatientAddress.kt
```diff
@@ -18,8 +18,12 @@ data class PatientAddress(
     @PrimaryKey
     val uuid: UUID,
 
+    val streetAddress: String?,
```
vinaysshenoy: Reword the commit message to reflect the change in the field name?

app/src/androidTest/java/org/simple/clinic/DatabaseMigrationAndroidTest.kt
ragunathjawahar: Done.

app/src/main/java/org/simple/clinic/storage/Migration_52_53.kt
ragunathjawahar: Done

app/src/main/java/org/simple/clinic/patient/sync/PatientPayload.kt
ragunathjawahar: Nice catch. Added.

PR #939, opened 2019-11-08
Use the selected country to display country dialling codes

vinaysshenoy:
https://www.pivotaltracker.com/story/show/169472237

### DIFF
--- a/app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
+++ b/app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
@@ -14,6 +14,7 @@ import androidx.transition.TransitionManager
 import com.jakewharton.rxbinding2.view.RxView
 import com.jakewharton.rxbinding2.widget.RxTextView
 import io.reactivex.Observable
+import kotlinx.android.synthetic.main.screen_enterotp.view.*
 import kotterknife.bindView
 import org.simple.clinic.R
 import org.simple.clinic.main.TheActivity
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
```diff
@@ -34,15 +35,6 @@ class EnterOtpScreen(context: Context, attributeSet: AttributeSet) : RelativeLay
   @Inject
   lateinit var screenRouter: ScreenRouter
 
-  private val userPhoneNumberTextView by bindView<TextView>(R.id.enterotp_phonenumber)
```
ragunathjawahar: ✂️

app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
```diff
@@ -75,7 +72,7 @@ class EnterOtpScreen(context: Context, attributeSet: AttributeSet) : RelativeLay
   fun showUserPhoneNumber(phoneNumber: String) {
     val phoneNumberWithCountryCode = resources.getString(
         R.string.enterotp_phonenumber,
-        resources.getString(R.string.country_dialing_code),
```
ragunathjawahar: Did we also remove this string resource?

app/src/main/res/layout/sheet_phone_mask.xml
```diff
@@ -28,7 +28,7 @@
     android:gravity="center"
     android:maxLines="1"
     android:textAppearance="@style/Clinic.V2.TextAppearance.Body2Left.Numeric.Grey1"
-    tools:text="+91 9876543210" />
```
ragunathjawahar: Curious, why are we removing this?

app/src/main/res/values/strings.xml
```diff
@@ -247,6 +245,7 @@
   <string name="registration_loggedout_dialog_title">You have been logged out</string>
   <string name="registration_loggedout_dialog_message">Please login again to continue</string>
   <string name="registration_loggedout_dialog_confirm">Okay</string>
+  <string name="registrationphone_country_code">+%1$s</string>
```
ragunathjawahar: Add the `translatable="false"` attribute?

vinaysshenoy: Good point, I'll make this change.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
vinaysshenoy: Yes, in a later commit.

app/src/main/res/layout/sheet_phone_mask.xml
vinaysshenoy: Because this is not in sync with the actual implementation. In the production code, we do not prepend the country dialling code to the number. I'm guessing this was changed after the initial feature was built, but we did not change the layout placeholder.

PR #943, opened 2019-11-13
Prepare patient entry screen for the new address model

ragunathjawahar:


app/src/main/java/org/simple/clinic/newentry/clearbutton/ClearFieldImageButton.kt
```diff
@@ -60,14 +62,21 @@ class ClearFieldImageButton(context: Context, attrs: AttributeSet) : AppCompatIm
 
     fieldOriginalPaddingEnd = field.paddingEnd
 
-    Observable.merge(clearableFieldTextChanges(), clearableFieldFocusChanges())
+    disposable = Observable.merge(clearableFieldTextChanges(), clearableFieldFocusChanges())
```
vinaysshenoy: Shouldn't `.takeUntil(RxView.detaches(this))` take care of disposing the stream?

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
```diff
@@ -44,6 +44,7 @@ class PatientEntryEffectHandler(
     private val patientRepository: PatientRepository,
     private val patientRegisteredCount: Preference<Int>,
     private val ui: PatientEntryUi,
+    private val validationActions: PatientEntryValidationActions,
```
vinaysshenoy: I am assuming that the future goal is to have this interface take in a single method that has a `Field[]` as a parameter, and it's up to the implementation to decide how to show/hide the errors for a field.

Otherwise, this doesn't give us that much benefit.

app/src/main/java/org/simple/clinic/newentry/clearbutton/ClearFieldImageButton.kt
ragunathjawahar: This just makes it explicit.

vinaysshenoy: Right, do we still need to `takeUntil` operator then?

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
ragunathjawahar: Yes, that is correct. Since the validation logic is going to become complicated. We can have a validation renderer component that just takes care of showing and hiding errors. This is a reason why we are moving in this direction.

app/src/main/java/org/simple/clinic/newentry/clearbutton/ClearFieldImageButton.kt
ragunathjawahar: I don't know if this behavior could influence other parts in which this component is being used.

vinaysshenoy: Alright, let's leave this in for now and then and decide later.

PR #944, opened 2019-11-13
Use the selected country endpoint for setting up the country specific Retrofit instance

vinaysshenoy:

## Blockers

- [x] Refactor login flows out of `UserSession` (#948)
- [x] Refactor find user flow out of `UserSession` (#950)
- [x] Refactor refresh user flow out of `UserSession` (#950)
- [x] Refactor register user flow out of `UserSession` (#951)
- [x] Refactor reset PIN flow out of `UserSession` (#953)
- [x] Refactor sync everything out of `UserSession` (#955)


app/src/main/java/org/simple/clinic/ClinicApp.kt
```diff
@@ -103,8 +81,6 @@ abstract class ClinicApp : Application() {
   abstract fun buildDaggerGraph(): AppComponent
 
   protected fun setupSync() {
-    syncScheduler.schedule().subscribe()
-    syncIndicatorStatusCalculator.updateSyncResults()
-    dataSync.syncTheWorld().subscribeOn(schedulersProvider.io()).subscribe()
+    syncSetup.run()
```
sanchita: Nice!

app/src/main/java/org/simple/clinic/di/network/RetrofitModule.kt
```diff
@@ -38,12 +40,18 @@ class RetrofitModule {
...
-    val baseUrl = BuildConfig.API_ENDPOINT
-    val currentApiVersion = "v3"
+  fun retrofit(
+      country: Country,
+      commonRetrofitBuilder: Retrofit.Builder
+  ): Retrofit {
+    // Since the endpoint is not under our control, and is defined at the server level,
+    // this is a safety check that will generate the right endpoint regardless of whether the
+    // endpoint defined in the manifest has a trailing slash or not.
+    val baseUrl = country.endpoint.toString().removeSuffix("/")
```
sanchita: Pruning the endpoint looks like an overkill.
The endpoint should either have a slash or not and this should be a contract. 
I don't think we need a fallback for this.

vinaysshenoy: @ragunathjawahar and I discussed this at length. Since we do not have any way to enforce this contract from our side and it's dependent on humans to configure, we decided to add this check.

In addition, the ones who are configuring this endpoint are people who may not be aware that Retrofit requires the endpoint to end in a trailing slash, so this is currently some tribal knowledge that it might be prudent to account for.

sanchita: Question: When we are talking about people configuring the endpoint, are they people outside the Simple team?

vinaysshenoy: Not at the moment. But it's still tribal knowledge even within the mobile team since this is something that is not done frequently.

sanchita: The endpoint being configured incorrectly points to a different problem. This can happen irrespective of the slash. The app should still be able to recover from this issue. If not, we should talk to server devs about documenting this maybe.
I am merging this PR for now. Meanwhile let's discuss this elsewhere.

PR #948, opened 2019-11-15
Refactor the login flow out of the User Session class

vinaysshenoy:

### DIFF
--- a/app/src/main/java/org/simple/clinic/login/OtpSmsReceiver.kt
+++ b/app/src/main/java/org/simple/clinic/login/OtpSmsReceiver.kt
@@ -11,7 +11,6 @@ import com.google.android.gms.common.api.CommonStatusCodes
 import com.google.android.gms.common.api.Status
 import org.simple.clinic.ClinicApp
 import org.simple.clinic.R
-import org.simple.clinic.user.LoginUserWithOtp
 import org.simple.clinic.user.UserSession
 import org.simple.clinic.util.scheduler.SchedulersProvider
 import timber.log.Timber
--- a/app/src/main/java/org/simple/clinic/login/OtpSmsReceiver.kt
+++ b/app/src/main/java/org/simple/clinic/login/OtpSmsReceiver.kt
@@ -4,16 +4,16 @@ import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
-import androidx.annotation.StringRes
 import android.widget.Toast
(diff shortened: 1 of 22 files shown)


app/src/main/java/org/simple/clinic/login/OtpSmsReceiver.kt
```diff
@@ -31,6 +31,14 @@ class OtpSmsReceiver : BroadcastReceiver() {
   @Inject
   lateinit var userSession: UserSession
 
+  // This is intentionally lazy so that the country specific retrofit endpoint
+  // is not instantiated until the SMS is received.
+  @Inject
+  lateinit var loginUserWithOtp: dagger.Lazy<LoginUserWithOtp>
```
vinaysshenoy: This is necessary for now because the `OtpSmsReceiver` is registered in the Android Manifest and we do not want to instantiate this class until the country is selected. We should figure out a better way to fix this.

app/src/main/java/org/simple/clinic/user/LoginUserWithOtp.kt
```diff
@@ -0,0 +1,106 @@
...
+  private fun syncOnLoginResult() {
+    dataSync
+        .get()
+        .syncTheWorld()
+        .subscribeOn(schedulersProvider.io())
+        .onErrorComplete()
+        .subscribe()
+  }
+
+  private fun <T : Any> readErrorResponseJson(error: HttpException, clazz: KClass<T>): T {
```
rakshakhegde: We should probably extract this function out into another Utility class as this is a general purpose function. Also while doing that it can be converted to a `reified` function that can take in just the generics without specifying the `KClass` explicitly.

app/src/main/java/org/simple/clinic/login/OtpSmsReceiver.kt
rakshakhegde: This should probably get fixed when we create the app and country scope right?

vinaysshenoy: That's the plan!

app/src/main/java/org/simple/clinic/user/LoginUserWithOtp.kt
vinaysshenoy: Good point, let me see if I can do this.

vinaysshenoy: Done.

PR #950, opened 2019-11-18
Refactor registration flows out of user session

vinaysshenoy:

### DIFF
--- a/app/src/main/java/org/simple/clinic/user/UserSession.kt
+++ b/app/src/main/java/org/simple/clinic/user/UserSession.kt
@@ -33,8 +33,6 @@ import org.simple.clinic.user.User.LoggedInStatus.RESET_PIN_REQUESTED
 import org.simple.clinic.user.User.LoggedInStatus.UNAUTHORIZED
 import org.simple.clinic.user.UserStatus.ApprovedForSyncing
 import org.simple.clinic.user.UserStatus.WaitingForApproval
-import org.simple.clinic.user.finduser.FindUserWithPhoneNumber
-import org.simple.clinic.user.refreshuser.RefreshCurrentUser
 import org.simple.clinic.util.Just
 import org.simple.clinic.util.None
 import org.simple.clinic.util.Optional
@@ -52,7 +50,6 @@ import javax.inject.Named
 class UserSession @Inject constructor(
     private val loginApi: LoginApi,
     private val registrationApi: RegistrationApi,
-    private val refreshCurrentUser: RefreshCurrentUser,
     private val facilityRepository: FacilityRepository,
     private val sharedPreferences: SharedPreferences,
     // This is Lazy to work around a cyclic dependency between
@@ -107,10 +104,6 @@ class UserSession @Inject constructor(
               loggedInStatus = NOT_LOGGED_IN)
           storeUser(user, entry.facilityId!!)
         }
-  }
-
-  fun refreshLoggedInUser(): Completable {
-    return refreshCurrentUser.refresh()
   }
 
   fun register(): Single<RegistrationResult> {
--- a/app/src/main/java/org/simple/clinic/user/UserSession.kt
+++ b/app/src/main/java/org/simple/clinic/user/UserSession.kt
@@ -17,9 +17,8 @@ import org.simple.clinic.facility.FacilityRepository
 import org.simple.clinic.forgotpin.ForgotPinResponse
 import org.simple.clinic.forgotpin.ResetPinRequest
 import org.simple.clinic.login.LoginApi
-import org.simple.clinic.login.LoginResponse
 import org.simple.clinic.patient.PatientRepository
-import org.simple.clinic.registration.FindUserResult
+import org.simple.clinic.user.finduser.FindUserResult
 import org.simple.clinic.registration.RegistrationApi
 import org.simple.clinic.registration.RegistrationRequest
 import org.simple.clinic.registration.RegistrationResponse
--- a/app/src/main/java/org/simple/clinic/user/UserSession.kt
+++ b/app/src/main/java/org/simple/clinic/user/UserSession.kt
@@ -18,7 +18,6 @@ import org.simple.clinic.forgotpin.ForgotPinResponse
 import org.simple.clinic.forgotpin.ResetPinRequest
 import org.simple.clinic.login.LoginApi
 import org.simple.clinic.patient.PatientRepository
-import org.simple.clinic.user.finduser.FindUserResult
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/simple/clinic/user/finduser/FindUserWithPhoneNumber.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.simple.clinic.user.finduser.FindUserResult.NetworkError
+import org.simple.clinic.user.finduser.FindUserResult.NotFound
+import org.simple.clinic.user.finduser.FindUserResult.UnexpectedError
+import retrofit2.HttpException
+import timber.log.Timber
+import java.io.IOException
+
+class FindUserWithPhoneNumber(private val registrationApi: RegistrationApi) {
+
+  fun find(phoneNumber: String): Single<FindUserResult> {
```
vinaysshenoy: This was copied from `UserSession#findExistingUser`.

app/src/main/java/org/simple/clinic/user/refreshuser/RefreshCurrentUser.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.simple.clinic.user.finduser.FindUserWithPhoneNumber
+import org.simple.clinic.util.mapType
+import timber.log.Timber
+
+class RefreshCurrentUser(
+    private val userDao: RoomDao,
+    private val findUserWithPhoneNumber: FindUserWithPhoneNumber
+) {
+
+  fun refresh(): Completable {
```
vinaysshenoy: This was copied from `UserSession#refreshLoggedInUser`

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -34,10 +33,13 @@ import org.simple.clinic.user.User.LoggedInStatus.RESET_PIN_REQUESTED
 import org.simple.clinic.user.User.LoggedInStatus.UNAUTHORIZED
 import org.simple.clinic.user.UserStatus.ApprovedForSyncing
 import org.simple.clinic.user.UserStatus.WaitingForApproval
+import org.simple.clinic.user.finduser.FindUserResult
+import org.simple.clinic.user.finduser.FindUserWithPhoneNumber
```
ragunathjawahar: Question, this naming abstracts away the information that we are making a network call (which is okay and is desirable in several cases). But, as convention dictates in the current project, should we not be using a more familiar convention. Second, should `FindUserWithPhoneNumber` be a noun, currently it sounds "verbish" like a method call. Can we call it `UserWithPhoneNumberFinder`?

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -108,27 +111,16 @@ class UserSession @Inject constructor(
...
+    return findUserWithPhoneNumber.find(phoneNumber)
   }
 
   fun refreshLoggedInUser(): Completable {
     return requireLoggedInUser()
         .firstOrError()
         .doOnSuccess { Timber.i("Refreshing logged-in user") }
         .flatMapCompletable { loggedInUser ->
-          registrationApi.findUser(loggedInUser.phoneNumber)
+          findUserWithPhoneNumber.find(loggedInUser.phoneNumber)
```
ragunathjawahar: Hrm… here we are finding a user by his phone number. Should this be,

`userFinder.findByPhoneNumber(String)` / `userApi.findByPhoneNumber(String)`?

app/src/main/java/org/simple/clinic/user/refreshuser/RefreshCurrentUser.kt
```diff
@@ -0,0 +1,68 @@
...
+    return Single.fromCallable { userDao.userImmediate() }
+        .doOnSuccess { Timber.i("Refreshing logged-in user") }
+        .flatMapCompletable { user ->
+          findUserWithPhoneNumber.find(user.phoneNumber)
+              .mapType<Found, LoggedInUserPayload> { it.user }
```
ragunathjawahar: Curious, are error cases handled by the caller?

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -53,12 +52,12 @@ import javax.inject.Named
 class UserSession @Inject constructor(
     private val loginApi: LoginApi,
     private val registrationApi: RegistrationApi,
-    private val findUserWithPhoneNumber: FindUserWithPhoneNumber,
+    private val refreshCurrentUser: RefreshCurrentUser,
```
ragunathjawahar: Not really sure with the direction we are moving towards, perhaps more context could help?

app/src/main/java/org/simple/clinic/user/refreshuser/RefreshCurrentUser.kt
vinaysshenoy: Currently, yes. In our current implementation, the only caller (`PatientsScreenController`) ignores any error that was generated since this is a fire-and-forget operation.

ragunathjawahar: Fire and forget isn't explicit?

vinaysshenoy: It's explicit at the usage site, i.e., in the `PatientsScreenController`.

app/src/main/java/org/simple/clinic/user/UserSession.kt
vinaysshenoy: Discussed offline: the purpose of this PR is to remove all network related code out of the `UserSession` in order to support the country specific endpoint. We will revisit the final design after doing a few more things:

- Breaking the DI graph into multiple scopes.
- Injecting the user directly into classes that require it instead of having them query it from the `UserSession`.

app/src/main/java/org/simple/clinic/user/UserSession.kt
vinaysshenoy: How about if I call it `UserLookup` instead? This way, I can change the method name to `findByPhoneNumber`, and this can become an ideal place to add future methods like `findByUserId` if needed.

app/src/main/java/org/simple/clinic/user/UserSession.kt
vinaysshenoy: See earlier comment about calling the class `UserLookup`?

PR #951, opened 2019-11-20
Refactor register user flow out of User session

vinaysshenoy:


app/src/main/java/org/simple/clinic/user/registeruser/RegisterUser.kt
```diff
@@ -0,0 +1,92 @@
...
+        .doOnSubscribe { Timber.i("Registering user") }
+        .map(::RegistrationRequest)
+        .flatMap(registrationApi::createUser)
+        .flatMap(::storeUserAndAccessToken)
+        .doOnSuccess(Analytics::setNewlyRegisteredUser)
+        .map { RegistrationResult.Success as RegistrationResult }
+        .onErrorReturn(::mapErrorToRegistrationResult)
+        .doOnSuccess { Timber.i("Registration result: $it") }
+
+  }
```
ragunathjawahar: WDYT?

```kotlin
  fun registerUserAtFacility(user: User, facility: Facility): Single<RegistrationResult> {
    val registrationRequest = RegistrationRequest(userToPayload(user, facility.uuid))
    return registrationApi
        .createUser(registrationRequest)
        .doOnSubscribe { Timber.i("Registering user") }
        .flatMap(::storeUserAndAccessToken)
        .doOnSuccess(Analytics::setNewlyRegisteredUser)
        .map { RegistrationResult.Success as RegistrationResult }
        .onErrorReturn(::mapErrorToRegistrationResult)
        .doOnSuccess { Timber.i("Registration result: $it") }
  }
```

app/src/main/java/org/simple/clinic/user/registeruser/RegisterUser.kt
```diff
@@ -0,0 +1,92 @@
...
+
+  private fun mapErrorToRegistrationResult(e: Throwable): RegistrationResult {
+    return when (e) {
+      is IOException -> RegistrationResult.NetworkError
+      else -> {
+        Timber.e(e)
+        RegistrationResult.UnexpectedError
+      }
+    }
+  }
```
ragunathjawahar:
```kotlin
  private fun mapErrorToRegistrationResult(e: Throwable): RegistrationResult {
    return when (e) {
      is IOException -> RegistrationResult.NetworkError
      else -> RegistrationResult.UnexpectedError
    }
  }
```

app/src/main/java/org/simple/clinic/user/registeruser/RegisterUser.kt
ragunathjawahar: I just made sure,
1. All the side-effect operators `doOn*` are at the same level.
2. Got rid of `Single.fromCallable()` and start with the meat `registrationApi.createUser(registrationRequest)`.

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -108,56 +107,15 @@ class UserSession @Inject constructor(
 
   fun register(): Single<RegistrationResult> {
     val user: Single<User> = loggedInUser()
-        .map { (user) -> user!! }
+        .filterAndUnwrapJust()
         .firstOrError()
         .cache()
 
     val currentFacility = user
-        .flatMap { facilityRepository.facilityUuidsForUser(it).firstOrError() }
```
ragunathjawahar: Curious, did the earlier case address the situation where a user may belong to multiple facilities? Trying to understand why this list of `UUIDs` were replaced with just the current facility?

app/src/test/java/org/simple/clinic/registration/register/RegistrationLoadingScreenControllerTest.kt
```diff
@@ -21,61 +19,84 @@ import org.simple.clinic.user.registeruser.RegistrationResult.UnexpectedError
 import org.simple.clinic.widgets.ScreenCreated
 import org.simple.clinic.widgets.UiEvent
 
-@RunWith(JUnitParamsRunner::class)
```
ragunathjawahar: Nice!

app/src/test/java/org/simple/clinic/registration/register/RegistrationLoadingScreenControllerTest.kt
```diff
@@ -21,61 +19,84 @@ import org.simple.clinic.user.registeruser.RegistrationResult.UnexpectedError
...
         .compose(controller)
         .subscribe { uiChange -> uiChange(screen) }
   }
 
-  @Parameters(method = "params for testing user registration")
   @Test
-  fun `when screen is created then user registration api should be called`(results: RegistrationResult) {
-    whenever(userSession.register()).doReturn(Single.just(results))
-    whenever(userSession.clearOngoingRegistrationEntry()).doReturn(Completable.complete())
+  fun `when retry button is clicked, then loader should show and register api should be called`() {
```
ragunathjawahar: The test implies that the loader should show. However, the verifications do not assert a loader being displayed. Am I missing something here?

app/src/test/java/org/simple/clinic/registration/register/RegistrationLoadingScreenControllerTest.kt
```diff
@@ -21,61 +19,84 @@ import org.simple.clinic.user.registeruser.RegistrationResult.UnexpectedError
...
+    whenever(userSession.register()).doReturn(Single.just<RegistrationResult>(Success))
+    whenever(userSession.clearOngoingRegistrationEntry()).doReturn(Completable.never())
```
ragunathjawahar: Curious, why a `Completable.never()` as opposed to a `Completable.complete()`?

app/src/test/java/org/simple/clinic/registration/register/RegistrationLoadingScreenControllerTest.kt
```diff
@@ -21,61 +19,84 @@ import org.simple.clinic.user.registeruser.RegistrationResult.UnexpectedError
...
-      }
-      NetworkError -> verify(screen).showNetworkError()
-      UnexpectedError -> verify(screen).showUnexpectedError()
-    }
+    verifyNoMoreInteractions(userSession)
+    verifyZeroInteractions(screen)
+  }
+
+  @Test
+  fun `when the register user call succeeds, then clear registration entry and go to home screen`() {
```
ragunathjawahar: The test implies that the registration entry should be cleared, but we don't have a verification to assert this. Are we missing something?

app/src/main/java/org/simple/clinic/registration/register/RegistrationLoadingScreenController.kt
```diff
@@ -31,9 +37,17 @@ class RegistrationLoadingScreenController @Inject constructor(
...
+              .loggedInUser()
+              .filterAndUnwrapJust()
+              .take(1)
+
+          val facilityToRegisterAt = savedUser
+              .flatMap { facilityRepository.currentFacility(it) }
+              .take(1)
+
+          Observables.combineLatest(savedUser, facilityToRegisterAt)
+              .flatMapSingle { (user, facility) -> registerUser.registerUserAtFacility(user, facility) }
```
ragunathjawahar: 👍

app/src/test/java/org/simple/clinic/registration/register/RegistrationLoadingScreenControllerTest.kt
```diff
@@ -36,34 +55,39 @@ class RegistrationLoadingScreenControllerTest {
 
   @Test
   fun `when retry button is clicked, then loader should show and register api should be called`() {
-    whenever(userSession.register()).doReturn(Single.just<RegistrationResult>(Success))
+    whenever(userSession.loggedInUser()) doReturn Observable.just<Optional<User>>(Just(user))
+    whenever(facilityRepository.currentFacility(user)) doReturn Observable.just(facility)
+    whenever(registerUser.registerUserAtFacility(user, facility)).doReturn(Single.just<RegistrationResult>(Success))
```
ragunathjawahar: `doReturn` instead of `.doReturn`? Mismatch in convention?

app/src/main/java/org/simple/clinic/user/UserSession.kt
vinaysshenoy: Ah, this is leftover from very old code. Initially, we left it to the user to decide which facility they wanted to be a part of so they would be select multiple facilities and we would send them all to the server.

However, this was later replaced by the server controlled facility group feature. The user would always register at only one facility and they would automatically get access to all facilities in that group. The api contracts were also changed to reflect this so the user registration api would accept a single facility ID instead of a list of facility IDs.

However, the local code still has some of this leftover behaviour (like being able to get a list of facility IDs for a user) which will always be a list of only one ID in reality, and we always use `.first()` or something to get the user's facility.

This goes one step towards removing that older code.

app/src/test/java/org/simple/clinic/registration/register/RegistrationLoadingScreenControllerTest.kt
vinaysshenoy: This is something explicit since the `openHomeScreen` method won't be invoked if the completable returned from `UserSession#clearOngoingRegistrationEntry` does not complete. But you're right about this not being explicit. I'll add a verification.

app/src/test/java/org/simple/clinic/registration/register/RegistrationLoadingScreenControllerTest.kt
vinaysshenoy: I didn't write this test right now, but I think the code got rearranged to show this as a change. I'll take a look at this test and see what's up.

app/src/test/java/org/simple/clinic/registration/register/RegistrationLoadingScreenControllerTest.kt
vinaysshenoy: Let me recheck this. I don't exactly remember why I made this a `Completable.never`.

app/src/main/java/org/simple/clinic/user/registeruser/RegisterUser.kt
vinaysshenoy: Looks good, I'll use this.

app/src/test/java/org/simple/clinic/registration/register/RegistrationLoadingScreenControllerTest.kt
vinaysshenoy: `doReturn` is an infix function. I've generally been using `doReturn` in the infix style when writing new stubs but using the normal method style when converting older `thenReturn` statements. I can change this to match the convention.

I could change all of them to use the infix style, or change the newer infix calls to the normal method styles. Do you have any particular preference?

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -105,19 +100,6 @@ class UserSession @Inject constructor(
         }
   }
 
-  fun register(): Single<RegistrationResult> {
```
ragunathjawahar: 👌

app/src/androidTest/java/org/simple/clinic/rules/ServerAuthenticationRule.kt
```diff
@@ -165,15 +173,12 @@ class ServerAuthenticationRule : TestRule {
   }
 
   private fun registerUserAtFacility(facility: Facility): RegistrationResult {
-    val registrationEntry = testData.ongoingRegistrationEntry(
-        phoneNumber = faker.number.number(10),
-        pin = testData.qaUserPin(),
-        registrationFacility = facility)
-
-    return userSession.saveOngoingRegistrationEntry(registrationEntry)
```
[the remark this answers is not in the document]
vinaysshenoy: Yeah, I was surprised myself just by how deeply coupled all this was. 😓

app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreenController.kt
```diff
@@ -18,7 +18,7 @@ import org.simple.clinic.user.LoggedInUserPayload
 import org.simple.clinic.user.OngoingLoginEntry
 import org.simple.clinic.user.OngoingRegistrationEntry
 import org.simple.clinic.user.UserSession
-import org.simple.clinic.user.finduser.FindUserWithPhoneNumber
+import org.simple.clinic.user.finduser.UserLookup
```
ragunathjawahar: 👌

app/src/test/java/org/simple/clinic/registration/register/RegistrationLoadingScreenControllerTest.kt
ragunathjawahar: I don't really have a preference. Just the first two calls have an `infix` syntax and the third one has the `.` syntax. We could just make them consistent.

app/src/main/java/org/simple/clinic/user/registeruser/RegisterUser.kt
vinaysshenoy: I could use `.doOnError(Timber::e)` because that changes the behaviour. Earlier, we would not report `IOException` errors since those are expected, but this would end up reporting those as well.

I added another method to report errors which would filter them out before reporting.

app/src/test/java/org/simple/clinic/registration/register/RegistrationLoadingScreenControllerTest.kt
vinaysshenoy: The `showLoader` function is being triggered on the screen via a side-effect when the retry button is clicked, but the function is private. I am guessing this was initially a part of the controller but was changed later without updating the test.

Will reword the test description to match.

app/src/test/java/org/simple/clinic/registration/register/RegistrationLoadingScreenControllerTest.kt
vinaysshenoy: The way the implementation is designed, clearing of ongoing entry does not impact going to the home screen after registration happens.

I had initially deleted the `Completable.complete` stub, but this broke the test because it needed a `Completable` to be returned from the `UserSession#clearOngoingRegistrationEntry` method.

I'll change this back to `Completable.complete` since it doesn't actually affect anything.

PR #952, opened 2019-11-21
Begin adding configurable fields for new patient entry screen

ragunathjawahar:
This is an evolving design at the moment and contains duplication that will be addressed in the future PRs.

### DIFF
--- a/app/src/main/java/org/simple/clinic/newentry/form/DateOfBirthField.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/form/DateOfBirthField.kt
@@ -1,0 +1,20 @@
+package org.simple.clinic.newentry.form
+
+import org.simple.clinic.newentry.form.ValidationError.InvalidDateFormat
+import org.simple.clinic.newentry.form.ValidationError.MissingValue
+import org.threeten.bp.LocalDate
+import org.threeten.bp.format.DateTimeParseException
+
+class DateOfBirthField : InputField<String>() {
+  override fun validate(value: String): List<ValidationError> {
+    if (value.isBlank()) return listOf(MissingValue)
+
+    try {
+      LocalDate.parse(value)
+    } catch (e: DateTimeParseException) {
+      return listOf(InvalidDateFormat)
+    }
+
+    TODO()
+  }
+}
--- a/app/src/main/java/org/simple/clinic/newentry/form/DateOfBirthField.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/form/DateOfBirthField.kt
@@ -14,13 +14,15 @@ class DateOfBirthField(
   override fun validate(value: String): List<ValidationError> {
     if (value.isBlank()) return listOf(MissingValue)
 
-    try {
+    return try {
       val parsedDate = LocalDate.parse(value, dateTimeFormatter)
-      if (parsedDate > today) return listOf(DateIsInFuture)
+      if (parsedDate > today) {
+        listOf(DateIsInFuture)
+      } else {
+        emptyList()
+      }
     } catch (e: DateTimeParseException) {
-      return listOf(InvalidDateFormat)
+      listOf(InvalidDateFormat)
     }
-
-    TODO()
   }
 }
--- a/app/src/main/java/org/simple/clinic/newentry/form/DateOfBirthField.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/form/DateOfBirthField.kt
@@ -1,16 +1,22 @@
 package org.simple.clinic.newentry.form
 
+import org.simple.clinic.newentry.form.ValidationError.DateIsInFuture
 import org.simple.clinic.newentry.form.ValidationError.InvalidDateFormat
 import org.simple.clinic.newentry.form.ValidationError.MissingValue
 import org.threeten.bp.LocalDate
+import org.threeten.bp.format.DateTimeFormatter
 import org.threeten.bp.format.DateTimeParseException
 
-class DateOfBirthField : InputField<String>() {
+class DateOfBirthField(
+    private val dateTimeFormatter: DateTimeFormatter,
+    private val today: LocalDate
+) : InputField<String>() {
   override fun validate(value: String): List<ValidationError> {
     if (value.isBlank()) return listOf(MissingValue)
 
     try {
-      LocalDate.parse(value)
+      val parsedDate = LocalDate.parse(value, dateTimeFormatter)
+      if (parsedDate > today) return listOf(DateIsInFuture)
     } catch (e: DateTimeParseException) {
       return listOf(InvalidDateFormat)
     }
--- a/app/src/main/java/org/simple/clinic/newentry/form/GenderField.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/form/GenderField.kt
@@ -1,10 +1,10 @@
 package org.simple.clinic.newentry.form
 
-import org.simple.clinic.newentry.form.ValidationError.FieldIsBlankEmpty
+import org.simple.clinic.newentry.form.ValidationError.MissingValue
 import org.simple.clinic.patient.Gender
 
 class GenderField : InputField<Gender?>() {
   override fun validate(value: Gender?): List<ValidationError> {
-    return if (value == null) listOf(FieldIsBlankEmpty) else emptyList()
+    return if (value == null) listOf(MissingValue) else emptyList()
   }
 }
--- a/app/src/main/java/org/simple/clinic/newentry/form/GenderField.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/form/GenderField.kt
@@ -1,0 +1,10 @@
+package org.simple.clinic.newentry.form
+
+import org.simple.clinic.newentry.form.ValidationError.FieldIsBlankEmpty
+import org.simple.clinic.patient.Gender
+
+class GenderField : InputField<Gender?>() {
+  override fun validate(value: Gender?): List<ValidationError> {
+    return if (value == null) listOf(FieldIsBlankEmpty) else emptyList()
+  }
+}
--- a/app/src/main/java/org/simple/clinic/newentry/form/InputField.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/form/InputField.kt
@@ -1,0 +1,3 @@
+package org.simple.clinic.newentry.form
+
+abstract class InputField
--- a/app/src/main/java/org/simple/clinic/newentry/form/InputField.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/form/InputField.kt
@@ -1,3 +1,5 @@
 package org.simple.clinic.newentry.form
 
-abstract class InputField
+abstract class InputField<T> {
+  abstract fun validate(value: T): List<ValidationError>
+}
--- a/app/src/test/java/org/simple/clinic/newentry/form/DateOfBirthFieldTest.kt
+++ b/app/src/test/java/org/simple/clinic/newentry/form/DateOfBirthFieldTest.kt
@@ -1,0 +1,22 @@
+package org.simple.clinic.newentry.form
+
+import com.google.common.truth.Truth.assertThat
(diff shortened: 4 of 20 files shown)


app/src/main/java/org/simple/clinic/newentry/form/GenderField.kt
```diff
@@ -0,0 +1,10 @@
+package org.simple.clinic.newentry.form
+
+import org.simple.clinic.newentry.form.ValidationError.FieldIsBlankEmpty
+import org.simple.clinic.patient.Gender
+
+class GenderField : InputField<Gender?>() {
```
vinaysshenoy: Why is this a `nullable` type?

app/src/main/java/org/simple/clinic/newentry/form/InputField.kt
```diff
@@ -1,3 +1,5 @@
 package org.simple.clinic.newentry.form
 
-abstract class InputField
+abstract class InputField<T> {
```
vinaysshenoy: Should this be an interface instead?

app/src/main/java/org/simple/clinic/newentry/form/GenderField.kt
vinaysshenoy: If this is to handle the case where the patient's gender has not yet been selected, shouldn't this be handled at the form level rather than here?

app/src/main/java/org/simple/clinic/newentry/form/InputField.kt
```diff
@@ -1,3 +1,5 @@
...
+  abstract fun validate(value: T): List<ValidationError>
```
vinaysshenoy: Should this be returning `Set<ValidationError>` instead?

app/src/main/java/org/simple/clinic/newentry/form/DateOfBirthField.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.simple.clinic.newentry.form.ValidationError.MissingValue
+import org.threeten.bp.LocalDate
+import org.threeten.bp.format.DateTimeParseException
+
+class DateOfBirthField : InputField<String>() {
+  override fun validate(value: String): List<ValidationError> {
+    if (value.isBlank()) return listOf(MissingValue)
+
+    try {
+      LocalDate.parse(value)
```
vinaysshenoy: 👏

vinaysshenoy: My concern with doing this in this way is that we are tying the entered value (which will generally be entered by a user) to the date format which `LocalDate` expects, which is `yyyy-MM-dd` which may not be what user's are used to.

We should try to change this method to accept a `org.threeten.bp.format.DateTimeFormatter` instead so that we can decouple the format from the validation.

app/src/test/java/org/simple/clinic/newentry/form/DateOfBirthFieldTest.kt
```diff
@@ -0,0 +1,22 @@
+package org.simple.clinic.newentry.form
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+import org.simple.clinic.newentry.form.ValidationError.InvalidDateFormat
+import org.simple.clinic.newentry.form.ValidationError.MissingValue
+
+class DateOfBirthFieldTest {
```
vinaysshenoy: Should we also add a test that entering a date in a valid format does not throw any validation error?

app/src/main/java/org/simple/clinic/newentry/form/InputField.kt
ragunathjawahar: I don't know that yet,
1. I might pull out all the validation logic outside the classes and use just composition instead of inheritance.
2. There are special cases around validation across multiple fields and country-specific fields.

The way I am approaching this problem right now is to build it for the current form and then scale it across other countries. I do have a high-level design of how this is going to look like, but we have to wait a see in a couple of PRs.

app/src/main/java/org/simple/clinic/newentry/form/InputField.kt
ragunathjawahar: Yes, you are right. Will make this change in the upcoming PR?

app/src/main/java/org/simple/clinic/newentry/form/DateOfBirthField.kt
ragunathjawahar: Yes, that is happening in one of the following commits.

app/src/test/java/org/simple/clinic/newentry/form/DateOfBirthFieldTest.kt
ragunathjawahar: Yes, that is happening in one of the upcoming commits.

app/src/test/java/org/simple/clinic/newentry/form/DateOfBirthFieldTest.kt
```diff
@@ -2,11 +2,17 @@ package org.simple.clinic.newentry.form
...
 import org.simple.clinic.newentry.form.ValidationError.InvalidDateFormat
 import org.simple.clinic.newentry.form.ValidationError.MissingValue
+import org.threeten.bp.LocalDate
+import org.threeten.bp.format.DateTimeFormatter
 
 class DateOfBirthFieldTest {
-  private val dateOfBirthField = DateOfBirthField()
+  private val dateOfBirthField = DateOfBirthField(
+      DateTimeFormatter.ofPattern("dd/MM/yyyy"),
+      LocalDate.of(2019, 11, 20)
```
vinaysshenoy: I generally prefer to use `LocalDate.parse("2019-11-20")` since I feel the timestamp version is more human readable as a standard date format. There's a live template, `date`, in the plugin which should make adding this easier.

app/src/test/java/org/simple/clinic/newentry/form/DateOfBirthFieldTest.kt
```diff
@@ -19,4 +25,10 @@ class DateOfBirthFieldTest {
     assertThat(dateOfBirthField.validate("7/6"))
         .containsExactly(InvalidDateFormat)
   }
+
+  @Test
+  fun `it returns date is in future when date of birth is in the future`() {
+    assertThat(dateOfBirthField.validate("12/12/2020"))
```
vinaysshenoy: When it comes to tests like these, I prefer to keep the date exactly one minimal unit after the limit to verify that it's working as expected.

i.e, when verifying that a given `LocalDate` (say `a`) is not more than another `LocalDate` (say `b`), I change `a` to be exactly one day after `b`. If it was an `Instant`, it would be 1 millisecond instead.

And then have other tests to verify that the check succeeds on the same/earlier timestamp. What do you think?

app/src/main/java/org/simple/clinic/newentry/form/DateOfBirthField.kt
```diff
@@ -14,13 +14,15 @@ class DateOfBirthField(
   override fun validate(value: String): List<ValidationError> {
     if (value.isBlank()) return listOf(MissingValue)
```
vinaysshenoy: I think this entire code block could be a lot simpler if it used a `MutableList` internally and converted to a `List` at the point of return. What do you think?

app/src/main/java/org/simple/clinic/newentry/form/InputField.kt
vinaysshenoy: Works.

app/src/test/java/org/simple/clinic/newentry/form/DistrictFieldTest.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import com.google.common.truth.Truth
+import org.junit.Test
+
+class DistrictFieldTest {
+  private val districtField = DistrictField()
+
+  @Test
+  fun `it returns a missing value error when district is empty`() {
+    Truth.assertThat(districtField.validate(""))
```
vinaysshenoy: Add an import for `Truth#assertThat`?

app/src/test/java/org/simple/clinic/newentry/form/StreetAddressFieldTest.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.google.common.truth.Truth
+import org.junit.Test
+import org.simple.clinic.newentry.form.ValidationError.MissingValue
+
+class StreetAddressFieldTest {
+  private val streetAddressField = StreetAddressField()
+
+  @Test
+  fun `it returns a missing value error when street address is empty`() {
+    Truth.assertThat(streetAddressField.validate(""))
```
vinaysshenoy: Add an import for `Truth#assertThat`?

app/src/main/java/org/simple/clinic/newentry/form/DateOfBirthField.kt
ragunathjawahar: We aren't accumulating errors, this is what I came up with.

```kotlin
    override fun validate(value: String): List<ValidationError> {
    val validationErrors = mutableListOf<ValidationError>()
    if (value.isBlank()) validationErrors.add(MissingValue)

    try {
      val parsedDate = LocalDate.parse(value, dateTimeFormatter)
      if (parsedDate > today) {
        validationErrors.add(DateIsInFuture)
      }
    } catch (e: DateTimeParseException) {
      validationErrors.add(InvalidDateFormat)
    }

    return validationErrors.toList()
  }
```

I guess we could wait until we are returning more than one error. WDYT?

app/src/test/java/org/simple/clinic/newentry/form/DistrictFieldTest.kt
ragunathjawahar: Fixed.

app/src/test/java/org/simple/clinic/newentry/form/StreetAddressFieldTest.kt
ragunathjawahar: Fixed.

app/src/main/java/org/simple/clinic/newentry/form/DateOfBirthField.kt
vinaysshenoy: Yeah, not something we need to do right now.

app/src/main/java/org/simple/clinic/newentry/form/GenderField.kt
ragunathjawahar: Clarification, are you talking about the impending `Form` object or the UI layer? If we are discussing the former, it will show up in an upcoming PR. If we are discussing the later, then we longer have a single-source of truth.

app/src/test/java/org/simple/clinic/newentry/form/DateOfBirthFieldTest.kt
ragunathjawahar: Yes, that is how it looks like in one of the upcoming commits,
`LocalDate.parse("20/11/2019", dateTimeFormatter)`

app/src/test/java/org/simple/clinic/newentry/form/DateOfBirthFieldTest.kt
ragunathjawahar: That makes sense. Fixed it!

app/src/main/java/org/simple/clinic/newentry/form/GenderField.kt
vinaysshenoy: I meant the `Form` object.

vinaysshenoy: I am assuming that the `GenderField` will become an `InputField<Gender>`, and the impending `PatientEntryForm` object should potentially have either an `GenderField?` or an `Optional<GenderField>` property.

Does this make sense?

PR #955, opened 2019-11-22
Refactor syncing and clearing of patient data out of UserSession

vinaysshenoy:

### DIFF
--- a/app/src/main/java/org/simple/clinic/forgotpin/confirmpin/ForgotPinConfirmPinScreenController.kt
+++ b/app/src/main/java/org/simple/clinic/forgotpin/confirmpin/ForgotPinConfirmPinScreenController.kt
@@ -28,7 +28,6 @@ typealias UiChange = (Ui) -> Unit
 class ForgotPinConfirmPinScreenController @Inject constructor(
     private val userSession: UserSession,
     private val facilityRepository: FacilityRepository,
-    private val patientRepository: PatientRepository,
     private val resetUserPin: ResetUserPin,
     private val syncAndClearPatientData: SyncAndClearPatientData
 ) : ObservableTransformer<UiEvent, UiChange> {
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/simple/clinic/forgotpin/confirmpin/ForgotPinConfirmPinScreenController.kt
```diff
@@ -102,7 +101,7 @@ class ForgotPinConfirmPinScreenController @Inject constructor(
         .share()
 
     val makeResetPinCall = validPin
-        .flatMapSingle { newPin -> syncAndClearPatientData().toSingleDefault(newPin) }
+        .flatMapSingle { newPin -> syncAndClearPatientData.run().toSingleDefault(newPin) }
```
ragunathjawahar: Curious, why do we have to clear patient data while resetting the user's pin?

app/src/test/java/org/simple/clinic/user/clearpatientdata/SyncAndClearPatientDataTest.kt
```diff
@@ -34,20 +34,6 @@ class SyncAndClearPatientDataTest {
   private val prescriptionPullToken = mock<Preference<Optional<String>>>()
   private val medicalHistoryPullToken = mock<Preference<Optional<String>>>()
 
-  private val syncAndClearPatientData = SyncAndClearPatientData(
```
ragunathjawahar: Curious, why are we doing this?

app/src/main/java/org/simple/clinic/login/LoginUserWithOtp.kt
```diff
@@ -22,7 +22,7 @@ import javax.inject.Named
 
 class LoginUserWithOtp @Inject constructor(
     private val loginApi: LoginApi,
-    private val dataSync: Lazy<DataSync>,
```
ragunathjawahar: 🚀

app/src/main/java/org/simple/clinic/forgotpin/confirmpin/ForgotPinConfirmPinScreenController.kt
vinaysshenoy: This is a product decision. We don't know if the user who is attempting to reset the PIN is a legitimate user who has genuinely forgotten their PIN, or someone who is trying to reset the PIN to get access to the data on the phone. To be safe, we clear the patient data stored locally when a user tries to reset their a PIN.

This is expected to be paired with a phone call from the user's supervisor to verify that they are the real user who is trying to reset the PIN before they get access to the patient data again.

app/src/test/java/org/simple/clinic/user/clearpatientdata/SyncAndClearPatientDataTest.kt
vinaysshenoy: In a later commit, I need to override one of the properties passed to the constructor. So I made this change.

PR #956, opened 2019-11-22
Change overdue appointments list

sanchita:
https://www.pivotaltracker.com/story/show/169913323


app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
```diff
@@ -133,6 +133,7 @@ class AppointmentRepository @Inject constructor(
                   facilityUuid = facility.uuid,
                   scheduledStatus = Scheduled,
                   scheduledBefore = since,
+                  scheduledAfter = since.minusMonths(12),
```
vinaysshenoy: I don't think the repository is the right place to make business logic discussions, and this should be configured via the `AppointmentConfig`. Can we add this as a property to the config and read it from there?

PR #960, opened 2019-11-25
Simplify querying of patient data in Patient Summary Screen controller

vinaysshenoy:
https://www.pivotaltracker.com/story/show/169944431


app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -562,4 +519,29 @@ class PatientSummaryScreenController @Inject constructor(
 
     return durationSinceBpCreated <= bpEditableFor
   }
+
+  @WorkerThread
```
ragunathjawahar: Curious, does this throw a runtime exception if invoked from the main thread?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -544,4 +535,14 @@ class PatientSummaryScreenController @Inject constructor(
           .or(medicalHistoryChangedSince)
     }.blockingGet()
   }
+
+  @WorkerThread
```
ragunathjawahar: We are already using strict mode and room to ensure these functions are not called on the main thread, do we still need the annotation?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
vinaysshenoy: Nope, it's just an annotation to Android Lint which will show IDE warnings if the method is inferred to be running on the main thread.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
vinaysshenoy: This is a good point, I just started adding them out of habit since this is no longer running on Room's internal thread pool.

I still would like to start adding it out of habit since we might add methods that do not use Room in the repository (say file I/O), and would want this to be a convention that is followed.

PR #964, opened 2019-11-28
Update default homescreen illustrations

rakshakhegde:
https://www.pivotaltracker.com/story/show/168980970


app/src/main/java/org/simple/clinic/appconfig/Country.kt
```diff
@@ -21,4 +21,11 @@ data class Country(
 
     @Json(name = "isd_code")
     val isdCode: String
-) : Parcelable
+) : Parcelable {
+
+    companion object {
+        const val INDIA = "IN"
+        const val BANGLADESH = "BD"
+        const val ETHIOPIA = "ET"
```
sanchita: Are we reusing these constants elsewhere?
If not, then we can inline them instead of creating global values.

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
```diff
@@ -215,8 +229,9 @@ open class PatientsScreen(context: Context, attrs: AttributeSet) : RelativeLayou
     // Hard-coding to show this simple video view exists because, as of now,
     // we are not sure if we will have variations of this training video.
     // We should make the title, duration and video thumbnail configurable in order to improve this.
-    simpleVideoDuration.text = resources.getString(R.string.simple_video_duration, "5:07")
-    showHomeScreenBackground(R.id.simpleVideoLayout)
+    //    simpleVideoDuration.text = resources.getString(R.string.simple_video_duration, "5:07")
+    //    showHomeScreenBackground(R.id.simpleVideoLayout)
```
sanchita: Is this code commented out by mistake?

app/src/main/java/org/simple/clinic/appconfig/Country.kt
rakshakhegde: I have seen Ragunath's PR too and this will be reused in other places too. Also, I find it nicer to have literal constants stored in one place than freely mentioning them all over our code?

app/src/main/java/org/simple/clinic/home/patients/PatientsScreen.kt
rakshakhegde: Whoops yes. Will fix this.

rakshakhegde: Done

app/src/main/java/org/simple/clinic/appconfig/Country.kt
sanchita: It make sense to do this once we need to reuse it not before. Let's not code ahead.
We can keep this as is for now because this should change once we come up with a way to set all country-specific configuration in one place.

sanchita: I have created a pivotol story to track this: https://www.pivotaltracker.com/story/show/170045234

PR #970, opened 2019-12-02
Refactor BP view on Patient Summary screen into its own widget

vinaysshenoy:


vinaysshenoy:
Closing this because I am not happy with the direction of the refactoring. Will try a different approach.

PR #975, opened 2019-12-04
Introduce Assisted Injection for effect handlers

vinaysshenoy:
Until now, we were manually creating effect handlers in the screen when connecting the Mobius loop. This is cumbersome since it requires us to inject the dependencies which the effect handler requires into the screen first and then construct the effect handler.

We should inject the effect handlers directly, but in most cases, the effect handler also needs dynamic dependencies at runtime (for example, a reference to the view to perform UI actions), which prevents us from doing constructor injection.

This PR uses [AssistedInject](https://github.com/square/AssistedInject) as a way to reduce this boilerplate around constructing effect handlers. This PR does not convert all the effect handlers at the moment, only one for an `Activity` and another for a `Screen` to show how this integration would work.

### DIFF
--- a/app/src/main/java/org/simple/clinic/setup/SetupActivity.kt
+++ b/app/src/main/java/org/simple/clinic/setup/SetupActivity.kt
@@ -5,7 +5,6 @@ import android.content.Intent
 import android.os.Bundle
 import android.view.WindowManager
 import androidx.appcompat.app.AppCompatActivity
-import com.f2prateek.rx.preferences2.Preference
 import io.github.inflationx.viewpump.ViewPumpContextWrapper
 import io.reactivex.Observable
 import org.simple.clinic.BuildConfig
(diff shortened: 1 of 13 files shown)


app/src/main/java/org/simple/clinic/setup/SetupActivity.kt
```diff
@@ -29,39 +26,24 @@ import org.simple.clinic.router.screen.NestedKeyChanger
...
+  private lateinit var component: SetupActivityComponent
 
   @Inject
-  lateinit var appConfigRepository: AppConfigRepository
-
-  @field:[Inject Named("fallback")]
-  lateinit var fallbackCountry: Country
-
-  private lateinit var component: SetupActivityComponent
+  lateinit var effectHandler: SetupActivityEffectHandler.Factory
```
sanchita: Rename to `effectHandlerFactory`

sanchita:
Left one small comment.
Rest of the PR lgtm.

app/src/main/java/org/simple/clinic/setup/SetupActivity.kt
vinaysshenoy: Done.

PR #977, opened 2019-12-05
Setup Patient transfer facility change activity

sanchita:
Pivotol story:
https://www.pivotaltracker.com/story/show/169753540

This PR has only the setup related code for the new facility change activity `PatientFacilityChangeActivity`. It does not implement any new functionality yet.
Also, the PR is largish because of copying code from `FacilityChangeScreen`

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -87,7 +87,10 @@
       android:name=".home.overdue.phonemask.PhoneMaskBottomSheet"
       android:theme="@style/Clinic.V2.Theme.BottomSheetActivity"
       android:windowSoftInputMode="adjustResize" />
-    <activity android:name=".scheduleappointment.patientFacilityTransfer.PatientFacilityChangeActivity" />
+
+    <activity
+      android:name=".scheduleappointment.patientFacilityTransfer.PatientFacilityChangeActivity"
+      android:windowSoftInputMode="adjustResize" />
 
     <receiver
       android:name=".login.OtpSmsReceiver"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -87,7 +87,7 @@
       android:name=".home.overdue.phonemask.PhoneMaskBottomSheet"
       android:theme="@style/Clinic.V2.Theme.BottomSheetActivity"
       android:windowSoftInputMode="adjustResize" />
-    <activity android:name=".scheduleappointment.PatientFacilityChangeActivity" />
+    <activity android:name=".scheduleappointment.patientFacilityTransfer.PatientFacilityChangeActivity" />
 
     <receiver
       android:name=".login.OtpSmsReceiver"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -89,7 +89,7 @@
       android:windowSoftInputMode="adjustResize" />
 
     <activity
-      android:name=".scheduleappointment.patientFacilityTransfer.PatientFacilityChangeActivity"
+      android:name=".scheduleappointment.facilityselection.PatientFacilityChangeActivity"
       android:windowSoftInputMode="adjustResize" />
 
     <receiver
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -87,6 +87,7 @@
       android:name=".home.overdue.phonemask.PhoneMaskBottomSheet"
       android:theme="@style/Clinic.V2.Theme.BottomSheetActivity"
       android:windowSoftInputMode="adjustResize" />
+    <activity android:name=".scheduleappointment.PatientFacilityChangeActivity" />
 
     <receiver
       android:name=".login.OtpSmsReceiver"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -89,7 +89,7 @@
       android:windowSoftInputMode="adjustResize" />
 
     <activity
-      android:name=".scheduleappointment.facilityselection.PatientFacilityChangeActivity"
+      android:name=".scheduleappointment.facilityselection.FacilitySelectionActivity"
       android:windowSoftInputMode="adjustResize" />
 
     <receiver
--- a/app/src/main/java/org/simple/clinic/di/AppComponent.kt
+++ b/app/src/main/java/org/simple/clinic/di/AppComponent.kt
@@ -5,7 +5,7 @@ import org.simple.clinic.ClinicApp
 import org.simple.clinic.login.OtpSmsReceiver
 import org.simple.clinic.main.TheActivityComponent
 import org.simple.clinic.newentry.clearbutton.ClearFieldImageButton
-import org.simple.clinic.scheduleappointment.facilityselection.PatientFacilityChangeComponent
+import org.simple.clinic.scheduleappointment.facilityselection.FacilitySelectionActivityComponent
 import org.simple.clinic.setup.SetupActivityComponent
 import org.simple.clinic.sync.DataSync
 import org.simple.clinic.sync.SyncWorker
@@ -25,7 +25,7 @@ interface AppComponent {
 
   fun theActivityComponentBuilder(): TheActivityComponent.Builder
   fun setupActivityComponentBuilder(): SetupActivityComponent.Builder
-  fun patientFacilityChangeComponentBuilder() : PatientFacilityChangeComponent.Builder
+  fun patientFacilityChangeComponentBuilder() : FacilitySelectionActivityComponent.Builder
 }
 
 @Scope
--- a/app/src/main/java/org/simple/clinic/di/AppComponent.kt
+++ b/app/src/main/java/org/simple/clinic/di/AppComponent.kt
@@ -5,7 +5,7 @@ import org.simple.clinic.ClinicApp
 import org.simple.clinic.login.OtpSmsReceiver
 import org.simple.clinic.main.TheActivityComponent
 import org.simple.clinic.newentry.clearbutton.ClearFieldImageButton
-import org.simple.clinic.scheduleappointment.PatientFacilityChangeActivity
+import org.simple.clinic.scheduleappointment.patientFacilityTransfer.PatientFacilityChangeActivity
 import org.simple.clinic.setup.SetupActivityComponent
 import org.simple.clinic.sync.DataSync
 import org.simple.clinic.sync.SyncWorker
--- a/app/src/main/java/org/simple/clinic/di/AppComponent.kt
+++ b/app/src/main/java/org/simple/clinic/di/AppComponent.kt
@@ -5,8 +5,7 @@ import org.simple.clinic.ClinicApp
 import org.simple.clinic.login.OtpSmsReceiver
 import org.simple.clinic.main.TheActivityComponent
 import org.simple.clinic.newentry.clearbutton.ClearFieldImageButton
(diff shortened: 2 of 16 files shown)


app/src/main/res/layout/screen_patient_facility_change.xml
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
```
vinaysshenoy: Can we squash the commit with a later commit (`Add UI functions to PatientFacilityChangeActivity`) since this change seems to have been reverted in that commit.

app/src/main/AndroidManifest.xml
```diff
@@ -87,7 +87,7 @@
       android:name=".home.overdue.phonemask.PhoneMaskBottomSheet"
       android:theme="@style/Clinic.V2.Theme.BottomSheetActivity"
       android:windowSoftInputMode="adjustResize" />
-    <activity android:name=".scheduleappointment.PatientFacilityChangeActivity" />
+    <activity android:name=".scheduleappointment.patientFacilityTransfer.PatientFacilityChangeActivity" />
```
vinaysshenoy: Accidental break in package naming conventions? (`patientFacilityTransfer` is in camel-case, which is not the usual convention).

app/src/main/java/org/simple/clinic/scheduleappointment/patientFacilityTransfer/PatientFacilityChangeActivity.kt
```diff
@@ -21,8 +22,31 @@ class PatientFacilityChangeActivity : AppCompatActivity() {
   }
 
   override fun attachBaseContext(baseContext: Context) {
-    ClinicApp.appComponent.inject(this)
```
vinaysshenoy: If we remove this line, the class will break at runtime at this commit since `locale` would not have been initialized. This is fine for now since the activity is not being launched, though.

app/src/main/java/org/simple/clinic/scheduleappointment/patientFacilityTransfer/PatientFacilityChangeActivity.kt
```diff
@@ -16,14 +19,27 @@ class PatientFacilityChangeActivity : AppCompatActivity() {
...
+    component = ClinicApp
+        .appComponent
+        .patientFacilityChangeComponentBuilder()
+        .activity(this)
+        .build()
+    component.inject(this)
+
+    val wrappedContext = baseContext
+        .wrap { LocaleOverrideContextWrapper.wrap(it, locale) }
+        .wrap { InjectorProviderContextWrapper.wrap(it, component) }
```
vinaysshenoy: This context wrapper is only necessary if there are nested screens/views within `PatientFacilityChangeActivity` that need to be field-injected. Since this activity does not do this at the moment, we can remove this.

app/src/main/java/org/simple/clinic/di/AppComponent.kt
```diff
@@ -22,10 +23,10 @@ interface AppComponent {
   fun inject(target: OtpSmsReceiver)
   fun inject(target: DataSync)
   fun inject(target: BottomSheetActivity)
-  fun inject(target: PatientFacilityChangeActivity)
 
   fun theActivityComponentBuilder(): TheActivityComponent.Builder
   fun setupActivityComponentBuilder(): SetupActivityComponent.Builder
+  fun patientFacilityChangeComponentBuilder() : PatientFacilityChangeComponent.Builder
```
vinaysshenoy: Can we rename this to retain the convention of the rest of the components?

`PatientFacilityChangeComponent` -> `PatientFacilityChangeActivityComponent`?
`patientFacilityChangeComponentBuilder` -> `patientFacilityChangeActivityComponentBuilder`

app/src/main/java/org/simple/clinic/scheduleappointment/patientFacilityTransfer/PatientFacilityChangeActivity.kt
```diff
@@ -1,29 +1,70 @@
...
-  
+
+  @Inject
+  lateinit var controller: PatientFacilityChangeController
+
+  private val recyclerViewAdapter = FacilitiesAdapter()
+
+  private val onDestroys = PublishSubject.create<ScreenDestroyed>()
+
+  private val screenResults = ScreenResultBus()
```
vinaysshenoy: The primary use of this class is to forward activity and permission results to screens when they are being managed with a `ScreenRouter`. Since this activity does not manage multiple screens, this is not necessary.

app/src/main/AndroidManifest.xml
sanchita: Oops. Will fix this.

app/src/main/java/org/simple/clinic/scheduleappointment/patientFacilityTransfer/PatientFacilityChangeActivity.kt
sanchita: Good catch!

vinaysshenoy:
@sanchita I also had a thought about the overall feature:

This is yet another screen we will have to migrate to Mobius at some point. I don't want to introduce more code that uses the older architecture, but it doesn't seem like it's avoidable at the moment. In addition, every change that happens in `FacilityChangeScreen` should also be duplicated here, which is painful and added maintenance overhead.

I have a suggestion to fix this in the long term:
1. Change `PatientsScreen` to use this screen itself for changing the user facility, i.e, open this screen and get a result.
2. Observe that result in the `PatientsScreen` and update the user's facility in `PatientsScreenController` instead of doing it in `FacilityChangeScreenController`.
3. Anything extra that `FacilityChangeScreen` does can be moved to the `PatientsScreenController`.
4. Remove the `FacilityChangeScreen`.

What do you think? This will let us remove the maintenance overhead of having to migrate another screen and maintaining feature parity between these two screens.

app/src/main/java/org/simple/clinic/scheduleappointment/PatientFacilityChangeActivity.kt
```diff
@@ -0,0 +1,28 @@
...
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatActivity
+import io.github.inflationx.viewpump.ViewPumpContextWrapper
+import org.simple.clinic.ClinicApp
+import org.simple.clinic.R
+import org.simple.clinic.util.LocaleOverrideContextWrapper
+import java.util.Locale
+import javax.inject.Inject
+
+class PatientFacilityChangeActivity : AppCompatActivity() {
```
vinaysshenoy: I am not too sure about the name `PatientFacilityChangeActivity` since this class has nothing to do with changing the patient's facility at the moment. What it does is just lets the user pick a facility and returns the result. Should we just rename it to something like `FacilityPickerActivity` instead?

sanchita:
> @sanchita I also had a thought about the overall feature:
> 
> This is yet another screen we will have to migrate to Mobius at some point. I don't want to introduce more code that uses the older architecture, but it doesn't seem like it's avoidable at the moment. In addition, every change that happens in `FacilityChangeScreen` should also be duplicated here, which is painful and added maintenance overhead.
> 
> I have a suggestion to fix this in the long term:
> 
>     1. Change `PatientsScreen` to use this screen itself for changing the user facility, i.e, open this screen and get a result.
> 
>     2. Observe that result in the `PatientsScreen` and update the user's facility in `PatientsScreenController` instead of doing it in `FacilityChangeScreenController`.
> 
>     3. Anything extra that `FacilityChangeScreen` does can be moved to the `PatientsScreenController`.
> 
>     4. Remove the `FacilityChangeScreen`.
> 
> 
> What do you think? This will let us remove the maintenance overhead of having to migrate another screen and maintaining feature parity between these two screens.

I did think about this as well. It does sound like a good plan. We can revisit the whole facility change thing including registration as well, maybe?

app/src/main/java/org/simple/clinic/di/AppComponent.kt
sanchita: I wanted to reduce the number of words here. I can rename it this way I guess.

vinaysshenoy:
> > @sanchita I also had a thought about the overall feature:
> > This is yet another screen we will have to migrate to Mobius at some point. I don't want to introduce more code that uses the older architecture, but it doesn't seem like it's avoidable at the moment. In addition, every change that happens in `FacilityChangeScreen` should also be duplicated here, which is painful and added maintenance overhead.
> > I have a suggestion to fix this in the long term:
> > ```
> > 1. Change `PatientsScreen` to use this screen itself for changing the user facility, i.e, open this screen and get a result.
> > 
> > 2. Observe that result in the `PatientsScreen` and update the user's facility in `PatientsScreenController` instead of doing it in `FacilityChangeScreenController`.
> > 
> > 3. Anything extra that `FacilityChangeScreen` does can be moved to the `PatientsScreenController`.
> > 
> > 4. Remove the `FacilityChangeScreen`.
> > ```
> > 
> > 
> > What do you think? This will let us remove the maintenance overhead of having to migrate another screen and maintaining feature parity between these two screens.
> 
> I did think about this as well. It does sound like a good plan. We can revisit the whole facility change thing including registration as well, maybe?

Sounds good, I'll raise a ticket for this.

app/src/main/java/org/simple/clinic/scheduleappointment/PatientFacilityChangeActivity.kt
sanchita: I will rename it to `FacilitySelectionActivity`.

vinaysshenoy:
> > > @sanchita I also had a thought about the overall feature:
> > > This is yet another screen we will have to migrate to Mobius at some point. I don't want to introduce more code that uses the older architecture, but it doesn't seem like it's avoidable at the moment. In addition, every change that happens in `FacilityChangeScreen` should also be duplicated here, which is painful and added maintenance overhead.
> > > I have a suggestion to fix this in the long term:
> > > ```
> > > 1. Change `PatientsScreen` to use this screen itself for changing the user facility, i.e, open this screen and get a result.
> > > 
> > > 2. Observe that result in the `PatientsScreen` and update the user's facility in `PatientsScreenController` instead of doing it in `FacilityChangeScreenController`.
> > > 
> > > 3. Anything extra that `FacilityChangeScreen` does can be moved to the `PatientsScreenController`.
> > > 
> > > 4. Remove the `FacilityChangeScreen`.
> > > ```
> > > 
> > > 
> > > What do you think? This will let us remove the maintenance overhead of having to migrate another screen and maintaining feature parity between these two screens.
> > 
> > 
> > I did think about this as well. It does sound like a good plan. We can revisit the whole facility change thing including registration as well, maybe?
> 
> Sounds good, I'll raise a ticket for this.

Raised a ticket: https://www.pivotaltracker.com/story/show/170126581

app/src/main/java/org/simple/clinic/di/AppComponent.kt
vinaysshenoy: IMO, it's fine to be more verbose. This is something that won't get typed very often anyway/

app/src/main/res/layout/screen_patient_facility_change.xml
sanchita: Done

app/src/main/java/org/simple/clinic/scheduleappointment/patientFacilityTransfer/PatientFacilityChangeActivity.kt
sanchita: Removed

PR #979, opened 2019-12-05
Add the blood sugar measurement database table

vinaysshenoy:
https://www.pivotaltracker.com/story/show/169676393


vinaysshenoy:
@sanchita Leaving this PR to you to get merged in.

app/src/androidTest/java/org/simple/clinic/storage/migrations/BaseDatabaseMigrationTest.kt
```diff
@@ -0,0 +1,37 @@
...
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+import org.junit.Before
+import org.junit.Rule
+import org.junit.rules.ExpectedException
+import org.simple.clinic.MigrationTestHelperWithForeignKeyConstraints
+import org.simple.clinic.TestClinicApp
+import javax.inject.Inject
+
+abstract class BaseDatabaseMigrationTest(
```
vinaysshenoy: `DatabaseMigrationAndroidTest` has been growing massively in size and getting slower and slower to work with over time. So I introduced this base class so we can have one test class per migration.

This base class provides two properties, `before` and `after`, which will give access to the databases before and after migration.

app/src/main/java/org/simple/clinic/storage/Timestamps.kt
```diff
@@ -0,0 +1,42 @@
...
+ * - `createdAt`: When the database record was created.
+ * - `updatedAt`: When the database record was updated.
+ * - `deletedAt`: When the database record was deleted.
+ *
+ * All timestamps are in UTC.
+ *
+ * This class is meant to encapsulate these timestamps as an [Embedded] model in the database
+ * entities.
+ **/
+data class Timestamps(
```
vinaysshenoy: At the moment, we can have upto four different timestamps on any domain model: `createdAt`, `updatedAt`, `deletedAt`, and `recordedAt`. They are all of the same type, `Instant. 

However, if we look at them from a domain perspective, only `recordedAt has an actual meaning at the domain level (When did this patient visit? When was this blood pressure recorded?). The others are all related to the application logic (when was this database record created? When was it changed? Was it ever deleted?).

This was causing trouble when working with records because it is up to us to remember what fields to change when something happens and it is easily missable. I've been thinking about trying to introduce a `Timestamps` notion for a while now (https://www.pivotaltracker.com/story/show/162636638) which will separate these two concepts cleanly. Decided to try it out here.

app/src/main/java/org/simple/clinic/bloodsugar/BloodSugarReading.kt
```diff
@@ -0,0 +1,3 @@
+package org.simple.clinic.bloodsugar
+
+data class BloodSugarReading(val value: Int, val type: BloodSugarMeasurementType)
```
vinaysshenoy: I am bringing the reading value and type together to give them meaning at a domain level. This lets us avoid [Primitive Obsession](https://enterprisecraftsmanship.com/posts/functional-c-primitive-obsession/) and lets us avoid unnecessary creation of validator classes like we currently do for `BloodPressureMeasurement`.

With this approach, this class can:
- Be responsible for its own validation.
- Be instantiated and passed around independently and expose intent through its type rather than the primitive parameter names.

app/src/main/java/org/simple/clinic/storage/Timestamps.kt
```diff
@@ -0,0 +1,42 @@
...
+    val createdAt: Instant,
+    val updatedAt: Instant,
+    val deletedAt: Instant?
+) {
+
+  companion object {
+
+    fun now(clock: UtcClock): Timestamps {
```
sanchita: This method should be called `create` since this will create timestamps for a new entry.

app/src/main/java/org/simple/clinic/storage/Timestamps.kt
```diff
@@ -0,0 +1,42 @@
...
+      val instant = Instant.now(clock)
+
+      return Timestamps(createdAt = instant, updatedAt = instant, deletedAt = null)
+    }
+  }
+
+  fun updated(clock: UtcClock): Timestamps {
```
sanchita: We should be using present tense for these methods.
Like `update` and `delete`.
Also, since we are not using these yet, we could have added these methods later.

app/src/main/java/org/simple/clinic/bloodsugar/BloodSugarReading.kt
sanchita: Nicely done :)

app/src/androidTest/java/org/simple/clinic/SqliteFunctions.kt
```diff
@@ -0,0 +1,101 @@
+package org.simple.clinic
```
sanchita: Rename to SqliteHelperFunctions?

app/src/androidTest/java/org/simple/clinic/storage/migrations/BaseDatabaseMigrationTest.kt
```diff
@@ -0,0 +1,37 @@
...
+  @Before
+  @CallSuper
+  open fun setUp() {
+    TestClinicApp.appComponent().inject(this)
+    helper.migrations = migrations
+  }
+
+  protected val before: SupportSQLiteDatabase by lazy(LazyThreadSafetyMode.NONE) { helper.createDatabase(fromVersion) }
+
+  protected val after: SupportSQLiteDatabase by lazy(LazyThreadSafetyMode.NONE) { helper.migrateTo(toVersion) }
```
sanchita: In my opinion, abstracting `before` and `after` in tests adds a bit of cognitive complexity in understanding the migrations tests themselves. We will have to go and look what `before` and `after` means. Also that would abstract the `given` and `when` while writing a test. Specially the `given` when we are not performing assertions on it.

sanchita:
@rakshakhegde please review this PR once.

app/src/androidTest/java/org/simple/clinic/storage/migrations/BaseDatabaseMigrationTest.kt
rakshakhegde: I find this to be quite neatly done. One would have to look at the source probably only once to understand what `before` and `after` is as they are just one-line definitions. So IMO, we can leave this as is.
I don't think I fully understand what the "abstraction of `given` and `when`" implies.

sanchita: @rakshakhegde To answer your [question](https://github.com/simpledotorg/simple-android/pull/979#discussion_r359188926):
We follow the `given` and `when` convention to write tests.
In our migration tests, the `given` is what `before` does here and `when` is what `after` does. We will have to read `before` and `after` every time we write these tests. That is what I mean by "abstract the given and when".

sanchita: But if you think that this is fine, then I will keep this as is. It was a comment for @vinaysshenoy to see later.

app/src/main/java/org/simple/clinic/storage/Timestamps.kt
vinaysshenoy: True, will rename.

app/src/main/java/org/simple/clinic/storage/Timestamps.kt
vinaysshenoy: I will rename the methods to update the tense.

Regd. whether it was necessary to add the methods now, you're right that we could have added them later. I intentionally added them now for two reasons:

- We know for a fact that we will need them because blood pressure readings can be both edited and deleted, and blood sugars will have the same features
- I wanted to give an indication to the reviewer that we can encapsulate the implementation details of updating a timestamp for editing or deleting.

app/src/androidTest/java/org/simple/clinic/SqliteFunctions.kt
vinaysshenoy: Will do.

app/src/androidTest/java/org/simple/clinic/storage/migrations/BaseDatabaseMigrationTest.kt
vinaysshenoy: My intention was to hide the details of the database migration itself. If you look at how we currently write a migration test, it goes something like:

```kotlin
val db_v1 = /* setup database */

val db_v2 = /* migrate db_v1 */

// verify db_v2
```

It is not actually important to know which version is being migrated to which. What is important is the state of the database **before** migration and the state of the database **after** migration. This was the reason I chose to name them `before` and `after` so that we can focus on the migration itself instead of anything else.

I felt that it is not necessary to know what `before` and `after` actually do, since they are quite evident. Since this test is written specifically to test database migrations, someone can just use `before` and `after` to setup and verify the migrations without needing to write the boilerplate to do the migrations themselves.

Does this make sense?

app/src/main/java/org/simple/clinic/storage/Timestamps.kt
vinaysshenoy: Never mind, I saw that you addressed the feedback yourself. Thanks!

PR #983, opened 2019-12-06
Select facility and set it in activity result

sanchita:
https://www.pivotaltracker.com/story/show/169753540

Part 1: https://github.com/simpledotorg/simple-android/pull/977
Part 2: https://github.com/simpledotorg/simple-android/pull/983
Part 3: [WIP]( https://www.pivotaltracker.com/story/show/169808154)


app/src/main/java/org/simple/clinic/scheduleappointment/facilityselection/PatientFacilityChangeEvents.kt
```diff
@@ -0,0 +1,26 @@
...
+data class PatientFacilityLocationPermissionChanged(val result: RuntimePermissionResult) : UiEvent {
+  override val analyticsName = "Facility Change:Location Permission Changed to $result"
+}
+
+data class PatientFacilitySearchQueryChanged(val query: String) : UiEvent {
+  override val analyticsName = "Facility Change:Facility Search Query Changed"
+}
+
+data class PatientFacilityUserLocationUpdated(val location: LocationUpdate) : UiEvent {
+  override val analyticsName = "Facility Change:Location Updated ${when (location) {
```
rakshakhegde: This can be made easier to read by extracting the when block into a separate property. But if you would like to keep it like this, that is fine with me too.

app/src/main/java/org/simple/clinic/scheduleappointment/facilityselection/FacilitySelectionActivityController.kt
```diff
@@ -184,17 +180,7 @@ class FacilitySelectionActivityController @Inject constructor(
                     .associateUserWithFacility(it, facility)
                     .andThen(facilityRepository.setCurrentFacility(it, facility))
               }
-              .doOnComplete { clearAndSyncReports() }
```
rakshakhegde: I don't really understand why the reports clearing is deleted here. Let's discuss in person?

app/src/main/java/org/simple/clinic/scheduleappointment/facilityselection/FacilitySelectionActivity.kt
```diff
@@ -28,6 +28,7 @@ import org.simple.clinic.widgets.ScreenDestroyed
 import org.simple.clinic.widgets.UiEvent
```
rakshakhegde: Please reword this commit?

app/src/main/java/org/simple/clinic/scheduleappointment/facilityselection/FacilitySelectionActivityController.kt
sanchita: In case of changing facility of a user, we need to delete existing reports since the reports are generated based on a particular facility.
Whereas in this case, facility change is for a patient and not a user. Hence we do not need to delete reports.

app/src/main/java/org/simple/clinic/scheduleappointment/facilityselection/PatientFacilityChangeEvents.kt
sanchita: I think this is fine for now. We are not going to add anything else to this `data class` anyway.

app/src/main/java/org/simple/clinic/scheduleappointment/facilityselection/FacilitySelectionActivity.kt
sanchita: Done

PR #991, opened 2019-12-19
User must be able to select a type of Blood Sugar to add

msasikanth:
https://www.pivotaltracker.com/story/show/170099068

### DIFF
--- a/app/src/main/java/org/simple/clinic/bloodsugar/selection/type/BloodSugarTypeAdapter.kt
+++ b/app/src/main/java/org/simple/clinic/bloodsugar/selection/type/BloodSugarTypeAdapter.kt
@@ -1,0 +1,59 @@
+package org.simple.clinic.bloodsugar.selection.type
+
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.recyclerview.widget.DiffUtil
+import androidx.recyclerview.widget.ListAdapter
+import androidx.recyclerview.widget.RecyclerView
+import kotlinx.android.extensions.LayoutContainer
+import kotlinx.android.synthetic.main.blood_sugar_type_list_item.*
+import org.simple.clinic.R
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/simple/clinic/bloodsugar/selection/type/BloodSugarTypeAdapter.kt
```diff
@@ -0,0 +1,59 @@
...
+
+class BloodSugarTypeAdapter(
+    private val onBloodSugarTypeSelected: OnBloodSugarTypeSelected
+) : ListAdapter<BloodSugarTypeListItem, BloodSugarTypeViewHolder>(BloodSugarTypeDiffer()) {
+
+  override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): BloodSugarTypeViewHolder {
+    val layout = LayoutInflater.from(parent.context).inflate(R.layout.blood_sugar_type_list_item, parent, false)
+    val viewHolder = BloodSugarTypeViewHolder(layout)
+    viewHolder.itemView.setOnClickListener {
+      val bloodSugarTypeListItem = getItem(viewHolder.adapterPosition)
```
sanchita: Can we not get this value from `viewHolder.bloodSugarTypeItem` itself instead of calling `getItem` again?

msasikanth: We can, but that would have to be a `lateinit` variable. I can change it to that, then we would be assigning the variable in `onBindViewHolder`. Shall I do that?

sanchita: It is already a `lateinit var`. So please go ahead and do this.

sanchita:
Merging @msasikanth 's first contribution to Simple :)

PR #1012, opened 2020-01-01
Replace Stetho with Flipper

vinaysshenoy:
We had integrated [Stetho](http://facebook.github.io/stetho/) for debugging the app during development. Stetho is not being actively developed, and they have an alternate tool called [Flipper](https://fbflipper.com/), which they recommend instead.

This PR replaces the usage of Stetho in the app with Flipper, along with integration for the following plugins:

### Layout Inspector
layout_inspector

### Network Inspector
network

### Database Inspector
database

### Shared preferences Inspector
shared_preferences

### DIFF
--- a/app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
+++ b/app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
@@ -63,9 +63,9 @@ class DebugClinicApp : ClinicApp() {
   }
 
   private fun setupFlipper() {
-    with(AndroidFlipperClient.getInstance(this)) {
-      val context = this@DebugClinicApp
+    val context = this
 
+    with(AndroidFlipperClient.getInstance(this)) {
       addPlugin(InspectorFlipperPlugin(context, DescriptorMapping.withDefaults()))
       addPlugin(networkFlipperPlugin)
 
--- a/app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
+++ b/app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
@@ -3,6 +3,10 @@ package org.simple.clinic
 import android.annotation.SuppressLint
 import android.app.Activity
 import android.os.StrictMode
+import com.facebook.flipper.android.AndroidFlipperClient
+import com.facebook.flipper.plugins.inspector.DescriptorMapping
+import com.facebook.flipper.plugins.inspector.InspectorFlipperPlugin
+import com.facebook.soloader.SoLoader
 import com.tspoon.traceur.Traceur
 import io.github.inflationx.viewpump.ViewPump
 import org.simple.clinic.activity.SimpleActivityLifecycleCallbacks
@@ -31,6 +35,9 @@ class DebugClinicApp : ClinicApp() {
     Traceur.enableLogging()
     super.onCreate()
 
+    SoLoader.init(this, false)
+    setupFlipper()
+
     appComponent().inject(this)
 
     Timber.plant(Timber.DebugTree())
@@ -41,6 +48,13 @@ class DebugClinicApp : ClinicApp() {
         .build())
 
     signature = AppSignature(this)
+  }
+
+  private fun setupFlipper() {
+    with(AndroidFlipperClient.getInstance(this)) {
+      addPlugin(InspectorFlipperPlugin(this@DebugClinicApp, DescriptorMapping.withDefaults()))
+      start()
+    }
   }
 
   private fun showDebugNotification() {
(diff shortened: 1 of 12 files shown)


app/src/debug/java/org/simple/clinic/ReadOnlySqliteDatabaseDriver.kt
```diff
@@ -0,0 +1,54 @@
...
+  }
+}
+
+private class ReadOnlySqliteDatabaseConnectionProvider : SqliteDatabaseConnectionProvider {
+
+  override fun openDatabase(databaseFile: File): SQLiteDatabase {
+    // We are setting this to READONLY for a couple of reasons:
+
+    // 1. Room sets all databases to WAL by default, which means that they need to
+    // shared a connection to the underlying database via Room's SQLiteOpenHelper.
```
sanchita: Typo? "shared" -> "share"

app/src/debug/java/org/simple/clinic/ReadOnlySqliteDatabaseDriver.kt
```diff
@@ -0,0 +1,54 @@
...
+        .databaseList()
+        .map(context::getDatabasePath)
+        .toMutableList()
+  }
+}
+
+private class ReadOnlySqliteDatabaseConnectionProvider : SqliteDatabaseConnectionProvider {
+
+  override fun openDatabase(databaseFile: File): SQLiteDatabase {
+    // We are setting this to READONLY for a couple of reasons:
```
sanchita: Are these three points, three different reasons? 
I think its more like one detailed reason. Maybe consider removing the numbering and just keep the paragraphs.

app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
```diff
@@ -67,8 +68,14 @@ class DebugClinicApp : ClinicApp() {
 
   private fun setupFlipper() {
     with(AndroidFlipperClient.getInstance(this)) {
-      addPlugin(InspectorFlipperPlugin(this@DebugClinicApp, DescriptorMapping.withDefaults()))
+      val context = this@DebugClinicApp
```
sanchita: Maybe move this out of the `with` block and remove the `@DebugClinicApp`?

app/src/main/java/org/simple/clinic/bp/entry/confirmremovebloodpressure/ConfirmRemoveBloodPressureDialog.kt
```diff
@@ -87,7 +87,7 @@ class ConfirmRemoveBloodPressureDialog : AppCompatDialogFragment() {
     screenDestroys.onNext(ScreenDestroyed())
   }
 
-  override fun onAttach(context: Context?) {
+  override fun onAttach(context: Context) {
```
sanchita: Can you check if these are happening in this PR only?

sanchita:
Left minor comments. PTAL.

app/src/debug/java/org/simple/clinic/ReadOnlySqliteDatabaseDriver.kt
vinaysshenoy: Good point, I'll do this.

app/src/debug/java/org/simple/clinic/ReadOnlySqliteDatabaseDriver.kt
vinaysshenoy: Whoops, fixed!

app/src/debug/java/org/simple/clinic/ReadOnlySqliteDatabaseDriver.kt
vinaysshenoy: Done.

app/src/debug/java/org/simple/clinic/DebugClinicApp.kt
vinaysshenoy: Looks better, will make this change.

app/src/main/java/org/simple/clinic/bp/entry/confirmremovebloodpressure/ConfirmRemoveBloodPressureDialog.kt
vinaysshenoy: I checked, it seems that on my machine, the Gradle task is using `androidx.fragment:1.1.0` to resolve the classes, which has added nullability annotations. I do not know where this dependency is coming from.

Can we merge this in for the moment? It will not cause any problems in the build since they are just added nullability annotations. This should get fixed if we migrate the app to explicitly depend on `androidx` v1.1.0.

PR #1016, opened 2020-01-02
Change address model to have 5 fields

rakshakhegde:
https://www.pivotaltracker.com/story/show/168980965

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
+++ b/app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
@@ -103,11 +103,17 @@ data class OngoingNewPatientEntry(
   fun withBangladeshNationalId(bangladeshNationalId: Identifier): OngoingNewPatientEntry =
       copy(bangladeshNationalId = bangladeshNationalId)
 
+  fun withStreetAddress(streetAddress: String): OngoingNewPatientEntry =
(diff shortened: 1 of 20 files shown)


app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
```diff
@@ -225,15 +225,17 @@ data class OngoingNewPatientEntry(
 
   @Parcelize
   data class Address(
-      val colonyOrVillage: String, // TODO (rj) 8-Nov-19: Add `streetAddress` and `zone` fields.
+      val colonyOrVillage: String,
       val district: String,
-      val state: String
+      val state: String,
+      val streetAddress: String?,
```
vinaysshenoy: Are these nullable because they are optional fields? If so, might make more sense to explicitly use `Optional<String>` instead?

rakshakhegde: In Kotlin, I don't find it useful to use `Optional`. Also I find null handling syntax is much better than using Optionals. So let's keep it as a nullable field?

vinaysshenoy: It might make sense to declare this as `Optional` just to expose the meaning at the UI level that the fields are optional.

But I don't feel strongly about this either way. Just wanted to confirm. We can revisit this in the future if needed.

PR #1018, opened 2020-01-02
Use creation facility ID instead of facility ID for recent patients

rakshakhegde:
https://www.pivotaltracker.com/story/show/170485623

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/RecentPatient.kt
+++ b/app/src/main/java/org/simple/clinic/patient/RecentPatient.kt
@@ -54,9 +54,9 @@ data class RecentPatient(
               GROUP BY patientUuid
           ) PD ON P.uuid = PD.patientUuid
           LEFT JOIN (
-            SELECT MAX(createdAt) latestCreatedAt, uuid, patientUuid, facilityUuid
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/patient/RecentPatient.kt
```diff
@@ -56,7 +56,7 @@ data class RecentPatient(
           LEFT JOIN (
             SELECT MAX(createdAt) latestCreatedAt, uuid, patientUuid, facilityUuid
               FROM Appointment
-              WHERE facilityUuid = :facilityUuid
+              WHERE creationFacilityUuid = :facilityUuid
```
sanchita: Please add a test for this.

sanchita:
Moving this to WIP as the query is not correct.
We should update the `WHERE` clause to `AP.creationFacilityUuid = :facilityUuid`

rakshakhegde:
`creationFacilityUuid` that is being used is that of `Appointment`.

app/src/main/java/org/simple/clinic/patient/RecentPatient.kt
rakshakhegde: done

PR #1008, opened 2019-12-30
User must be able to enter a blood sugar reading - Part 2

msasikanth:
https://www.pivotaltracker.com/story/show/169676359

- `BloodSugarEntryInit` & `BloodSugarEntryUiRenderer` implementation. 
- Inital implementation of `BloodSugarEntryEffectHandler`

PR #1019, opened 2020-01-02
User must be able to enter a blood sugar reading - Part 3

msasikanth:
https://www.pivotaltracker.com/story/show/169676359

Completed `BloodSugarEntryEffectHandler` implementation

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/bloodsugar/BloodSugarRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/bloodsugar/BloodSugarRepositoryAndroidTest.kt
@@ -7,7 +7,9 @@ import org.junit.Test
 import org.junit.rules.RuleChain
 import org.simple.clinic.TestClinicApp
 import org.simple.clinic.TestData
+import org.simple.clinic.patient.SyncStatus
 import org.simple.clinic.rules.LocalAuthenticationRule
+import org.simple.clinic.storage.Timestamps
 import org.simple.clinic.util.RxErrorsRule
 import org.simple.clinic.util.TestUtcClock
 import org.threeten.bp.Instant
@@ -48,22 +50,37 @@ class BloodSugarRepositoryAndroidTest {
 
   @Test
   fun saving_a_blood_sugar_reading_should_work_correctly() {
-    //given
-    val bloodSugarReading = BloodSugarReading(value = 10, type = Random)
+    // given
+    val bloodSugarUuid = UUID.fromString("e35dea7a-b13b-4ab1-aaa4-0c5998e5c79f")
+    val bloodSugarReading = BloodSugarReading(value = 50, type = Random)
+    val now = Instant.now(clock)
     val patientUuid = UUID.fromString("a5921ec9-5c70-421a-bb0b-1291364683f6")
+    val facility = testData.qaFacility()
     val user = testData.qaUser()
 
-    //when
-    val testObserver = repository.saveMeasurement(
+    val expectedBloodSugarMeasurement = BloodSugarMeasurement(
+        uuid = bloodSugarUuid,
+        reading = bloodSugarReading,
+        recordedAt = now,
+        patientUuid = patientUuid,
+        userUuid = user.uuid,
+        facilityUuid = facility.uuid,
+        timestamps = Timestamps.create(clock),
+        syncStatus = SyncStatus.PENDING
+    )
+
+    // when
+    val savedBloodSugar = repository.saveMeasurement(
+        uuid = bloodSugarUuid,
         reading = bloodSugarReading,
         patientUuid = patientUuid,
         loggedInUser = user,
-        facility = testData.qaFacility(),
-        recordedAt = Instant.now(clock)
-    ).test()
+        facility = facility,
+        recordedAt = now)
+        .blockingGet()
 
-    //then
-    testObserver.assertComplete().assertNoErrors()
+    // then
+    assertThat(savedBloodSugar).isEqualTo(expectedBloodSugarMeasurement)
   }
 
   @Test
--- a/app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandler.kt
@@ -3,14 +3,27 @@ package org.simple.clinic.bloodsugar.entry
 import com.spotify.mobius.rx2.RxMobius
 import io.reactivex.ObservableTransformer
 import io.reactivex.Scheduler
+import io.reactivex.Single
+import io.reactivex.rxkotlin.cast
+import org.simple.clinic.ReportAnalyticsEvents
+import org.simple.clinic.bloodsugar.BloodSugarMeasurement
+import org.simple.clinic.bloodsugar.BloodSugarReading
+import org.simple.clinic.bloodsugar.BloodSugarRepository
 import org.simple.clinic.bloodsugar.entry.BloodSugarValidator.Result.ErrorBloodSugarEmpty
 import org.simple.clinic.bloodsugar.entry.BloodSugarValidator.Result.ErrorBloodSugarTooHigh
 import org.simple.clinic.bloodsugar.entry.BloodSugarValidator.Result.ErrorBloodSugarTooLow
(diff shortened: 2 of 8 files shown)


sanchita:
@msasikanth Please address feedback in https://github.com/simpledotorg/simple-android/pull/1005, as discussed before, in this PR.

msasikanth:
> @msasikanth Please address feedback in #1005, as discussed before, in this PR.

I have already made those changes here: https://github.com/simpledotorg/simple-android/pull/1008

Please let me know if I missed any

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandler.kt
```diff
@@ -66,4 +72,12 @@ class BloodSugarEntryEffectHandler(
 
   private fun getYear(date: LocalDate): String =
       date.year.toString().substring(startIndex = 2, endIndex = 4)
+
+  private fun showDateValidationError(result: UserInputDateValidator.Result) {
+    when (result) {
+      InvalidPattern -> ui.showInvalidDateError()
+      DateIsInFuture -> ui.showDateIsInFutureError()
+      is Valid -> throw IllegalStateException("Date validation error cannot be $result")
```
vinaysshenoy: Since this `when` block does not return anything, adding a new type to `UserInputDateValidator.Result` will result in no effect.

Should we change this to an `else` instead so that the app will fail during development? The other option is to use the `.exhaustive()` extension function to ensure that we have added all types to this block.

```kotlin
when (result) {
  InvalidPattern -> ui.showInvalidDateError()
  DateIsInFuture -> ui.showDateIsInFutureError()
  is Valid -> throw IllegalStateException("Date validation error cannot be $result")
}.exhaustive()
```

app/src/test/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandlerTest.kt
```diff
@@ -129,4 +131,24 @@ class BloodSugarEntryEffectHandlerTest {
     verify(ui).showDateOnDateButton(entryDate)
     verifyNoMoreInteractions(ui)
   }
+
+  @Test
+  fun `show invalid blood sugar error when show date validation error is received with validation invalid pattern`() {
```
vinaysshenoy: Check the name please? Is it `invalid blood sugar error` here?

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandler.kt
```diff
@@ -7,7 +7,6 @@ import org.simple.clinic.bloodsugar.entry.BloodSugarValidator.Result.ErrorBloodS
 import org.simple.clinic.bloodsugar.entry.BloodSugarValidator.Result.ErrorBloodSugarTooHigh
```
vinaysshenoy: Can we reword the commit message to describe what is actually happening?

Ex: "Finish the blood sugar entry activity after the blood sugar is saved"

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandler.kt
msasikanth: Added `.exhaustive()` extension. Forgot Kotlin doesn't fail build for missing branches in `when` block when we don't return it.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEvent.kt
```diff
@@ -26,12 +26,7 @@ data class YearChanged(
 
 data class ScreenChanged(val type: ScreenType) : BloodSugarEntryEvent()
 
-data class BloodSugarSaved(val wasDateChanged: Boolean) : BloodSugarEntryEvent() {
-  override val analyticsName: String = when {
```
vinaysshenoy: Was this removed for a reason? This will no longer get pushed to analytics now. It is important for us to detect in reports if readings are being saved with the current dates or older dates.

app/src/androidTest/java/org/simple/clinic/bloodsugar/BloodSugarRepositoryAndroidTest.kt
```diff
@@ -48,22 +48,23 @@ class BloodSugarRepositoryAndroidTest {
```
vinaysshenoy: Reword the commit message?

Ex: "Return `BloodSugarMeasurement` after creating it"

app/src/androidTest/java/org/simple/clinic/bloodsugar/BloodSugarRepositoryAndroidTest.kt
```diff
@@ -48,22 +48,23 @@ class BloodSugarRepositoryAndroidTest {
...
         facility = testData.qaFacility(),
-        recordedAt = Instant.now(clock)
-    ).test()
+        recordedAt = now)
+        .blockingGet()
 
-    //then
-    testObserver.assertComplete().assertNoErrors()
+    // then
+    assertThat(savedBloodSugar.recordedAt).isEqualTo(now)
```
vinaysshenoy: We should assert the complete `BloodSugarMeasurement` value and not just the `recorded` at timestamp.

vinaysshenoy: We can change the `BloodSugarRepository#saveMeasurement` function to accept the `UUID` as a parameter and then provide a `UUID` here to actually verify the complete measurement.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEvent.kt
msasikanth: `wasDateChanged` was never used other than for the analytics. Should I add it back?

vinaysshenoy: Yes, if is being used for analytics, we should add it back. I think undoing this commit should be quite easy.

app/src/test/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandlerTest.kt
```diff
@@ -161,4 +188,33 @@ class BloodSugarEntryEffectHandlerTest {
...
+    whenever(facilityRepository.currentFacility(user)).doReturn(Observable.just(facility))
+    whenever(bloodSugarRepository.saveMeasurement(bloodSugar.reading, bloodSugar.patientUuid, user, facility, date.toUtcInstant(userClock))).doReturn(Single.just(bloodSugar))
+    whenever(patientRepository.compareAndUpdateRecordedAt(bloodSugar.patientUuid, date.toUtcInstant(userClock))).doReturn(Completable.complete())
+    whenever(appointmentRepository.markAppointmentsCreatedBeforeTodayAsVisited(bloodSugar.patientUuid)).doReturn(Completable.complete())
+
+    // when
+    testCase.dispatch(createNewBloodSugarEntry)
+
+    // then
+    testCase.assertOutgoingEvents(BloodSugarSaved)
```
vinaysshenoy: Should we add a `veriifyZeroInteractions(ui)` here?

app/src/androidTest/java/org/simple/clinic/bloodsugar/BloodSugarRepositoryAndroidTest.kt
```diff
@@ -48,22 +48,25 @@ class BloodSugarRepositoryAndroidTest {
...
         facility = testData.qaFacility(),
-        recordedAt = Instant.now(clock)
-    ).test()
+        recordedAt = now)
+        .blockingGet()
 
-    //then
-    testObserver.assertComplete().assertNoErrors()
+    // then
+    assertThat(savedBloodSugar.uuid).isEqualTo(bloodSugarUuid)
```
vinaysshenoy: We should verify the entire `BloodSugarMeasurement` model here and not just the UUID.

PR #1020, opened 2020-01-02
Implement Blood sugar Summary EffectHandler

sanchita:
https://www.pivotaltracker.com/story/show/169675670
We are not implementing `Edit` for now.

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandler.kt
@@ -2,15 +2,25 @@ package org.simple.clinic.summary.bloodsugar
 
 import com.spotify.mobius.rx2.RxMobius
 import io.reactivex.ObservableTransformer
+import org.simple.clinic.bloodsugar.BloodSugarRepository
+import org.simple.clinic.util.scheduler.SchedulersProvider
 
-class BloodSugarSummaryViewEffectHandler {
+object BloodSugarSummaryViewEffectHandler {
 
-  companion object {
-    fun create(): ObservableTransformer<BloodSugarSummaryViewEffect, BloodSugarSummaryViewEvent> {
-      return RxMobius
-          .subtypeEffectHandler<BloodSugarSummaryViewEffect, BloodSugarSummaryViewEvent>()
-          .addTransformer(FetchBloodSugarSummary::class.java) { effect -> effect.map { BloodSugarSummaryFetched } }
-          .build()
-    }
+  fun create(
+      bloodSugarRepository: BloodSugarRepository,
+      schedulersProvider: SchedulersProvider
+  ): ObservableTransformer<BloodSugarSummaryViewEffect, BloodSugarSummaryViewEvent> {
+    return RxMobius
+        .subtypeEffectHandler<BloodSugarSummaryViewEffect, BloodSugarSummaryViewEvent>()
+        .addTransformer(FetchBloodSugarSummary::class.java) { effect ->
+          effect
+              .flatMap {
+                bloodSugarRepository.latestMeasurements(it.patientUuid, 100)
+                    .subscribeOn(schedulersProvider.io())
+              }
+              .map { BloodSugarSummaryFetched(it) }
+        }
+        .build()
   }
 }
--- a/app/src/main/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandler.kt
@@ -2,6 +2,7 @@ package org.simple.clinic.summary.bloodsugar
 
 import com.spotify.mobius.rx2.RxMobius
 import io.reactivex.ObservableTransformer
+import io.reactivex.Scheduler
 import org.simple.clinic.bloodsugar.BloodSugarRepository
 import org.simple.clinic.util.scheduler.SchedulersProvider
 
@@ -9,18 +10,28 @@ object BloodSugarSummaryViewEffectHandler {
 
   fun create(
       bloodSugarRepository: BloodSugarRepository,
-      schedulersProvider: SchedulersProvider
+      schedulersProvider: SchedulersProvider,
+      uiActions: UiActions
   ): ObservableTransformer<BloodSugarSummaryViewEffect, BloodSugarSummaryViewEvent> {
     return RxMobius
         .subtypeEffectHandler<BloodSugarSummaryViewEffect, BloodSugarSummaryViewEvent>()
-        .addTransformer(FetchBloodSugarSummary::class.java) { effect ->
-          effect
-              .flatMap {
-                bloodSugarRepository.latestMeasurements(it.patientUuid, 100)
-                    .subscribeOn(schedulersProvider.io())
-              }
-              .map { BloodSugarSummaryFetched(it) }
-        }
+        .addTransformer(FetchBloodSugarSummary::class.java, fetchBloodSugarMeasurements(bloodSugarRepository, schedulersProvider.ui()))
+        .addAction(OpenBloodSugarTypeSelector::class.java, uiActions::showBloodSugarTypeSelector, schedulersProvider.ui())
         .build()
+  }
+
+  private fun fetchBloodSugarMeasurements(
+      bloodSugarRepository: BloodSugarRepository,
+      scheduler: Scheduler
+  ): ObservableTransformer<FetchBloodSugarSummary, BloodSugarSummaryViewEvent> {
+    return ObservableTransformer { effect ->
+      effect
+          .flatMap {
+            bloodSugarRepository
+                .latestMeasurements(it.patientUuid, 100)
+                .subscribeOn(scheduler)
+          }
+          .map { BloodSugarSummaryFetched(it) }
+    }
   }
 }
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandler.kt
```diff
@@ -2,15 +2,25 @@ package org.simple.clinic.summary.bloodsugar
 
 import com.spotify.mobius.rx2.RxMobius
 import io.reactivex.ObservableTransformer
+import org.simple.clinic.bloodsugar.BloodSugarRepository
+import org.simple.clinic.util.scheduler.SchedulersProvider
 
-class BloodSugarSummaryViewEffectHandler {
+object BloodSugarSummaryViewEffectHandler {
```
vinaysshenoy: We initially started with using `object` declarations for effect handlers, but that became too verbose over time, so we changed to declaring them as classes and providing the dependencies as class properties. This lets us inject the class when setting up the mobius loop as well.

Could you please change this to a class declaration? `SetupActivityEffectHandler` is a good example.

Note: it is true that some of our effect handlers are still declared as `object` types. We should work towards migrating them as well.

app/src/main/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandler.kt
```diff
@@ -2,15 +2,25 @@ package org.simple.clinic.summary.bloodsugar
...
+  fun create(
+      bloodSugarRepository: BloodSugarRepository,
+      schedulersProvider: SchedulersProvider
+  ): ObservableTransformer<BloodSugarSummaryViewEffect, BloodSugarSummaryViewEvent> {
+    return RxMobius
+        .subtypeEffectHandler<BloodSugarSummaryViewEffect, BloodSugarSummaryViewEvent>()
+        .addTransformer(FetchBloodSugarSummary::class.java) { effect ->
+          effect
+              .flatMap {
+                bloodSugarRepository.latestMeasurements(it.patientUuid, 100)
```
vinaysshenoy: Instead of a magic number, the value should be read from a config. Is this going to happen in another PR?

app/src/test/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandlerTest.kt
```diff
@@ -1,20 +1,30 @@
...
 class BloodSugarSummaryViewEffectHandlerTest {
 
   @Test
   fun `when fetch blood sugar effect is received then blood sugar should be fetched`() {
     //given
-    val testCase = EffectHandlerTestCase(BloodSugarSummaryViewEffectHandler.create())
+    val bloodSugarRepository = mock<BloodSugarRepository>()
+    val testCase = EffectHandlerTestCase(BloodSugarSummaryViewEffectHandler.create(bloodSugarRepository, TrampolineSchedulersProvider()))
+    val measurements = listOf<BloodSugarMeasurement>()
+    whenever(bloodSugarRepository.latestMeasurements(any(), any())).thenReturn(Observable.just(measurements))
```
vinaysshenoy: Instead of using catch-all matchers (`any()`), we should be using the values that we expect to be called. Otherwise, this might end up passing when it should fail.

An example would be if the limit to read blood sugars was being passed via a config, but the production source was not actually using the config and just hardcoding the value. Letting it be `any()` would make the test pass regardless of whether the parameter was read from the config or not and will hide a bug.

We could make this `whenever(bloodSugarRepository.latestMeasurements(patientUuid, 100)).thenReturn(Observable.just(measurements))` and then change it later to the config value when we implement the config.

app/src/main/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewModel.kt
```diff
@@ -1,16 +1,20 @@
 package org.simple.clinic.summary.bloodsugar
 
+import org.simple.clinic.bloodsugar.BloodSugarMeasurement
 import java.util.UUID
 
-data class BloodSugarSummaryViewModel(val patientUuid: UUID) {
+data class BloodSugarSummaryViewModel(
+    val patientUuid: UUID,
+    val measurements: List<BloodSugarMeasurement>
```
vinaysshenoy: When I see an empty list here, it could possibly be for two reasons:

- Either the data was never queried
- Or the data was queried, and was found to be empty

Which means that treating these two cases separately is not possible when representing in this manner. We've been using nullable lists in models for this reason. Can we make this a nullable list instead?

app/src/main/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandler.kt
```diff
@@ -2,25 +2,36 @@ package org.simple.clinic.summary.bloodsugar
...
         .subtypeEffectHandler<BloodSugarSummaryViewEffect, BloodSugarSummaryViewEvent>()
-        .addTransformer(FetchBloodSugarSummary::class.java) { effect ->
-          effect
-              .flatMap {
-                bloodSugarRepository.latestMeasurements(it.patientUuid, 100)
-                    .subscribeOn(schedulersProvider.io())
-              }
-              .map { BloodSugarSummaryFetched(it) }
-        }
+        .addTransformer(FetchBloodSugarSummary::class.java, fetchBloodSugarMeasurements(bloodSugarRepository, schedulersProvider.ui()))
```
vinaysshenoy: This commit is doing two things:

- Extracting this transformer into a separate method, and
- Adding an action to open the sheet

It is not necessary to change this, but we should try to restrict each commit to only one change.

app/src/test/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandlerTest.kt
```diff
@@ -27,4 +30,14 @@ class BloodSugarSummaryViewEffectHandlerTest {
...
   }
+
+  @Test
+  fun `when open blood sugar type selector effect is received then type selector sheet should be opened`() {
+    //when
+    testCase.dispatch(OpenBloodSugarTypeSelector)
+
+    //then
+    testCase.assertNoOutgoingEvents()
+    verify(uiActions).showBloodSugarTypeSelector()
```
vinaysshenoy: We should also use `verifyNoMoreActions(ui)` here to check that nothing else got triggered accidentally.

app/src/test/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandlerTest.kt
```diff
@@ -27,4 +30,14 @@ class BloodSugarSummaryViewEffectHandlerTest {
     //then
     testCase.assertOutgoingEvents(BloodSugarSummaryFetched(measurements))
```
vinaysshenoy: Can we add a `verifyZeroInteractions(ui)` here?

app/src/main/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandler.kt
sanchita: Yes. That will be handled in the next PR.

app/src/main/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandler.kt
sanchita: Since the EffectHandlers are written differently in different places, I wasn't sure which one to follow. I will make this change.

app/src/main/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewModel.kt
sanchita: I would want to do this when I write the `UiRenderer` if that makes sense.
I will have to handle the `No blood sugars` state.

app/src/main/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandler.kt
sanchita: I am not adding the transformer in this commit. I am just refactoring the code to a separate method since there are now more than one effects being handled.

vinaysshenoy: Yeah, the transformer was added in a previous commit. What I meant was in this commit, we are both extracting the method as well as adding another action.

It's not necessary to split this commit right now. Just a comment for future PRs.

app/src/test/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandlerTest.kt
sanchita: Done

app/src/test/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewEffectHandlerTest.kt
sanchita: Done

PR #1029, opened 2020-01-06
Add CI step for verifying release builds

vinaysshenoy:


vinaysshenoy:
Moving this back to WIP because the release verification takes too long. Will look at providing an alternate proguard configuration just for running verification checks and then will update this.

PR #1033, opened 2020-01-06
Add BloodSugarSummaryViewUiRenderer

sanchita:
https://www.pivotaltracker.com/story/show/169675670


app/src/test/java/org/simple/clinic/summary/bloodsugar/BloodSugarSummaryViewUiRendererTest.kt
```diff
@@ -0,0 +1,49 @@
...
+    //when
+    renderer.render(defaultModel.summaryFetched(bloodSugars))
+
+    //then
+    verify(ui).showBloodSugarSummary(bloodSugars)
+    verifyNoMoreInteractions(ui)
+  }
+
+  @Test
+  fun `when blood sugar measurement is empty then show empty ui state`() {
```
msasikanth: Should we rename it to `when blood sugar summary is empty, then show empty ui state`?

PR #1035, opened 2020-01-07
Migrate patient summary screen to Mobius #3: Extract medical history summary view controller

vinaysshenoy:
https://www.pivotaltracker.com/story/show/170538616

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
@@ -21,7 +21,6 @@ import org.simple.clinic.drugs.selection.PrescribedDrugsScreenKey
 import org.simple.clinic.editpatient.EditPatientScreenKey
 import org.simple.clinic.home.HomeScreenKey
 import org.simple.clinic.main.TheActivity
-import org.simple.clinic.medicalhistory.MedicalHistory
 import org.simple.clinic.patient.DateOfBirth
 import org.simple.clinic.patient.Gender
 import org.simple.clinic.patient.Patient
@@ -45,7 +44,6 @@ import org.simple.clinic.summary.updatephone.UpdatePhoneNumberDialog
 import org.simple.clinic.util.Just
 import org.simple.clinic.util.None
 import org.simple.clinic.util.Optional
-import org.simple.clinic.util.RelativeTimestampGenerator
 import org.simple.clinic.util.Truss
 import org.simple.clinic.util.Unicode
 import org.simple.clinic.util.UserClock
@@ -60,7 +58,7 @@ import java.util.UUID
 import javax.inject.Inject
 import javax.inject.Named
 
-class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayout(context, attrs), PatientSummaryScreenUi, MedicalHistorySummaryUi {
+class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayout(context, attrs), PatientSummaryScreenUi {
 
   @Inject
   lateinit var screenRouter: ScreenRouter
@@ -79,9 +77,6 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
 
   @field:[Inject Named("exact_date")]
   lateinit var exactDateFormatter: DateTimeFormatter
-
-  @Inject
-  lateinit var timestampGenerator: RelativeTimestampGenerator
 
   @Deprecated("""
     ~ DOA ~
@@ -290,14 +285,6 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
     )
   }
 
-  override fun populateMedicalHistory(medicalHistory: MedicalHistory) {
-    medicalHistorySummaryView.bind(
-        medicalHistory = medicalHistory,
-        lastUpdatedAt = timestampGenerator.generate(medicalHistory.updatedAt, userClock),
-        dateFormatter = exactDateFormatter
-    )
-  }
-
   override fun showScheduleAppointmentSheet(patientUuid: UUID) {
     val intent = ScheduleAppointmentSheet.intent(context, patientUuid)
     activity.startActivityForResult(intent, SUMMARY_REQCODE_SCHEDULE_APPOINTMENT)
@@ -340,7 +327,7 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
   }
 
   override fun medicalHistorySummaryUi(): MedicalHistorySummaryUi {
-    return this
+    return medicalHistorySummaryView
   }
 }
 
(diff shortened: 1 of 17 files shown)


app/src/main/java/org/simple/clinic/summary/medicalhistory/MedicalHistorySummaryUiController.kt
```diff
@@ -21,6 +21,10 @@ class MedicalHistorySummaryUiController(
 ) : ObservableTransformer<UiEvent, UiChange> {
 
   override fun apply(events: Observable<UiEvent>): ObservableSource<UiChange> {
+    return displayMedicalHistory(events)
```
vinaysshenoy: This commit is actually reordered to be before the previous one. This is not reflected in the PR since it was force-pushed later.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -281,16 +281,15 @@ class PatientSummaryScreen(context: Context, attrs: AttributeSet) : RelativeLayo
     }
   }
 
-  override fun populateList(
-      prescribedDrugs: List<PrescribedDrug>,
-      medicalHistory: MedicalHistory
-  ) {
+  override fun populateList(prescribedDrugs: List<PrescribedDrug>) {
```
sanchita: Rename to `populateDrugs()`

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
```diff
@@ -157,7 +157,7 @@ class PatientSummaryScreenController @AssistedInject constructor(
         .ofType<PatientSummaryItemChanged>()
```
sanchita: Rename this method to `populateDrugs`

app/src/test/java/org/simple/clinic/summary/medicalhistory/MedicalHistorySummaryUiControllerTest.kt
```diff
@@ -0,0 +1,65 @@
...
+    // then
+    verify(medicalHistorySummaryUi).populateMedicalHistory(medicalHistory)
+    verifyNoMoreInteractions(medicalHistorySummaryUi)
+  }
+
+  private fun setupController() {
+    controller = MedicalHistorySummaryUiController(patientUuid, repository)
+    controllerSubscription = events.compose(controller).subscribe { it.invoke(ui) }
+
+    events.onNext(ScreenCreated())
```
sanchita: I feel the events being triggered should be part of the tests themselves and not the setupController() method.
As the name of the method suggests, I would only expect some initialization to happen and not some events being sent.

vinaysshenoy: You are right about the `setupController` being method being expected to do only initialization, and no events being triggered here.

However, I feel that the `ScreenCreated` event is kind of a special case. The only reason it exists is because our v1 architecture does not have any way to be notified when the view is "ready" since it is modelled on a stream. The only reason that `ScreenCreated` exists is so that we can send this to the stream as a "trigger".

If we consider the Mobius architecture, we do not need to use `ScreenCreated` here because that job is being done by the `Init` function. Regardless of whether screen is written in Mobius or the architecture v1, the (controller) tests for them will end up having a general structure like so:


```kotlin
fun test() {
  // given
  // set-up mocks here

  // when
  // simulate screen set-up here
  // trigger user events

  // then
  // verify
}
```

If we move the emission of the `ScreenCreated` event outside the controller setup and into the `trigger user events` section, then we will have to continuously remove this line for every test that is migrated to Mobius. This is an indication that it is not truly a "user" event but something that is specific to the v1 controller.

Does this make sense?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenController.kt
vinaysshenoy: I intentionally did not name this because this method will be removed in #1038.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: I intentionally did not rename this because this method will be removed in #1038.

sanchita: Yea I got that. But think it is always a good idea to ensure every PR is at least complete in itself for other people who are working on the code.
It is okay to not change this for now.

app/src/test/java/org/simple/clinic/summary/medicalhistory/MedicalHistorySummaryUiControllerTest.kt
sanchita: Given that we are not re-writing tests for now. We are still going to continue reading the existing `LogicTests`. I agree that `ScreenCreated` is not relevant for Mobius architecture. But given that these tests are not written in the Mobius format either, I think it is ok to be explicit that this event is being triggered after the controller is setup. That is how we are creating this event in the older architecture. 
My point is that this is sort of a hidden and confusing behavior for someone reading the tests. Also why do we need to setup the controller inside every test? Is it because of the event?
We can extract this to the `setup` method of of the test suite since we are testing the controller, it is safe to assume we would need it to be setup for every test. Any specific reason for doing so?

vinaysshenoy: > I agree that ScreenCreated is not relevant for Mobius architecture. But given that these tests are not written in the Mobius format either, I think it is ok to be explicit that this event is being triggered after the controller is setup. That is how we are creating this event in the older architecture.

I forgot to mention that `ScreenCreated` will end up getting removed from the older style controller tests when we migrate the controller to Mobius. If you look at `BloodPressureEntrySheetLogicTest`, the controller was migrated to Mobius. We no longer have any `ScreenCreated` events here.

> Also why do we need to setup the controller inside every test?
We can extract this to the setup method of of the test suite since we are testing the controller, it is safe to assume we would need it to be setup for every test. Any specific reason for doing so?

Yes. Our current style of setting up controllers in the `@Before` fixture does not allow us to setup stubs before the controller is initialized. Since the controller subscribes to the streams before the mocks are setup in many cases, this makes testing in a lot of cases hard and we need to jump through hoops (like using a `PublishSubject`, for instance) to setup the test correctly. This is why as a matter of policy, we will no longer setup the controller in the `@Before` fixture and will manually do this in every test before refactoring to Mobius.

If you look at all the older controller tests where they've been moved to Mobius, we've done this all the time to make the refactoring and testing easier.

These are some of the older style tests that we have migrated:
- `BloodPressureEntrySheetLogicTest`
- `PatientEntryScreenLogicTest`
- `AllPatientsInFaciltyLogicTest`

sanchita: So the reason for using this method and calling it in each test is that in the future once we move this controller to Mobius, we would need to do instantiate the controller in each test?

vinaysshenoy: Yup, we'll do this in future PRs.

PR #1039, opened 2020-01-08
Automate code formatting fixes and raising PR using GitHub Action

msasikanth:
`./gradlew spotlessApply` will run daily at 5 AM on `master` branch and raise a PR if there are any changes

### DIFF
--- a/.github/workflows/code_formatting_checks.yml
+++ b/.github/workflows/code_formatting_checks.yml
@@ -1,0 +1,42 @@
+name: Code formatting
+
+# Scheduled at 11:30 PM UTC (05:00 AM IST)
+# TODO: Replace on: push with cron job
+on: push
+
+jobs:
+  fix_code_formatting:
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Fix code formatting issues
+        run: ./gradlew spotlessApply
+
+      - name: Commit and Push
+        run: ./scripts/commit_and_push.sh ${{secrets.USER_NAME}} ${{secrets.USER_EMAIL}}
+
+      - name: Create PR
+        uses: actions/github-script@0.3.0
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/spotless_checks.yml
```diff
@@ -0,0 +1,17 @@
+name: Spotless Checks
+
+on:
+  schedule:
+    - cron: "0 5 * * *"
```
vinaysshenoy: When I said `5 AM`, I meant it in IST. I think this will run at 5 AM UTC. Could you update this so it runs in IST, please?

.github/workflows/spotless_checks.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  build:
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v1
+      - name: set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Spotless Apply
```
vinaysshenoy: This is exposing implementation detail that we are using Spotless to run the checks. Could we change this name to something that is tooling agnostic, like `Run code formatting checks`?

.github/workflows/spotless_checks.yml
```diff
@@ -0,0 +1,17 @@
...
+        run: ./gradlew spotlessCheck
```
vinaysshenoy: This action does not contribute anything useful at the moment because it neither raises a PR with the updated formatting, nor pushes the result of the checks somewhere. Does `spotlessCheck` allow us to generate and upload a report of some sort?

.github/workflows/spotless_checks.yml
msasikanth: At this moment there is no time zone parameter. So I can just change the time to 11:30 PM UTC

vinaysshenoy: Yup, that works.

.github/workflows/spotless_checks.yml
msasikanth: I am marking this PR as WIP for now. I will see if I can add the PR creation as well

PR #1040, opened 2020-01-08
Fix flaky test when recording a patient

vinaysshenoy:
The test generator would use a random number in the range [0, 100] for the age when creating a patient. However, we changed the age limit to have a minimum value of 1. This
ended up causing flakiness in some test runs where the random age
generator would create a patient with an age = 0.

This commit fixes the issue by extracting the minimum and maximum age
limits to global constants and using that in the test data generation.


msasikanth:
Can you rebase it onto master

vinaysshenoy:
> Can you rebase it onto`master `

Done.

PR #1043, opened 2020-01-09
Add Mobius update tests for showing validation errors

janhavisinghh:
https://www.pivotaltracker.com/story/show/169030605

### DIFF
--- a/app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
+++ b/app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
@@ -64,7 +64,7 @@ class PatientEntryUpdateTest {
   fun `when the user leaves full name field empty, then show error`() {
     //given
     val errors: List<PatientEntryValidationError> = listOf(PatientEntryValidationError.FULL_NAME_EMPTY)
-    val givenModel = defaultModel
+    val model = defaultModel
         .fullNameChanged("")
         .ageChanged("21")
         .genderChanged(Just(Gender.Male))
@@ -76,7 +76,7 @@ class PatientEntryUpdateTest {
         .zoneChanged("zone")
 
     updateSpec
-        .given(givenModel)
+        .given(model)
         .`when`(SaveClicked)
         .then(
             assertThatNext(
@@ -90,7 +90,7 @@ class PatientEntryUpdateTest {
   fun `when the user enters phone number which is too short, then show error`() {
     //given
     val errors: List<PatientEntryValidationError> = listOf(PatientEntryValidationError.PHONE_NUMBER_LENGTH_TOO_SHORT)
-    val givenModel = defaultModel
+    val model = defaultModel
         .fullNameChanged("Name")
         .ageChanged("21")
         .genderChanged(Just(Gender.Male))
@@ -102,7 +102,7 @@ class PatientEntryUpdateTest {
         .zoneChanged("zone")
 
     updateSpec
-        .given(givenModel)
+        .given(model)
         .`when`(SaveClicked)
         .then(
             assertThatNext(
@@ -116,7 +116,7 @@ class PatientEntryUpdateTest {
   fun `when the user enters phone number which is too long, then show error`() {
     //given
     val errors: List<PatientEntryValidationError> = listOf(PatientEntryValidationError.PHONE_NUMBER_LENGTH_TOO_LONG)
-    val givenModel = defaultModel
+    val model = defaultModel
         .fullNameChanged("Name")
         .ageChanged("21")
         .genderChanged(Just(Gender.Male))
@@ -128,7 +128,7 @@ class PatientEntryUpdateTest {
         .zoneChanged("zone")
 
     updateSpec
-        .given(givenModel)
+        .given(model)
         .`when`(SaveClicked)
         .then(
             assertThatNext(
@@ -143,7 +143,7 @@ class PatientEntryUpdateTest {
   fun `when the user doesn't enter date of birth and age, then show error`() {
     //given
     val errors: List<PatientEntryValidationError> = listOf(PatientEntryValidationError.BOTH_DATEOFBIRTH_AND_AGE_ABSENT)
-    val givenModel = defaultModel
+    val model = defaultModel
         .fullNameChanged("Name")
         .genderChanged(Just(Gender.Male))
         .phoneNumberChanged("7721084840")
@@ -154,7 +154,7 @@ class PatientEntryUpdateTest {
         .zoneChanged("zone")
 
     updateSpec
-        .given(givenModel)
+        .given(model)
         .`when`(SaveClicked)
         .then(
             assertThatNext(
@@ -168,7 +168,7 @@ class PatientEntryUpdateTest {
   fun `when the user enters invalid date of birth, then show error`() {
     //given
     val errors: List<PatientEntryValidationError> = listOf(PatientEntryValidationError.INVALID_DATE_OF_BIRTH)
-    val givenModel = defaultModel
+    val model = defaultModel
         .fullNameChanged("Name")
         .dateOfBirthChanged("02-19-2000")
         .genderChanged(Just(Gender.Male))
@@ -180,7 +180,7 @@ class PatientEntryUpdateTest {
         .zoneChanged("zone")
 
     updateSpec
-        .given(givenModel)
+        .given(model)
         .`when`(SaveClicked)
         .then(
             assertThatNext(
@@ -194,7 +194,7 @@ class PatientEntryUpdateTest {
   fun `when the user enters date of birth in future, then show error`() {
     //given
     val errors: List<PatientEntryValidationError> = listOf(PatientEntryValidationError.DATE_OF_BIRTH_IN_FUTURE)
-    val givenModel = defaultModel
+    val model = defaultModel
         .fullNameChanged("Name")
         .dateOfBirthChanged("02/02/4000")
         .genderChanged(Just(Gender.Male))
@@ -206,7 +206,7 @@ class PatientEntryUpdateTest {
         .zoneChanged("zone")
 
     updateSpec
-        .given(givenModel)
+        .given(model)
         .`when`(SaveClicked)
         .then(
             assertThatNext(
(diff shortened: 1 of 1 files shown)


app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
```diff
@@ -54,4 +58,32 @@ class PatientEntryUpdateTest {
...
+        .phoneNumberChanged("7721084840")
+        .streetAddressChanged("street")
+        .colonyOrVillageChanged("village")
+        .districtChanged("district")
+        .stateChanged("state")
+        .zoneChanged("zone")
+
+    updateSpec
+        .given(givenModel)
+        .`when`(SaveClicked)
```
vinaysshenoy: Can we use `whenEvent` instead of `when` please?

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
```diff
@@ -54,4 +58,32 @@ class PatientEntryUpdateTest {
             )
         )
   }
+
+  //Tests for validation errors in Patient Entry Screen
+  @Test
+  fun `when the user leaves full name field empty, then show error`() {
+    //given
+    val errors: List<PatientEntryValidationError> = listOf(PatientEntryValidationError.FULL_NAME_EMPTY)
+    val givenModel = defaultModel
```
vinaysshenoy: We should just rename this to `model` since `givenModel` is specific to the `UpdateSpec`.

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
```diff
@@ -86,4 +86,30 @@ class PatientEntryUpdateTest {
         )
   }
 
+  @Test
+  fun `when the user enters phone number which is too short, then show error`() {
+    //given
```
vinaysshenoy: Not necessary to change this, but we can skip the `// given`, `// when`, and `// then` sections for mobius `Init` and `Update` tests since they are a part of the `Spec` test helpers.

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
```diff
@@ -190,4 +190,30 @@ class PatientEntryUpdateTest {
         )
   }
 
+  @Test
+  fun `when the user enters date of birth in future, then show error`() {
+    //given
+    val errors: List<PatientEntryValidationError> = listOf(PatientEntryValidationError.DATE_OF_BIRTH_IN_FUTURE)
+    val givenModel = defaultModel
+        .fullNameChanged("Name")
+        .dateOfBirthChanged("02/02/4000")
```
vinaysshenoy: Since the validation is done using the `ageValidator`, which takes in a `UserClock`, we can set the date in the clock to a particular `LocalDate` (the default is the Unix EPOCH) and then set the date in the model to one which is after that to validate this.

I feel that this exposes the intent in this test more than just setting a date at some distant future.

We can pair on this if you're unclear about what to do.

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
```diff
@@ -241,4 +241,54 @@ class PatientEntryUpdateTest {
         )
   }
 
+  @Test
```
vinaysshenoy: Can we reword the commit message to better reflect that we've done more than just add the district field error validation test?

Ex: `Add test for address fields validation checks`

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
```diff
@@ -316,4 +316,30 @@ class PatientEntryUpdateTest {
...
+  fun `when the age exceeds max limit, then show error`() {
+    //given
+    val errors: List<PatientEntryValidationError> = listOf(PatientEntryValidationError.AGE_EXCEEDS_MAX_LIMIT)
+    val givenModel = defaultModel
+        .fullNameChanged("Name")
+        .ageChanged("130")
```
vinaysshenoy: This can cause problems if we change the max allowed age in the app to more than 130. The test will have to be updated because of a hidden dependency which is not exposed in this test. 

We have declared a global constant for the max age (`MAX_ALLOWED_PATIENT_AGE`). Could you please use this value instead?

```kotlin
.ageChanged(MAX_ALLOWED_PATIENT_AGE.plus(1).toString())
```

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
```diff
@@ -342,4 +342,30 @@ class PatientEntryUpdateTest {
         )
   }
 
+  @Test
+  fun `when the age exceeds min limit, then show error`() {
+    //given
+    val errors: List<PatientEntryValidationError> = listOf(PatientEntryValidationError.AGE_EXCEEDS_MIN_LIMIT)
+    val givenModel = defaultModel
+        .fullNameChanged("Name")
+        .ageChanged("0")
```
vinaysshenoy: Use the static constant for minimum allowed patient age here?

```kotlin
.ageChanged(MIN_ALLOWED_PATIENT_AGE.minus(1).toString())
```

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
```diff
@@ -368,4 +368,30 @@ class PatientEntryUpdateTest {
         )
   }
 
+  @Test
+  fun `when the date of birth exceeds max limit, then show error`() {
+    //given
+    val errors: List<PatientEntryValidationError> = listOf(PatientEntryValidationError.DOB_EXCEEDS_MAX_LIMIT)
+    val givenModel = defaultModel
+        .fullNameChanged("Name")
+        .dateOfBirthChanged("01/02/1800")
```
vinaysshenoy: Since the `userInputAgeValidator` takes in a `UserClock` to compare dates, we can expose the intent more clearly in this test by using:

```kotlin
val enteredDate = formatter.format(LocalDate.now(userClock).minusYears(MAX_ALLOWED_PATIENT_AGE))

.dateOfBirthChanged(enteredDate)
```

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
vinaysshenoy: This applies to all the tests in this PR.

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
vinaysshenoy: This applies to all the tests in this PR.

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
vinaysshenoy: This applies to all the tests in this PR.

janhavisinghh: Resolved this issue in the next pr

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
janhavisinghh: Resolved

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
janhavisinghh: Okay resolved

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
janhavisinghh: resolved

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
janhavisinghh: resolved

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
janhavisinghh: Resolved

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
janhavisinghh: Resolved

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
janhavisinghh: resolved

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
janhavisinghh: resolved

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
janhavisinghh: resolved

app/src/test/java/org/simple/clinic/newentry/PatientEntryUpdateTest.kt
janhavisinghh: I have added a commit for each test that's why I wrote specific error names. the address field will have to combine 3 different tests and commits. Is that what you want me to do here? @vinaysshenoy

vinaysshenoy: If you look at the contents of this commit, it adds tests for both the `Colony/Village` field as well as the `District` field. The commit message just mentions that the test for the `District` field was added.

My intention is to make the commit message and the commit contents make sense when taken together.

janhavisinghh: Ok, resolved

janhavisinghh:
> PR looks good, overall. Left some comments, please take a look?

I have resolved all the issues, please check @vinaysshenoy

PR #1044, opened 2020-01-09
Automate code formatting

msasikanth:
- Runs daily at 5 AM IST / 11:30 PM UTC
- Automatically fixes code formatting issues using `./gradlew spotlessApply`
- Commit the changes to a new branch with format `bot/MMM-dd-yyyy/code-formatting`
- Raise a PR with `master` with title `[BOT] Code formatting fixes MMM-dd-yyyy`

**REQUIRED**:
- `USER_NAME` and `USER_EMAIL` have to be set in GitHub secrets

### DIFF
--- a/.github/scripts/commit_and_push.sh
+++ b/.github/scripts/commit_and_push.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+readonly GIT_HAS_CHANGES=1 # 0 = clean, 1 = dirty
+
+USER_NAME="$1"
+USER_EMAIL="$2"
+
+git config user.name "$USER_NAME"
+git config user.email "$USER_EMAIL"
+
+# Following git command checks the git diff of staged files
+# and assigns the exit code to GIT_DIFF_STATUS variable.
+git diff --quiet; GIT_DIFF_STATUS=$?
+
+# If the GIT_DIFF_STATUS is 1 (GIT_HAS_CHANGES) then we know
+# there are changes and we can commit the changes and push
+# them to new branch
+if [[ ${GIT_DIFF_STATUS} -eq ${GIT_HAS_CHANGES} ]]; then
+    DAY=$( date +%b-%d-%Y )
+    BRANCH_NAME="bot/$DAY/code-formatting"
+
+    git checkout -b "$BRANCH_NAME"
+    git commit -am "[BOT] Code formatting fixes"
+    git push -u origin "$BRANCH_NAME"
+
+    echo "Git changes pushed to $BRANCH_NAME"
+else
+    echo "No git changes"
+fi
--- a/.github/workflows/code_formatting_checks.yml
+++ b/.github/workflows/code_formatting_checks.yml
@@ -1,0 +1,47 @@
+name: Code formatting
+
+# Scheduled at 11:30 PM UTC (05:00 AM IST)
+on:
+  schedule:
+    - cron: "30 23 * * *"
+
+jobs:
+  fix_code_formatting:
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Fix code formatting issues
+        run: ./gradlew spotlessApply
+
+      - name: Commit and Push
+        run: .github/scripts/commit_and_push.sh "${{secrets.USER_NAME}}" "${{secrets.USER_EMAIL}}"
+
+      - name: Create PR
+        uses: actions/github-script@0.3.0
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            var monthNames = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
+            var date = new Date();
+            var monthName = monthNames[date.getMonth()];
+            var day = `${date.getDate()}`.padStart(2, 0);
+            day = `${monthName}-${day}-${date.getFullYear()}`
+            var branchName = `bot/${day}/code-formatting`
+
+            const branch = await github.repos.getBranch({
+                ...context.repo,
+                branch: branchName
+            });
+            await github.pulls.create({
+                ...context.repo,
+                title: `[BOT] Code formatting fixes ${day}`,
+                head: branch.data.name,
+                base: 'master'
+            });
+


.github/workflows/code_formatting_checks.yml
```diff
@@ -0,0 +1,55 @@
+name: Code formatting
+
+# Scheduled at 11:30 PM UTC (05:00 AM IST)
+on:
+  schedule:
+    - cron: "30 23 * * *"
```
vinaysshenoy: Since the minute is 0-based, this will actually run at 05:01 AM IST and not 05:00 AM.

.github/workflows/code_formatting_checks.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Fix code formatting issues
+        run: ./gradlew spotlessApply
+
+      - name: Commit and Push
+        run: ./scripts/commit_and_push.sh "${{secrets.USER_NAME}}" "${{secrets.USER_EMAIL}}"
```
vinaysshenoy: Can we move the scripts inside the `.github` folder since it is currently used only for the workflow?

.github/workflows/code_formatting_checks.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Create PR
+        uses: actions/github-script@0.3.0
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            var monthNames = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
+            var date = new Date();
+            var monthName = monthNames[date.getMonth()];
+            var day = `${date.getDate()}`;
+            if (day.length < 2) {
```
vinaysshenoy: JavaScript has a [padStart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart) method that we might be able to use instead of this.

.github/workflows/code_formatting_checks.yml
```diff
@@ -0,0 +1,55 @@
...
+                day = `0${date.getDate()}`;
+            }
+            day = `${monthName}-${day}-${date.getFullYear()}`
+            var branchName = `ga/${day}/code-formatting`
+            console.log(`Branch name is ${branchName}`)
```
vinaysshenoy: Is this log statement necessary anymore?

.github/workflows/code_formatting_checks.yml
```diff
@@ -0,0 +1,55 @@
...
+          script: |
+            var monthNames = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
+            var date = new Date();
+            var monthName = monthNames[date.getMonth()];
+            var day = `${date.getDate()}`;
+            if (day.length < 2) {
+                day = `0${date.getDate()}`;
+            }
+            day = `${monthName}-${day}-${date.getFullYear()}`
+            var branchName = `ga/${day}/code-formatting`
```
vinaysshenoy: Instead of `ga/` can we change the prefix to something that is platform agnostic? Say `bot`?

.github/workflows/code_formatting_checks.yml
vinaysshenoy: Never mind, this is my mistake.

.github/workflows/code_formatting_checks.yml
vinaysshenoy: We are depending on an implementation detail that the `commit_and_push` script sets the branch to this specific name. If we change the branch formatting there, we will also have to remember to change this here. I think we can set outputs from a step and then consume that as inputs in other steps (https://help.github.com/en/actions/automating-your-workflow-with-github-actions/metadata-syntax-for-github-actions#outputs).

Can we look at this to pass the created branch name to the `create PR` step?

.github/workflows/code_formatting_checks.yml
```diff
@@ -0,0 +1,49 @@
...
+            var branchName = `bot/${day}/code-formatting`
+
+            try {
+                const branch = await github.repos.getBranch({
+                    ...context.repo,
+                    branch: branchName
+                });
+                await github.pulls.create({
+                    ...context.repo,
+                    title: `Code formatting fixes on ${day}`,
```
vinaysshenoy: Can we prefix the PR title with `[BOT]` to indicate that this is an automated check? Maybe we can also skip adding the 'on' in the title.

This would make the title look something like `[BOT] Code formatting fixes Jan-05-2019`. What do you think?

.github/workflows/code_formatting_checks.yml
msasikanth: At the moment GitHub script action doesn't allow taking input. One possible way is writing our own javascript action that takes input and uses in the script

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-a-javascript-action

.github/workflows/code_formatting_checks.yml
```diff
@@ -0,0 +1,49 @@
...
+                    head: branch.data.name,
+                    base: 'master'
+                });
+            } catch (e) {
+                console.log(`${branchName} is not present`)
```
vinaysshenoy: Does this fail the action? If not, we should probably not wrap it in the `try-catch` block so that we know that the workflow failed for some reason.

.github/workflows/code_formatting_checks.yml
msasikanth: Yeah, this makes more sense. Changed it

.github/workflows/code_formatting_checks.yml
msasikanth: Removed the `try-catch`. Yes, when a PR with the title already exists or authentication fails the action will fail

.github/workflows/code_formatting_checks.yml
vinaysshenoy: I think we might be able to use it. Anyway, we can merge this in now and check back on this later, not a blocker to merge.

.github/scripts/commit_and_push.sh
```diff
@@ -0,0 +1,25 @@
...
+USER_EMAIL="$2"
+
+git config --global user.name "$USER_NAME"
+git config --global user.email "$USER_EMAIL"
+
+git diff --quiet; GIT_DIFF_STATUS=$?
+
+if [[ ${GIT_DIFF_STATUS} -eq ${GIT_HAS_CHANGES} ]]; then
+    DAY=$( date +%b-%d-%Y )
+    BRANCH_NAME="ga/$DAY/code-formatting"
```
vinaysshenoy: This should be `bot/`.

.github/scripts/commit_and_push.sh
```diff
@@ -0,0 +1,25 @@
...
+    echo "Branch name is $BRANCH_NAME"
+
+    git checkout -b "$BRANCH_NAME"
+    git commit -am "Code formatting fixes"
```
vinaysshenoy: Can we follow the convention for commit messages?

`[bot] Fix code formatting?`

.github/scripts/commit_and_push.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+readonly GIT_HAS_CHANGES=1 # 0 = clean, 1 = dirty
+
+USER_NAME="$1"
+USER_EMAIL="$2"
+
+git config --global user.name "$USER_NAME"
```
vinaysshenoy: Does this need to be a global git config? Will it work with a local git config as well?

.github/scripts/commit_and_push.sh
```diff
@@ -0,0 +1,25 @@
...
+git config --global user.email "$USER_EMAIL"
+
+git diff --quiet; GIT_DIFF_STATUS=$?
```
vinaysshenoy: Nice! Can we add a small comment that explains this?

msasikanth: Done

.github/scripts/commit_and_push.sh
msasikanth: It doesn't have to be since it's just a VM I set as global. But changed it to repository level.

PR #1046, opened 2020-01-10
Get branch name from `Commit and Push` step

msasikanth:
Instead of creating the branch name in `Create PR` step. Get the already created branch name from `Commit and Push` step and set it as env variable for `Create PR` step.

### DIFF
--- a/.github/workflows/code_formatting_checks.yml
+++ b/.github/workflows/code_formatting_checks.yml
@@ -20,28 +20,33 @@ jobs:
         run: ./gradlew spotlessApply
 
       - name: Commit and Push
+        id: commit_and_push
         run: .github/scripts/commit_and_push.sh "${{secrets.USER_NAME}}" "${{secrets.USER_EMAIL}}"
 
       - name: Create PR
         uses: actions/github-script@0.3.0
+        env:
+          branch_name: ${{steps.commit_and_push.outputs.branch_name}}
         with:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
-            var monthNames = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
-            var date = new Date();
-            var monthName = monthNames[date.getMonth()];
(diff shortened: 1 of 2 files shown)


.github/workflows/code_formatting_checks.yml
```diff
@@ -20,28 +20,35 @@ jobs:
...
+                    branch: branchName
+                });
+                await github.pulls.create({
+                    ...context.repo,
+                    title: `[BOT] Code formatting fixes ${day}`,
+                    head: branch.data.name,
+                    base: 'master'
+                });
+            } else {
+                console.log("Branch name is empty!")
```
vinaysshenoy: Is this necessary? Can we just remove the `else` block?

.github/workflows/code_formatting_checks.yml
```diff
@@ -20,28 +20,35 @@ jobs:
...
           script: |
-            var monthNames = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
-            var date = new Date();
-            var monthName = monthNames[date.getMonth()];
-            var day = `${date.getDate()}`.padStart(2, 0);
+            const monthNames = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
+            const date = new Date();
+            const monthName = monthNames[date.getMonth()];
+            let day = `${date.getDate()}`.padStart(2, '0');
             day = `${monthName}-${day}-${date.getFullYear()}`
```
vinaysshenoy: Can we create separate variables to track the `day` and the `date`? This will allow us to make all this variables a `const`.

.github/workflows/code_formatting_checks.yml
msasikanth: Sure

.github/workflows/code_formatting_checks.yml
msasikanth: Updated it. Please check now

PR #1047, opened 2020-01-10
Add Diabetes widget to PatientSummaryScreen

sanchita:
This is the last PR in the story -> https://www.pivotaltracker.com/story/show/169675670

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/bloodsugar/view/BloodSugarSummaryView.kt
+++ b/app/src/main/java/org/simple/clinic/summary/bloodsugar/view/BloodSugarSummaryView.kt
@@ -1,6 +1,8 @@
 package org.simple.clinic.summary.bloodsugar.view
 
+import android.annotation.SuppressLint
 import android.content.Context
+import android.content.Intent
 import android.os.Parcelable
 import android.util.AttributeSet
 import android.view.LayoutInflater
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/simple/clinic/summary/bloodsugar/view/BloodSugarSummaryView.kt
```diff
@@ -110,10 +118,12 @@ class BloodSugarSummaryView(
     if (isInEditMode) {
       return
     }
-
     context.injector<BloodSugarSummaryViewInjector>().inject(this)
 
     delegate.prepare()
+
+    val screenDestroys: Observable<ScreenDestroyed> = detaches().map { ScreenDestroyed() }
```
vinaysshenoy: Not necessary to change this, but `ScreenDestroyed` is only used as a signal to `ReplayUntilScreenIsDestroyed` to end the stream. Since this screen is not using Mobius, we can just use `detaches()` instead.

app/src/main/java/org/simple/clinic/summary/bloodsugar/view/BloodSugarSummaryView.kt
```diff
@@ -147,7 +157,23 @@ class BloodSugarSummaryView(
...
+  private fun openEntrySheetAfterTypeIsSelected(onDestroys: Observable<ScreenDestroyed>) {
+    screenRouter.streamScreenResults()
+        .ofType<ActivityResult>()
+        .filter { it.requestCode == TYPE_PICKER_SHEET && it.succeeded() }
+        .takeUntil(onDestroys)
+        .map { showBloodSugarEntrySheet(it) }
+        .subscribe()
+  }
+
+  private fun showBloodSugarEntrySheet(it: ActivityResult) {
```
vinaysshenoy: I would change the parameter of this method to accept the blood sugar measurement type directly, and then extract this out at the method call site. Otherwise, this method now knows the implementation detail that the type is selected via the `BloodSugarTypePickerSheet`.

PR #1056, opened 2020-01-13
Add migration to fix blank identifiers and invalid patients

vinaysshenoy:


app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration57AndroidTest.kt
```diff
@@ -0,0 +1,343 @@
...
+
+  @Before
+  override fun setUp() {
+    super.setUp()
+    TestClinicApp.appComponent().inject(this)
+  }
+
+  @Suppress("IllegalIdentifier")
+  @Test
+  fun migrating_to_57_should_mark_all_patients_with_blank_identifiers_as_sync_pending() {
```
rakshakhegde: This just needs one more patient with ID that has non-blank value.

vinaysshenoy: Done.

PR #1060, opened 2020-01-14
`BloodPressureHistoryScreen` - Part 2

msasikanth:
https://www.pivotaltracker.com/story/show/169677099

- Initial implementation of `BloodPressureHistoryScreen`

### DIFF
--- a/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
@@ -3,14 +3,30 @@ package org.simple.clinic.bp.history
 import android.content.Context
 import android.util.AttributeSet
 import androidx.constraintlayout.widget.ConstraintLayout
+import androidx.recyclerview.widget.LinearLayoutManager
+import kotlinx.android.synthetic.main.screen_bp_history.view.*
 import org.simple.clinic.bp.BloodPressureMeasurement
+import org.simple.clinic.bp.history.adapter.BloodPressureHistoryListItem
+import org.simple.clinic.bp.history.adapter.BloodPressureHistoryListItemDiffCallback
 import org.simple.clinic.di.injector
+import org.simple.clinic.summary.PatientSummaryConfig
+import org.simple.clinic.util.UtcClock
+import org.simple.clinic.widgets.ItemAdapter
 import java.util.UUID
+import javax.inject.Inject
 
 class BloodPressureHistoryScreen(
     context: Context,
     attrs: AttributeSet
 ) : ConstraintLayout(context, attrs), BloodPressureHistoryScreenUi, BloodPressureHistoryScreenUiActions {
+
+  @Inject
+  lateinit var utcClock: UtcClock
+
+  @Inject
+  lateinit var config: PatientSummaryConfig
+
+  private val bloodPressureHistoryAdapter = ItemAdapter(BloodPressureHistoryListItemDiffCallback())
 
   override fun onFinishInflate() {
     super.onFinishInflate()
@@ -18,9 +34,20 @@ class BloodPressureHistoryScreen(
       return
     }
     context.injector<BloodPressureHistoryScreenInjector>().inject(this)
+
+    setupBloodPressureHistoryList()
+  }
+
+  private fun setupBloodPressureHistoryList() {
+    bpHistoryList.apply {
+      setHasFixedSize(true)
+      layoutManager = LinearLayoutManager(context)
+      adapter = bloodPressureHistoryAdapter
+    }
   }
 
   override fun showBloodPressureHistory(bloodPressures: List<BloodPressureMeasurement>) {
+    bloodPressureHistoryAdapter.submitList(BloodPressureHistoryListItem.from(bloodPressures, config.bpEditableDuration, utcClock))
   }
 
   override fun openBloodPressureEntrySheet() {
--- a/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
@@ -1,0 +1,31 @@
+package org.simple.clinic.bp.history
+
+import android.content.Context
+import android.util.AttributeSet
+import androidx.constraintlayout.widget.ConstraintLayout
+import org.simple.clinic.bp.BloodPressureMeasurement
+import org.simple.clinic.di.injector
+import java.util.UUID
+
+class BloodPressureHistoryScreen(
+    context: Context,
+    attrs: AttributeSet
+) : ConstraintLayout(context, attrs), BloodPressureHistoryScreenUi, BloodPressureHistoryScreenUiActions {
+
+  override fun onFinishInflate() {
+    super.onFinishInflate()
+    if (isInEditMode) {
+      return
+    }
+    context.injector<BloodPressureHistoryScreenInjector>().inject(this)
+  }
+
+  override fun showBloodPressureHistory(bloodPressures: List<BloodPressureMeasurement>) {
+  }
+
+  override fun openBloodPressureEntrySheet() {
+  }
+
+  override fun openBloodPressureUpdateSheet(bpUuid: UUID) {
+  }
+}
--- a/app/src/main/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItem.kt
+++ b/app/src/main/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItem.kt
@@ -1,0 +1,73 @@
+package org.simple.clinic.bp.history.adapter
+
+import io.reactivex.subjects.Subject
+import kotlinx.android.synthetic.main.list_bp_history_item.*
+import org.simple.clinic.R
+import org.simple.clinic.bp.BloodPressureMeasurement
+import org.simple.clinic.bp.history.adapter.Event.BloodPressureClicked
+import org.simple.clinic.util.UtcClock
+import org.simple.clinic.widgets.ItemAdapter
+import org.simple.clinic.widgets.recyclerview.ViewHolderX
+import org.simple.clinic.widgets.visibleOrGone
+import org.threeten.bp.Duration
+import org.threeten.bp.Instant
+
+// TODO (sm) : Add `NewBpButtonItem`
+data class BloodPressureHistoryListItem(
+    val measurement: BloodPressureMeasurement,
+    val isBpEditable: Boolean,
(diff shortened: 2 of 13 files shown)


app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
```diff
@@ -0,0 +1,31 @@
+package org.simple.clinic.bp.history
```
vinaysshenoy: Reword the commit message to match our conventions?

Ex: `Set up the blood pressure history screen`

app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
```diff
@@ -0,0 +1,31 @@
...
+    context: Context,
+    attrs: AttributeSet
+) : ConstraintLayout(context, attrs), BloodPressureHistoryScreenUi, BloodPressureHistoryScreenUiActions {
+
+  override fun onFinishInflate() {
+    super.onFinishInflate()
+    if (isInEditMode) {
+      return
+    }
+    TheActivity.component.inject(this)
```
vinaysshenoy: We should use the newer way to inject screens instead of the older way.

```kotlin
context.injector<BloodPressureHistoryScreenInjector>().inject(this)
```

app/src/main/res/layout/list_bp_history_item.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:app="http://schemas.android.com/apk/res-auto"
+  xmlns:tools="http://schemas.android.com/tools"
+  android:layout_width="match_parent"
+  android:layout_height="wrap_content">
+
+  <LinearLayout
```
vinaysshenoy: Do we need another `LinearLayout` here given that the parent is a `ConstraintLayout`?

app/src/main/res/layout/list_bp_history_item.xml
```diff
@@ -0,0 +1,63 @@
...
+      android:layout_width="0dp"
+      android:layout_height="wrap_content"
+      android:layout_weight="1"
+      android:textAlignment="viewEnd"
+      android:textAppearance="@style/Clinic.V2.TextAppearance.Body2Center.Grey1"
+      tools:text="10-Aug-2019" />
+
+  </LinearLayout>
+
+  <View
```
vinaysshenoy: I've been steadily getting annoyed with having to manage dividers everywhere manually. I'm wondering if we should investigate using the [`DividerItemDecoration`](https://developer.android.com/reference/androidx/recyclerview/widget/DividerItemDecoration?hl=en) to do this automatically.

app/src/main/res/values/styles_v2.xml
```diff
@@ -728,6 +728,16 @@
     <item name="android:paddingStart">@dimen/spacing_16</item>
     <item name="android:paddingTop">@dimen/spacing_16</item>
   </style>
+
+  <style name="Clinic.V2.HistoryScreen.CardView" parent="Clinic.V2.CardView">
```
vinaysshenoy: I feel that layout attributes like margin, width, and height are specific to the screen the component is used in and hence should be part of the attributes in the parent screen. Whereas things like padding, background, etc are fine to place in the styles since they are common to the component no matter which screen it is placed in.

What do you think?

app/src/main/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItem.kt
```diff
@@ -0,0 +1,69 @@
...
+  override fun render(holder: ViewHolderX, subject: Subject<Event>) {
+    val context = holder.itemView.context
+
+    if (isHighBloodPressure) {
+      holder.heartImageView.setImageResource(R.drawable.bp_reading_high)
+    } else {
+      holder.heartImageView.setImageResource(R.drawable.bp_reading_normal)
+    }
+
+    if (isBpEditable) {
```
vinaysshenoy: We will need to clear the click listener if the BP is not editable since this view might be recycled.

app/src/main/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItem.kt
```diff
@@ -0,0 +1,69 @@
...
+import org.simple.clinic.widgets.recyclerview.ViewHolderX
+import org.threeten.bp.Duration
+import org.threeten.bp.Instant
+
+// TODO (sm) : Add `NewBpButtonItem`
+data class BloodPressureHistoryListItem(
+    val measurement: BloodPressureMeasurement,
+    val isBpEditable: Boolean,
+    val isHighBloodPressure: Boolean,
+    val showDivider: Boolean
```
vinaysshenoy: This property doesn't seem to be used anywhere.

app/src/test/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItemTest.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.simple.clinic.util.TestUtcClock
+import org.threeten.bp.Duration
+import org.threeten.bp.Instant
+import org.threeten.bp.temporal.ChronoUnit.MINUTES
+import java.util.UUID
+
+class BloodPressureHistoryListItemTest {
+
+  @Test
+  fun `if blood pressure is editable, then user can edit blood pressure`() {
```
vinaysshenoy: Reword this test?

Ex: `if the blood pressure was created within the BP editable duration, then the bp list item must be editable`

app/src/test/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItemTest.kt
```diff
@@ -0,0 +1,81 @@
...
+    // when
+    val createdAt = Instant.parse("2020-01-01T00:00:00Z")
+    val bloodPressureNow = PatientMocker.bp(
+        uuid = UUID.fromString("78e876ae-3055-43d6-a132-7ad5dd930e23"),
+        systolic = 120,
+        diastolic = 70,
+        createdAt = createdAt
+    )
+    val bloodPressureADayInFuture = PatientMocker.bp(
```
vinaysshenoy: The name does not match the time that was used to create the BP (`aDayInTheFuture` vs `15 minutes`).

app/src/test/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItemTest.kt
```diff
@@ -0,0 +1,81 @@
...
+        uuid = UUID.fromString("2b929a33-e543-4a4f-a98e-5d0d3e8e1e03"),
+        systolic = 120,
+        diastolic = 70,
+        createdAt = createdAt.minus(15, MINUTES)
+    )
+    val utcClock = TestUtcClock(Instant.parse("2020-01-01T00:00:00Z"))
```
vinaysshenoy: We can use `createdAt` here instead of instantiating the same `Instant` again.

app/src/test/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItemTest.kt
```diff
@@ -0,0 +1,81 @@
...
+    // then
+    assertThat(listItems)
+        .containsExactly(
+            BloodPressureHistoryListItem(measurement = bloodPressureNow, isBpEditable = true, isHighBloodPressure = false, showDivider = true),
+            BloodPressureHistoryListItem(measurement = bloodPressureADayInFuture, isBpEditable = false, isHighBloodPressure = false, showDivider = false)
+        )
+  }
+
+  @Test
+  fun `if blood pressure is high, then show high blood pressure`() {
```
vinaysshenoy: Reword this test?

Ex: `if the blood pressure reading is high, then it must be displayed as a high BP`?

app/src/test/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItemTest.kt
```diff
@@ -0,0 +1,81 @@
...
+    // then
+    assertThat(listItems)
+        .containsExactly(
+            BloodPressureHistoryListItem(measurement = bloodPressureNormal, isBpEditable = true, isHighBloodPressure = false, showDivider = true),
+            BloodPressureHistoryListItem(measurement = bloodPressureHigh, isBpEditable = true, isHighBloodPressure = true, showDivider = false)
+        ).inOrder()
+  }
+
+  @Test
+  fun `if there is only one list item, then divider should not be shown`() {
```
vinaysshenoy: We should probably also add another test to verify that dividers are added between each item if there are more than two BPs.

app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
```diff
@@ -2,25 +2,58 @@ package org.simple.clinic.bp.history
...
+  @Inject
+  lateinit var utcClock: UtcClock
+
+  @Inject
+  lateinit var config: PatientSummaryConfig
+
+  private val bloodPressureHistoryAdapter = ItemAdapter(BloodPressureHistoryListItemDiffCallback())
+
+  init {
+    LayoutInflater.from(context).inflate(R.layout.screen_bp_history, this, true)
```
vinaysshenoy: This should not be necessary since `screenRouter` will automatically inflate the view.

app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
msasikanth: Didn't knew we switched to an extension. Fixed it 👍

app/src/main/res/layout/list_bp_history_item.xml
msasikanth: We can use that, but we may have to extend it to take `marginStart` and `marginEnd`

app/src/main/res/values/styles_v2.xml
msasikanth: Moved margin, height and width specific attributes to layout

app/src/main/res/layout/list_bp_history_item.xml
vinaysshenoy: Alright, let's explore this once the feature is complete.

PR #1064, opened 2020-01-17
Update Firebase Remote Config logging

vinaysshenoy:
https://www.pivotaltracker.com/story/show/170741951

Updated Firebase release notes: https://firebase.google.com/support/release-notes/android#latest_sdk_versions


app/src/main/java/org/simple/clinic/remoteconfig/firebase/FirebaseRemoteConfigService.kt
```diff
@@ -2,34 +2,28 @@ package org.simple.clinic.remoteconfig.firebase
...
+    private val crashReporter: dagger.Lazy<CrashReporter>
 ) : RemoteConfigService {
 
   override fun reader(): ConfigReader = FirebaseConfigReader(firebaseRemoteConfig)
 
   override fun update(): Completable {
     return firebaseRemoteConfig
-        .fetch(cacheExpirationDuration.seconds)
+        .fetchAndActivate()
         .toCompletable { Timber.w("Failed to update Firebase remote config") }
```
rakshakhegde: Let's change it to this? Because right now, it looks like, when task completes it means that the firebase remot econfig failed to update, which is not the case.
`.toCompletable(onTaskUnsuccessful = { Timber.w("Failed to update Firebase remote config") })`

app/src/debug/java/org/simple/clinic/remoteconfig/firebase/FirebaseRemoteConfigModule.kt
```diff
@@ -15,23 +16,14 @@ class FirebaseRemoteConfigModule {
   @AppScope
   fun remoteConfig(): FirebaseRemoteConfig {
     return FirebaseRemoteConfig.getInstance().apply {
-      // Enable developer mode so that calls are not throttled during dev.
-      // More details in FirebaseRemoteConfigCacheExpiration.
-      setConfigSettings(FirebaseRemoteConfigSettings.Builder()
-          .setDeveloperModeEnabled(true)
+      val settings = FirebaseRemoteConfigSettings
+          .Builder()
+          .setMinimumFetchIntervalInSeconds(Duration.ofMinutes(1).seconds)
```
rakshakhegde: Let's change it to this just as it was before?
`Duration.ofMinutes(0)`

vinaysshenoy: Will do.

app/src/main/java/org/simple/clinic/remoteconfig/firebase/FirebaseRemoteConfigService.kt
vinaysshenoy: Good point, I'll make this change.

PR #1068, opened 2020-01-20
`BloodPressureHistoryScreen` - Part 3

msasikanth:
https://www.pivotaltracker.com/story/show/169677099

- Wire Mobius Loop with UI

### DIFF
--- a/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
@@ -4,13 +4,20 @@ import android.content.Context
 import android.util.AttributeSet
 import androidx.constraintlayout.widget.ConstraintLayout
 import androidx.recyclerview.widget.LinearLayoutManager
+import io.reactivex.Observable
+import io.reactivex.rxkotlin.cast
+import io.reactivex.rxkotlin.ofType
 import kotlinx.android.synthetic.main.screen_bp_history.view.*
+import org.simple.clinic.ReportAnalyticsEvents
 import org.simple.clinic.bp.BloodPressureMeasurement
 import org.simple.clinic.bp.history.adapter.BloodPressureHistoryListItem
 import org.simple.clinic.bp.history.adapter.BloodPressureHistoryListItemDiffCallback
+import org.simple.clinic.bp.history.adapter.Event.AddNewBpClicked
+import org.simple.clinic.bp.history.adapter.Event.BloodPressureHistoryItemClicked
 import org.simple.clinic.di.injector
 import org.simple.clinic.summary.PatientSummaryConfig
 import org.simple.clinic.util.UtcClock
+import org.simple.clinic.util.unsafeLazy
 import org.simple.clinic.widgets.ItemAdapter
 import java.util.UUID
 import javax.inject.Inject
@@ -27,6 +34,16 @@ class BloodPressureHistoryScreen(
   lateinit var config: PatientSummaryConfig
 
   private val bloodPressureHistoryAdapter = ItemAdapter(BloodPressureHistoryListItemDiffCallback())
+
+  private val events: Observable<BloodPressureHistoryScreenEvent> by unsafeLazy {
+    Observable
+        .merge(
+            addNewBpClicked(),
+            bloodPressureClicked()
+        )
+        .compose(ReportAnalyticsEvents())
+        .cast<BloodPressureHistoryScreenEvent>()
+  }
 
   override fun onFinishInflate() {
     super.onFinishInflate()
@@ -54,5 +71,20 @@ class BloodPressureHistoryScreen(
   }
 
   override fun openBloodPressureUpdateSheet(bpUuid: UUID) {
+  }
+
+  private fun addNewBpClicked(): Observable<BloodPressureHistoryScreenEvent> {
+    return bloodPressureHistoryAdapter
+        .itemEvents
+        .ofType<AddNewBpClicked>()
+        .map { NewBloodPressureClicked }
+  }
+
+  private fun bloodPressureClicked(): Observable<BloodPressureHistoryScreenEvent> {
+    return bloodPressureHistoryAdapter
+        .itemEvents
+        .ofType<BloodPressureHistoryItemClicked>()
+        .map { it.measurement }
+        .map(::BloodPressureClicked)
   }
 }
--- a/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
@@ -11,6 +11,7 @@ import io.reactivex.rxkotlin.ofType
 import kotlinx.android.synthetic.main.screen_bp_history.view.*
 import org.simple.clinic.ReportAnalyticsEvents
 import org.simple.clinic.bp.BloodPressureMeasurement
+import org.simple.clinic.bp.entry.BloodPressureEntrySheet
 import org.simple.clinic.bp.history.adapter.BloodPressureHistoryListItem
 import org.simple.clinic.bp.history.adapter.BloodPressureHistoryListItemDiffCallback
 import org.simple.clinic.bp.history.adapter.Event.AddNewBpClicked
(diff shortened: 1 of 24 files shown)


app/src/main/res/layout/list_bp_history_item.xml
```diff
@@ -3,7 +3,10 @@
   xmlns:app="http://schemas.android.com/apk/res-auto"
   xmlns:tools="http://schemas.android.com/tools"
   android:layout_width="match_parent"
-  android:layout_height="wrap_content">
+  android:layout_height="wrap_content"
+  android:clickable="true"
+  android:focusable="true"
+  android:background="?attr/selectableItemBackground">
```
vinaysshenoy: Not necessary to fix in this PR, but this commit is actually doing two things:

- Adds a fix for the edit button being clickable even for non-editable BPs
- Adds a touch selector to the item

Since we need to keep commits small and focused (one commit, one change), this should ideally be another commit.

app/src/main/res/layout/list_new_bp_button.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:app="http://schemas.android.com/apk/res-auto"
+  android:layout_width="match_parent"
+  android:layout_height="wrap_content">
+
+  <androidx.appcompat.widget.AppCompatButton
```
vinaysshenoy: As far as I know, we don't explicitly need to use `AppCompatButton` right? If we use `Button`, the framework automatically takes care of using `AppCompatButton` internally.

app/src/main/res/layout/list_new_bp_button.xml
```diff
@@ -0,0 +1,32 @@
...
+    android:id="@+id/newBp"
+    style="@style/Widget.AppCompat.Button.Borderless.Colored"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:contentDescription="@string/bloodpressurehistory_add_bp_button"
```
vinaysshenoy: We are currently not doing any work on accessibility. Do we need to add this attribute?

app/src/main/res/layout/list_new_bp_button.xml
msasikanth: I will change it to `Button`, I have gotten used to using `AppCompatButton` or `MaterialButton`

app/src/main/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItem.kt
```diff
@@ -27,6 +30,9 @@ sealed class BloodPressureHistoryListItem : ItemAdapter.Item<Event> {
                 showDivider = !isLastMeasurement
             )
           }
+          .let {
+            listOf(NewBpButton) + it
+          }
```
vinaysshenoy: Instead of doing this, I'd recommend extracting the history items to a separate variable for better readability.

```kotlin
val bpHistoryItems = measurements
  .mapIndexed { /* Generate measurement items */ }

return listOf(NewBpButton) + bpHistoryItems
```

app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
```diff
@@ -4,13 +4,20 @@ import android.content.Context
 import android.util.AttributeSet
```
vinaysshenoy: Reword the commit message to be more descriptive?

Ex: `Forward adapter events to Mobius loop`

app/src/main/res/layout/list_bp_history_item.xml
msasikanth: Split this commit into 2

app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
```diff
@@ -111,14 +113,24 @@ class BloodPressureHistoryScreen(
...
+      screenRouter.pop()
+    }
+  }
+
   override fun showBloodPressureHistory(bloodPressures: List<BloodPressureMeasurement>) {
     bloodPressureHistoryAdapter.submitList(BloodPressureHistoryListItem.from(bloodPressures, config.bpEditableDuration, utcClock))
   }
 
   override fun openBloodPressureEntrySheet() {
+    val intent = BloodPressureEntrySheet.intentForNewBp(context, delegate.model.patientUuid)
```
vinaysshenoy: Can we pass the `patientUuid` as a parameter to this function instead of reading it from the `MobiusDelegate`. The `model` is currently exposed from the delegate only to support refactoring the `PatientSummaryScreen` to mobius.

app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreenEffect.kt
```diff
@@ -1,10 +1,12 @@
 package org.simple.clinic.bp.history
```
vinaysshenoy: The commit message indicates that we are loading both the patient details as well as the list of BPs when this screen is created. However, that is not the case. What is actually happening is that this commit changes the screen so that the patient details are loaded in addition to the list of BPs which were already being done.

Can we reword this commit message? For ex: `Load patient details when the BP history screen is created`

app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreenInit.kt
```diff
@@ -6,6 +6,6 @@ import org.simple.clinic.mobius.first
 
 class BloodPressureHistoryScreenInit : Init<BloodPressureHistoryScreenModel, BloodPressureHistoryScreenEffect> {
   override fun init(model: BloodPressureHistoryScreenModel): First<BloodPressureHistoryScreenModel, BloodPressureHistoryScreenEffect> {
-    return first(model, LoadBloodPressureHistory(model.patientUuid))
+    return first(model, LoadPatient(model.patientUuid), LoadBloodPressureHistory(model.patientUuid))
```
vinaysshenoy: Not necessary to do in this PR, but since the model is being saved and restored, we should also add checks to avoid loading the patient details if the screen is restored and the patient details already exist in the model.

It might make sense to continue loading the BP history even when it is restored because it is more likely to change frequently than the patient details and it is a reactive query. This means we can avoid parcelling them in the model and mark them as `@IgnoredOnParcel`.

app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreenEffectHandler.kt
```diff
@@ -41,4 +45,19 @@ class BloodPressureHistoryScreenEffectHandler @AssistedInject constructor(
...
+
+  private fun loadPatient(
+      scheduler: Scheduler
+  ): ObservableTransformer<LoadPatient, BloodPressureHistoryScreenEvent> {
+    return ObservableTransformer { effect ->
+      effect
+          .switchMap {
+            patientRepository
+                .patient(it.patientUuid)
+                .subscribeOn(scheduler)
```
vinaysshenoy: I feel maintaining an ongoing subscription to the patient query is unnecessary because the patient details aren't likely to change while the user is on this screen. What do you think of adding a `.take(1)` here after the `.patient()` call to ensure that this patient is only queried once?

app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
```diff
@@ -133,6 +147,11 @@ class BloodPressureHistoryScreen(
     context.startActivity(intent)
   }
 
+  private fun displayNameGenderAge(name: String, gender: Gender, age: Int) {
+    val genderLetter = resources.getString(gender.displayLetterRes)
+    toolbar.title = resources.getString(R.string.patientsummary_name, name, genderLetter, age)
```
vinaysshenoy: We try to avoid using string resources across features even though the actual string values might be same. Could you please duplicate the string resource and name it scoped to this screen (ex: `bphistory_toolbar_title`)?

vinaysshenoy:
@rakshakhegde I just saw that this PR was assigned to you. I ended up reviewing it without looking at who it was assigned to. Sorry about this.

PR #1071, opened 2020-01-22
New blood pressure summary view - Part 1

msasikanth:
https://www.pivotaltracker.com/story/show/170816330

- `NewBloodPressureSummaryViewUpdate `
- `NewBloodPressureSummaryEffectHandler`

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewConfig.kt
+++ b/app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewConfig.kt
@@ -1,0 +1,3 @@
+package org.simple.clinic.summary.bloodpressures.newbpsummary
+
+data class NewBloodPressureSummaryViewConfig(val numberOfBpsToDisplay: Int)
--- a/app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffect.kt
+++ b/app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffect.kt
@@ -1,0 +1,9 @@
+package org.simple.clinic.summary.bloodpressures.newbpsummary
+
+import java.util.UUID
+
+sealed class NewBloodPressureSummaryViewEffect
+
+data class LoadBloodPressures(val patientUuid: UUID, val numberOfBpsToDisplay: Int) : NewBloodPressureSummaryViewEffect()
+
+data class LoadBloodPressuresCount(val patientUuid: UUID) : NewBloodPressureSummaryViewEffect()
--- a/app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffect.kt
+++ b/app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffect.kt
@@ -7,3 +7,5 @@ sealed class NewBloodPressureSummaryViewEffect
 data class LoadBloodPressures(val patientUuid: UUID, val numberOfBpsToDisplay: Int) : NewBloodPressureSummaryViewEffect()
 
 data class LoadBloodPressuresCount(val patientUuid: UUID) : NewBloodPressureSummaryViewEffect()
+
+data class OpenBloodPressureEntrySheet(val patientUuid: UUID) : NewBloodPressureSummaryViewEffect()
--- a/app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffect.kt
+++ b/app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffect.kt
@@ -12,3 +12,5 @@ data class LoadBloodPressuresCount(val patientUuid: UUID) : NewBloodPressureSumm
 data class OpenBloodPressureEntrySheet(val patientUuid: UUID) : NewBloodPressureSummaryViewEffect()
 
 data class OpenBloodPressureUpdateSheet(val measurement: BloodPressureMeasurement) : NewBloodPressureSummaryViewEffect()
+
+data class ShowBloodPressureHistoryScreen(val patientUuid: UUID) : NewBloodPressureSummaryViewEffect()
--- a/app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffect.kt
+++ b/app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffect.kt
@@ -1,5 +1,6 @@
 package org.simple.clinic.summary.bloodpressures.newbpsummary
 
+import org.simple.clinic.bp.BloodPressureMeasurement
 import java.util.UUID
 
 sealed class NewBloodPressureSummaryViewEffect
@@ -9,3 +10,5 @@ data class LoadBloodPressures(val patientUuid: UUID, val numberOfBpsToDisplay: I
 data class LoadBloodPressuresCount(val patientUuid: UUID) : NewBloodPressureSummaryViewEffect()
 
 data class OpenBloodPressureEntrySheet(val patientUuid: UUID) : NewBloodPressureSummaryViewEffect()
+
+data class OpenBloodPressureUpdateSheet(val measurement: BloodPressureMeasurement) : NewBloodPressureSummaryViewEffect()
--- a/app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffectHandler.kt
@@ -16,6 +16,7 @@ class NewBloodPressureSummaryViewEffectHandler @Inject constructor(
     return RxMobius
         .subtypeEffectHandler<NewBloodPressureSummaryViewEffect, NewBloodPressureSummaryViewEvent>()
         .addTransformer(LoadBloodPressures::class.java, loadBloodPressureHistory(schedulersProvider.io()))
+        .addTransformer(LoadBloodPressuresCount::class.java, loadBloodPressuresCount(schedulersProvider.io()))
         .build()
   }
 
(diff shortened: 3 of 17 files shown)


app/src/test/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewInitTest.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.spotify.mobius.test.InitSpec.assertThatFirst
+import org.junit.Test
+import java.util.UUID
+
+class NewBloodPressureSummaryViewInitTest {
+  @Test
+  fun `when widget is created, then load blood pressures for patient`() {
+    val patientUuid = UUID.fromString("33d7f6d7-39f9-46a2-a104-457e5f77dc20")
+    val defaultModel = NewBloodPressureSummaryViewModel.create(patientUuid)
+    val initSpec = InitSpec<NewBloodPressureSummaryViewModel, NewBloodPressureSummaryViewEffect>(NewBloodPressureSummaryViewInit())
```
vinaysshenoy: Can you add a line break here so that there's a clear separation between the variable declarations and the actual test setup please?

app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffect.kt
```diff
@@ -10,3 +10,5 @@ data class LoadBloodPressures(val patientUuid: UUID) : NewBloodPressureSummaryVi
 data class OpenBloodPressureEntrySheet(val patientUuid: UUID) : NewBloodPressureSummaryViewEffect()
```
vinaysshenoy: Minor grammar correction in the commit message:

- "See all" button is **only shown** if there are more than 3 blood pressures

app/src/test/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewUpdateTest.kt
```diff
@@ -59,4 +59,33 @@ class NewBloodPressureSummaryViewUpdateTest {
             hasEffects(OpenBloodPressureUpdateSheet(bloodPressureMeasurement) as NewBloodPressureSummaryViewEffect)
         ))
   }
+
+  @Test
+  fun `when see all is clicked, then show blood pressure history screen`() {
+    val bloodPressure1 = PatientMocker.bp(
```
vinaysshenoy: This test is deeply coupled to the implementation detail that the number of BPs that will be shown in the list will be 3. In addition, this number is defined as a magic number in the effect handler and is not configurable in any fashion.

This also merges both the fetching of number of BPs and showing the "See All" button to the same flow, which will become hard to maintain and understand going forward.

What I recommend we should do is this:

- Create a `NewBloodPressureSummaryViewConfig` class which will have a single property, say, `numberOfBpsToDisplay`.
- Provide instances of this config object to the `Init`,`Update`, and `ViewRenderer` classes when setting up the test fixtures and mobius delegate.
- Pass in the number of BPs to load as a property of the load BPs effect and use this in the effect handler to fetch the number of BPs.
- Separate out the flow to hide/show the "See All" button by having a separate effect to load the total count of BPs for a given patient. This can then become a new event that is passed to the `Update` class which can then update the model with the total count of BPs.
- This can then be used in the UI renderer to compare with the total number of BPs to display to decide to hide/show the "See All" button.

I understand that we are making two different SQL queries for essentially the same data, but in terms of real world performance impact, this will be less than a couple of hundred ms on most devices. What we gain instead is:

- Clear, separate, flows for both displaying the BPs and toggling the "See All" button
- Discrete tests for both of these flows so that they can be tested independently of each other
- Better separation of both the flows since they are easier to remove/change since they do not affect each other

app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffectHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  private fun loadBloodPressureHistory(
+      scheduler: Scheduler
+  ): ObservableTransformer<LoadBloodPressures, NewBloodPressureSummaryViewEvent> {
+    return ObservableTransformer { effect ->
+      effect
+          .switchMap {
+            bloodPressureRepository
+                .newestMeasurementsForPatient(it.patientUuid, 4)
```
vinaysshenoy: This comes back to the earlier comment I left about separating the flows for toggling the "See All" button and loading the list of BPs to display.

One problem with using magic values in codebases is that they lose meaning over time. Second is, they make testing hard since they are hardcoded and difficult to change for tests. This is an "easy" solution, but it eventually leads to making the codebase hard to maintain and understand.

I am making an educated guess here that you will display the first three BPs in this list in the UI renderer, and then check if the count is 4 or more to decide whether to toggle the "See All" button. The problem here is that the intention of displaying BPs and toggling the "See All" button are both lost to anyone who is trying to maintain the software going forward. Instead, we should attempt to make the code as explicit as possible, and make separate flows for displaying the list of BPs and toggling the "See All" button.

app/src/test/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffectHandlerTest.kt
```diff
@@ -0,0 +1,46 @@
...
+  ).build()
+  private val testCase = EffectHandlerTestCase(effectHandler)
+
+  @After
+  fun tearDown() {
+    testCase.dispose()
+  }
+
+  @Test
+  fun `when load blood pressures is received, then load blood pressures`() {
```
vinaysshenoy: Are we missing an "effect" here in the test method name?

app/src/test/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewUpdateTest.kt
msasikanth: I have couple of questions of about this can we discuss this in person?

app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffectHandler.kt
msasikanth: I was planning on moving to a config class in later commits, but you're right I think having a config class from start will be much more meaningful. I will separate the "see all" button and loading the list of BPs effects.

app/src/test/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewUpdateTest.kt
msasikanth: In this PR `ViewRenderer` is not implemented, It will be implemented in the later PR

app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewConfig.kt
```diff
@@ -0,0 +1,5 @@
+package org.simple.clinic.summary.bloodpressures.newbpsummary
+
+data class NewBloodPressureSummaryViewConfig(
+    val numberOfBpsToDisplay: Int = 3
```
vinaysshenoy: We try to avoid using default parameters because it hides details. Our convention, especially for configuration objects, is to set the default values when instantiating the instance in the Dagger graph. See `PatientSummaryConfig` for an example.

This is fine for now, and we can make this change in the future PR when adding the view to the screen.

app/src/test/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffectHandlerTest.kt
```diff
@@ -0,0 +1,47 @@
...
+  @Test
+  fun `when load blood pressures effect is received, then load blood pressures`() {
+    // given
+    val patientUuid = UUID.fromString("6b00207f-a613-4adc-9a72-dff68481a3ff")
+    val bloodPressure = PatientMocker.bp(
+        UUID.fromString("51ac042d-2f70-495c-a3e3-2599d8990da2"),
+        patientUuid
+    )
+    val bloodPressures = listOf(bloodPressure)
+    whenever(bloodPressureRepository.newestMeasurementsForPatient(patientUuid = patientUuid, limit = config.numberOfBpsToDisplay)) doReturn Observable.just(bloodPressures)
```
vinaysshenoy: The effect handler itself doesn't need the config since the number of BPs comes from the effect. To the effect handler, this config is irrelevant. In this particular test we should only care about the number of BPs to show from the effect, not where that value comes from.

Not necessary to fix this in this PR, we can change this in the next.

app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewModel.kt
```diff
@@ -5,12 +5,16 @@ import java.util.UUID
 
 data class NewBloodPressureSummaryViewModel(
     val patientUuid: UUID,
-    val bloodPressures: List<BloodPressureMeasurement>?
+    val bloodPressures: List<BloodPressureMeasurement>?,
+    val bloodPressuresCount: Int
```
vinaysshenoy: This should be a nullable property, since it is not clear from a value of `0` if we have loaded the BPs and found there to be none, or if we have never loaded them at all.

vinaysshenoy: It is not clear from the names of these two properties that they are actually different. Someone might look at this model in the future and wonder why we are tracking the count of BPs separately when we already have the list of BPs available.

I think we can name these properties better in order to expose these details. What do you think of:
- `bloodPressures` -> `latestBloodPressuresToDisplay`
- `bloodPressuresCount` -> `totalRecordedBloodPressureCount`

app/src/test/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewEffectHandlerTest.kt
msasikanth: I have fixed it in this PR. Instead of using config to get the `numberOfBpsToDisplay`, I have created a variable `numberOfBpsToDisplay ` in the test.

app/src/main/java/org/simple/clinic/summary/bloodpressures/newbpsummary/NewBloodPressureSummaryViewConfig.kt
msasikanth: Fixed it

PR #1074, opened 2020-01-23
Add instructions for setting up the JDK

vinaysshenoy:
The `README` was missing setup instructions for the JDK. Added this in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,56 @@
 
 An Android app for recording blood pressure measurements.
 
+## Pre-requisites
+
+The application currently requires JDK 1.8 to build. If you already have JDK 1.8 installed, skip this step.
+
+**Check if the right JDK is already available**
+
+Run the command `java -version`. If you have the right version of the JDK installed, you should see something like:
+```sh
+openjdk version "1.8.0_232"
+OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_232-b09)
+OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.232-b09, mixed mode)
+```
+
+If this command has an error, or shows a different version, you can follow the instructions below to install the JDK.
+
+**Install the JDK**
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,6 +5,56 @@
 
 An Android app for recording blood pressure measurements.
 
+## Pre-requisites
+
+The application currently requires JDK 1.8 to build. If you already have JDK 1.8 installed, skip this step.
+
+**I am not sure if I have the JDK installed, or I am not sure if I have the right version installed**
```
sanchita: Can we add a title here which says `FAQs`?

README.md
```diff
@@ -5,6 +5,56 @@
...
+
+Run the command `java -version`. If you have the right version of the JDK installed, you should see something like:
+```sh
+openjdk version "1.8.0_232"
+OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_232-b09)
+OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.232-b09, mixed mode)
+```
+
+If this command has an error, or shows a different version, you can follow the instructions below to install the JDK.
```
sanchita: I think the answers to the questions do not need a margin on the left.
Wdyt?

README.md
sanchita: This looks like a question. Should we add a `?` to all of them?
Or we can just rename them to 
1. `How to check JDK version?`
2. `How to install JDK?`

vinaysshenoy: Hmm, this is not exactly an FAQ, but sort of a POV of a reader who might be in a state of uncertainty about the JDK installation. I'll reword this to match your suggestions.

README.md
vinaysshenoy: I have not added any margins, I think this is how it's getting rendered. Could you please check the raw file and confirm?

sanchita: Ok. No worries then.

README.md
vinaysshenoy: Done.

PR #1076, opened 2020-01-23
Preserve validation errors in Patient entry screen

janhavisinghh:
Preserve validation errors in Patient entry screen when the screen is restored from background.

https://www.pivotaltracker.com/story/show/169030605

//Description
This has been the case for both old and new architectures. We can fix this by modeling the validation errors as part of the model and rendering it via the view renderer, rather than modelling it as a side effect.


sanchita:
@rakshakhegde Can we close this PR?
You can create a new PR I guess once you start working on this?

sanchita:
Since this requires considerable re-work, we can open a separate PR when we get to fixing this.

PR #1077, opened 2020-01-23
Fix Patient SQL query to include house address and zone

rakshakhegde:
https://www.pivotaltracker.com/story/show/170696612

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -2784,4 +2784,30 @@ class PatientRepositoryAndroidTest {
     assertThat(patientRepository.hasMedicalHistoryForPatientChangedSince(medicalHistoryForSomeOtherPatient.patientUuid, oneSecondEarlier)).isFalse()
   }
 
+  @Test
+  fun when_patient_address_details_are_saved_then_all_patient_address_fields_should_be_retrieved_as_expected() {
+    val patientUuid = UUID.fromString("a0ac466f-5a8d-436b-8845-4412e7f86cdd")
+    val patientAddressUuid = UUID.fromString("a0ac466f-5a8d-436b-8845-4412e7f86cdd")
+    val streetAddress = "Streetz"
+    val zone = "Zonez"
+
+    val patientProfile = testData.patientProfile(
+        patientUuid = patientUuid,
+        patientAddressUuid = patientAddressUuid,
+        syncStatus = PENDING
+    ).run {
+      copy(address = testData.patientAddress(
+          uuid = patientAddressUuid,
+          streetAddress = streetAddress,
+          zone = zone
+      ))
+    }
+    patientRepository.save(listOf(patientProfile)).blockingAwait()
+
+    val patients = patientRepository.recordsWithSyncStatus(PENDING).blockingGet()
+    assertThat(patients.size).isEqualTo(1)
+
+    assertThat(patients.first().address.streetAddress).isEqualTo(streetAddress)
+    assertThat(patients.first().address.zone).isEqualTo(zone)
+  }
 }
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -288,6 +288,7 @@ data class Patient(
           PA.uuid addr_uuid, PA.colonyOrVillage addr_colonyOrVillage, PA.district addr_district,
           PA.state addr_state, PA.country addr_country,
           PA.createdAt addr_createdAt,PA.updatedAt addr_updatedAt, PA.deletedAt addr_deletedAt,
```
vinaysshenoy: Can we add a failing test for this, please?

rakshakhegde:
@vinaysshenoy Can you please help finish this tomorrow? This has been pending for a while.

vinaysshenoy:
> @vinaysshenoy Can you please help finish this tomorrow? This has been pending for a while.

Yup, we can pair on this as soon as you get into work.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -2784,4 +2784,30 @@ class PatientRepositoryAndroidTest {
...
+      ))
+    }
+    patientRepository.save(listOf(patientProfile)).blockingAwait()
+
+    val patients = patientRepository.recordsWithSyncStatus(PENDING).blockingGet()
+    assertThat(patients.size).isEqualTo(1)
+
+    assertThat(patients.first().address.streetAddress).isEqualTo(streetAddress)
+    assertThat(patients.first().address.zone).isEqualTo(zone)
+  }
```
vinaysshenoy: 👍

app/src/androidTest/java/org/simple/clinic/storage/migrations/MigrationHelper.kt
```diff
@@ -0,0 +1,133 @@
...
+import org.simple.clinic.patient.businessid.BusinessId
+import org.simple.clinic.patient.businessid.Identifier
+
+fun SupportSQLiteDatabase.savePatientProfile(patientProfile: PatientProfile) {
+  savePatientAddress(patientProfile, this)
+  savePatient(patientProfile, this)
+  savePatientBusinessIds(patientProfile, this)
+}
+
+private fun savePatientAddress(patientProfile: PatientProfile, database: SupportSQLiteDatabase) {
```
vinaysshenoy: These were intentionally left specific to the migration class. The reason being that the database fields which map to column names and which fields need to be written to are specific to a given database schema version and cannot be reused across migration tests since the schemas will be different in each one.

This is fine for now, but let's revert this change later.

app/src/main/java/org/simple/clinic/storage/migrations/Migration_58_59.kt
```diff
@@ -0,0 +1,25 @@
...
+class Migration_58_59 @Inject constructor() : Migration(58, 59) {
+
+  override fun migrate(database: SupportSQLiteDatabase) {
+    database.inTransaction {
+      execSQL("""
+        UPDATE "Patient" SET "syncStatus" = 'PENDING' WHERE "uuid" IN (
+          SELECT P.uuid FROM "Patient" P
+          LEFT JOIN (
+            SELECT * FROM "PatientAddress"
+              WHERE LENGTH("zone") > 0
```
vinaysshenoy: I just realised that the fields shown for India will never have Zone since it's a Bangladesh only field. This means that the value of this field will always be blank and this migration will end up forcing **ALL** patients to sync.

I will make the change to remove the zone from this check and force-push.

rakshakhegde: @vinaysshenoy It is correct that it'll will be blank in India and that is expected. This query is only checking for length **greater** than zero which will be false for patients registered in India. This query is correct as it is. Please revert back to including the zone too?

vinaysshenoy: You... are right. Brain is fried. Shall revert. 😅

rakshakhegde: This surfaces the problem... maybe this PR is missing a couple of tests too.

vinaysshenoy: Reverted. I also update the tests to cover all cases.

PR #1079, opened 2020-01-27
New Medical History screen - cleanup #3

vinaysshenoy:
https://www.pivotaltracker.com/story/show/169674952

This moves the feature of registering a patient and opening the summary screen to the Mobius loop.

### DIFF
--- a/app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryEffectHandler.kt
@@ -1,13 +1,25 @@
 package org.simple.clinic.medicalhistory.newentry
 
 import com.spotify.mobius.rx2.RxMobius
+import com.squareup.inject.assisted.Assisted
+import com.squareup.inject.assisted.AssistedInject
 import io.reactivex.ObservableTransformer
+import org.simple.clinic.util.scheduler.SchedulersProvider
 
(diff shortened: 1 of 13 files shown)


vinaysshenoy:
@rakshakhegde Removed your request for review since Sasikanth was available and we are running on a tight schedule to get this feature out. Maybe we can pair review on a future PR in this feature later.

app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryEffectHandler.kt
```diff
@@ -19,7 +31,42 @@ class NewMedicalHistoryEffectHandler @AssistedInject constructor(
...
+                .flatMap { registeredPatient ->
+                  medicalHistoryRepository
+                      .save(registeredPatient.uuid, ongoingMedicalHistoryEntry)
+                      .toSingleDefault(PatientRegistered(registeredPatient.uuid))
+                }
+          }
+    }
+  }
+
+  private data class RegisterPatientData(
```
msasikanth: Can't this data class be a separate file? or will this be moved later?

vinaysshenoy: This was intentionally left in this class since it is not required anywhere else. The only reason this exists is so that we can group the required data together to pass to the next operator in the Rx chain.

msasikanth:
Rebase the PR please

PR #1094, opened 2020-01-29
Fix empty blood pressures state in BP history screen

msasikanth:
- If we check for both null and empty list, edge case of removing all the BPs is not being handled. In order to fix that only null check is being made and the empty list is being rendered on the UI.

### DIFF
--- a/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreenModel.kt
+++ b/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreenModel.kt
@@ -18,7 +18,7 @@ data class BloodPressureHistoryScreenModel(
   }
 
   val hasBloodPressures: Boolean
-    get() = bloodPressures.isNullOrEmpty().not()
+    get() = bloodPressures != null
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreenModel.kt
```diff
@@ -17,9 +17,6 @@ data class BloodPressureHistoryScreenModel(
     fun create(patientUuid: UUID) = BloodPressureHistoryScreenModel(patientUuid, null, null)
   }
 
-  val hasBloodPressures: Boolean
-    get() = bloodPressures.isNullOrEmpty().not()
```
vinaysshenoy: The advantage of having a separate function to check if the model has loaded BPs is that we can change the implementation in one place.

So, if leave this method as is, but change the getter method to `get() = bloodPressures != null`, we can make this fix without needing to change multiple places.

msasikanth: Done

PR #1096, opened 2020-01-29
Change `timeDateTextView` string format in `list_bp_history_item`

msasikanth:
- Design updated to be similar as the summary view

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -528,5 +528,5 @@
   <string name="bloodpressurehistory_bp_reading" translatable="false">%1$d / %2$d</string>
   <string name="bloodpressurehistory_add_bp_button">Blood Pressure</string>
   <string name="bloodpressurehistory_toolbar_title" translatable="false">%1$s, %2$s, %3$d</string>
-  <string name="bloodpressurehistory_bp_time_date">%1$s, %2$s</string>
+  <string name="bloodpressurehistory_bp_time_date" translatable="false">%1$s\n%2$s</string>
 </resources>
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -528,5 +528,5 @@
   <string name="bloodpressurehistory_bp_reading" translatable="false">%1$d / %2$d</string>
   <string name="bloodpressurehistory_add_bp_button">Blood Pressure</string>
   <string name="bloodpressurehistory_toolbar_title" translatable="false">%1$s, %2$s, %3$d</string>
-  <string name="bloodpressurehistory_bp_time_date">%1$s, %2$s</string>
+  <string name="bloodpressurehistory_bp_date_time" translatable="false">%1$s\n%2$s</string>
```
vinaysshenoy: It is not necessary to change the names of resource IDs to match how we're displaying a certain value. The only time we will need to change the IDs is if thing we are displaying here itself has changed (say, from a timestamp to something else).

app/src/main/res/layout/list_bp_history_item.xml
```diff
@@ -70,12 +70,12 @@
...
     android:text="@string/patientsummary_edit"
     app:layout_constraintBottom_toTopOf="@id/bottomGuideline"
-    app:layout_constraintEnd_toStartOf="@id/timeDateTextView"
+    app:layout_constraintEnd_toStartOf="@id/dateTimeTextView"
     app:layout_constraintStart_toEndOf="@id/readingsTextView"
     app:layout_constraintTop_toBottomOf="@id/topGuideline" />
 
   <TextView
-    android:id="@+id/timeDateTextView"
+    android:id="@+id/dateTimeTextView"
```
vinaysshenoy: It is not necessary to change the names of view IDs to match how we're displaying a certain value. The only time we will need to change the IDs is if thing we are displaying here itself has changed (say, from a timestamp to something else).

PR #1104, opened 2020-01-30
`BloodSugarHistoryScreen` - Part 2

msasikanth:
https://www.pivotaltracker.com/story/show/169677139

### DIFF
--- a/app/src/main/java/org/simple/clinic/bloodsugar/history/BloodSugarHistoryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/bloodsugar/history/BloodSugarHistoryScreen.kt
@@ -1,23 +1,34 @@
 package org.simple.clinic.bloodsugar.history
 
+import android.annotation.SuppressLint
 import android.content.Context
+import android.content.Intent
 import android.util.AttributeSet
+import androidx.appcompat.app.AppCompatActivity
 import androidx.constraintlayout.widget.ConstraintLayout
 import androidx.recyclerview.widget.LinearLayoutManager
+import com.jakewharton.rxbinding3.view.detaches
+import io.reactivex.Observable
+import io.reactivex.rxkotlin.ofType
 import kotlinx.android.synthetic.main.screen_blood_sugar_history.view.*
 import org.simple.clinic.R
 import org.simple.clinic.bloodsugar.BloodSugarMeasurement
+import org.simple.clinic.bloodsugar.entry.BloodSugarEntrySheet
 import org.simple.clinic.bloodsugar.history.adapter.BloodSugarHistoryListItem
 import org.simple.clinic.bloodsugar.history.adapter.BloodSugarHistoryListItemDiffCallback
+import org.simple.clinic.bloodsugar.selection.type.BloodSugarTypePickerSheet
 import org.simple.clinic.di.injector
 import org.simple.clinic.patient.DateOfBirth
 import org.simple.clinic.patient.Gender
 import org.simple.clinic.patient.Patient
 import org.simple.clinic.patient.displayLetterRes
+import org.simple.clinic.router.screen.ActivityResult
 import org.simple.clinic.router.screen.ScreenRouter
+import org.simple.clinic.summary.TYPE_PICKER_SHEET
 import org.simple.clinic.util.UserClock
 import org.simple.clinic.widgets.DividerItemDecorator
 import org.simple.clinic.widgets.ItemAdapter
+import org.simple.clinic.widgets.ScreenDestroyed
 import org.simple.clinic.widgets.dp
 import org.threeten.bp.format.DateTimeFormatter
 import java.util.UUID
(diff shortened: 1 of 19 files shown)


app/src/test/java/org/simple/clinic/bloodsugar/history/adapter/BloodSugarHistoryListItemTest.kt
```diff
@@ -0,0 +1,116 @@
...
+import java.util.UUID
+
+class BloodSugarHistoryListItemTest {
+  private val userClock = TestUserClock()
+  private val locale = Locale.ENGLISH
+  private val dateFormatter = DateTimeFormatter.ofPattern("d-MMM-yyyy", locale)
+  private val timeFormatter = DateTimeFormatter.ofPattern("h:mm a", locale)
+
+  @Test
+  fun `if there is only one list item, then divider should be shown`() {
```
vinaysshenoy: It made sense to test the generation of dividers earlier because we were showing and hiding them manually. Given that we now will use the item decoration instead, does it still make to test all of these?

app/src/main/java/org/simple/clinic/bloodsugar/history/BloodSugarHistoryScreen.kt
```diff
@@ -3,30 +3,94 @@ package org.simple.clinic.bloodsugar.history
...
     attrs: AttributeSet
 ) : ConstraintLayout(context, attrs), BloodSugarHistoryScreenUi, BloodSugarHistoryScreenUiActions {
 
+  @Inject
+  lateinit var screenRouter: ScreenRouter
+
+  @Inject
+  lateinit var userClock: UserClock
+
+  @field:[Inject Named("date_for_bp_history")]
```
vinaysshenoy: The name indicates this is only for `bp` history while this is being used in the blood sugar screen. Since it's the same format in both screens, maybe we can find a common name for them, say `measurement_*` instead of `bp_*`.

What do you think?

app/src/main/java/org/simple/clinic/bloodsugar/history/BloodSugarHistoryScreen.kt
```diff
@@ -3,30 +3,94 @@ package org.simple.clinic.bloodsugar.history
 import android.content.Context
```
vinaysshenoy: Reword the commit message? Ex: `Implement UI methods in BloodSugarHistoryScreen`

app/src/main/java/org/simple/clinic/bloodsugar/history/BloodSugarHistoryScreen.kt
```diff
@@ -1,23 +1,34 @@
 package org.simple.clinic.bloodsugar.history
 
+import android.annotation.SuppressLint
 import android.content.Context
+import android.content.Intent
```
vinaysshenoy: Reword the commit message? Ex: `Implement UI action methods in BloodSugarHistoryScreen`

app/src/main/java/org/simple/clinic/bloodsugar/history/BloodSugarHistoryScreen.kt
```diff
@@ -57,6 +61,12 @@ class BloodSugarHistoryScreen(
 
   private val bloodSugarHistoryAdapter = ItemAdapter(BloodSugarHistoryListItemDiffCallback())
 
+  private val events: Observable<BloodSugarHistoryScreenEvent> by unsafeLazy {
+    addNewBloodSugarClicked()
```
vinaysshenoy: Not necessary to do in this PR, but we will also need to add `analyticsName` values to all user generated events in `BloodSugarHistoryScreenEvent` so that they get reported to analytics.

app/src/main/java/org/simple/clinic/bloodsugar/history/BloodSugarHistoryScreen.kt
```diff
@@ -59,6 +59,15 @@ class BloodSugarHistoryScreen(
   @field:[Inject Named("time_for_bp_history")]
   lateinit var timeFormatter: DateTimeFormatter
 
+  @Inject
+  lateinit var screenRouter: ScreenRouter
+
+  @Inject
+  lateinit var effectHandleFactory: BloodSugarHistoryScreenEffectHandler.Factory
```
vinaysshenoy: Typo? Missing an 'r' in the 'effectHandle**r**'.

app/src/main/res/layout/list_blood_sugar_history_item.xml
```diff
@@ -3,10 +3,7 @@
   xmlns:app="http://schemas.android.com/apk/res-auto"
   xmlns:tools="http://schemas.android.com/tools"
   android:layout_width="match_parent"
-  android:layout_height="wrap_content"
-  android:background="?attr/selectableItemBackground"
```
vinaysshenoy: If this is removed, we won't have any touch feedback, right?

msasikanth: It will be, but since there is no edit feature in this PR touch feedback doesn't seem to be needed. what do you think?

vinaysshenoy: Ah, that makes sense. We can add this when we implement the Edit feature.

app/src/main/java/org/simple/clinic/bloodsugar/history/BloodSugarHistoryScreen.kt
msasikanth: I was hoping to discuss it later and implement it in later PR. But I will change it now 👍

PR #1109, opened 2020-01-30
Fix translation issue in numerals

sanchita:
https://www.pivotaltracker.com/story/show/170946758

### DIFF
--- a/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
@@ -9,6 +9,7 @@ import io.reactivex.Observable
 import io.reactivex.rxkotlin.cast
 import io.reactivex.rxkotlin.ofType
 import kotlinx.android.synthetic.main.screen_bp_history.view.*
+import org.simple.clinic.R
 import org.simple.clinic.ReportAnalyticsEvents
 import org.simple.clinic.bp.BloodPressureMeasurement
 import org.simple.clinic.bp.entry.BloodPressureEntrySheet
@@ -169,7 +170,7 @@ class BloodPressureHistoryScreen(
 
   private fun displayNameGenderAge(name: String, gender: Gender, age: Int) {
     val genderLetter = resources.getString(gender.displayLetterRes)
-    toolbar.title = "$name, $genderLetter, $age"
+    toolbar.title = resources.getString(R.string.bloodpressurehistory_toolbar_title, name, genderLetter, age.toString())
   }
 
   private fun addNewBpClicked(): Observable<BloodPressureHistoryScreenEvent> {
--- a/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
@@ -9,7 +9,6 @@ import io.reactivex.Observable
 import io.reactivex.rxkotlin.cast
 import io.reactivex.rxkotlin.ofType
 import kotlinx.android.synthetic.main.screen_bp_history.view.*
-import org.simple.clinic.R
 import org.simple.clinic.ReportAnalyticsEvents
 import org.simple.clinic.bp.BloodPressureMeasurement
 import org.simple.clinic.bp.entry.BloodPressureEntrySheet
@@ -170,7 +169,7 @@ class BloodPressureHistoryScreen(
 
   private fun displayNameGenderAge(name: String, gender: Gender, age: Int) {
     val genderLetter = resources.getString(gender.displayLetterRes)
-    toolbar.title = resources.getString(R.string.bloodpressurehistory_toolbar_title, name, genderLetter, age)
+    toolbar.title = "$name, $genderLetter, $age"
   }
 
   private fun addNewBpClicked(): Observable<BloodPressureHistoryScreenEvent> {
--- a/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
@@ -126,7 +126,7 @@ data class OverdueAppointmentRow(
     holder.patientNameTextView.setCompoundDrawableStart(gender.displayIconRes)
 
     holder.patientBPTextView.text = context.resources.getQuantityString(
-        R.plurals.overdue_list_item_patient_bp,
+        R.plurals.overdue_list_item_patient_bp_days_ago,
         bpDaysAgo,
         "$bpSystolic/$bpDiastolic, $bpDaysAgo"
     )
@@ -137,7 +137,7 @@ data class OverdueAppointmentRow(
     holder.isAtHighRiskTextView.visibility = if (isAtHighRisk) VISIBLE else GONE
 
     holder.overdueDaysTextView.text = context.resources.getQuantityString(
-        R.plurals.overdue_list_item_overdue_days,
+        R.plurals.overdue_list_item_appointment_overdue_days,
         overdueDays,
         "$overdueDays"
     )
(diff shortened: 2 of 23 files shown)


app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
```diff
@@ -139,7 +137,7 @@ data class OverdueAppointmentRow(
     holder.overdueDaysTextView.text = context.resources.getQuantityString(
         R.plurals.overdue_list_item_overdue_days,
         overdueDays,
-        overdueDays
+        "$overdueDays"
```
vinaysshenoy: This is un unnecessary template since it has nothing else apart from the variable. Can't this just be `overdueDays.toString()`?

app/src/main/res/values-bn-rIN/strings.xml
```diff
@@ -239,12 +239,12 @@
   <string name="overdue_no_patients">সময় পার হয়ে যাওয়া কোন রোগী নেই</string>
   <string name="overdue_empty_desc">যে রোগীরা ফলো আপ রক্তচাপ মাপার জন্য তাদের পূর্বনির্ধারিত সাক্ষাৎকার মিস করেছে আপনি সেই রোগীদের ফোন করতে পারবেন।</string>
   <plurals name="overdue_list_item_patient_bp">
-    <item quantity="one">%1$d/%2$d, %3$d দিন আগে</item>
-    <item quantity="other">%1$d/%2$d/ %3$d/.......... দিন আগে</item>
```
vinaysshenoy: Ugh, I think I missed this when merging the new translations in.

app/src/main/java/org/simple/clinic/security/pin/PinEntryCardView.kt
```diff
@@ -138,11 +138,11 @@ class PinEntryCardView(context: Context, attrs: AttributeSet) : CardView(context
   }
 
   fun showIncorrectPinErrorOnSubsequentAttempts(remaining: Int) {
-    showError(resources.getString(R.string.pinentry_error_incorrect_pin_on_subsequent_attempts, remaining))
+    showError(resources.getString(R.string.pinentry_error_incorrect_pin_on_subsequent_attempts, "$remaining"))
   }
 
   fun showIncorrectAttemptsLimitReachedError(attemptsMade: Int) {
-    showError(resources.getString(R.string.pinentry_error_incorrect_attempts_limit_reached, attemptsMade))
+    showError(resources.getString(R.string.pinentry_error_incorrect_attempts_limit_reached, "$attemptsMade"))
```
vinaysshenoy: Both the string templates are unnecessary since there's only the variable. Can we use `.toString()` instead?

app/src/main/res/values-bn-rBD/strings.xml
```diff
@@ -193,8 +193,8 @@
   <!-- PIN entry -->
   <string name="pinentry_forgotpin">পিন ভুলে গেছেন?</string>
   <string name="pinentry_error_incorrect_pin_on_first_attempt">পিনটি আসল পিনটির সাথে মেলে না</string>
-  <string name="pinentry_error_incorrect_pin_on_subsequent_attempts" tools:ignore="PluralsCandidate">ভুল PIN, আরো %d বার চেষ্টা করতে পারবেন</string>
-  <string name="pinentry_error_incorrect_attempts_limit_reached" tools:ignore="PluralsCandidate">আপনি %d বার ভুল প্রচেষ্টা করেছেন</string>
+  <string name="pinentry_error_incorrect_pin_on_subsequent_attempts" tools:ignore="PluralsCandidate">ভুল PIN, আরো %s বার চেষ্টা করতে পারবেন</string>
```
vinaysshenoy: I think we can remove the `tools:ignore="PluralsCandidate"` now since that will only be checked for strings which contain an integer placeholder.

app/src/main/java/org/simple/clinic/bp/history/BloodPressureHistoryScreen.kt
```diff
@@ -170,7 +169,7 @@ class BloodPressureHistoryScreen(
 
   private fun displayNameGenderAge(name: String, gender: Gender, age: Int) {
     val genderLetter = resources.getString(gender.displayLetterRes)
-    toolbar.title = resources.getString(R.string.bloodpressurehistory_toolbar_title, name, genderLetter, age)
+    toolbar.title = "$name, $genderLetter, $age"
```
vinaysshenoy: We should not be adding the comma in code. Reasons:

- There are languages which do not have commas
- There are languages where the comma glyph should be replaced by another character (The Ge'ez script used for Ethiopia languages has an alternate symbol that serves the same purpose as a comma), so we should use the correct glyph in those languages.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
```diff
@@ -77,7 +78,7 @@ class SyncIndicatorView(context: Context, attrs: AttributeSet) : LinearLayout(co
         syncStatusTextView.text = if (durationToMinsAgo == 0) {
           context.getString(R.string.syncindicator_status_synced_just_now)
         } else {
-          context.getString(R.string.syncindicator_status_synced, durationToMinsAgo)
+          context.getString(R.string.syncindicator_status_synced, "$durationToMinsAgo")
```
vinaysshenoy: Unnecessary template?

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheet.kt
```diff
@@ -133,7 +133,7 @@ class ScheduleAppointmentSheet : BottomSheetActivity() {
       is TimeToAppointment.Months -> R.plurals.scheduleappointment_appointmentin_months
     }
 
-    return resources.getQuantityString(quantityStringResourceId, timeToAppointment.value, timeToAppointment.value)
+    return resources.getQuantityString(quantityStringResourceId, timeToAppointment.value, "${timeToAppointment.value}")
```
vinaysshenoy: Unnecessary template?

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
```diff
@@ -126,7 +126,7 @@ data class OverdueAppointmentRow(
     holder.patientNameTextView.setCompoundDrawableStart(gender.displayIconRes)
```
vinaysshenoy: Typo in commit message: `Transifix` should be `Transifex`.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
sanchita: It can be. But I figured that this is as a consistent way to convert to strings. 
Any particular reason for suggesting this since both look the same to me.

app/src/main/java/org/simple/clinic/security/pin/PinEntryCardView.kt
sanchita: See comment above.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
vinaysshenoy: No specific reason, I personally use templates only when the variable is used for formatting. If you prefer this, that's fine as well.

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheet.kt
sanchita: As discussed, I am not making this change.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
sanchita: As discussed, I am not making this change.

app/src/main/res/values-bn-rBD/strings.xml
sanchita: As discussed, I am not making this change. This will be handled by Transifex. If not, we will make this change later.

PR #1110, opened 2020-01-31
Implement the Diagnosis view in the new medical history screen

vinaysshenoy:
https://www.pivotaltracker.com/story/show/169674952

Sorry about the large-ish PR, but most of this is test code which tests a bunch of similar cases and involves some repetition. Let me know if you need help pairing on the review at any point.


app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryEffectHandler.kt
```diff
@@ -90,6 +91,18 @@ class NewMedicalHistoryEffectHandler @AssistedInject constructor(
     }
   }
 
+  private fun setupUiForDiabetesManagement(diabetesManagementEnabled: Boolean) {
+    uiActions.apply {
+      if (diabetesManagementEnabled) {
+        showDiagnosisView()
```
msasikanth: Should we have these as part of UI renderer as well instead of effect handler?

vinaysshenoy: Good point, these should be in the UI renderer. I will make this change.

vinaysshenoy: Done.

msasikanth:
Can you rebase this PR, please?

PR #1114, opened 2020-02-03
Unit for period is not translated in any language when you set appointment reminder for overdue patients

msasikanth:
https://www.pivotaltracker.com/story/show/170310730

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/appointmentreminder/AppointmentReminderSheet.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/appointmentreminder/AppointmentReminderSheet.kt
@@ -11,8 +11,8 @@ import io.reactivex.Observable
 import io.reactivex.subjects.PublishSubject
 import kotterknife.bindView
 import org.simple.clinic.R
-import org.simple.clinic.main.TheActivity
 import org.simple.clinic.bindUiToController
+import org.simple.clinic.main.TheActivity
 import org.simple.clinic.widgets.BottomSheetActivity
 import org.simple.clinic.widgets.ScreenDestroyed
 import org.simple.clinic.widgets.UiEvent
@@ -99,7 +99,8 @@ class AppointmentReminderSheet : BottomSheetActivity() {
 
   fun updateDisplayedDate(newIndex: Int) {
     currentIndex = newIndex
-    currentDateTextView.text = possibleDates[currentIndex].displayText
+    val appointmentReminder = possibleDates[currentIndex]
+    currentDateTextView.text = textForAppointmentReminder(appointmentReminder.timeAmount, appointmentReminder.chronoUnit)
   }
 
   fun enableIncrementButton(state: Boolean) {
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -319,6 +319,14 @@
...
   <string name="appointmentreminder_call_later_in">Call later in</string>
   <string name="appointmentreminder_done">Done</string>
+  <plurals name="appointmentreminder_days">
+    <item quantity="one">%d day</item>
+    <item quantity="other">%d days</item>
+  </plurals>
+  <plurals name="appointmentreminder_weeks">
+    <item quantity="one">%d week</item>
+    <item quantity="other">%d weeks</item>
+  </plurals>
```
vinaysshenoy: We're going to be forcing arabic numerals in the UI going forward. This means that the placeholders should be using string placeholders instead of integer placeholders. Please update this?

See: https://docs.google.com/document/d/1v5siRSZD3YgnWpPJ9dgEvt7Z7x_O-9mc0mh78y0kUD0/edit#

app/src/main/java/org/simple/clinic/home/overdue/appointmentreminder/AppointmentReminderSheet.kt
```diff
@@ -110,6 +111,19 @@ class AppointmentReminderSheet : BottomSheetActivity() {
...
 
+  private fun textForAppointmentReminder(timeAmount: Int, chronoUnit: ChronoUnit): String {
+    val quantityStringResourceId = when (chronoUnit) {
+      ChronoUnit.DAYS -> {
+        R.plurals.appointmentreminder_days
+      }
+      ChronoUnit.WEEKS -> {
+        R.plurals.appointmentreminder_weeks
+      }
+      else -> throw AssertionError()
```
vinaysshenoy: Assertion error is meant to be thrown in tests. Can you instead use something that exposes the actual problem and add a message?

Ex:
```kotlin
else -> throw IllegalArgumentException("$chronoUnit is unsupported type for appointment reminders")
```

vinaysshenoy:
@rakshakhegde In the interest of time, I am reviewing this PR.

app/src/main/res/values/strings.xml
msasikanth: Done. Forgot about this change.

app/src/main/java/org/simple/clinic/home/overdue/appointmentreminder/AppointmentReminderSheet.kt
msasikanth: Done

app/src/main/java/org/simple/clinic/home/overdue/appointmentreminder/AppointmentReminderSheet.kt
```diff
@@ -110,6 +111,19 @@ class AppointmentReminderSheet : BottomSheetActivity() {
     decrementDateButton.isEnabled = state
   }
 
+  private fun textForAppointmentReminder(timeAmount: Int, chronoUnit: ChronoUnit): String {
+    val quantityStringResourceId = when (chronoUnit) {
+      ChronoUnit.DAYS -> {
```
vinaysshenoy: Forgot to leave this comment in the last review. These braces are unnecessary since the block is only one line. I think we can remove them. What do you think?

app/src/main/java/org/simple/clinic/home/overdue/appointmentreminder/AppointmentReminderSheet.kt
```diff
@@ -110,6 +111,19 @@ class AppointmentReminderSheet : BottomSheetActivity() {
...
+        R.plurals.appointmentreminder_days
+      }
+      ChronoUnit.WEEKS -> {
+        R.plurals.appointmentreminder_weeks
+      }
+      else -> throw IllegalArgumentException("$chronoUnit is unsupported type for appointment reminders")
+    }
+    return resources.getQuantityString(quantityStringResourceId, timeAmount, timeAmount)
```
vinaysshenoy: This might cause a crash since we are passing `Int` values to the format string while it expects `String`. Could you please test this out and verify that it renders as expected?

msasikanth: I have tested it on my phone, it's working. I will also try it on API 21 and see if it crashes.

PR #1123, opened 2020-02-04
Add an additional security environment for security auditing

vinaysshenoy:
https://www.pivotaltracker.com/story/show/171075080


vinaysshenoy:
We tested the SMS hash generation tool on @harimohanraj89's machine, and it generates the hashes correctly.

PR #1128, opened 2020-02-06
Fix edit patient issue

sanchita:
https://www.pivotaltracker.com/story/show/171126418

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
@@ -43,17 +43,8 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
             uiWorkScheduler = schedulersProvider.ui()
         ))
         .addTransformer(LoadCurrentFacility::class.java, loadCurrentFacility(schedulersProvider.io()))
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
```diff
@@ -43,9 +43,19 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
             uiWorkScheduler = schedulersProvider.ui()
         ))
         .addTransformer(LoadCurrentFacility::class.java, loadCurrentFacility(schedulersProvider.io()))
+        .addTransformer(HandleEditClick::class.java, handleEditClick(uiWorkScheduler = schedulersProvider.ui()))
```
vinaysshenoy: This can be replaced with a `consumer` instead.

```kotlin
.addConsumer(HandleEditClick::class.java, { uiActions.showPatientEditScreen(it.patientSummaryProfile) }, schedulersProvider.ui())
```

vinaysshenoy: Can we add a test in `PatientSummaryEffectHandlerTest` to capture this behaviour, please? Currently, there's no regression test for this since this click was handled entirely on the UI earlier.

PR #1136, opened 2020-02-11
Add ADR for Mobius

vinaysshenoy:
https://www.pivotaltracker.com/story/show/170146761

### DIFF
--- a/doc/arch/008-screen-architecture-v3.md
+++ b/doc/arch/008-screen-architecture-v3.md
@@ -1,0 +1,99 @@
+# ADR 008: Screen Architecture (v3)
+
+## Status
+Accepted (Supersedes [001](./001-screen-controllers.md), [007](./007-screen-architecture-v2.md)) on 2019-10-01
+
+## Context
+Our [intermediate architecture](./007-screen-architecture-v2.md) was a step above the [v1 screen architecture](./001-screen-controllers.md). However, it still has problems of its own:
+
+- The `UiStateProducer` is still some sort of a **God Class** since it is reponsible for both business logic (what needs to happen) and effects (changes to the system). This makes testing it cumbersome since the tests are more of integration tests which end up verifying both business logic and implementation details.
+- The architecture is heavily dependent on [RxJava](https://github.com/ReactiveX/RxJava/). While `RxJava` is good for a lot of things, the current architecture encourages us to use it for everything and this eventually leads to the tests and production code becoming increasingly harder to maintain and refactor. 
+- In addition, `RxJava` has a steep learning curve which requires significant onboarding effort before a new contributor is able to reach an acceptable level of productivity with the current architecture.
+
+We need a scalable framework to build screens that lets us manage business logic and effects, and which is receptive to change.
+
+## Decision
+
+### Goals
+- Separate business logic, presentation logic, and effects so that they can tested independently of each other.
+- Make UI state explicit and save/restore it manually instead of depending on hidden behaviour.
+- Restrict `RxJava` to managing events and effects, and let the business logic be implemented as pure functions.
+
+### Choosing a framework
+We evaluated many patterns and frameworks that are common in the industry, including but not limited to:
+
+##### [Android recommended architecture](https://developer.android.com/jetpack/docs/guide)
+The Android recommended architecture depends on a few architecture components to work together:
+
+- [`ViewModel`](https://developer.android.com/topic/libraries/architecture/viewmodel): To retain the state across configuration changes.
+- [`LiveData`](https://developer.android.com/reference/androidx/lifecycle/LiveData): To provide lifecycle-aware reactive notifications to the UI.
+- [`Lifecyle`](https://developer.android.com/reference/androidx/lifecycle/Lifecycle): To automatically manage subscriptions to the reactive notifications from the `LiveData` instances provided via the `ViewModel`.
+
+This architecture makes sense for new codebases, but this is a codebase with an already existing architecture which does not lend itself well to this specific setup. Some of the problems are:
+
+- The `Lifecycle` component is designed to work with screens that are built on top of `Activity` or `Fragment` classes. However, our current architecture uses a single `Activity` setup, where the individual screens are implemented using `View` subclasses. The `LiveData` component will not give us much benefit unless we transition all the `View` based screens to be `Fragment` instances.
+- We already use `RxJava` for reactive notifications. Using `LiveData` would mean that we would either have to replace usages of `RxJava` with `LiveData` which is neither feasible not desirable at this point.
+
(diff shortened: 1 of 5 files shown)


doc/arch/008-screen-architecture-v3.md
```diff
@@ -0,0 +1,759 @@
...
+
+```kotlin
+interface UiActions {
+  fun goBackToPreviousScreen()
+  fun restartActivity()
+}
+```
+
+### Wiring it all together
+We have implemented a class, `MobiusDelegate`, which is a light wrapper around `Mobius` and `RxJava`, and is designed to encapsulate the work of tying the `Mobius` loop to the Android component lifecycle.
```
msasikanth: Should we link `MobiusDelegate` here?

vinaysshenoy: Good point, I'll do this.

vinaysshenoy: Done.

timcheadle:
I'm reviewing this now. One thing I'd like to see is a brief summary of why we didn't choose the 3 other alternative architectures. That's the kind of thing that helps future decisions.

vinaysshenoy:
> I'm reviewing this now. One thing I'd like to see is a brief summary of why we didn't choose the 3 other alternative architectures. That's the kind of thing that helps future decisions.

Yup, I'm planning to update the ADR later with this information.

doc/arch/008-screen-architecture-v3.md
```diff
@@ -0,0 +1,759 @@
...
+#### Reference implementation
+A reference implementation of the complete architecture can be found at [this commit](https://github.com/simpledotorg/simple-android/tree/e51a6b2fad295f9da4064b95f51d95e3ab2f0e6c/app/src/main/java/org/simple/clinic/settings/changelanguage).
+
+#### Core components
+It is recommended to read the [Concepts](https://github.com/spotify/mobius/wiki/Concepts) page on the Mobius wiki before proceeding.
+
+We will take an example feature along with sample implementations for the various components and key classes:
+
+##### `Model`
+[**DOC**](https://github.com/spotify/mobius/wiki/Concepts#model)
```
sanchita: From what I understand of an ADR, it should talk about why we chose Mobius and how it solves for those reasons we have mentioned before. This document should be a justification for moving to Mobius.

I don't think we need to capture the implementation details and what do the different components look like and how does the architecture work, rather we should capture how those different components help us achieve what the old arch didn't.
Let's talk about this in person if you have doubts.

doc/arch/008-screen-architecture-v3.md
```diff
@@ -0,0 +1,759 @@
...
+We need a scalable framework to build screens that lets us manage business logic and effects, and which is receptive to change.
+
+## Decision
+
+### Goals
+- Fulfill all the requirements of the [v2 architecture](./007-screen-architecture-v2.md#goals).
+- Separate business logic and effects.
+- Separate tests for business logic and effects so that we can manage them more easily.
+- Have a way to migrate from the v1 architecture to the v3 for existing screens.
+
```
sanchita: We should also add the following:
1. Increase our TAT for feature delivery before and how its impacting the overall project
2. Refactoring code is an increasingly hard problem
3. Onboarding engineers to our old arch required sufficient effort in understanding complexity of RxJava.

doc/arch/008-screen-architecture-v3.md
vinaysshenoy: These are all excellent points. My intention in capturing these implementation details is because we have some classes that we have written on top of Mobius to support our new architecture (like the `MobiusDelegate`, for example) and how to test them because we don't have any more documentation on this.

I think we can:
- Split the implementation details out into a separate doc which will live alongside the Mobius migration doc (so we'll have one doc for new screens and another for migrating older screens side-by-side)
- Update the ADR to only capture the decision and the other metrics you talked about.

What do you think?

doc/arch/008-screen-architecture-v3.md
```diff
@@ -0,0 +1,759 @@
...
+We need a scalable framework to build screens that lets us manage business logic and effects, and which is receptive to change.
+
+## Decision
+
+### Goals
+- Fulfill all the requirements of the [v2 architecture](./007-screen-architecture-v2.md#goals).
+- Separate business logic and effects.
+- Separate tests for business logic and effects so that we can manage them more easily.
+- Have a way to migrate from the v1 architecture to the v3 for existing screens.
```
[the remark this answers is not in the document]
vinaysshenoy: They might be. I mostly copied this from the ADR for the intermediate architecture, which had the goals defined here. If we move goals into context here, then we should probably do it there as well for consistency.

doc/arch/008-screen-architecture-v3.md
sanchita: Yea. Sounds good. We can even have a small 30 min discussion tomorrow to make sure our understanding of an ADR is aligned. And I would also love for you to talk about this decision since you have the most context. We can pull-in Pratul so that he can share what writing ADRs mean. Wdyt?

doc/arch/008-screen-architecture-v3.md
```diff
@@ -0,0 +1,759 @@
...
+
+## Consequences
+- In legacy controllers, events were cached and replayed as soon as the screen was inflated without waiting for it to be attached to the view hierarchy. In the v3 architecture, any events which are forwarded to the binding before the screen is attached to the view hierarchy, will be lost and ignored. This is the correct way to implement the events, but some of the older screen controllers which were dependent on that behaviour might take more effort to migrate to the current architecture.
+- Migrating to the v3 architecture from the v1 architecture is a process that takes some time to understand and get used to. Until this migration completes, new developers will need to be onboarded onto both the v1 and v3 architectures so that the codebase can be maintained.
```
sanchita: We should think and add more `Consequences` since this was a huge decision which has to have quite a bit of impact to justify why we chose it. We can discuss this. 
This could even mean that we will have to figure out a way to measure some of the goals we set in the beginning.

doc/arch/008-screen-architecture-v3.md
vinaysshenoy: Works.

vinaysshenoy:
> > I'm reviewing this now. One thing I'd like to see is a brief summary of why we didn't choose the 3 other alternative architectures. That's the kind of thing that helps future decisions.
> 
> Yup, I'm planning to update the ADR later with this information.

Updated.

doc/arch/008-screen-architecture-v3.md
vinaysshenoy: I have made these changes. Can we review this again?

PR #1141, opened 2020-02-14
Allow user to edit a Blood Sugar within 60 minutes of it being added - Part 2

msasikanth:
https://www.pivotaltracker.com/story/show/169676402

### DIFF
--- a/app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryInit.kt
+++ b/app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryInit.kt
@@ -7,8 +7,8 @@ import com.spotify.mobius.Init
 class BloodSugarEntryInit : Init<BloodSugarEntryModel, BloodSugarEntryEffect> {
   override fun init(model: BloodSugarEntryModel): First<BloodSugarEntryModel, BloodSugarEntryEffect> {
     return when (model.openAs) {
-      is New -> first(model, setOf(PrefillDate.forNewEntry()))
-      is Update -> first(model, setOf(FetchBloodSugarMeasurement(model.openAs.bloodSugarMeasurementUuid) as BloodSugarEntryEffect))
+      is OpenAs.New -> first(model, setOf(PrefillDate.forNewEntry()))
+      is OpenAs.Update -> first(model, setOf(FetchBloodSugarMeasurement(model.openAs.bloodSugarMeasurementUuid) as BloodSugarEntryEffect))
     }
   }
 }
--- a/app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
+++ b/app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
@@ -298,6 +298,10 @@ class BloodSugarEntrySheet : BottomSheetActivity(), BloodSugarEntryUi {
     editBloodSugarTitleTextView.text = getString(R.string.bloodsugarentry_edit_fasting_title)
   }
 
+  override fun showRemoveButton() {
+    removeBloodSugarButton.visibility = View.VISIBLE
+  }
+
   override fun dismiss() {
     finish()
   }
--- a/app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
+++ b/app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
@@ -216,14 +216,20 @@ class BloodSugarEntrySheet : BottomSheetActivity(), BloodSugarEntryUi {
   }
 
   override fun showRandomBloodSugarTitle() {
+    enterBloodSugarTitleTextView.visibility = View.VISIBLE
+    editBloodSugarTitleTextView.visibility = View.GONE
     enterBloodSugarTitleTextView.text = getString(R.string.bloodsugarentry_random_title)
   }
 
   override fun showPostPrandialBloodSugarTitle() {
+    enterBloodSugarTitleTextView.visibility = View.VISIBLE
+    editBloodSugarTitleTextView.visibility = View.GONE
     enterBloodSugarTitleTextView.text = getString(R.string.bloodsugarentry_post_prandial_title)
   }
 
   override fun showFastingBloodSugarTitle() {
+    enterBloodSugarTitleTextView.visibility = View.VISIBLE
+    editBloodSugarTitleTextView.visibility = View.GONE
     enterBloodSugarTitleTextView.text = getString(R.string.bloodsugarentry_fasting_title)
   }
 
@@ -272,6 +278,24 @@ class BloodSugarEntrySheet : BottomSheetActivity(), BloodSugarEntryUi {
 
   override fun showDateOnDateButton(date: LocalDate) {
     bloodSugarDateButton.text = dateFormatter.format(date)
+  }
+
+  override fun showEditRadomBloodSugarTitle() {
+    enterBloodSugarTitleTextView.visibility = View.GONE
+    editBloodSugarTitleTextView.visibility = View.VISIBLE
+    editBloodSugarTitleTextView.text = getString(R.string.bloodsugarentry_edit_random_title)
+  }
+
+  override fun showEditPostPrandialBloodSugarTitle() {
+    enterBloodSugarTitleTextView.visibility = View.GONE
+    editBloodSugarTitleTextView.visibility = View.VISIBLE
+    editBloodSugarTitleTextView.text = getString(R.string.bloodsugarentry_edit_post_prandial_title)
+  }
+
+  override fun showEditFastingBloodSugarTitle() {
+    enterBloodSugarTitleTextView.visibility = View.GONE
+    editBloodSugarTitleTextView.visibility = View.VISIBLE
+    editBloodSugarTitleTextView.text = getString(R.string.bloodsugarentry_edit_fasting_title)
   }
 
   override fun dismiss() {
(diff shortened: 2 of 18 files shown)


app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUi.kt
```diff
@@ -18,5 +18,8 @@ interface BloodSugarEntryUi {
   fun hideDateErrorMessage()
   fun setDateOnInputFields(dayOfMonth: String, month: String, twoDigitYear: String)
   fun showDateOnDateButton(date: LocalDate)
+  fun showEditRadomBloodSugarTitle()
+  fun showEditPostPrandialBloodSugarTitle()
+  fun showEditFastingBloodSugarTitle()
```
vinaysshenoy: Can we rename the corresponding methods for setting the title when entering a new blood sugar reading as well?

Example:
`showRandomBloodSugarTitle` -> `showEnterRandomBloodSugarTitle`.

app/src/test/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRendererTest.kt
```diff
@@ -58,4 +59,46 @@ class BloodSugarEntryUiRendererTest {
...
+    // given
+    val testUserClock = TestUserClock()
+    val year = LocalDate.now(testUserClock).year
+    val bloodSugarEntryModel = BloodSugarEntryModel.create(year, Update(bloodSugarUuid, Fasting))
+
+    // when
+    bloodSugarEntryUiRenderer.render(bloodSugarEntryModel)
+
+    // then
+    verify(ui).showEditFastingBloodSugarTitle()
```
vinaysshenoy: Should we also add `verifyNoMoreInteractions(ui)` after the assertions in these methods?

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRenderer.kt
```diff
@@ -35,6 +35,7 @@ class BloodSugarEntryUiRenderer(
 
   private fun showEditBloodSugarTitle(measurementType: BloodSugarMeasurementType) {
     with(ui) {
+      showRemoveButton()
```
vinaysshenoy: The name of this method is `showEditBloodSugarTitle`, but it also does showing of the remove button. Given that the showing of the remove button is not conditionally triggered based on the `measurementType`, it seems to me that this method call here is because of convenience and not because it actually belongs here.

What do you think about moving this into the `render()` method instead? We can just decide to show/hide this based on the `openAs` property.

app/src/test/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRendererTest.kt
```diff
@@ -85,6 +86,7 @@ class BloodSugarEntryUiRendererTest {
     bloodSugarEntryUiRenderer.render(bloodSugarEntryModel)
 
     // then
+    verify(ui).showRemoveButton()
```
vinaysshenoy: Should we add separate tests for this?

`When the sheet is opened for editing a blood sugar measurement, the remove button must be shown`, and `When the sheet is opened for recording a new blood sugar measurement, the remove button must not be shown`

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRenderer.kt
```diff
@@ -25,6 +25,7 @@ class BloodSugarEntryUiRenderer(
 
   private fun showEnterNewBloodSugarTitle(measurementType: BloodSugarMeasurementType) {
     with(ui) {
+      hideRemoveButton()
```
vinaysshenoy: Same comment as earlier about the method being here because of convenience, and moving it into the `render` function.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUpdate.kt
```diff
@@ -33,7 +32,7 @@ class BloodSugarEntryUpdate @AssistedInject constructor(
     return when (event) {
       is ScreenChanged -> next(model.screenChanged(event.type))
       is DatePrefilled -> next(model.datePrefilled(event.prefilledDate))
-      is BloodSugarMeasurementFetched -> noChange()
+      is BloodSugarMeasurementFetched -> onBloodSugarMeasurementFetched(model, event)
```
vinaysshenoy: The prefix `on` indicates that this is some form of callback method, while this is a method that returns a `Next`. I feel we can just drop the `on` prefix.

What do you think?

app/src/test/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandlerTest.kt
```diff
@@ -162,6 +162,22 @@ class BloodSugarEntryEffectHandlerTest {
     verifyNoMoreInteractions(ui)
   }
 
+  @Test
+  fun `set specific date on input fields and date button when prefill date for update effect is received`() {
+    // given
+    val entryDate = LocalDate.of(2020, 2, 14)
+    userClock.setDate(LocalDate.of(2020, 2, 14))
```
vinaysshenoy: Can we reuse `entryDate` for setting the date for `userClock`?

app/src/test/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandlerTest.kt
```diff
@@ -162,6 +162,22 @@ class BloodSugarEntryEffectHandlerTest {
...
+
+    // when
+    testCase.dispatch(PrefillDate.forUpdateEntry(userClock.instant()))
```
vinaysshenoy: We generally use `Instant.now(clock)` instead of `clock.instant()` throughout the codebase. Could we follow this convention here as well, please?

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
```diff
@@ -324,6 +324,12 @@ class BloodSugarEntrySheet : BottomSheetActivity(), BloodSugarEntryUi {
     finish()
```
vinaysshenoy: Was "blood" in the commit message meant to be "blood sugar"?

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryInit.kt
```diff
@@ -7,8 +7,8 @@ import com.spotify.mobius.Init
 class BloodSugarEntryInit : Init<BloodSugarEntryModel, BloodSugarEntryEffect> {
```
vinaysshenoy: Can we add a line to the commit message as to why this change was made?

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUi.kt
msasikanth: In later commits, I am planning on replacing these methods with a single method with just the measurement type parameter. Should I still replace these methods names in this PR?

app/src/test/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRendererTest.kt
msasikanth: Yes, forgot about that

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRenderer.kt
msasikanth: Moved the `showRemoveButton` inside the `setupUi` method

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRenderer.kt
msasikanth: Done 👍

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUpdate.kt
msasikanth: Yeah 👍

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
msasikanth: Yes, reworded the commit message `Disable background click of sheet if blood sugar reading is not blank`

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUi.kt
vinaysshenoy: If you're going to do this, then it's fine. Let's just take a look at it in the end after the feature is complete.

app/src/test/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRendererTest.kt
vinaysshenoy: Discussed offline: the tests are already present and were written earlier.

PR #1143, opened 2020-02-17
Allow user to edit a Blood Sugar within 60 minutes of it being added - Part 3

msasikanth:
https://www.pivotaltracker.com/story/show/169676402


app/src/test/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUpdateTest.kt
```diff
@@ -209,6 +210,33 @@ class BloodSugarEntryUpdateTest {
         ))
   }
 
+  @Test
+  fun `when save is clicked when updating blood sugar and there are no validation errors, then update blood sugar`() {
```
sanchita: Reword this to `when save is clicked after the blood sugar is updated, without any validation errors, then update the blood sugar`?

app/src/test/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUpdateTest.kt
```diff
@@ -209,6 +210,33 @@ class BloodSugarEntryUpdateTest {
...
+        .whenEvent(SaveClicked)
+        .then(assertThatNext(
+            hasNoModel(),
+            hasEffects(UpdateBloodSugarEntry(
+                bloodSugarMeasurementUuid,
+                validBloodSugar.toInt(),
+                Random,
+                validBloodSugarDate,
+                validBloodSugarDate
+            ) as BloodSugarEntryEffect)
```
sanchita: Can we extract this to an `expectedEffect` variable?

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
```diff
@@ -337,6 +337,10 @@ class BloodSugarEntrySheet : BottomSheetActivity(), BloodSugarEntryUi {
     }
   }
 
+  override fun showConfirmRemoveBloodSugarDialog(bloodSugarMeasurementUuid: UUID) {
+
```
sanchita: Please add a TODO to methods that you would add implementation to later.
This helps the reviewer know why the method body is empty.

sanchita: Oh. There is no implementation at all. I was expecting it in the later commits :)
Can you please add a TODO here?

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandler.kt
```diff
@@ -67,6 +67,7 @@ class BloodSugarEntryEffectHandler @AssistedInject constructor(
         .addTransformer(CreateNewBloodSugarEntry::class.java, createNewBloodSugarEntryTransformer())
         .addTransformer(FetchBloodSugarMeasurement::class.java, fetchBloodSugarMeasurement(schedulersProvider.io()))
         .addConsumer(SetBloodSugarReading::class.java, { ui.setBloodSugarReading(it.bloodSugarReading) }, schedulersProvider.ui())
+        .addAction(UpdateBloodSugarEntry::class.java, {})
```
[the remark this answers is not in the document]
msasikanth: If we don't handle a effect that is being dispatched from the update function, effect handler would throw a error when that effect is received. To prevent that I have used a no-op condition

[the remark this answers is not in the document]
sanchita: I can see that you are implementing the operation in the next commit. 
Instead of this being empty which means that this `Effect` has no action which is not true, can we add `TODO`s wherever the implementation is in progress?

app/src/androidTest/java/org/simple/clinic/bloodsugar/BloodSugarRepositoryAndroidTest.kt
```diff
@@ -187,4 +188,35 @@ class BloodSugarRepositoryAndroidTest {
...
+        reading = bloodSugarReading,
+        createdAt = Instant.now(clock),
+        updatedAt = Instant.now(clock),
+        syncStatus = SyncStatus.DONE
+    )
+
+    appDatabase.bloodSugarDao().save(listOf(bloodSugar))
+
+    val durationToAdvanceBy = Duration.ofMinutes(15L)
+    clock.advanceBy(durationToAdvanceBy)
```
sanchita: Advancing the clock should be part of the `when` block. Wdyt?

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandler.kt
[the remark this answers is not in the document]
msasikanth: Done

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandler.kt
```diff
@@ -181,4 +184,53 @@ class BloodSugarEntryEffectHandler @AssistedInject constructor(
...
+                .toSingleDefault(BloodSugarSaved(wasDateChanged))
+          }
+          .compose(reportAnalyticsEvents)
+          .cast<BloodSugarEntryEvent>()
+    }
+  }
+
+  private fun getUpdatedBloodSugarMeasurement(updateBloodSugarEntry: UpdateBloodSugarEntry): BloodSugarMeasurement {
+    val bloodSugarMeasurement = getExistingBloodSugarMeasurement(updateBloodSugarEntry.bloodSugarMeasurementUuid)!!
+    val (user, facility) = userAndCurrentFacility().blockingGet()
```
sanchita: `blockingGet` should not be used in production code as it blocks the thread until something is received.
Let's pair on this and see if we can extract this in the Rx chain itself.

msasikanth: Yes, but that entire block of code is called on io thread. So it should be good?

sanchita: Yea. But using `blockingGet` is a non-reactive way of handling events. This takes away the advantage of using Rx in the first place. We should extract these values inside the stream itself.

PR #1153, opened 2020-02-27
Use ndk folder if ndk-bundle doesn’t exist

msasikanth:


app/ndk-versions.gradle
```diff
@@ -1,10 +1,25 @@
 def installedNdkVersion() {
-  def androidNdkPropertiesLocation = "$System.env.ANDROID_HOME/ndk-bundle/source.properties"
+  def androidNdk = file("$System.env.ANDROID_HOME/ndk")
```
vinaysshenoy: Rename this to `sideBySideNdkFolder`?

app/ndk-versions.gradle
```diff
@@ -1,10 +1,25 @@
...
+  def androidNdkProperties = file("$System.env.ANDROID_HOME/ndk-bundle/source.properties")
```
vinaysshenoy: Rename this to `ndkBundlePropertiesFile`?

PR #1155, opened 2020-02-28
POC: Detect unsubscribed completables in tests

vinaysshenoy:
We currently use `io.reactivex.Completable` instances in a lot of places to perform side-effects. While this works well with our use-case, a common problem is that we sometimes do not subscribe to the instantiated `Completable`.

```kotlin
Observable.just(1)
  .doOnNext { repository.saveToDatabase(it) }

interface Repository {
  fun saveToDatabase(value: Int): Completable
}
```

A test to verify this might be:
```kotlin
verify(repository).saveToDatabase(value)
```

Here, the `doOnNext` should actually have been a `flatMapCompletable` because the created `Completable` from `saveToDatabase(Int)` never gets subscribed to. However, the test never fails because we only verify that `saveToDatabase` gets called, but not that the returned `Completable` gets subscribed to.

This PR attemps to add a check for this. To do this, we

- Hook into the `Completable` assembly to verify that every `Completable` that is instantiated also gets subscribed to.
- Use a jUnit rule (currently piggybacks on `RxErrorsRule`, which is already present in all our test classes) to set up and tear this down in tests.

The end result is that we get an error message like this in tests:
```shell
Found unsubscribed Completables!
--------------------------------
Assembled: 3
Subscribed: 0
Expected to be not subscribed to: 0
--------------------------------
#00 - org.simple.clinic.user.UserSessionTest.setUp(UserSessionTest.kt:89)
#01 - org.simple.clinic.user.UserSessionTest.setUp(UserSessionTest.kt:91)
#02 - org.simple.clinic.user.UserSessionTest.setUp(UserSessionTest.kt:93)
--------------------------------
```

This PR also includes a meta-annotation, `ExpectUnsubscribed`, that we can annotate test methods with to handle cases where the correct behaviour is never to subscribe to a created `Completable`.

### DIFF
--- a/router/build.gradle
+++ b/router/build.gradle
@@ -30,9 +30,6 @@ dependencies {
   implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$versions.kotlin"
   implementation "io.reactivex.rxjava2:rxjava:$versions.rxJava"
   implementation "com.jakewharton.timber:timber:$versions.timber"
-  testImplementation "pl.pragmatists:JUnitParams:$versions.junitParams"
+  testImplementation "com.github.vinaysshenoy:JUnitParams:$versions.junitParams"
   testImplementation "com.google.truth:truth:$versions.truth"
-}
-repositories {
-  mavenCentral()
 }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -9,10 +9,6 @@ apply plugin: 'io.sentry.android.gradle'
 
 apply from: "../quality/install-git-hook.gradle"
 
-repositories {
-  maven { url 'https://jitpack.io' }
-}
-
 sentry {
   autoProguardConfig true
   // We can disable auto upload of proguard mappings since we use -dontobfuscate when running proguard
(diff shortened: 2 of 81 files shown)


vinaysshenoy:
Waiting until #1174 gets merged in so that we add a separate rule for this.

saket:
💯

router/build.gradle
```diff
@@ -30,9 +30,6 @@ dependencies {
   implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$versions.kotlin"
   implementation "io.reactivex.rxjava2:rxjava:$versions.rxJava"
   implementation "com.jakewharton.timber:timber:$versions.timber"
-  testImplementation "pl.pragmatists:JUnitParams:$versions.junitParams"
+  testImplementation "com.github.vinaysshenoy:JUnitParams:$versions.junitParams"
```
saket: Why was forking needed?

vinaysshenoy: The reasoning is in the commit message.

Copying it over here:
```
The main dependency has a bug where annotations added on test methods
are not forwarded to any test rules (Pragmatists/JUnitParams#151),
which breaks the `ExpectedUnsubscribed` annotation used for hinting to
the `RxErrorsRule` that we expect a certain number of `Completable`
instances to not be subscribed to.

There is a PR to fix it (Pragmatists/JUnitParams#158),
but this has been inactive for over a year, so I forked the repository
and applied the fix myself and published it via Jitpack.
```

vinaysshenoy:
Closing this because there's an issue with this rule interacts with Traceur. Will revisit this later.

PR #1157, opened 2020-03-03
Fix crash when editing the date of birth for a patient

vinaysshenoy:
https://www.pivotaltracker.com/story/show/171209605


app/src/main/java/org/simple/clinic/di/DateFormatterModule.kt
```diff
@@ -23,7 +23,7 @@ class DateFormatterModule {
 
   @Provides
   @Named("exact_date")
-  fun providesTimeFormatterForRecentPatientsHeader(locale: Locale): DateTimeFormatter = DateTimeFormatter.ofPattern("d MMMM, yyyy", locale)
+  fun providesTimeFormatterForFullReadableDate(locale: Locale): DateTimeFormatter = DateTimeFormatter.ofPattern("d MMMM, yyyy", locale)
```
msasikanth: Can we rename the method name?
```suggestion
  fun providesDateFormatterForFullReadableDate(locale: Locale): DateTimeFormatter = DateTimeFormatter.ofPattern("d MMMM, yyyy", locale)
```

vinaysshenoy: Done.

PR #1163, opened 2020-03-05
Migrate patient summary to Mobius: epilogue

vinaysshenoy:
https://www.pivotaltracker.com/story/show/170538616

This PR attempts to clean up the `PatientSummaryEffectHandler` in order to make it more "Mobius idiomatic". I managed to get a good bit of the module, but there are other parts left over which I haven't gotten around to yet. I'll leave for later.

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
@@ -715,5 +715,9 @@ class PatientRepository @Inject constructor(
         .or(bloodSugarsChangedSince)
   }
 
+  fun patientProfileImmediate(patientUuid: UUID): Optional<PatientProfile> {
+    return database.patientDao().patientProfileImmediate(patientUuid).toOptional()
+  }
+
   private data class BusinessIdMetaAndVersion(val metaData: String, val metaDataVersion: MetaDataVersion)
 }
--- a/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
@@ -452,6 +452,10 @@ class PatientRepository @Inject constructor(
         }
   }
 
+  fun latestPhoneNumberForPatient(patientUuid: UUID): Optional<PatientPhoneNumber> {
+    return database.phoneNumberDao().latestPhoneNumber(patientUuid).toOptional()
+  }
+
   fun compareAndUpdateRecordedAt(patientUuid: UUID, instantToCompare: Instant): Completable {
     return Completable.fromAction {
       database.patientDao().compareAndUpdateRecordedAt(
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
@@ -77,6 +77,7 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
             { uiActions.showScheduleAppointmentSheet(it.patientUuid, it.sheetOpenedFrom) },
             schedulersProvider.ui()
         )
+        .addTransformer(LoadDataForBackClick::class.java, loadDataForBackClick(schedulersProvider.io()))
         .build()
   }
 
@@ -223,6 +224,24 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
                 .markReminderAsShownFor(effect.patientUuid)
                 .subscribeOn(scheduler)
                 .andThen(Observable.empty<PatientSummaryEvent>())
+          }
+    }
+  }
+
+  private fun loadDataForBackClick(
+      scheduler: Scheduler
+  ): ObservableTransformer<LoadDataForBackClick, PatientSummaryEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(scheduler)
+          .map { loadDataForBackClick ->
+            val patientUuid = loadDataForBackClick.patientUuid
+            val timestamp = loadDataForBackClick.screenCreatedTimestamp
+            DataForBackClickLoaded(
+                hasPatientDataChangedSinceScreenCreated = patientRepository.hasPatientDataChangedSince(patientUuid, timestamp),
+                noBloodPressuresRecordedForPatient = doesNotHaveBloodPressures(patientUuid),
+                noBloodSugarsRecordedForPatient = doesNotHaveBloodSugars(patientUuid)
+            )
           }
     }
   }
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
@@ -18,7 +18,6 @@ import org.simple.clinic.overdue.Appointment
 import org.simple.clinic.overdue.AppointmentCancelReason
 import org.simple.clinic.overdue.AppointmentRepository
 import org.simple.clinic.patient.PatientRepository
-import org.simple.clinic.summary.AppointmentSheetOpenedFrom.DONE_CLICK
 import org.simple.clinic.summary.addphone.MissingPhoneReminderRepository
 import org.simple.clinic.user.UserSession
 import org.simple.clinic.util.Just
@@ -48,10 +47,6 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
     return RxMobius
         .subtypeEffectHandler<PatientSummaryEffect, PatientSummaryEvent>()
         .addTransformer(LoadPatientSummaryProfile::class.java, loadPatientSummaryProfile(schedulersProvider.io()))
-        .addTransformer(HandleDoneClick::class.java, handleDoneClick(
-            backgroundWorkScheduler = schedulersProvider.io(),
-            uiWorkScheduler = schedulersProvider.ui()
-        ))
         .addTransformer(LoadCurrentFacility::class.java, loadCurrentFacility(schedulersProvider.io()))
         .addConsumer(HandleEditClick::class.java, { uiActions.showEditPatientScreen(it.patientSummaryProfile) }, schedulersProvider.ui())
         .addAction(HandleLinkIdCancelled::class.java, { uiActions.goToPreviousScreen() }, schedulersProvider.ui())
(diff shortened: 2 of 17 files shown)


app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
```diff
@@ -68,40 +67,28 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
...
-            }
-            .take(1)
-            .map(::PatientSummaryProfileLoaded)
-            .cast<PatientSummaryEvent>()
-      }
+      effects
+          .observeOn(scheduler)
+          .map { patientRepository.patientProfileImmediate(it.patientUuid) }
+          .filterAndUnwrapJust()
+          .map { it.withoutDeletedBusinessIds().withoutDeletedPhoneNumbers() }
```
sanchita: Do we need to filter this here or we can do it at the query level? 
In what scenario will we need deleted values?

app/src/main/java/org/simple/clinic/summary/PatientSummaryInit.kt
```diff
@@ -8,11 +8,7 @@ import org.simple.clinic.summary.OpenIntention.LinkIdWithPatient
 class PatientSummaryInit : Init<PatientSummaryModel, PatientSummaryEffect> {
 
   override fun init(model: PatientSummaryModel): First<PatientSummaryModel, PatientSummaryEffect> {
-    val effects = mutableSetOf<PatientSummaryEffect>()
-
-    if (!model.hasLoadedPatientSummaryProfile) {
-      effects.add(LoadPatientSummaryProfile(model.patientUuid))
-    }
+    val effects = mutableSetOf<PatientSummaryEffect>(LoadPatientSummaryProfile(model.patientUuid))
```
sanchita: Can you please specify as to why we needed to make this change?
If I understand correctly, it is just to avoid delays in loading the patient profile in this view or is it something else?
We were already taking care of showing edited patient info via screen key?

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -452,6 +452,10 @@ class PatientRepository @Inject constructor(
         }
   }
 
+  fun latestPhoneNumberForPatient(patientUuid: UUID): Optional<PatientPhoneNumber> {
+    return database.phoneNumberDao().latestPhoneNumber(patientUuid).toOptional()
+  }
```
sanchita: Since we will be doing this for a lot more repository methods now, can we start marking the reactive versions of these methods as `Deprecated` yet?

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
```diff
@@ -209,15 +205,20 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
...
-  private fun hasInvalidPhone(patientUuid: UUID): Observable<Boolean> {
-    return patientRepository.phoneNumber(patientUuid)
-        .filterAndUnwrapJust()
-        .map { phoneNumber -> phoneNumber to appointmentRepository.lastCreatedAppointmentForPatient(patientUuid) }
-        .filter { (_, appointmentOptional) -> appointmentOptional is Just }
-        .map { (number, appointmentOptional) -> number to (appointmentOptional as Just).value }
-        .filter { (_, appointment) -> appointment.wasCancelledBecauseOfInvalidPhoneNumber() }
-        .map { (number, appointment) -> appointment.updatedAt > number.updatedAt }
+  private fun hasInvalidPhone(patientUuid: UUID): Boolean {
+    val number = patientRepository.latestPhoneNumberForPatient(patientUuid)
```
sanchita: Please rename to `phoneNumber`

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
```diff
@@ -209,15 +205,20 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
...
+    val appointment = appointmentRepository.lastCreatedAppointmentForPatient(patientUuid)
+
+    return when {
+      number is None || appointment is None -> false
```
sanchita: Replace with `phoneNumber.isEmpty() || appointment.isEmpty()`

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
```diff
@@ -224,6 +225,23 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
...
+  ): ObservableTransformer<LoadDataForBackClick, PatientSummaryEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(scheduler)
+          .map { loadDataForBackClick ->
+            val patientUuid = loadDataForBackClick.patientUuid
+            val timestamp = loadDataForBackClick.screenCreatedTimestamp
+            DataForBackClickLoaded(
+                hasPatientDataChangedSinceScreenCreated = patientRepository.hasPatientDataChangedSince(patientUuid, timestamp),
+                noBloodPressuresRecordedForPatient = doesNotHaveBloodPressures(patientUuid)
```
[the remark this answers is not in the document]
vinaysshenoy: 1. We cannot make `hasPatientDataChangedSinceScreenCreated` part of the model because it involves the database.
2. I initially thought of making them a part of the model, but then I realised that they should not be there.
  - If we look at how these values are being used in `PatientSummaryUpdate`, they never lead to a change in the screen state because they always result in a `dispatch` and not in a `next`. They are never used in the `ViewRenderer`.
  - We use them to decide navigation which should not be handled as part of the regular screen flow and should be handled elsewhere. The only reason we have them here is because we are not going to make too many changes at once (limiting ourselves to breaking out a separate `UiActions` only).

Does this make sense?

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
vinaysshenoy: The scenario in which we need deleted values is because this specific query is also used for fetching data for syncing, in which case we should not exclude deleted values.

One option is that we can add an alternate repository method that lets us drop these records at the query level, but I thought that this PR was quite large enough already without adding it.

This is why I wanted to add this here first and see what someone else has to say about it.

app/src/main/java/org/simple/clinic/summary/PatientSummaryInit.kt
vinaysshenoy: If we edit the patient and come back to the patient summary screen, the details on the header do not get updated immediately because we restore the last saved model instead of fetching the latest.

A "correct" approach is to detect from a screen result whether the patient has been edited and then requery it, but it seemed like a lot of maintenance effort for a local database call so I decided to query the patient details again.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: I do not specifically want to mark them as `Deprecated` because there might be use-cases where we actually want them to be reactive streams. What I am thinking is that we should prefer the non-Rx versions by default but then use the Rx versions if we really need to.

Does this make sense?

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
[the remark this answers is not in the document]
sanchita: >     1. We cannot make `hasPatientDataChangedSinceScreenCreated` part of the model because it involves the database.
> 
This is why I was recommending to move this logic to the `Model` so that we don't put this in the database. Also, this would mean that we don't use `Effects` to handle navigation either.
>     2. I initially thought of making them a part of the model, but then I realised that they should not be there.
> 
> 
>     * If we look at how these values are being used in `PatientSummaryUpdate`, they never lead to a change in the screen state because they always result in a `dispatch` and not in a `next`. They are never used in the `ViewRenderer`.
> 
That's right and that is the point I am making. Because we rely on a database call to decide whether user should navigate to `ScheduleAppointmentSheet` or to exit, we end up dispatching an effect. I am suggesting that we use `UiRenderer` for this change by making it part of the `Model`.
>     * We use them to decide navigation which should not be handled as part of the regular screen flow and should be handled elsewhere. The only reason we have them here is because we are not going to make too many changes at once (limiting ourselves to breaking out a separate `UiActions` only).
> 
> 
> Does this make sense?
I just wanted to call out that we can utilize Mobius to achieve the navigation logic rather than the database state.
Let's discuss this in person. As of now I do not expect you to change this implementation.

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
sanchita: I would recommend using either a query to filter this out or the repository would make more sense. But this is alright for now. I just wanted to know the reason.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
sanchita: Things like these end up becoming tribal knowledge. I wanted to avoid that. 
But it does make sense to keep the Rx versions for now in case we need them.

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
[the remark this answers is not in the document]
vinaysshenoy: The fact that they're being used only for navigation is precisely why they should not be added to the `Model`. The `Model` is meant to record only the data that is needed to render the screen and nothing else.

Let's discuss this in person tomorrow.

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
vinaysshenoy: Let's take a look at the rest of the screens if we need this model (`PatientProfile`) there and if yes, we'll add this as a separate repository method.

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
vinaysshenoy: Done

PR #1165, opened 2020-03-05
Support blood sugars for overdue high risk measurement

rakshakhegde:
https://www.pivotaltracker.com/story/show/170067831

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
@@ -17,6 +17,9 @@ import org.simple.clinic.TestData
 import org.simple.clinic.bloodsugar.BloodSugarMeasurement
 import org.simple.clinic.bloodsugar.BloodSugarReading
 import org.simple.clinic.bloodsugar.BloodSugarRepository
+import org.simple.clinic.bloodsugar.Fasting
+import org.simple.clinic.bloodsugar.HbA1c
+import org.simple.clinic.bloodsugar.PostPrandial
 import org.simple.clinic.bloodsugar.Random
 import org.simple.clinic.bp.BloodPressureMeasurement
 import org.simple.clinic.bp.BloodPressureRepository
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
```diff
@@ -52,9 +52,9 @@ data class OverdueAppointmentRow(
           gender = overdueAppointment.gender,
           age = DateOfBirth.fromOverdueAppointment(overdueAppointment, clock).estimateAge(clock),
           phoneNumber = overdueAppointment.phoneNumber?.number,
-          bpSystolic = overdueAppointment.bloodPressure.systolic,
-          bpDiastolic = overdueAppointment.bloodPressure.diastolic,
-          bpDaysAgo = calculateDaysAgoFromInstant(overdueAppointment.bloodPressure.recordedAt, clock),
+          bpSystolic = overdueAppointment.bloodPressure?.systolic ?: 0,
```
msasikanth: I don't think we should use an elvis operator here, wdyt?

rakshakhegde: We'll keep this PR from being merged, until the PR on showing last seen gets merged.

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -1818,6 +1821,228 @@ class AppointmentRepositoryAndroidTest {
     assertThat(overdueAppointments).containsExactlyElementsIn(expectedAppointments)
   }
 
+  @Test
+  fun patients_with_different_combinations_of_blood_pressure_and_blood_sugars_should_be_included_in_overdue_list() {
```
msasikanth: We can maybe add blood sugar to the existing test `high_risk_patients_should_be_present_at_the_top_when_loading_overdue_appointments` instead of a new test, wdyt?

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
```diff
@@ -62,6 +58,10 @@ data class OverdueAppointment(
 
           (
             CASE
+              WHEN BloodSugar.reading_type = "fasting" AND CAST (BloodSugar.reading_value AS REAL) >= 200 THEN 1
+              WHEN BloodSugar.reading_type = "random" AND CAST (BloodSugar.reading_value AS REAL) >= 300 THEN 1
+              WHEN BloodSugar.reading_type = "post_prandial" AND CAST (BloodSugar.reading_value AS REAL) >= 300 THEN 1
```
msasikanth:
```suggestion
              WHEN (BloodSugar.reading_type = "random" OR BloodSugar.reading_type == "post_prandial") AND CAST (BloodSugar.reading_value AS REAL) >= 300 THEN 1
```

We can maybe replace both cases with this, wdyt?

rakshakhegde: To maintain consistency, let's keep this separate?

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
rakshakhegde: I tried doing this, but the function did not have a way to add blood sugars. Also, the test is already quite long I feel.

PR #1168, opened 2020-03-06
Show confirmation alert on switching facility

sanchita:
https://www.pivotaltracker.com/story/show/171623909


app/src/main/java/org/simple/clinic/facility/change/confirm/ConfirmFacilityChangeSheet.kt
```diff
@@ -0,0 +1,30 @@
...
+
+  companion object {
+    lateinit var component: ConfirmFacilityChangeComponent
+  }
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+
+    setContentView(R.layout.sheet_confirm_facility_change)
+    setupDi()
```
msasikanth: ~I think we are supposed to do it in `attachBaseContext`, can you PTAL?~ 
NVM it has been added in a later commit.

app/src/main/java/org/simple/clinic/facility/change/confirm/ConfirmFacilityChangeSheet.kt
```diff
@@ -1,15 +1,29 @@
...
+
+    private const val SELECTED_FACILITY = "seleceted_facility"
+
+    fun intent(
+        context: Context,
+        facility: Facility
+    ): Intent {
+      val intent = Intent(context, ConfirmFacilityChangeSheet::class.java)
+      intent.putExtra(SELECTED_FACILITY, facility)
+      return intent
```
msasikanth: Should we inline the return, wdyt?
```suggestion
      return Intent(context, ConfirmFacilityChangeSheet::class.java)
          .putExtra(SELECTED_FACILITY, facility)
```

app/src/main/java/org/simple/clinic/facility/change/confirm/ConfirmFacilityChangeSheet.kt
```diff
@@ -39,11 +45,39 @@ class ConfirmFacilityChangeSheet : BottomSheetActivity() {
...
   }
 
+  private val delegate by unsafeLazy {
+    MobiusDelegate.forActivity(
+        events.ofType(),
+        ConfirmFacilityChangeModel(),
+        ConfirmFacilityChangeUpdate(),
+        effectHandlerFactory.create(this).build(),
+        ConfirmFacilityChangeInit(),
+        { /* No-op, there's nothing to render */ }
```
msasikanth: We don't have to provide no-op for `modelUpdateListener` in Mobius loop anymore. `MobiusDelegate.forActivity` has default parameter for `modelUpdateListener` so you can remove the no-op.

app/src/main/java/org/simple/clinic/facility/change/confirm/ConfirmFacilityChangeSheet.kt
```diff
@@ -24,7 +24,6 @@ class ConfirmFacilityChangeSheet : BottomSheetActivity(), ConfirmFacilityChangeU
     lateinit var component: ConfirmFacilityChangeComponent
```
msasikanth: There is a typo in the commit message
"Close facility selection when **cancel** is clicked"

app/src/main/java/org/simple/clinic/facility/change/confirm/ConfirmFacilityChangeUpdate.kt
```diff
@@ -10,7 +10,8 @@ class ConfirmFacilityChangeUpdate : Update<ConfirmFacilityChangeModel, ConfirmFa
       event: ConfirmFacilityChangeEvent
   ): Next<ConfirmFacilityChangeModel, ConfirmFacilityChangeEffect> {
     return when (event) {
-      FacilityChangeConfirmed -> dispatch(ChangeFacilityEffect)
+      is FacilityChangeConfirmed -> dispatch(ChangeFacilityEffect(event.selectedFacility))
+      FacilityChanged -> TODO()
```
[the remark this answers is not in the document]
sanchita: This is a `TODO` and not a `noChange`. It will be implemented in the next commits.
Also, if I make it a `noChange()` that would be an incorrect implementation.
Since the sheet is not being called from anywhere yet, it's better to keep it as a `TODO`.

app/src/main/java/org/simple/clinic/facility/change/confirm/ConfirmFacilityChangeSheet.kt
sanchita: Yea. I forgot that the first time.

app/src/main/java/org/simple/clinic/facility/change/confirm/ConfirmFacilityChangeSheet.kt
sanchita: I can. But I find this to be more readable. Plus we are following this everywhere else. So unless this is something that you feel will be useful, I'd rather not change it.

app/src/main/java/org/simple/clinic/facility/change/confirm/ConfirmFacilityChangeSheet.kt
msasikanth: Did you forget to remove the no-op for `modelUpdateListener` here? 😅

PR #1173, opened 2020-03-10
Add alternate triggers for syncing the frequent syncs

vinaysshenoy:
https://www.pivotaltracker.com/story/show/171718198

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
@@ -22,7 +22,7 @@ class PatientSummaryUpdate : Update<PatientSummaryModel, PatientSummaryEvent, Pa
       is CurrentFacilityLoaded -> next(model.currentFacilityLoaded(event.facility))
       PatientSummaryEditClicked -> dispatch(HandleEditClick(model.patientSummaryProfile!!))
       is PatientSummaryLinkIdCancelled -> dispatch(HandleLinkIdCancelled)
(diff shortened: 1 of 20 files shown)


app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
```diff
@@ -22,7 +22,7 @@ class PatientSummaryUpdate : Update<PatientSummaryModel, PatientSummaryEvent, Pa
       is CurrentFacilityLoaded -> next(model.currentFacilityLoaded(event.facility))
       PatientSummaryEditClicked -> dispatch(HandleEditClick(model.patientSummaryProfile!!))
       is PatientSummaryLinkIdCancelled -> dispatch(HandleLinkIdCancelled)
-      is ScheduledAppointment -> scheduleAppointmentSheetClosed(event, model)
+      is ScheduledAppointment -> dispatch(TriggerSync(event.sheetOpenedFrom))
```
msasikanth: This will get triggered even if the user doesn't schedule an appointment but the schedule sheet is closed right?

vinaysshenoy: It won't. The `ScheduledAppointment` event is triggered only if an appointment is scheduled.

msasikanth: What if a BP or Blood sugar is recorded and the appointment is not scheduled? Will we sync at regular intervals like before?

vinaysshenoy: Yes, the regular syncs will continue to happen. This PR only adds additional sync triggers.

PR #1174, opened 2020-03-11
Integrate tool to detect flaky tests

vinaysshenoy:
# Problem
We currently have a problem (mostly with the instrumented tests) with some tests being flaky. If tests sometimes fail, and sometimes pass, we end up re-running the entire test suite, which takes time and is wasteful.

This tool is mostly an effort to get ongoing visibility into
- Which of our tests are the *most* flaky?
- Which of the tests can start causing more trouble going forward?

This will give us insights into what tests need to be fixed on priority.

Dashboard: https://flakiness.dx.obvious.in/simple-android-instrumented-tests/reports


vinaysshenoy:
> Overall PR looks good.
> Can you confirm that the CI check is enabled to just report flaky tests right now and nothing else right?
> Also if you could add a link to the dashboard?

It does one more thing: If a test is detected as flaky in this run (i.e., if it fails and then passes upon repetition, it marks the test as `PASSED`). This is intentionally done because we don't want  to re-run on flaky tests.

vinaysshenoy:
> Overall PR looks good.
> Can you confirm that the CI check is enabled to just report flaky tests right now and nothing else right?
> Also if you could add a link to the dashboard?

Added a link to dashboard in the PR.

vinaysshenoy:
> Why are we removing `RunWith` annotation for Android tests?

Because it is unnecessary. By default, it already runs with our runner.

PR #1179, opened 2020-03-15
Show diagnosis label in overdue list

rakshakhegde:
https://www.pivotaltracker.com/story/show/170068614

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
@@ -32,7 +32,11 @@ data class OverdueAppointment(
 
     val isAtHighRisk: Boolean,
 
-    val patientLastSeen: Instant
+    val patientLastSeen: Instant,
+
+    val diagnosedWithDiabetes: Answer?,
+
+    val diagnosedWithHypertension: Answer?
 ) {
 
   @Dao
@@ -47,7 +51,9 @@ data class OverdueAppointment(
 
           PPN.uuid phone_uuid, PPN.patientUuid phone_patientUuid, PPN.number phone_number, PPN.phoneType phone_phoneType, PPN.active phone_active,
           PPN.createdAt phone_createdAt, PPN.updatedAt phone_updatedAt,
-          
+
+          MH.hasDiabetes diagnosedWithDiabetes, MH.diagnosedWithHypertension diagnosedWithHypertension,
+
           (
             CASE
                 WHEN BP.uuid IS NULL THEN BloodSugar.recordedAt
@@ -84,7 +90,7 @@ data class OverdueAppointment(
           ) BloodSugar ON BloodSugar.patientUuid = P.uuid
 
           WHERE 
-            P.deletedAt IS NULL 
+            P.deletedAt IS NULL
             AND A.deletedAt IS NULL
             AND A.facilityUuid = :facilityUuid
             AND A.status = :scheduledStatus
--- a/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
@@ -1,12 +1,14 @@
 package org.simple.clinic.home.overdue
 
 import android.annotation.SuppressLint
+import android.content.Context
 import android.view.View.GONE
 import android.view.View.VISIBLE
 import androidx.recyclerview.widget.DiffUtil
 import io.reactivex.subjects.Subject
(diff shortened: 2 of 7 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -296,6 +296,7 @@
     <item quantity="one">%s day overdue</item>
     <item quantity="other">%s days overdue</item>
   </plurals>
+  <string name="overduelist_diagnosis_label">Dx:</string>
   <string name="overduelist_last_seen">Last seen:</string>
```
sanchita: Please update these labels as per new designs.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
```diff
@@ -131,10 +136,23 @@ data class OverdueAppointmentRow(
...
+    holder.diagnosisTextView.text = diagnosisText()
+
     updateBottomLayoutVisibility(holder)
     updatePhoneNumberViewVisibility(holder)
   }
 
+  private fun diagnosisText(): CharSequence {
+    return listOf(
+        diagnosedWithDiabetes to "Diabetes",
+        diagnosedWithHypertension to "Hypertension"
```
sanchita: These need to be translated. Please add them in `strings.xml`

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
```diff
@@ -131,10 +136,23 @@ data class OverdueAppointmentRow(
...
+    )
+        .filter { (answer, _) -> answer is Answer.Yes }
+        .map { (_, diagnosisTitle) -> diagnosisTitle }
+        .ifEmpty { listOf("None") }
```
sanchita: Please check if this is the expected behavior. If so, please add this string to `strings.xml`

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
```diff
@@ -32,7 +32,11 @@ data class OverdueAppointment(
 
     val isAtHighRisk: Boolean,
 
-    val patientLastSeen: Instant
+    val patientLastSeen: Instant,
+
+    val diagnosedWithDiabetes: Answer?,
+
+    val diagnosedWithHypertension: Answer?
```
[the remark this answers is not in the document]
rakshakhegde: There are a lot more fields in `MedicalHistory` that are not relevant right now and the query becomes unnecessarily huge if we do include them. When we need more fields, we can make this change as and when needed.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
```diff
@@ -32,7 +32,11 @@ data class OverdueAppointment(
 
     val isAtHighRisk: Boolean,
 
-    val patientLastSeen: Instant
+    val patientLastSeen: Instant,
+
+    val diagnosedWithDiabetes: Answer?,
```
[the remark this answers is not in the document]
rakshakhegde: We are performing a left join on `MedicalHistory`, so it can potentially be `null`. This is similar to `PatientPhoneNumber`.

app/src/main/res/values/strings.xml
rakshakhegde: Done

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
rakshakhegde: Done

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
rakshakhegde: I had spoken to Pragati and we had decided that we'll show "None". But now that Pragati will show us new designs this might change again tomorrow.

rakshakhegde: But I have put it under string.xml regardless

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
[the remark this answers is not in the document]
sanchita: The problem then is that happens when the value is `Unanswered`?
You will have to handle that case explicitly.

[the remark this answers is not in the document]
rakshakhegde: I've included Unanswered too

PR #1180, opened 2020-03-16
Unify the conditions for checking blood pressure and blood sugar on "Back" and "Done" events in patient summary screen

vinaysshenoy:
https://www.pivotaltracker.com/story/show/171762828


app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
```diff
@@ -32,8 +32,7 @@ class PatientSummaryUpdate : Update<PatientSummaryModel, PatientSummaryEvent, Pa
       is DataForBackClickLoaded -> dataForHandlingBackLoaded(
```
msasikanth: Can you reword this commit to fit in 1 line maybe?

"Change `PatientSummaryUpdate` to use count of recorded measurements on back click"

PR #1182, opened 2020-03-17
Use dedicated DI components for all activities

vinaysshenoy:
https://www.pivotaltracker.com/story/show/166179356

# Issue
We currently use `TheActivity` to inject all other activities in the app. This causes crashes in the case of process death and restoration if one of these other activities is in the foreground when the process gets killed because when the restoration happens:

- Activity 2 which was on top of the stack gets created first and tries to inject itself using `TheActivity`.
- `TheActivity` hasn't been restored yet, so we get a `NullPointerException` when trying to fetch `TheActivityComponent`.

# Fix
This PR fixes the issue by creating dedicated Dagger components for each activity.


vinaysshenoy:
@msasikanth The commit ordering in the PR in correct. Please check out the branch locally for the actual ordering.

msasikanth:
LGTM. Can you rebase the PR, please?

- Probably not for this PR, but maybe can start moving towards `@Component.Factory` and avoid having build method for every subcomponent. We can move the create/build to `BindsActivity`. wdyt?

vinaysshenoy:
> LGTM. Can you rebase the PR, please?
> 
> * Probably not for this PR, but maybe can start moving towards `@Component.Factory` and avoid having build method for every subcomponent. We can move the create/build to `BindsActivity`. wdyt?

I haven't looked at `@Component.Factory` yet. Will check this out and evaluate.

PR #1183, opened 2020-03-18
Revisit logic for scheduling an automatic appointment when the user has not scheduled one

msasikanth:
https://www.pivotaltracker.com/story/show/170068992

In the previous PR we are only checking if the blood sugar value is not null. Deepa has added the blood sugar cut off values in the PRD. I have updated the query to include those values.

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/Patient.kt
+++ b/app/src/main/java/org/simple/clinic/patient/Patient.kt
@@ -245,7 +245,9 @@ data class Patient(
             CASE
               WHEN
                 (BP.systolic >= 140 OR BP.diastolic >= 90
-                OR BloodSugar.bloodSugarValue IS NOT NULL
+                OR (BloodSugar.type = "random" AND BloodSugar.value >= 300)
+                OR (BloodSugar.type = "fasting" AND BloodSugar.value >= 200)
+                OR (BloodSugar.type = "hba1c" AND BloodSugar.value >= 9)
                 OR MH.hasHadHeartAttack = :yesAnswer
                 OR MH.hasHadStroke = :yesAnswer
                 OR PD.uuid IS NOT NULL)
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -245,7 +245,9 @@ data class Patient(
             CASE
               WHEN
                 (BP.systolic >= 140 OR BP.diastolic >= 90
-                OR BloodSugar.bloodSugarValue IS NOT NULL
+                OR (BloodSugar.type = "random" AND BloodSugar.value >= 200)
+                OR (BloodSugar.type = "fasting" AND BloodSugar.value >= 300)
```
sanchita: There is a mistake here. The `fasting` and `random` values are switched.
Please refer the PRD and correct these.
Also update the tests accordingly.

msasikanth: Fixed it 👍

app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -245,7 +245,9 @@ data class Patient(
             CASE
               WHEN
                 (BP.systolic >= 140 OR BP.diastolic >= 90
-                OR BloodSugar.bloodSugarValue IS NOT NULL
+                OR (BloodSugar.type = "random" AND BloodSugar.value >= 300)
+                OR (BloodSugar.type = "fasting" AND BloodSugar.value >= 200)
```
sanchita: Can you add test for high `Random`?

msasikanth: I added `1 Normal Random Blood Sugar`. Should I add a blood sugar which crosses the cut off as well?

msasikanth: Added a test for high `Random Blood Sugar`

PR #1186, opened 2020-03-18
Fix: Add phone dialog gets shown when the patient has a phone number

vinaysshenoy:
https://www.pivotaltracker.com/story/show/171689706

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/Patient.kt
+++ b/app/src/main/java/org/simple/clinic/patient/Patient.kt
@@ -9,12 +9,16 @@ import androidx.room.Index
 import androidx.room.PrimaryKey
 import androidx.room.Query
 import io.reactivex.Flowable
+import io.reactivex.Observable
 import kotlinx.android.parcel.Parcelize
 import org.intellij.lang.annotations.Language
 import org.simple.clinic.medicalhistory.Answer
 import org.simple.clinic.overdue.Appointment
 import org.simple.clinic.patient.businessid.BusinessId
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/simple/clinic/summary/PatientSummaryProfile.kt
```diff
@@ -14,4 +14,8 @@ data class PatientSummaryProfile(
     val phoneNumber: PatientPhoneNumber?,
```
msasikanth: Can you change the commit message to fit in one line, please?

app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -186,9 +190,11 @@ data class Patient(
           .map(::queryModelsToPatientProfiles)
     }
 
-    fun patientProfile(patientUuid: UUID): Flowable<PatientProfile> {
+    fun patientProfile(patientUuid: UUID): Observable<Optional<PatientProfile>> {
```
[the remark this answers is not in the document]
vinaysshenoy: Earlier, that method was only being used in tests and we were forcefully getting a `PatientProfile` (by using `.first()` on the returned list).

Now since we are using it in the production source set, changed it to an `Optional` since we don't know in what use cases we will be using this method.

app/src/main/java/org/simple/clinic/summary/PatientSummaryProfile.kt
vinaysshenoy: Done.

PR #1187, opened 2020-03-18
User must select a diagnosis for a patient from the patient summary screen

msasikanth:
https://www.pivotaltracker.com/story/show/169676456

### DIFF
--- a/app/src/main/java/org/simple/clinic/medicalhistory/MedicalHistory.kt
+++ b/app/src/main/java/org/simple/clinic/medicalhistory/MedicalHistory.kt
@@ -8,8 +8,9 @@ import androidx.room.OnConflictStrategy
 import androidx.room.PrimaryKey
 import androidx.room.Query
 import io.reactivex.Flowable
-import org.simple.clinic.medicalhistory.MedicalHistoryQuestion.DIAGNOSED_WITH_HYPERTENSION
+import org.simple.clinic.medicalhistory.Answer.Unanswered
 import org.simple.clinic.medicalhistory.MedicalHistoryQuestion.DIAGNOSED_WITH_DIABETES
+import org.simple.clinic.medicalhistory.MedicalHistoryQuestion.DIAGNOSED_WITH_HYPERTENSION
 import org.simple.clinic.medicalhistory.MedicalHistoryQuestion.HAS_HAD_A_HEART_ATTACK
 import org.simple.clinic.medicalhistory.MedicalHistoryQuestion.HAS_HAD_A_KIDNEY_DISEASE
 import org.simple.clinic.medicalhistory.MedicalHistoryQuestion.HAS_HAD_A_STROKE
@@ -44,6 +45,8 @@ data class MedicalHistory(
 
     val deletedAt: Instant?
 ) {
+  val diagnosisRecorded: Boolean
+    get() = diagnosedWithHypertension != Unanswered && diagnosedWithDiabetes != Unanswered
 
   fun answered(question: MedicalHistoryQuestion, answer: Answer): MedicalHistory {
     return when (question) {
--- a/app/src/main/java/org/simple/clinic/medicalhistory/MedicalHistory.kt
+++ b/app/src/main/java/org/simple/clinic/medicalhistory/MedicalHistory.kt
@@ -101,6 +101,14 @@ data class MedicalHistory(
     fun historyForPatient(patientUuid: PatientUuid): Flowable<List<MedicalHistory>>
 
     @Query("""
+      SELECT * FROM MedicalHistory
+      WHERE patientUuid = :patientUuid AND deletedAt IS NULL
+      ORDER BY updatedAt DESC
+      LIMIT 1
+    """)
+    fun historyForPatientImmediate(patientUuid: PatientUuid): MedicalHistory?
+
+    @Query("""
         SELECT (
             CASE
                 WHEN (COUNT(uuid) > 0) THEN 1
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
@@ -175,8 +175,9 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
           .observeOn(scheduler)
           .map { loadDataForBackClick ->
             val patientUuid = loadDataForBackClick.patientUuid
+            val medicalHistory = (medicalHistoryRepository.historyForPatient(patientUuid) as Just<MedicalHistory>).value
 
-            DataForDoneClickLoaded(countOfRecordedMeasurements = countOfRecordedMeasurements(patientUuid), diagnosisRecorded = true)
+            DataForDoneClickLoaded(countOfRecordedMeasurements = countOfRecordedMeasurements(patientUuid), diagnosisRecorded = medicalHistory.diagnosisRecorded)
           }
     }
   }
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
@@ -176,7 +176,7 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
           .map { loadDataForBackClick ->
             val patientUuid = loadDataForBackClick.patientUuid
 
-            DataForDoneClickLoaded(countOfRecordedMeasurements = countOfRecordedMeasurements(patientUuid))
+            DataForDoneClickLoaded(countOfRecordedMeasurements = countOfRecordedMeasurements(patientUuid), diagnosisRecorded = true)
           }
     }
   }
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
@@ -154,7 +154,8 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
 
             DataForBackClickLoaded(
                 hasPatientDataChangedSinceScreenCreated = patientRepository.hasPatientDataChangedSince(patientUuid, timestamp),
-                countOfRecordedMeasurements = countOfRecordedMeasurements(patientUuid)
+                countOfRecordedMeasurements = countOfRecordedMeasurements(patientUuid),
+                diagnosisRecorded = true
             )
           }
     }
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
@@ -72,6 +72,7 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
         .addTransformer(LoadDataForBackClick::class.java, loadDataForBackClick(schedulersProvider.io()))
         .addTransformer(LoadDataForDoneClick::class.java, loadDataForDoneClick(schedulersProvider.io()))
         .addTransformer(TriggerSync::class.java, triggerSync())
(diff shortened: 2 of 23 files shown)


app/src/test/java/org/simple/clinic/summary/PatientSummaryEffectHandlerTest.kt
```diff
@@ -189,4 +189,14 @@ class PatientSummaryEffectHandlerTest {
...
     verifyZeroInteractions(uiActions)
   }
+
+  @Test
+  fun `when show diagnosis error effect is received, then show diagnosis error`() {
+    // when
+    testCase.dispatch(ShowDiagnosisError)
+
+    // then
+    verify(uiActions).showDiagnosisError()
```
vinaysshenoy: Can we also assert no outgoing events on the `testCase`?

app/src/main/res/values-bn-rBD/strings.xml
```diff
@@ -481,7 +481,7 @@
   <string name="bloodsugarhistory_toolbar_title">%1$s, %2$s, %3$s</string>
   <!-- Summary CV History view -->
   <string name="medicalhistorysummaryview_history">ইতিহাস</string>
-  <string name="medicalhistorysummaryview_diagnosis">রোগ নির্ণয়</string>
+  <string name="medicalhistorysummaryview_diagnosis">রোগ নির্ণয় (প্রয়োজন)</string>
```
vinaysshenoy: How did the string resource for this change? Did you copy this from Transifex?

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
```diff
@@ -231,6 +229,11 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
     return countOfRecordedBloodPressures + countOfRecordedBloodSugars
   }
 
+  private fun diagnosisRecorded(patientUuid: UUID): Boolean {
+    val medicalHistory = medicalHistoryRepository.historyForPatient(patientUuid)
+    return (medicalHistory.diagnosedWithDiabetes != Unanswered && medicalHistory.diagnosedWithHypertension != Unanswered)
```
vinaysshenoy: Can we move this boolean logic into the `MedicalHistory` model as a computed property? That way, this gets defined in one place, and is a property of the domain instead of being in the effect handler.

vinaysshenoy: Once we do this, we can also add a few simple unit tests that `diagnosisRecorded` is computed correctly.

app/src/test/java/org/simple/clinic/summary/PatientSummaryEffectHandlerTest.kt
```diff
@@ -179,6 +188,32 @@ class PatientSummaryEffectHandlerTest {
     verifyZeroInteractions(uiActions)
   }
 
+  @Test
+  fun `when the load data for done click effect is received and diagnosis is not recorded, load the data`() {
```
vinaysshenoy: We're adding all this boilerplate just to verify that a boolean field is being computed correctly. The computation is essentially a pure function and should be part of business logic, but we are doing this because this is currently in the effect handler.

If we make that earlier change about moving the computation for `diagnosisRequired` into the `MedicalHistory` class, we can remove this test and have very simple tests elsewhere which just validate the computation of the property.

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
```diff
@@ -47,11 +48,16 @@ class PatientSummaryUpdate : Update<PatientSummaryModel, PatientSummaryEvent, Pa
...
       patientUuid: UUID,
-      countOfRecordedMeasurements: Int
+      countOfRecordedMeasurements: Int,
+      diagnosisRecorded: Boolean
   ): Next<PatientSummaryModel, PatientSummaryEffect> {
     val hasAtLeastOneMeasurementRecorded = countOfRecordedMeasurements > 0
 
-    val effect = if (hasAtLeastOneMeasurementRecorded) ShowScheduleAppointmentSheet(patientUuid, DONE_CLICK) else GoToHomeScreen
+    val effect = when {
+      hasAtLeastOneMeasurementRecorded && diagnosisRecorded.not() -> ShowDiagnosisError
```
vinaysshenoy: Extract this into a separate variable?

```kotlin
    val hasAtLeastOneMeasurementRecorded = countOfRecordedMeasurements > 0
    val shouldShowDiagnosisError = hasAtLeastOneMeasurementRecorded && diagnosisRecorded.not()
    
    val effect = when {
      shouldShowDiagnosisError -> ShowDiagnosisError
      hasAtLeastOneMeasurementRecorded -> ShowScheduleAppointmentSheet(patientUuid, DONE_CLICK)
      else -> GoToHomeScreen
    }
```

app/src/main/java/org/simple/clinic/medicalhistory/MedicalHistory.kt
```diff
@@ -100,6 +100,14 @@ data class MedicalHistory(
     """)
     fun historyForPatient(patientUuid: PatientUuid): Flowable<List<MedicalHistory>>
 
+    @Query("""
+      SELECT * FROM MedicalHistory
+      WHERE patientUuid = :patientUuid
```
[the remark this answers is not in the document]
msasikanth: Can medical histories for a patient be deleted? 🤔

app/src/main/res/values-bn-rBD/strings.xml
msasikanth: I got from new patient medical history

app/src/main/java/org/simple/clinic/medicalhistory/MedicalHistory.kt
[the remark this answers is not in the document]
vinaysshenoy: Not from the app, but we should act as if any record can be marked as deleted from the server side going forward.

app/src/main/res/values-bn-rBD/strings.xml
vinaysshenoy: What about the strings in other languages?

Ideally, we should *only* change the source file (english), unless we are renaming keys or deleting keys. Adding new strings should go through the standard translation workflow.

msasikanth: The text `Diagnosis (Required)` has already been translated. So I am just adding `Required` part to the existing string

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
```diff
@@ -38,7 +38,8 @@ class PatientSummaryUpdate : Update<PatientSummaryModel, PatientSummaryEvent, Pa
       )
```
vinaysshenoy: Typo in the commit message? 

"Show diagnosis error **if** diagnosis is not recorded and done is clicked"

app/src/main/java/org/simple/clinic/medicalhistory/MedicalHistory.kt
```diff
@@ -8,8 +8,9 @@ import androidx.room.OnConflictStrategy
 import androidx.room.PrimaryKey
```
vinaysshenoy: Check the commit message? Should it be `diagnosisRecorded`?

app/src/test/java/org/simple/clinic/medicalhistory/MedicalHistoryTest.kt
```diff
@@ -0,0 +1,55 @@
...
+  @Test
+  fun `if diabetes and hypertension is answered then diagnosis required is true`() {
+    val medicalHistory = TestData.medicalHistory(
+        uuid = UUID.fromString("0bd13198-2dca-4a06-b795-6ee45ef8572b"),
+        diagnosedWithHypertension = Yes,
+        hasDiabetes = Yes
+    )
+
+    assertThat(medicalHistory.diagnosisRecorded).isTrue()
+  }
```
vinaysshenoy: Nice! Can I ask you do a couple of small changes to these tests, please?

- Change all the test method names to "should be" instead of "is" when declaring should happen. Ex: `if diabetes and hypertension is answered then diagnosis required should be true`

app/src/test/java/org/simple/clinic/summary/PatientSummaryUpdateTest.kt
```diff
@@ -71,8 +71,10 @@ class PatientSummaryUpdateTest {
 
   @Test
   fun `when there are patient summary changes and at least one measurement is present, clicking on back must show the schedule appointment sheet`() {
+    val model = defaultModel.copy(currentFacility = facility)
```
vinaysshenoy: `PatientSummaryModel` already has a method, `currentFacilityLoaded()`. Could you use that instead since that is the canonical way to update the model with the facility?

app/src/test/java/org/simple/clinic/summary/PatientSummaryUpdateTest.kt
```diff
@@ -85,9 +87,14 @@ class PatientSummaryUpdateTest {
   }
 
   @Test
-  fun `when there are patient summary changes and at least one measurement is present and no diagnosis is recorded, then clicking on back must show diagnosis error`() {
+  fun `when there are patient summary changes and at least one measurement is present and no diagnosis is recorded and diabetes management is enabled, then clicking on back must show diagnosis error`() {
```
vinaysshenoy: Can we add a corresponding test to verify the behaviour is unchanged when the diabetes management flag is disabled?

app/src/test/java/org/simple/clinic/summary/PatientSummaryUpdateTest.kt
```diff
@@ -189,9 +208,14 @@ class PatientSummaryUpdateTest {
   }
 
   @Test
-  fun `when at least one measurement is present and diagnosis is not recorded, clicking on save must show diagnosis error`() {
+  fun `when at least one measurement is present and diagnosis is not recorded and diabetes management is enabled, clicking on save must show diagnosis error`() {
```
vinaysshenoy: Can we add a corresponding test to verify the behaviour is unchanged when the diabetes management flag is disabled?

app/src/test/java/org/simple/clinic/summary/PatientSummaryUpdateTest.kt
```diff
@@ -85,9 +87,14 @@ class PatientSummaryUpdateTest {
   }
 
   @Test
-  fun `when there are patient summary changes and at least one measurement is present and no diagnosis is recorded, then clicking on back must show diagnosis error`() {
+  fun `when there are patient summary changes and at least one measurement is present and no diagnosis is recorded and diabetes management is enabled, then clicking on back must show diagnosis error`() {
+    val facility = TestData.facility(
+        facilityConfig = FacilityConfig(diabetesManagementEnabled = true)
+    )
```
vinaysshenoy: We can create two facilities in the test class:

```kotlin
private val facilityWithDiabetesManagementEnabled = TestData.facility(
  uuid = UUID.fromString("abe86f8e-1828-48fe-afb5-d697b3ce36bb"),
  facilityConfig = FacilityConfig(diabetesManagementEnabled = true)
)

private val facilityWithDiabetesManagementDisabled = TestData.facility(
  uuid = UUID.fromString("abe86f8e-1828-48fe-afb5-d697b3ce36bb"),
  facilityConfig = FacilityConfig(diabetesManagementEnabled = false)
)
```

And then use these in tests as per requirement.

app/src/test/java/org/simple/clinic/summary/PatientSummaryUpdateTest.kt
msasikanth: I added them but looks like I forgot to commit the changes 😅

PR #1193, opened 2020-03-23
Update CHANGELOG

vinaysshenoy:


CHANGELOG.md
```diff
@@ -1,4 +1,26 @@
 # CHANGELOG
+
+## Next release
+### Feature
+- Add support for recording HbA1c blood sugars
+- Add a diagnosis label on the overdue appointments
```
msasikanth: We also started showing Last seen instead of Last BP. This is a user-facing feature change, right?

vinaysshenoy: Right, I'll update changelog.

vinaysshenoy: Or wait, it's not actually a feature. The user cannot use it at all. It's a change that we've made and it is captured in the "Internal" section.

PR #1198, opened 2020-03-23
Fix `HbA1c` naming

msasikanth:


app/src/main/java/org/simple/clinic/bloodsugar/history/adapter/BloodSugarHistoryListItem.kt
```diff
@@ -115,7 +115,11 @@ sealed class BloodSugarHistoryListItem : ItemAdapter.Item<Event> {
           .build()
       val isBloodSugarHigh = measurement.reading.isHigh
 
-      holder.readingsTextView.text = "${bloodSugarReading.displayValue} ${unitForReadingType(context, bloodSugarReading.type)} ${textForReadingType(context, bloodSugarReading.type)}"
+      holder.readingsTextView.text = if (bloodSugarReading.type == HbA1c) {
+        "${bloodSugarReading.displayValue}${unitForReadingType(context, bloodSugarReading.type)} ${textForReadingType(context, bloodSugarReading.type)}"
+      } else {
+        "${bloodSugarReading.displayValue} ${unitForReadingType(context, bloodSugarReading.type)} ${textForReadingType(context, bloodSugarReading.type)}"
+      }
```
vinaysshenoy: This is getting a little hard to read. Can we break these out into separate variables?

```kotlin
val readingPrefix = bloodSugarReading.displayValue
val readingSuffix = "${unitForReadingType(context, bloodSugarReading.type)} ${textForReadingType(context, bloodSugarReading.type)}"

holder.readingsTextView.text = if (bloodSugarReading.type == HbA1c) {
  "$readingPrefix$readingSuffix"}
else {
   "$readingPrefix $readingSuffix"      
}
```

One more option we can explore is adding the separator character to the reading:

```kotlin
data class BloodSugarReading {

val displayUnitSeparator: String
    get() {
      return when (type) {
        Random, PostPrandial, Fasting, is Unknown -> " "
        HbA1c -> ""
      }
    }
}
```
and then use the separator character when formatting.

What do you think?

msasikanth: Yeah, I was considering that as well. But it felt like it was a UI detail, should I go for it? 🤔

vinaysshenoy: It is a UI detail, but I think it's important to keep things clean here as well.

PR #1200, opened 2020-03-25
Add CI check to verify Room schemas are not changed

vinaysshenoy:


msasikanth:
Should we also add a check to verify if database version is increased when schema changes?

vinaysshenoy:
> Should we also add a check to verify if database version is increased when schema changes?

This check implicitly does that since if the schema changes, instead of a new schema file getting generated, it will update the existing one.

PR #1201, opened 2020-03-26
Design changes and fixes

rakshakhegde:
Confirmed with Pragati that this feature has to be changed to accept 4 digits. So I'll skip this commit and make this change in another PR.
~https://www.pivotaltracker.com/story/show/171296031~
https://www.pivotaltracker.com/story/show/171854372
https://www.pivotaltracker.com/story/show/170942753

### DIFF
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
@@ -28,6 +28,7 @@ import io.reactivex.rxkotlin.ofType
 import kotlinx.android.synthetic.main.screen_manual_patient_entry.view.*
 import org.simple.clinic.R
 import org.simple.clinic.ReportAnalyticsEvents
+import org.simple.clinic.appconfig.Country
 import org.simple.clinic.facility.FacilityRepository
 import org.simple.clinic.main.TheActivity
 import org.simple.clinic.medicalhistory.newentry.NewMedicalHistoryScreenKey
@@ -118,6 +119,9 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
   @Inject
   lateinit var inputFields: InputFields
 
+  @Inject
+  lateinit var country: Country
+
   // FIXME This is temporally coupled to `scrollToFirstFieldWithError()`.
   private val allTextInputFields: List<EditText> by unsafeLazy {
     listOf(
@@ -197,6 +201,7 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
     identifierTextView.setCompoundDrawableStartWithTint(R.drawable.patient_id_card, R.color.grey1)
 
     setConsentText()
+    setConsentLabelText()
 
     delegate.prepare()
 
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -519,7 +519,7 @@
   <string name="bloodsugarentry_error_higher_limit">Value cannot be more than 1000 mg/dl</string>
   <string name="bloodsugarentry_error_higher_limit_hba1c">Value cannot be more than 25%</string>
   <string name="bloodsugarentry_error_empty">Value cannot be empty</string>
-  <string name="bloodsugarentry_error_date_invalid_pattern">Enter a valid date in DD/MM/YY</string>
+  <string name="bloodsugarentry_error_date_invalid_pattern">Enter a valid date in DD/MM/YYYY</string>
```
sanchita: The pivotal story says that :
"Backdating a BP error message says the date must be DD/MM/YYYY, but it actually requires a 2 digit date".
It is a 2 digit date. Why are we changing it to 4 digits?

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
```diff
@@ -255,6 +260,14 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
     consentTextView.text = consentText
   }
 
+  private fun setConsentLabelText() {
+    val consentLabelTextResId: Int = when (country.isoCountryCode) {
+      Country.BANGLADESH -> R.string.patiententry_consent_sms_reminders
+      else -> R.string.patiententry_consent_whatsapp_sms_reminders
+    }
+    consentLabel.setText(consentLabelTextResId)
```
sanchita: Can we refactor this to remove the type `Int` and change `setText` to `text` in order to make it more Kotlin idiomatic?
```
    val consentLabelTextResId = when (country.isoCountryCode) {
      Country.BANGLADESH -> R.string.patiententry_consent_sms_reminders
      else -> R.string.patiententry_consent_whatsapp_sms_reminders
    }
    consentLabel.text = resources.getString(consentLabelTextResId)
```

app/src/main/res/values/strings.xml
rakshakhegde: The pivotal story says that right now the app has only 2 digits but it must actually be YYYY.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
rakshakhegde: This doesn't really change much. I prefer to keep it as it is.

app/src/main/res/values/strings.xml
sanchita: No. We enter 2 digits. Can you please check the app?

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
sanchita: It does not make sense to have the type specified as it is derived already.
Can you please make that change atleast?

app/src/main/res/values/strings.xml
rakshakhegde: You are right. Confirmed with Pragati that this feature has to be changed to accept 4 digits. So I'll skip this commit and make this change in another PR.

PR #1206, opened 2020-03-26
Alert users for facility change in patient registration

sanchita:
https://www.pivotaltracker.com/story/show/171623970

This PR adds the `AlertFacilityChangeSheet` and takes care of the navigation based on the flow. The next PR will have the implementation to show this sheet when the user switches facility.

### DIFF
--- a/app/src/main/java/org/simple/clinic/facility/alertchange/AlertFacilityChangeSheet.kt
+++ b/app/src/main/java/org/simple/clinic/facility/alertchange/AlertFacilityChangeSheet.kt
@@ -1,0 +1,49 @@
+package org.simple.clinic.facility.alertchange
+
+import android.content.Context
+import android.content.Intent
+import android.os.Bundle
+import kotlinx.android.synthetic.main.sheet_alert_facility_change.*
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/simple/clinic/facility/alertchange/AlertFacilityChangeSheet.kt
```diff
@@ -0,0 +1,49 @@
...
+    private const val CURRENT_FACILITY = "current_facility"
+
+    fun intent(context: Context, currentFacility: Facility): Intent {
+      val intent = Intent(context, AlertFacilityChangeSheet::class.java)
+      intent.putExtra(CURRENT_FACILITY, currentFacility)
+      return intent
+    }
+  }
+
+  private val currentFacility by lazy {
```
msasikanth: Maybe we should pass `Facility.name` instead of the entire `Facility` object? wdyt?

sanchita: Yes. I will do this.

PR #1209, opened 2020-03-27
Rename "Recent Patients" string to "Recent"

msasikanth:
Pragati asked for this change along with facility name truncate.


app/src/main/res/values/strings.xml
```diff
@@ -139,7 +139,7 @@
   <string name="patients_loggedoutalert_message">You have been logged out of your previous device.</string>
   <string name="patients_loggedoutalert_dismiss">Got it</string>
   <string name="patients_scan_simple_card">Scan BP Passport</string>
-  <string name="patients_recentpatients_title">Recent Patients</string>
+  <string name="patients_recentpatients_title">Recent</string>
```
sanchita: This change also needs to happen on the `SEE ALL` page. Can you change that string as well?

PR #1210, opened 2020-03-27
Enable incremental annotation processing

vinaysshenoy:
We had two libraries that were not supporting annotation processing.

- Room
- Moshi

I bumped them up to their latest versions and enabled it.


app/src/main/java/org/simple/clinic/patient/businessid/BusinessIdMetaData.kt
```diff
@@ -1,10 +1,12 @@
 package org.simple.clinic.patient.businessid
```
msasikanth: Should these changes be part of this PR?

vinaysshenoy: Yup, the moshi dependency requires this.

PR #1211, opened 2020-03-27
Alert to confirm facility selection during sign-up flow 

msasikanth:
https://www.pivotaltracker.com/story/show/171858938

### DIFF
--- a/app/src/main/java/org/simple/clinic/registration/confirmfacility/ConfirmFacilitySheet.kt
+++ b/app/src/main/java/org/simple/clinic/registration/confirmfacility/ConfirmFacilitySheet.kt
@@ -1,0 +1,13 @@
+package org.simple.clinic.registration.confirmfacility
+
+import android.os.Bundle
+import org.simple.clinic.R
+import org.simple.clinic.widgets.BottomSheetActivity
+
+class ConfirmFacilitySheet : BottomSheetActivity() {
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.sheet_registration_confirm_facility)
+  }
+}
--- a/app/src/main/java/org/simple/clinic/registration/confirmfacility/ConfirmFacilitySheet.kt
+++ b/app/src/main/java/org/simple/clinic/registration/confirmfacility/ConfirmFacilitySheet.kt
@@ -1,9 +1,12 @@
 package org.simple.clinic.registration.confirmfacility
 
+import android.app.Activity
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
+import kotlinx.android.synthetic.main.sheet_registration_confirm_facility.*
 import org.simple.clinic.R
+import org.simple.clinic.util.unsafeLazy
 import org.simple.clinic.widgets.BottomSheetActivity
 import java.util.UUID
 
@@ -29,8 +32,33 @@ class ConfirmFacilitySheet : BottomSheetActivity() {
     }
   }
 
+  private val facilityName: String by unsafeLazy {
+    intent.getStringExtra(EXTRA_FACILITY_NAME)
+  }
+
+  private val facilityUuid: UUID by unsafeLazy {
+    intent.getSerializableExtra(EXTRA_FACILITY_UUID) as UUID
+  }
+
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     setContentView(R.layout.sheet_registration_confirm_facility)
+
+    facilityNameTextView.text = facilityName
+
+    yesButton.setOnClickListener {
+      confirmFacilitySelection()
+    }
+
+    cancelButton.setOnClickListener {
+      finish()
+    }
+  }
+
+  private fun confirmFacilitySelection() {
+    val intent = Intent()
+    intent.putExtra(EXTRA_FACILITY_UUID, facilityUuid)
+    setResult(Activity.RESULT_OK, intent)
+    finish()
   }
 }
--- a/app/src/main/java/org/simple/clinic/registration/confirmfacility/ConfirmFacilitySheet.kt
+++ b/app/src/main/java/org/simple/clinic/registration/confirmfacility/ConfirmFacilitySheet.kt
@@ -1,10 +1,29 @@
 package org.simple.clinic.registration.confirmfacility
 
+import android.content.Context
+import android.content.Intent
 import android.os.Bundle
 import org.simple.clinic.R
 import org.simple.clinic.widgets.BottomSheetActivity
+import java.util.UUID
 
 class ConfirmFacilitySheet : BottomSheetActivity() {
+
+  companion object {
+    private const val EXTRA_FACILITY_UUID = "extra_facility_uuid"
+    private const val EXTRA_FACILITY_NAME = "extra_facility_name"
+
+    fun intentForConfirmFacilitySheet(
+        context: Context,
+        facilityUuid: UUID,
+        facilityName: String
+    ): Intent {
+      val intent = Intent(context, ConfirmFacilitySheet::class.java)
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenController.kt
```diff
@@ -207,11 +207,7 @@ class RegistrationFacilitySelectionScreenController @Inject constructor(
         .ofType<RegistrationFacilityClicked>()
         .map { it.facility }
         .flatMap { facility ->
-          userSession.ongoingRegistrationEntry()
-              .map { it.copy(facilityId = facility.uuid) }
-              .flatMapCompletable { userSession.saveOngoingRegistrationEntry(it) }
-              .andThen(userSession.saveOngoingRegistrationEntryAsUser())
-              .andThen(Observable.just { ui: Ui -> ui.openRegistrationScreen() })
+          Observable.just { ui: Ui -> ui.showConfirmFacilitySheet(facility.uuid, facility.name) }
```
vinaysshenoy: We can replace this with a `.map`.

```kotlin
return events
        .ofType<RegistrationFacilityClicked>()
        .map { it.facility }
        .map { facility -> { ui: Ui -> ui.showConfirmFacilitySheet(facility.uuid, facility.name) } }
```

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreen.kt
```diff
@@ -64,7 +67,8 @@ class RegistrationFacilitySelectionScreen(context: Context, attrs: AttributeSet)
             searchQueryChanges(),
             retryClicks(),
             facilityClicks(),
-            locationPermissionChanges()
+            locationPermissionChanges(),
+            activityResultStream()
         ),
         controller = controller,
         screenDestroys = RxView.detaches(this).map { ScreenDestroyed() }
```
vinaysshenoy: We can extract `RxView.detaches(this).map { ScreenDestroyed() }` as a separate variable here and then use that both while setting up the events and pass it as a parameter to `activityResultStream`.

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreen.kt
```diff
@@ -104,6 +108,21 @@ class RegistrationFacilitySelectionScreen(context: Context, attrs: AttributeSet)
           .facilityClicks
           .map(::RegistrationFacilityClicked)
 
+  private fun activityResultStream(): Observable<UiEvent> {
```
vinaysshenoy: Rename this to `registrationFacilityConfirmations`?

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenEvents.kt
```diff
@@ -27,3 +28,7 @@ data class RegistrationFacilityUserLocationUpdated(val location: LocationUpdate)
 data class RegistrationFacilityLocationPermissionChanged(val result: RuntimePermissionResult) : UiEvent {
   override val analyticsName = "Registration:Facility Selection:Location Permission Changed to $result"
 }
+
+data class RegistrationFacilityConfirmed(val facilityUuid: UUID) : UiEvent {
+  override val analyticsName: String = "Registration:Confirm Facility Selection"
```
vinaysshenoy: Can we change the `analyticsName` to be as an action that has happened?

Ex: `Registration:Confirmed Facility Selection`.

vinaysshenoy: We format all the analytics events within a screen in a specific way. Can you follow that convention, please? This is missing the `:Facility Selection:` part.

app/src/test/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenControllerTest.kt
```diff
@@ -408,6 +408,22 @@ class RegistrationFacilitySelectionScreenControllerTest {
     verify(screen).showConfirmFacilitySheet(facility1.uuid, facility1.name)
   }
 
+  @Test
+  fun `when a facility is confirmed then the ongoing entry should be updated with selected facility and the user should be logged in`() {
+    val ongoingEntry = OngoingRegistrationEntry(
+        uuid = UUID.randomUUID(),
```
vinaysshenoy: Can we fix this `UUID` please?

app/src/test/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenControllerTest.kt
```diff
@@ -408,6 +408,22 @@ class RegistrationFacilitySelectionScreenControllerTest {
...
+        phoneNumber = "1234567890",
+        fullName = "Ashok",
+        pin = "1234",
+        pinConfirmation = "5678",
+        createdAt = Instant.now())
```
vinaysshenoy: Can we fix the `Instant` instead of using the current time, please?

app/src/main/java/org/simple/clinic/registration/confirmfacility/ConfirmFacilitySheet.kt
```diff
@@ -29,8 +31,33 @@ class ConfirmFacilitySheet : BottomSheetActivity() {
     }
   }
 
+  private val facilityName: String by lazy {
+    intent.getStringExtra(EXTRA_FACILITY_NAME)
+  }
+
+  private val facilityUuid: UUID by lazy {
+    intent.getSerializableExtra(EXTRA_FACILITY_UUID) as UUID
+  }
```
vinaysshenoy: We can use `by unsafeLazy` here since we will only invoke this from the UI thread.

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreen.kt
```diff
@@ -109,20 +109,21 @@ class RegistrationFacilitySelectionScreen(context: Context, attrs: AttributeSet)
...
+        .takeUntil(onDestroy())
         .map { it.data!! }
         .map { intent ->
           val confirmedFacilityUuid = ConfirmFacilitySheet.confirmedFacilityUuid(intent)
           RegistrationFacilityConfirmed(confirmedFacilityUuid)
         }
   }
 
+  private fun onDestroy() =
+      detaches().map { ScreenDestroyed() }
```
vinaysshenoy: Since this stream is required only in `onFinishInflate()`, we can make this stream a variable instead of a separate method.

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreen.kt
msasikanth: I have done that in latter commits when migrating to rxbinding 3

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreen.kt
msasikanth: It's being called in `onFinishInflate()` & `registrationFacilityConfirmations()`

vinaysshenoy: Yeah, but `registrationFacilityConfirmations` is itself called inside `onFinishInflate`, so we can create a variable and pass it as a parameter.

msasikanth: Done. PTAL

PR #1213, opened 2020-03-30
Cancel patient link ID on back press

msasikanth:
Back press when showing `linkIdWithPatientView` should cancel the link id with the patient, this will take back the user to search results instead of the home screen.

https://www.pivotaltracker.com/story/show/171308016

Since `LinkIdWithPatientView` is a subview of `PatientSummaryScreen` I am not sure if we can handle the back clicks separately for that view, so instead I just went this approach. But maybe once we move towards the navigation component we can migrate `LinkIdWithPatientView` to a proper bottom sheet?

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
@@ -45,6 +45,7 @@ import org.simple.clinic.util.identifierdisplay.IdentifierDisplayAdapter
 import org.simple.clinic.util.unsafeLazy
 import org.simple.clinic.widgets.UiEvent
 import org.simple.clinic.widgets.hideKeyboard
+import org.simple.clinic.widgets.isVisible
 import org.simple.clinic.widgets.scrollToChild
 import org.simple.clinic.widgets.visibleOrGone
 import java.util.UUID
@@ -161,7 +162,13 @@ class PatientSummaryScreen(
 
     return backButton.clicks()
         .mergeWith(hardwareBackKeyClicks)
-        .map { PatientSummaryBackClicked(screenKey.patientUuid, screenKey.screenCreatedTimestamp) }
+        .map {
+          if (linkIdWithPatientView.isVisible) {
+            PatientSummaryLinkIdCancelled
+          } else {
+            PatientSummaryBackClicked(screenKey.patientUuid, screenKey.screenCreatedTimestamp)
+          }
+        }
   }
 
   private fun bloodPressureSaves(): Observable<PatientSummaryBloodPressureSaved> {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/simple/clinic/summary/linkId/LinkIdWithPatientView.kt
```diff
@@ -163,4 +163,6 @@ class LinkIdWithPatientView(
         endAction = runAfter
     )
   }
+
+  fun isVisible(): Boolean = (visibility == View.VISIBLE)
```
vinaysshenoy: This seems like a good candidate for an extension function for the `View` class. What do you think of moving this to `Views.kt` and making it an extension function?

vinaysshenoy:
> Back press when showing `linkIdWithPatientView` should cancel the link id with the patient, this will take back the user to search results instead of the home screen.
> 
> https://www.pivotaltracker.com/story/show/171308016
> 
> Since `LinkIdWithPatientView` is a subview of `PatientSummaryScreen` I am not sure if we can handle the back clicks separately for that view, so instead I just went this approach. But maybe once we move towards the navigation component we can migrate `LinkIdWithPatientView` to a proper bottom sheet?

Sounds good. It was made a subview because of some issue with `Flow` anyway (you can read about it in the javadoc for the class).

app/src/main/java/org/simple/clinic/summary/linkId/LinkIdWithPatientView.kt
msasikanth: So the reason I avoided it is, if we use androidx `core-ktx` package we can get several extension functions for view. Instead of trying to reimplement those I was hoping we can add it maybe. wdyt?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -6,6 +6,7 @@ import android.os.Parcelable
 import android.text.style.TextAppearanceSpan
 import android.util.AttributeSet
 import android.view.View
+import android.view.View.VISIBLE
```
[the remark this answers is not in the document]
msasikanth: Oops, forgot to remove unused import.

app/src/main/java/org/simple/clinic/widgets/Views.kt
```diff
@@ -279,3 +279,9 @@ fun ViewPropertyAnimator.setDuration(duration: Duration): ViewPropertyAnimator {
 
 val Int.dp: Int
   get() = (Resources.getSystem().displayMetrics.density * this).toInt()
+
+inline var View.isVisible: Boolean
```
vinaysshenoy: I didn't know vars could be inlined. 😮

vinaysshenoy: Can we make it only a computed `val` property which doesn't have a setter? A caller might assume setting `isVisible` to `true` makes it `INVISIBLE` from the name, but the implementation sets it to `GONE`.

Let callers use the platform `setVisibility` method for changing it as necessary.

msasikanth: Sure 👍

PR #1215, opened 2020-03-30
Change `PinEntryCardView` to be able to do both local and remote PIN verifications

vinaysshenoy:
https://www.pivotaltracker.com/story/show/171923997

# Changes
- Add a `PinVerificationMethod` and interfaces that implement the current behaviour
  - For verifying during login by reading the PIN digest from `OngoingLoginEntry`
  - For verifying during app unlock by reading the PIN digest from `User`
- Change `PinEntryEffectHandler` to dynamically select which method to use via an XML attribute which can be set by the screen it is embedded in.

### DIFF
--- a/app/src/main/java/org/simple/clinic/security/pin/PinEntryCardView.kt
+++ b/app/src/main/java/org/simple/clinic/security/pin/PinEntryCardView.kt
@@ -229,10 +229,6 @@ class PinEntryCardView(context: Context, attrs: AttributeSet) : CardView(context
     pinEditText.text = null
   }
 
-  override fun dispatchAuthenticatedCallback(enteredPin: String) {
-    downstreamUiEvents.onNext(PinAuthenticated(enteredPin))
-  }
-
   override fun pinVerified(data: Any?) {
     // TODO: Change the type to an `Any?`
     downstreamUiEvents.onNext(PinAuthenticated(data as String))
--- a/app/src/main/java/org/simple/clinic/security/pin/PinEntryCardView.kt
+++ b/app/src/main/java/org/simple/clinic/security/pin/PinEntryCardView.kt
@@ -19,6 +19,8 @@ import org.simple.clinic.R
 import org.simple.clinic.ReportAnalyticsEvents
 import org.simple.clinic.di.injector
 import org.simple.clinic.mobius.MobiusDelegate
+import org.simple.clinic.security.di.Method
+import org.simple.clinic.security.pin.verification.PinVerificationMethod
 import org.simple.clinic.util.UtcClock
 import org.simple.clinic.util.exhaustive
 import org.simple.clinic.util.unsafeLazy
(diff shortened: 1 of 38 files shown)


app/src/test/java/org/simple/clinic/security/pin/verification/OngoingLoginEntryPinVerificationMethodTest.kt
```diff
@@ -0,0 +1,52 @@
...
+    val entry = TestData.ongoingLoginEntry(
+        pinDigest = passwordHasher.hash(correctPassword)
+    )
+    whenever(repository.entryImmediate()) doReturn entry
+
+    // when
+    val result = pinVerificationMethod.verify(correctPassword)
+
+    // then
+    Truth.assertThat(result).isEqualTo(PinVerificationMethod.VerificationResult.Correct(correctPassword))
```
msasikanth: Can you add the import for `com.google.common.truth.Truth.assertThat`, like we do with the rest of the tests?

app/src/test/java/org/simple/clinic/security/pin/verification/OngoingLoginEntryPinVerificationMethodTest.kt
```diff
@@ -0,0 +1,52 @@
...
+        pinDigest = passwordHasher.hash(correctPassword)
+    )
+    whenever(repository.entryImmediate()) doReturn entry
+
+    // when
+    val enteredPassword = "1111"
+    val result = pinVerificationMethod.verify(enteredPassword)
+
+    // then
+    Truth.assertThat(result).isEqualTo(PinVerificationMethod.VerificationResult.Incorrect(enteredPassword))
```
msasikanth: Can you add the import for `com.google.common.truth.Truth.assertThat`, like we do with the rest of the tests?

app/src/main/java/org/simple/clinic/security/pin/PinEntryEffect.kt
```diff
@@ -26,4 +26,4 @@ object ShowProgress: PinEntryEffect()
 
 object ClearPin: PinEntryEffect()
 
-data class PinVerified(val pin: String): PinEntryEffect()
+data class DispatchPinVerified(val pin: String): PinEntryEffect()
```
msasikanth: ~Why are we renaming this effect?~ NVM we are deleting it in later commits.

app/src/main/java/org/simple/clinic/security/pin/PinEntryCardView.kt
```diff
@@ -233,6 +233,11 @@ class PinEntryCardView(context: Context, attrs: AttributeSet) : CardView(context
     downstreamUiEvents.onNext(PinAuthenticated(enteredPin))
   }
 
+  override fun pinVerified(data: Any?) {
+    // TODO: Change the type to an `Any?`
```
msasikanth: Should we remove this TODO, wdyt?

app/src/main/java/org/simple/clinic/security/pin/PinEntryEffect.kt
```diff
@@ -29,3 +29,5 @@ object ClearPin: PinEntryEffect()
 data class DispatchPinVerified(val pin: String): PinEntryEffect()
 
 data class VerifyPin(val pin: String): PinEntryEffect()
+
+data class DispatchCorrectPinEntered(val pinVerifiedData: Any?): PinEntryEffect()
```
msasikanth: Can we name it as `CorrectPinEntered` instead of `DispatchCorrectPinEntered`?

app/src/main/java/org/simple/clinic/security/pin/PinEntryCardView.kt
```diff
@@ -73,11 +72,7 @@ class PinEntryCardView(context: Context, attrs: AttributeSet) : CardView(context
   }
 
   private val events by unsafeLazy {
-    Observable
-        .merge(
-            pinTextChanges(),
-            upstreamUiEvents
-        )
+    pinTextChanges()
```
msasikanth: There seems to be an unused import for rx `Observable` when you did this, can you fix that, please?

app/src/main/java/org/simple/clinic/security/pin/PinEntryCardView.kt
vinaysshenoy: This will get changed in a later PR.

app/src/main/java/org/simple/clinic/security/pin/PinEntryCardView.kt
vinaysshenoy: I've fixed that in another PR.

PR #1218, opened 2020-03-31
Implement a remote PIN verification in the PIN entry component

vinaysshenoy:
https://www.pivotaltracker.com/story/show/171874779

### DIFF
--- a/app/src/test/java/org/simple/clinic/security/pin/PinEntryEffectHandlerTest.kt
+++ b/app/src/test/java/org/simple/clinic/security/pin/PinEntryEffectHandlerTest.kt
@@ -1,0 +1,61 @@
+package org.simple.clinic.security.pin
+
+import com.nhaarman.mockito_kotlin.mock
+import com.nhaarman.mockito_kotlin.verify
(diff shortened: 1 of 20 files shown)


vinaysshenoy:
Removing @sanchita and adding @msasikanth for review because she is on leave.

app/src/test/java/org/simple/clinic/security/pin/PinEntryEffectHandlerTest.kt
```diff
@@ -0,0 +1,58 @@
...
+    testCase.dispose()
+  }
+
+  @Test
+  fun `when the show network error effect is received, the network error must be shown`() {
+    // when
+    testCase.dispatch(ShowNetworkError)
+
+    // then
+    verify(uiActions).showNetworkError()
```
msasikanth: Should we also assert that there are no out going events?

vinaysshenoy: Good point, I'll do this.

PR #1222, opened 2020-04-01
Do not show Transgender option in gender for Ethiopia

msasikanth:
https://www.pivotaltracker.com/story/show/169392408

### DIFF
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
@@ -156,6 +156,7 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
 
     showOrHideInputFields()
     setInputFieldsHint()
+    showOrHideGenderRadioButtons()
   }
 
   private fun showOrHideInputFields() {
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -181,6 +181,17 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
...
+    val allGendersRadioButtons = mapOf(
+        Male to maleRadioButton,
+        Female to femaleRadioButton,
+        Transgender to transgenderRadioButton
+    )
+    val genderField = inputFields.fields.find { it is GenderField } as GenderField
+
+    allGendersRadioButtons.forEach { (gender, radioButton) ->
+      radioButton.visibleOrGone(gender in genderField.allowedGenders)
+    }
```
vinaysshenoy, line 193: This is currently hiding or showing the fields, but the name of the method is `setInputFieldsHint`. The implementation does something which is not clear from the name.

Can we either:
- Rename the method that makes sense for the implementation, or
- Move the hiding or showing of fields to a separate method?

msasikanth: Moved the hiding or showing gender radio buttons to `showOrHideGenderRadioButtons` method.

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -181,6 +182,17 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
   }
 
   private fun setInputFieldsHint() {
+    val allGendersRadioButtons = mapOf(
```
vinaysshenoy, line 185: Since this map is being used in two different methods, what do you think of extracting it out of the method and pass it as a parameter to both the methods?

msasikanth: Oh no, that's a mistake. I forgot to remove this in `setInputFieldsHunt`. I moved this logic into a separate method.

PR #1226, opened 2020-04-01
Change the Enter OTP screen to use the new Activate user API

vinaysshenoy:
https://www.pivotaltracker.com/story/show/171874779


vinaysshenoy:
Assigning to @msasikanth since @rakshakhegde is on leave.

app/src/main/java/org/simple/clinic/login/LoginOtpSmsListener.kt
```diff
@@ -5,4 +5,6 @@ import io.reactivex.Completable
 interface LoginOtpSmsListener {
 
   fun listenForLoginOtp(): Completable
+
+  fun listenForLoginOtpBlocking()
```
msasikanth: ~For all the blocking calls we are using immediate as suffix. Should we change this to match that, WDYT?~

msasikanth: NVM. This has been replaced in later commits by single method `listenForLoginOtp`

PR #1228, opened 2020-04-02
CameraX QR code scanner

msasikanth:


app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeAnalyzer.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.google.zxing.common.HybridBinarizer
+import java.nio.ByteBuffer
+
+private typealias OnQrCodeDetected = (String) -> Unit
+
+class QrCodeAnalyzer(
+    private val onQrCodeDetected: OnQrCodeDetected
+) : ImageAnalysis.Analyzer {
+
+  private fun ByteBuffer.toByteArray(): ByteArray {
```
vinaysshenoy, line 24: Can we make this a top-level function instead of a member of this class?

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeAnalyzer.kt
```diff
@@ -0,0 +1,74 @@
...
+    rewind()
+    val data = ByteArray(remaining())
+    get(data)
+    return data
+  }
+
+  private val reader = MultiFormatReader().apply {
+    val map = mapOf(
+        Pair(DecodeHintType.POSSIBLE_FORMATS, arrayListOf(BarcodeFormat.QR_CODE))
```
vinaysshenoy, line 33: I personally prefer the `to` infix function when constructing `Pair` instances.

```kotlin
val map = mapOf(
  DecodeHintType.POSSIBLE_FORMATS to arrayListOf(BarcodeFormat.QR_CODE)
)
```

You don't need to change this, just adding it as a suggestion.

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeAnalyzer.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.google.zxing.DecodeHintType
+import com.google.zxing.MultiFormatReader
+import com.google.zxing.NotFoundException
+import com.google.zxing.PlanarYUVLuminanceSource
+import com.google.zxing.common.HybridBinarizer
+import java.nio.ByteBuffer
+
+private typealias OnQrCodeDetected = (String) -> Unit
+
+class QrCodeAnalyzer(
```
vinaysshenoy, line 20: Do you think we should rename this based on the method used to analyze the image?

Ex: `ZxingQrCodeAnalyzer`.

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeAnalyzer.kt
```diff
@@ -0,0 +1,74 @@
...
+        Pair(DecodeHintType.POSSIBLE_FORMATS, arrayListOf(BarcodeFormat.QR_CODE))
+    )
+    setHints(map)
+  }
+
+  override fun analyze(image: ImageProxy) {
+    val yuvFormats = mutableListOf(YUV_420_888)
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+      yuvFormats.addAll(listOf(YUV_422_888, YUV_444_888))
+    }
```
vinaysshenoy, line 42: We can extract this into a class property since it will be constant and never change.

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeAnalyzer.kt
```diff
@@ -0,0 +1,74 @@
...
+    val isYuvFormat = yuvFormats.contains(image.format)
+    // We are using YUV format because, ImageProxy internally uses ImageReader to get the image.
+    // By default ImageReader uses YUV format unless changed.
+    // https://developer.android.com/reference/androidx/camera/core/ImageProxy.html#getImage()
+    // https://developer.android.com/reference/android/media/Image.html#getFormat()
+    if (!isYuvFormat) {
```
vinaysshenoy, line 48: We can simplify this by using `!in`.

```kotlin
if(image.format !in yuvFormats) {
  return
}
```

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -19,7 +22,10 @@ constructor(
     defStyleAttr: Int = 0
 ) : FrameLayout(context, attrs, defStyleAttr) {
 
+  val scans = PublishSubject.create<String>()
```
vinaysshenoy, line 36: Can we make this a private variable and expose it to callers as an `Observable` separately?

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -47,9 +53,18 @@ constructor(
 
     val preview = Preview.Builder().build()
 
+    val analyzer = ImageAnalysis.Builder()
+        .setBackpressureStrategy(ImageAnalysis.STRATEGY_KEEP_ONLY_LATEST)
+        .build()
+        .also {
```
vinaysshenoy, line 73: This can be an `.apply` block, right?

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -47,9 +53,18 @@ constructor(
...
+          it.setAnalyzer(cameraExecutor, QrCodeAnalyzer { scanResult ->
```
vinaysshenoy, line 74: We can convert this into a method reference.

```kotlin
it.setAnalyzer(cameraExecutor, QrCodeAnalyzer(scans::onNext))
```

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdConfig.kt
```diff
@@ -0,0 +1,14 @@
...
+
+data class ScanSimpleIdConfig(val useNewQrScanner: Boolean) {
+
+  companion object {
+
+    fun read(configReader: ConfigReader): ScanSimpleIdConfig {
+      val useNewQrScanner = configReader.boolean("use_new_qr_scanner", true)
+      return ScanSimpleIdConfig(useNewQrScanner = useNewQrScanner)
+    }
+  }
```
vinaysshenoy, line 13: I've been thinking of getting rid of companion functions for constructing config instances. We were using them earlier because we were injecting `Observable` instances, but now that we inject them directly, I think we can replace this with a secondary constructor.

```kotlin
@Inject constructor(configReader: ConfigReader): this(
      useNewQrScanner = configReader.boolean("use_new_qr_scanner", true)
)
```

With this, we can tell Dagger to use this constructor for injection and avoid having to construct a new module completely.

What do you think?

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
```diff
@@ -57,6 +61,9 @@ class ScanSimpleIdScreen(context: Context, attrs: AttributeSet) : ConstraintLayo
     toolBar.setNavigationOnClickListener { screenRouter.pop() }
     setupShortCodeTextField()
 
+    qrCodeScannerView.visibleOrGone(config.useNewQrScanner)
+    qrCodeScannerViewOld.visibleOrGone(!config.useNewQrScanner)
```
vinaysshenoy, line 65: Since the view is present in the layout, won't this activate the camera in both the views?

I was thinking we should manually inflate one or the other view based on the flag.

vinaysshenoy:
@msasikanth LeakCanary also detected a leak where the camera manager is retaining a reference to the activity in some form. Please take a look at this as well?

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
msasikanth: I don't think it will activate the camera views If the view is set to gone it won't inflate the views right?

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -13,12 +13,21 @@ constructor(
...
+
   init {
     View.inflate(context, R.layout.view_qrcode_scanner, this)
   }
 
   override fun onFinishInflate() {
     super.onFinishInflate()
     if (isInEditMode) return
+
+    qrCodeScannerLifecycle.bindCamera()
```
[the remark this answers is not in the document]
msasikanth: View references are not available in `onAttachedToWindow` right?

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -15,15 +21,36 @@ constructor(
...
   override fun onFinishInflate() {
     super.onFinishInflate()
     if (isInEditMode) return
 
     qrCodeScannerLifecycle.bindCamera()
+
+    cameraProviderFuture?.addListener(Runnable {
+      val cameraProvider = cameraProviderFuture?.get()
+      startCamera(cameraProvider!!)
+    }, ContextCompat.getMainExecutor(context))
```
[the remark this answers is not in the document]
msasikanth: View references are not available in `onAttachedToWindow` right?

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
msasikanth: We can use `.apply` as well, but since it's not a direct object configuration I personally prefer `.also` . Because we are taking the build `ImageAnalysis` as argument and using it to set the analyzer. Should I change this to `.apply`?

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
vinaysshenoy: The views will always be inflated regardless of the visibility status.

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
vinaysshenoy: It's up to you. If you want to leave it as is, that's fine as well.

build.gradle
```diff
@@ -73,6 +73,7 @@ buildscript {
       camerax             : '1.0.0-beta02',
       cameraView          : '1.0.0-alpha09',
       cameraLifecycle     : '1.0.0-alpha09',
+      zxing               : '3.4.0',
```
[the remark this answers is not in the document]
msasikanth: Yeah, it doesn't seem to be working. so switched to zxing android-core v3.3.0 instead.

PR #1238, opened 2020-04-07
Simplify alerting users for facility change 

sanchita:
https://www.pivotaltracker.com/story/show/169808600

### DIFF
--- a/app/src/main/java/org/simple/clinic/facility/alertchange/AlertFacilityChangeSheet.kt
+++ b/app/src/main/java/org/simple/clinic/facility/alertchange/AlertFacilityChangeSheet.kt
@@ -66,17 +66,17 @@ class AlertFacilityChangeSheet : BottomSheetActivity() {
 
     if (isFacilitySwitchedPreference.get().not()) {
       closeSheetWithContinuation()
-    }
+    } else {
+      setContentView(R.layout.sheet_alert_facility_change)
 
-    setContentView(R.layout.sheet_alert_facility_change)
+      facilityName.text = getString(R.string.alertfacilitychange_facility_name, currentFacilityName)
+      yesButton.setOnClickListener {
+        closeSheetWithResult(Activity.RESULT_OK)
+      }
 
-    facilityName.text = getString(R.string.alertfacilitychange_facility_name, currentFacilityName)
-    yesButton.setOnClickListener {
-      closeSheetWithResult(Activity.RESULT_OK)
-    }
-
-    changeButton.setOnClickListener {
-      openFacilityChangeScreen()
+      changeButton.setOnClickListener {
+        openFacilityChangeScreen()
+      }
     }
   }
 
--- a/app/src/main/java/org/simple/clinic/facility/alertchange/AlertFacilityChangeSheet.kt
+++ b/app/src/main/java/org/simple/clinic/facility/alertchange/AlertFacilityChangeSheet.kt
@@ -4,13 +4,17 @@ import android.app.Activity
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
+import android.os.Parcelable
+import android.view.ContextThemeWrapper
 import com.f2prateek.rx.preferences2.Preference
 import io.github.inflationx.viewpump.ViewPumpContextWrapper
 import kotlinx.android.synthetic.main.sheet_alert_facility_change.*
 import org.simple.clinic.ClinicApp
 import org.simple.clinic.R
 import org.simple.clinic.facility.change.FacilityChangeActivity
+import org.simple.clinic.router.screen.FullScreenKey
 import org.simple.clinic.util.LocaleOverrideContextWrapper
+import org.simple.clinic.util.unsafeLazy
 import org.simple.clinic.util.wrap
 import org.simple.clinic.widgets.BottomSheetActivity
 import java.util.Locale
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/facility/alertchange/AlertFacilityChangeSheet.kt
```diff
@@ -30,21 +33,47 @@ class AlertFacilityChangeSheet : BottomSheetActivity() {
   companion object {
     const val FACILITY_CHANGE = 101
     private const val CURRENT_FACILITY_NAME = "current_facility"
+    private const val CONTINUE_TO_SCREEN = "continue_to_screen"
+
+    const val EXTRA_CONTINUE_SCREEN = "extra_continue_screen"
```
vinaysshenoy, line 38: This need not be public, right?

app/src/main/java/org/simple/clinic/facility/alertchange/AlertFacilityChangeSheet.kt
```diff
@@ -30,21 +33,47 @@ class AlertFacilityChangeSheet : BottomSheetActivity() {
...
+        currentFacilityName: String,
+        continueToScreen: FullScreenKey
+    ): Intent {
+      val intent = Intent(context, AlertFacilityChangeSheet::class.java)
+      intent.putExtra(CURRENT_FACILITY_NAME, currentFacilityName)
+      intent.putExtra(CONTINUE_TO_SCREEN, continueToScreen)
+      return intent
+    }
+
+    fun <T : Parcelable> readExtra(intent: Intent): T {
```
vinaysshenoy, line 51: Should we rename this to something like `readContinuationExtra` since just `Extra` is also applicable to any of the extras in the `Intent`?

app/src/main/java/org/simple/clinic/facility/alertchange/AlertFacilityChangeSheet.kt
```diff
@@ -30,21 +33,47 @@ class AlertFacilityChangeSheet : BottomSheetActivity() {
...
+    }
   }
 
   private val currentFacilityName by lazy {
     intent.getStringExtra(CURRENT_FACILITY_NAME)!!
   }
 
+  private val continueToScreen by lazy {
+    intent.getParcelableExtra(CONTINUE_TO_SCREEN) as FullScreenKey
+  }
```
vinaysshenoy, line 62: These could be `unsafeLazy` since they will only be invoked from the UI thread, so we can avoid the synchronisation penalty.

app/src/main/java/org/simple/clinic/facility/alertchange/AlertFacilityChangeSheet.kt
```diff
@@ -91,6 +120,13 @@ class AlertFacilityChangeSheet : BottomSheetActivity() {
     finish()
   }
 
+  private fun closeSheetWithResult() {
```
vinaysshenoy, line 120: We now have a `closeSheetWithResult` and a `closeSheet(result)` and it is not clear from the name what is the difference between the two? What do you think about renaming `closeSheetWithResult` to `closeSheetWithContinuation` or something like that?

app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsController.kt
```diff
@@ -74,14 +74,8 @@ class PatientSearchResultsController @Inject constructor(
   }
 
   private fun saveEntryAndGoToRegisterPatientScreen(ongoingNewPatientEntry: OngoingNewPatientEntry, currentFacility: Facility): Observable<UiChange> {
-    val nextScreenChange = if (isFacilitySwitchedPreference.get()) {
```
vinaysshenoy, line 77: The `isFacilitySwitchedPreference` dependency is now unused.

app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsController.kt
```diff
@@ -74,14 +74,8 @@ class PatientSearchResultsController @Inject constructor(
...
-      { ui: Ui -> ui.openAlertFacilityChangeSheet(currentFacility) }
-    } else {
-      { ui: Ui -> ui.openPatientEntryScreen() }
-    }
-
     return patientRepository
         .saveOngoingEntry(ongoingNewPatientEntry)
-        .andThen(Observable.just(nextScreenChange))
+        .andThen(Observable.just { ui: Ui -> ui.openAlertFacilityChangeSheet(currentFacility) })
```
vinaysshenoy, line 79: I think we can continue calling `openPatientEntryScreen()` here and leave it up to a UI implementation to decide whether to invoke the facility change sheet or not. What do you think?

app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsScreen.kt
```diff
@@ -76,16 +77,15 @@ class PatientSearchResultsScreen(context: Context, attrs: AttributeSet) : Relati
     setupAlertResults(screenDestroys)
   }
 
-  @SuppressLint("CheckResult")
```
vinaysshenoy, line 79: Can we add this back? It now shows a warning that we are not handling the result of the subscription.

app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsScreen.kt
```diff
@@ -76,16 +77,15 @@ class PatientSearchResultsScreen(context: Context, attrs: AttributeSet) : Relati
...
   private fun setupAlertResults(screenDestroys: Observable<ScreenDestroyed>) {
     screenRouter.streamScreenResults()
         .ofType<ActivityResult>()
         .filter { it.requestCode == ALERT_FACILITY_CHANGE && it.succeeded() }
+        .map { AlertFacilityChangeSheet.readExtra<FullScreenKey>(it.data!!) }
         .takeUntil(screenDestroys)
-        .subscribe { openPatientEntryScreen() }
+        .subscribe { screenRouter.push(it) }
```
vinaysshenoy, line 86: This could be a method reference as well.

```kotlin
.subscribe(screenRouter::push)
```

app/src/main/java/org/simple/clinic/facility/alertchange/AlertFacilityChangeSheet.kt
sanchita: This will be refactored when we start using the intent with continuation.
I have made this change already in the next PR.

app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsController.kt
sanchita: Shouldn't this be part of the business logic and not the UI?
Also, it can be misleading.

app/src/main/java/org/simple/clinic/facility/alertchange/AlertFacilityChangeSheet.kt
sanchita: No. It doesn't. I will make this change.

app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsController.kt
vinaysshenoy: The way I am seeing it is that the business logic is to open the patient entry screen and that confirming the facility switch is something that is left up the UI to decide as an interception.

But if you would like to leave it like this, I am fine with this too.

app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsController.kt
sanchita: Good catch!

app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsController.kt
sanchita: Yes. I understand what you mean. I will make this change.

PR #1242, opened 2020-04-09
Implement initial logic for the call patient view

vinaysshenoy:

This PR does not include the UI. This sets up the Mobius classes and adds some of the business logic and presentation logic.

### DIFF
--- a/app/src/main/java/org/simple/clinic/patientcontact/PatientContactEvent.kt
+++ b/app/src/main/java/org/simple/clinic/patientcontact/PatientContactEvent.kt
@@ -1,0 +1,3 @@
+package org.simple.clinic.patientcontact
+
+sealed class PatientContactEvent
--- a/app/src/main/java/org/simple/clinic/patientcontact/PatientContactEvent.kt
+++ b/app/src/main/java/org/simple/clinic/patientcontact/PatientContactEvent.kt
@@ -1,3 +1,9 @@
 package org.simple.clinic.patientcontact
 
+import org.simple.clinic.patient.PatientProfile
+
 sealed class PatientContactEvent
+
+data class PatientProfileLoaded(
+    val patientProfile: PatientProfile
+): PatientContactEvent()
--- a/app/src/main/java/org/simple/clinic/patientcontact/PatientContactEvent.kt
+++ b/app/src/main/java/org/simple/clinic/patientcontact/PatientContactEvent.kt
@@ -1,9 +1,15 @@
 package org.simple.clinic.patientcontact
 
+import org.simple.clinic.home.overdue.OverdueAppointment
 import org.simple.clinic.patient.PatientProfile
+import org.simple.clinic.util.Optional
 
 sealed class PatientContactEvent
 
 data class PatientProfileLoaded(
     val patientProfile: PatientProfile
-): PatientContactEvent()
+) : PatientContactEvent()
+
+data class OverdueAppointmentLoaded(
+    val overdueAppointment: Optional<OverdueAppointment>
+) : PatientContactEvent()
--- a/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUi.kt
+++ b/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUi.kt
@@ -1,4 +1,7 @@
 package org.simple.clinic.patientcontact
 
+import org.simple.clinic.patient.Gender
+
 interface PatientContactUi {
+  fun renderPatientDetails(name: String, gender: Gender, age: Int, phoneNumber: String)
 }
--- a/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUi.kt
+++ b/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUi.kt
@@ -4,4 +4,6 @@ import org.simple.clinic.patient.Gender
 
 interface PatientContactUi {
   fun renderPatientDetails(name: String, gender: Gender, age: Int, phoneNumber: String)
+  fun showCallResultSection()
+  fun hideCallResultSection()
 }
--- a/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUi.kt
+++ b/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUi.kt
@@ -1,0 +1,4 @@
+package org.simple.clinic.patientcontact
+
+interface PatientContactUi {
+}
--- a/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUi.kt
+++ b/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUi.kt
@@ -4,6 +4,10 @@ import org.simple.clinic.patient.Gender
 
 interface PatientContactUi {
   fun renderPatientDetails(name: String, gender: Gender, age: Int, phoneNumber: String)
+
   fun showCallResultSection()
   fun hideCallResultSection()
+
+  fun showSecureCallUi()
+  fun hideSecureCallUi()
 }
--- a/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUpdate.kt
@@ -1,0 +1,15 @@
+package org.simple.clinic.patientcontact
+
+import com.spotify.mobius.Next
+import com.spotify.mobius.Next.noChange
+import com.spotify.mobius.Update
+
+class PatientContactUpdate : Update<PatientContactModel, PatientContactEvent, PatientContactEffect> {
+
+  override fun update(
+      model: PatientContactModel,
+      event: PatientContactEvent
+  ): Next<PatientContactModel, PatientContactEffect> {
+    return noChange()
+  }
+}
--- a/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUpdate.kt
@@ -3,6 +3,7 @@ package org.simple.clinic.patientcontact
 import com.spotify.mobius.Next
 import com.spotify.mobius.Next.noChange
 import com.spotify.mobius.Update
+import org.simple.clinic.mobius.next
 
 class PatientContactUpdate : Update<PatientContactModel, PatientContactEvent, PatientContactEffect> {
 
@@ -10,6 +11,8 @@ class PatientContactUpdate : Update<PatientContactModel, PatientContactEvent, Pa
       model: PatientContactModel,
       event: PatientContactEvent
   ): Next<PatientContactModel, PatientContactEffect> {
-    return noChange()
+    return when(event) {
+      is PatientProfileLoaded -> next(model.patientProfileLoaded(event.patientProfile))
+    }
   }
 }
--- a/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/patientcontact/PatientContactUpdate.kt
@@ -1,7 +1,6 @@
 package org.simple.clinic.patientcontact
 
 import com.spotify.mobius.Next
-import com.spotify.mobius.Next.noChange
 import com.spotify.mobius.Update
 import org.simple.clinic.mobius.next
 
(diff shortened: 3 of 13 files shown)


app/src/test/java/org/simple/clinic/patientcontact/PatientContactUpdateTest.kt
```diff
@@ -31,24 +31,36 @@ class PatientContactUpdateTest {
...
     val appointment = Just(overdueAppointment)
+    val model = defaultModel()
 
     spec
-        .given(defaultModel)
+        .given(model)
         .whenEvent(OverdueAppointmentLoaded(appointment))
         .then(assertThatNext(
-            hasModel(defaultModel.overdueAppointmentLoaded(appointment))
+            hasModel(model.overdueAppointmentLoaded(appointment))
```
sanchita, line 54: We can keep the name as `defaultModel`.

app/src/test/java/org/simple/clinic/patientcontact/PatientContactUiRendererTest.kt
```diff
@@ -107,10 +114,41 @@ class PatientContactUiRendererTest {
...
+
+  @Test
+  fun `if the secure call feature is disabled, hide the secure call button and message`() {
+    // when
+    val model = defaultModel(phoneMaskFeatureEnabled = false)
+    uiRenderer.render(model)
+
+    // then
+    verify(ui).hideSecureCallButton()
+    verify(ui).hideSecureCallMessage()
```
sanchita, line 143: Replace with `verifyStandardInteractions()`

app/src/test/java/org/simple/clinic/patientcontact/PatientContactUiRendererTest.kt
```diff
@@ -107,10 +114,41 @@ class PatientContactUiRendererTest {
...
+    verifyNoMoreInteractions(ui)
+  }
+
+  private fun verifyStandardInteractions() {
```
sanchita, line 147: Rename to `hideSecureCallUi()`?

app/src/main/java/org/simple/clinic/patientcontact/PatientContactUi.kt
```diff
@@ -4,6 +4,13 @@ import org.simple.clinic.patient.Gender
...
   fun renderPatientDetails(name: String, gender: Gender, age: Int, phoneNumber: String)
+
   fun showCallResultSection()
   fun hideCallResultSection()
+
+  fun showSecureCallButton()
+  fun hideSecureCallButton()
+
+  fun showSecureCallMessage()
+  fun hideSecureCallMessage()
```
sanchita, line 15: A doubt. Is the toggling logic for secure call button and message different?

app/src/test/java/org/simple/clinic/patientcontact/PatientContactUiRendererTest.kt
sanchita: I am not sure if we should even extract these to a separate method. 
Repetition in tests is never a problem and it is better to avoid a common pitfall while writing tests which is dependency of tests on each other.

app/src/main/java/org/simple/clinic/patientcontact/PatientContactEvent.kt
```diff
@@ -1,9 +1,15 @@
...
 
 sealed class PatientContactEvent
 
 data class PatientProfileLoaded(
     val patientProfile: PatientProfile
-): PatientContactEvent()
+) : PatientContactEvent()
+
+data class OverdueAppointmentLoaded(
+    val overdueAppointment: Optional<OverdueAppointment>
```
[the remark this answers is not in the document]
vinaysshenoy: Yes, this sheet can be invoked from the `PatientSummaryScreen` by clicking on the patient phone number. In this scenario, a patient could either have an overdue appointment or not.

If the patient does not have an overdue appointment, we will show the call buttons, but hide the call result options. Please see the following designs:

- [Without overdue appointment](https://app.zeplin.io/project/5b9a49113dbe7958c1d3fde6/screen/5e71cc19a72b0302e4e7c808/)
- [With overdue appointment](https://app.zeplin.io/project/5b9a49113dbe7958c1d3fde6/screen/5e71cc1882081dbcc848d34d/)

app/src/main/java/org/simple/clinic/patientcontact/PatientContactUpdate.kt
```diff
@@ -13,6 +13,8 @@ class PatientContactUpdate : Update<PatientContactModel, PatientContactEvent, Pa
   ): Next<PatientContactModel, PatientContactEffect> {
     return when(event) {
       is PatientProfileLoaded -> next(model.patientProfileLoaded(event.patientProfile))
+      is OverdueAppointmentLoaded -> next(model.overdueAppointmentLoaded(event.overdueAppointment))
+      else -> noChange()
```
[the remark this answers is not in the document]
vinaysshenoy: Not necessary, I added this while I was implementing the events. Will remove.

app/src/test/java/org/simple/clinic/patientcontact/PatientContactUiRendererTest.kt
```diff
@@ -0,0 +1,85 @@
...
+        age = null,
+        dateOfBirth = dateOfBirth,
+        gender = gender
+    )
+
+    // when
+    uiRenderer.render(defaultModel.patientProfileLoaded(patientProfile))
+
+    // then
+    val expectedAge = 48 // difference between clock date and DOB
```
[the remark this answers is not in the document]
vinaysshenoy: Yes. Plus calculating the diff here would mean that there could be scenarios where the test continues to since we have duplicated the logic in the test, but the production code has changed and is incorrect.

app/src/test/java/org/simple/clinic/patientcontact/PatientContactUiRendererTest.kt
vinaysshenoy: In this specific case, I do not want to extract these to a separate method because these are the interactions and assertions that are specific to this test.

app/src/test/java/org/simple/clinic/patientcontact/PatientContactUiRendererTest.kt
```diff
@@ -0,0 +1,85 @@
...
+  private val clock = TestUserClock(LocalDate.parse("2018-01-01"))
+  private val uiRenderer = PatientContactUiRenderer(ui, clock)
+
+  @Test
+  fun `when the model is not initialized, do nothing`() {
+    // when
+    uiRenderer.render(defaultModel)
+
+    // then
+    verifyZeroInteractions(ui)
```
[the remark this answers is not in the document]
vinaysshenoy: As in? Rename the test method?

app/src/test/java/org/simple/clinic/patientcontact/PatientContactUiRendererTest.kt
vinaysshenoy: My intention behind doing this was to separate the assertions that are not specific to the scenario being testing in the specific test outside so that it is clear when reading the test what assertions are part of this test vs what are not.

This is also why I called it `verifyStandardInteractions` since I do not want to expose what exactly the standard interactions are.

Does this make sense?

app/src/main/java/org/simple/clinic/patientcontact/PatientContactUi.kt
vinaysshenoy: It is not, I can unify them.

vinaysshenoy:
> Overall PR looks good. Just a few overall feedback points:
> 
> * The PRD and the state transition diagram for this is missing
> * Can we rename the whole feature from `PatientContact` -> `PatientFollowup` or `ContactPatient` or `CallPatient`?

> Overall PR looks good. Just a few overall feedback points:
> 
> * The PRD and the state transition diagram for this is missing
> * Can we rename the whole feature from `PatientContact` -> `PatientFollowup` or `ContactPatient` or `CallPatient`?

- I'll rename to `ContactPatient` in a later PR since I have built on top of this a lot and I don't want to struggle with rebase 😅.
- There is currently no PRD for this feature since it is mostly the same behaviour as the phone feature in the `OverdueScreen`. A small description of the feature can be found at https://www.pivotaltracker.com/story/show/170595045.
- I didn't create the state diagram yet because I am investigating some ways to create a state diagram without needing to draw things manually so that we can generate images and check them into version control. I will investigating these on the weekend, so I held off on making the state diagram till then.

sanchita:
> > Overall PR looks good. Just a few overall feedback points:
> > 
> > * The PRD and the state transition diagram for this is missing
> > * Can we rename the whole feature from `PatientContact` -> `PatientFollowup` or `ContactPatient` or `CallPatient`?
> 
>     * I'll rename to `ContactPatient` in a later PR since I have built on top of this a lot and I don't want to struggle with rebase 😅.
> 
That works too.
>     * There is currently no PRD for this feature since it is mostly the same behaviour as the phone feature in the `OverdueScreen`. A small description of the feature can be found at https://www.pivotaltracker.com/story/show/170595045.
> 
So the reason I was looking for a PRD was to understand what last overdue appointment means?
>     * I didn't create the state diagram yet because I am investigating some ways to create a state diagram without needing to draw things manually so that we can generate images and check them into version control. I will investigating these on the weekend, so I held off on making the state diagram till then.
We can do that too but we won't come back to drawing the state diagrams for existing things anytime soon. It is alright for now.

vinaysshenoy:
> So the reason I was looking for a PRD was to understand what last overdue appointment means?

It's the same thing as the overdue appointments in the `OverdueScreen` except that it is not tied to a specific facility. I will confirm this once more.

vinaysshenoy:
> > So the reason I was looking for a PRD was to understand what last overdue appointment means?
> 
> It's the same thing as the overdue appointments in the `OverdueScreen` except that it is not tied to a specific facility. I will confirm this once more.

Confirmed with Pragati.

app/src/test/java/org/simple/clinic/patientcontact/PatientContactUiRendererTest.kt
sanchita: The way I see it, this has two issues:
- There is nothing like a standard interaction here and we shouldn't assume this as a standard interaction for the screen. This is not like an error or unexpected state.
- By separating out this verification, we are assuming that this is not something we care about in this particular test which isn't true. Testing whether this view is hidden or not is equally important part of Ui rendering for this state. If it isn't then we shouldn't be testing this at all.

app/src/test/java/org/simple/clinic/patientcontact/PatientContactUiRendererTest.kt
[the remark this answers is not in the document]
sanchita: I just wanted to understand that what are we testing here? What is the expected behavior when the default model is loaded? 
If I am saying do nothing and still testing something being done, it is confusing.
Maybe renaming might help.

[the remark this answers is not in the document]
vinaysshenoy: Ah, this was originally meant to be a regression test to capture accidental calls for a default model. Since the default model actually has actions that are verified in another test, we can remove this.

I will remove this test.

app/src/test/java/org/simple/clinic/patientcontact/PatientContactUiRendererTest.kt
vinaysshenoy: Hmm, I see this point as well. I'll inline this method into the rest of the tests for now.

vinaysshenoy:
> > Overall PR looks good. Just a few overall feedback points:
> > 
> > * The PRD and the state transition diagram for this is missing
> > * Can we rename the whole feature from `PatientContact` -> `PatientFollowup` or `ContactPatient` or `CallPatient`?
> 
>     * I'll rename to `ContactPatient` in a later PR since I have built on top of this a lot and I don't want to struggle with rebase 😅.
> 
>     * There is currently no PRD for this feature since it is mostly the same behaviour as the phone feature in the `OverdueScreen`. A small description of the feature can be found at https://www.pivotaltracker.com/story/show/170595045.
> 
>     * I didn't create the state diagram yet because I am investigating some ways to create a state diagram without needing to draw things manually so that we can generate images and check them into version control. I will investigating these on the weekend, so I held off on making the state diagram till then.

Renaming the feature was done in https://github.com/simpledotorg/simple-android/pull/1270.

PR #1246, opened 2020-04-10
Add zxing core dependency

msasikanth:
We were using a transitive zxing dependency from the QR scanner library. Once we remove the old QR scanner the new one will be broken, adding the zxing dependency to the app gradle.

The zxing android-core dependency we added didn't internally contain the zxing core dependency.


vinaysshenoy:
Will review this since @sanchita is on leave.

PR #1251, opened 2020-04-14
Add custom `DatePickerDialog` subclass that works with threetenbp

vinaysshenoy:

### DIFF
--- a/app/src/main/java/org/simple/clinic/widgets/ThreeTenBpDatePickerDialog.kt
+++ b/app/src/main/java/org/simple/clinic/widgets/ThreeTenBpDatePickerDialog.kt
@@ -1,0 +1,43 @@
+package org.simple.clinic.widgets
+
+import android.app.DatePickerDialog
+import android.app.DatePickerDialog.OnDateSetListener
+import android.content.Context
+import org.simple.clinic.util.UserClock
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/widgets/ThreeTenBpDatePickerDialog.kt
```diff
@@ -0,0 +1,43 @@
...
+ * So when we convert from LocalDate to the DatePicker (and vice versa), we have to adjust the
+ * month accordingly.
+ **/
+class ThreeTenBpDatePickerDialog(
+    context: Context,
+    preselectedDate: LocalDate,
+    allowedDateRange: ClosedRange<LocalDate>,
+    clock: UserClock,
+    datePickedListener: (LocalDate) -> Unit
+) : DatePickerDialog(
```
msasikanth, line 26: Now that we have added MaterialComponents, should we try and use the material date picker, WDYT?

https://material.io/components/pickers/#

vinaysshenoy: I tried doing this, but the change is too invasive to use at the moment. I'll take a look at this later once I complete the feature.

PR #1260, opened 2020-04-16
Fix flaky test in `BloodPressureRepositoryAndroidTest`

vinaysshenoy:
`BloodPressureRepositoryAndroidTest#getting_the_blood_pressure_count_immediate_for_a_patient_should_work_correctly`
was detected as 10% flaky over the last few days. I investigated this
and found out the cause:

- `BloodPressureHistoryListItemDataSourceTest` was added which creates
some data in the `BloodPressureMeasurement` table, but does not clean
up after itself.
- The flaky test in question asserts the total count of BPs present
in the database before running its assertions.
- When the test was run by itself, it would work fine.
- When the test was run *after* `BloodPressureHistoryListItemDataSourceTest`,
it would first fail because the BP count was not what it expected, then
the test rule would clear the databases, then the flaky test rule would
execute it again and it would pass.

This commit fixes the issue by changing `BloodPressureHistoryListItemDataSourceTest`
to clean up after itself.

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/bp/BloodPressureHistoryListItemDataSourceTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/bp/BloodPressureHistoryListItemDataSourceTest.kt
@@ -5,14 +5,18 @@ import androidx.paging.PositionalDataSource.LoadInitialCallback
 import androidx.paging.PositionalDataSource.LoadInitialParams
 import androidx.paging.PositionalDataSource.LoadRangeParams
 import com.google.common.truth.Truth.assertThat
+import org.junit.After
 import org.junit.Before
+import org.junit.Rule
 import org.junit.Test
+import org.junit.rules.RuleChain
 import org.simple.clinic.AppDatabase
 import org.simple.clinic.TestClinicApp
 import org.simple.clinic.TestData
 import org.simple.clinic.bp.history.adapter.BloodPressureHistoryListItem
 import org.simple.clinic.bp.history.adapter.BloodPressureHistoryListItem.BloodPressureHistoryItem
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/org/simple/clinic/bp/BloodPressureHistoryListItemDataSourceTest.kt
```diff
@@ -41,6 +45,9 @@ class BloodPressureHistoryListItemDataSourceTest {
   @field:[Inject Named("time_for_measurement_history")]
   lateinit var timeFormatter: DateTimeFormatter
 
+  @get:Rule
+  val ruleChain: RuleChain = Rules.global()
```
rakshakhegde, line 49: As the database is getting cleared now, does the Quarantine rule (`Rules.global()`) still need to be used?

rakshakhegde:
@vinaysshenoy Left a comment PTAL and also please update branch.

app/src/androidTest/java/org/simple/clinic/bp/BloodPressureHistoryListItemDataSourceTest.kt
vinaysshenoy: That needs to be added for all instrumented tests.

vinaysshenoy: The point of the rule is to continuously observe all tests for flakiness, which could get introduced in the future.

rakshakhegde: Oh gotcha

PR #1265, opened 2020-04-20
Implement the Call patient UI

vinaysshenoy:

This PR still doesn't trigger the UI, this will be in the next PR.

### DIFF
--- a/app/src/main/java/org/simple/clinic/patientcontact/CallPatientView.kt
+++ b/app/src/main/java/org/simple/clinic/patientcontact/CallPatientView.kt
@@ -4,16 +4,37 @@ import android.content.Context
 import android.util.AttributeSet
 import android.view.View
 import androidx.constraintlayout.widget.ConstraintLayout
+import kotlinx.android.synthetic.main.patientcontact_view_callpatient.view.*
 import org.simple.clinic.R
+import org.simple.clinic.patient.Gender
+import org.simple.clinic.patient.displayLetterRes
 
 class CallPatientView(
     context: Context,
     attributeSet: AttributeSet
 ) : ConstraintLayout(context, attributeSet) {
 
+  var callResultSectionVisible: Boolean = false
+    set(value) {
+      field = value
+      callResultsGroup.visibility = if (field) View.VISIBLE else View.GONE
+    }
+
+  var secureCallingSectionVisible: Boolean = false
+    set(value) {
+      field = value
+      secureCallingGroup.visibility = if (field) View.VISIBLE else View.GONE
+    }
+
   override fun onFinishInflate() {
     super.onFinishInflate()
 
     View.inflate(context, R.layout.patientcontact_view_callpatient, this)
+  }
+
+  fun renderPatientDetails(name: String, gender: Gender, age: Int, phoneNumber: String) {
+    val genderLetter = resources.getString(gender.displayLetterRes)
+    nameTextView.text = resources.getString(R.string.patientcontact_patientdetails, name, genderLetter, age.toString())
+    phoneNumberTextView.text = phoneNumber
   }
 }
--- a/app/src/main/java/org/simple/clinic/patientcontact/CallPatientView.kt
+++ b/app/src/main/java/org/simple/clinic/patientcontact/CallPatientView.kt
@@ -3,11 +3,18 @@ package org.simple.clinic.patientcontact
 import android.content.Context
 import android.util.AttributeSet
 import android.view.View
+import android.widget.Toast
 import androidx.constraintlayout.widget.ConstraintLayout
 import kotlinx.android.synthetic.main.patientcontact_view_callpatient.view.*
 import org.simple.clinic.R
 import org.simple.clinic.patient.Gender
 import org.simple.clinic.patient.displayLetterRes
+
+private typealias AgreedToVisitClicked = () -> Unit
+private typealias RemindToCallLaterClicked = () -> Unit
+private typealias RemoveFromOverdueListClicked = () -> Unit
+private typealias NormalCallButtonClicked = () -> Unit
+private typealias SecureCallButtonClicked = () -> Unit
 
 class CallPatientView(
     context: Context,
@@ -26,10 +33,26 @@ class CallPatientView(
       secureCallingGroup.visibility = if (field) View.VISIBLE else View.GONE
     }
 
+  var agreedToVisitClicked: AgreedToVisitClicked? = { Toast.makeText(context, "WIP", Toast.LENGTH_SHORT).show() }
+
+  var remindToCallLaterClicked: RemindToCallLaterClicked? = { Toast.makeText(context, "WIP", Toast.LENGTH_SHORT).show() }
+
+  var removeFromOverdueListClicked: RemoveFromOverdueListClicked? = { Toast.makeText(context, "WIP", Toast.LENGTH_SHORT).show() }
+
+  var normalCallButtonClicked: NormalCallButtonClicked? = null
+
+  var secureCallButtonClicked: SecureCallButtonClicked? = null
+
   override fun onFinishInflate() {
     super.onFinishInflate()
 
     View.inflate(context, R.layout.patientcontact_view_callpatient, this)
+
+    agreedToVisitTextView.setOnClickListener { agreedToVisitClicked?.invoke() }
+    remindToCallLaterTextView.setOnClickListener { remindToCallLaterClicked?.invoke() }
+    removeFromOverdueListTextView.setOnClickListener { removeFromOverdueListClicked?.invoke() }
+    normalCallButton.setOnClickListener { normalCallButtonClicked?.invoke() }
+    secureCallButton.setOnClickListener { secureCallButtonClicked?.invoke() }
   }
 
   fun renderPatientDetails(name: String, gender: Gender, age: Int, phoneNumber: String) {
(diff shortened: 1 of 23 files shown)


app/src/main/java/org/simple/clinic/patientcontact/CallPatientView.kt
```diff
@@ -3,12 +3,19 @@ package org.simple.clinic.patientcontact
...
 import kotlinx.android.synthetic.main.patientcontact_view_callpatient.view.*
 import org.simple.clinic.R
 import org.simple.clinic.patient.Gender
 import org.simple.clinic.patient.displayLetterRes
 
+private typealias AgreedToVisitClicked = () -> Unit
+private typealias RemindToCallLaterClicked = () -> Unit
+private typealias RemoveFromOverdueListClicked = () -> Unit
+private typealias NormalCallButtonClicked = () -> Unit
+private typealias SecureCallButtonClicked = () -> Unit
```
sanchita, line 17: I do not understand the use of these lambas. Can you please explain?

app/src/main/java/org/simple/clinic/patientcontact/CallPatientView.kt
```diff
@@ -26,10 +33,26 @@ class CallPatientView(
...
 
+  var agreedToVisitClicked: AgreedToVisitClicked? = { Toast.makeText(context, "WIP", Toast.LENGTH_SHORT).show() }
+
+  var remindToCallLaterClicked: RemindToCallLaterClicked? = { Toast.makeText(context, "WIP", Toast.LENGTH_SHORT).show() }
+
+  var removeFromOverdueListClicked: RemoveFromOverdueListClicked? = { Toast.makeText(context, "WIP", Toast.LENGTH_SHORT).show() }
+
+  var normalCallButtonClicked: NormalCallButtonClicked? = null
+
+  var secureCallButtonClicked: SecureCallButtonClicked? = null
```
sanchita, line 44: Please make these `private`.

app/src/main/java/org/simple/clinic/patientcontact/PatientContactEvent.kt
```diff
@@ -13,3 +17,15 @@ data class PatientProfileLoaded(
 data class OverdueAppointmentLoaded(
     val overdueAppointment: Optional<OverdueAppointment>
 ) : PatientContactEvent()
+
+data class NormalCallClicked(
+    override var permission: Optional<RuntimePermissionResult> = None,
+    override val permissionRequestCode: Int = 1,
+    override val permissionString: String = Manifest.permission.CALL_PHONE
```
sanchita, line 25: Few questions here:
1. Are we going to be passing these values from somewhere else apart from the `permission` itself?
2. Why in `NormalCallClicked` do we need to pass a different `permissionString` than `CALL_PHONE`?
3. Why did we choose to set default values here rather than passing them from the relevant `Update` class?

app/src/main/java/org/simple/clinic/patientcontact/CallPatientView.kt
vinaysshenoy: This will become more clear in the future PRs, but I'll try to explain here as well as I can.

In order to make a future migration away from Flow easier, I decided to implement the entire set of features (Call patient, Set appointment reminder, and Remove from overdue list) as sub views within the parent `ContactPatientBottomSheet` and switch them based on UI events. 

However, there will still only be one Mobius loop which controls the overall feature, (`ContactPatientUpdate`).

So I broke these out into separate views which the `ContactPatientBottomSheet` can manage.

These lambdas are so that the `ContactPatientBottomSheet` can register listeners on the views so that it can forward these events to the `ContactPatientUpdate`.

app/src/main/java/org/simple/clinic/patientcontact/CallPatientView.kt
vinaysshenoy: They need to be public so that they can be accessed by the `ContactPatientBottomSheet` to register listeners.

app/src/main/java/org/simple/clinic/patientcontact/PatientContactEvent.kt
vinaysshenoy: Ah, this is because of the way we switched to doing runtime permissions requesting.

I found the older way of requesting permissions was leaking into business logic, and I wanted to move it out of this and manage it before it goes into the Mobius loop.

As a side-effect of doing this, I had to implement this in such a way that these events have these properties that have to be set in the constructor. Please see [this PR](https://github.com/simpledotorg/simple-android/pull/1243) for the relevant details.

I also think it's possible to do this via an annotation, but I haven't gotten time to explore this yet.

vinaysshenoy: The tl;dr is that these properties are required to be defined by any type that implements the `RequiresPermission` interface since it uses them to do the automatic permissions management.

I can also convert them to functions instead in another PR if you think that might be more explicit.

app/src/test/java/org/simple/clinic/patientcontact/PatientContactUpdateTest.kt
```diff
@@ -56,6 +61,66 @@ class PatientContactUpdateTest {
         ))
   }
 
+  @Test
+  fun `when normal call is selected the call permission is granted, directly call the patient with auto dial`() {
```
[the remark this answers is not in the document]
vinaysshenoy: Whoops, typo, then copy-pasta. Will fix this and the others.

app/src/main/java/org/simple/clinic/patientcontact/CallPatientView.kt
sanchita: Can we make that change when we get there since it isn't obvious as to why are they kept open?

vinaysshenoy: I'd rather not do this now because all the work is already done and making this change now will cause conflicts when I rebase my branch.

If you feel strongly about this, I can make this change though.

app/src/main/java/org/simple/clinic/patientcontact/PatientContactEvent.kt
sanchita: I understand now.
Although the thing that I want to point here is that while we are moving runtime permissions from business logic, we are adding them to `Event` instead and eventually this looks like we are using `Event` to do more than what it's expected to. I don't think it is a problem if the business logic knows it needs permission to execute something. This could very well have been a side effect as well. Wdyt?
Also, why did we choose to make `RequiresPermission` an `interface` and not a `data class` which can be passed around in an event? You can have default values there as well but they can be set in the `data class` itself instead of `Event`.

sanchita: I know that this is already implemented this way now so I don't expect it to be changed.
I am just trying to make some things clear for myself.

vinaysshenoy: > I understand now.
> Although the thing that I want to point here is that while we are moving runtime permissions from business logic, we are adding them to `Event` instead and eventually this looks like we are using `Event` to do more than what it's expected to. I don't think it is a problem if the business logic knows it needs permission to execute something. This could very well have been a side effect as well. Wdyt?
> Also, why did we choose to make `RequiresPermission` an `interface` and not a `data class` which can be passed around in an event? You can have default values there as well but they can be set in the `data class` itself instead of `Event`.

The reason is because it needs to be a type. If you look at the `RequestPermissions` observable transformer, this looks at incoming events to see which ones require permissions (which is does by examining the type) and then holds them in-flight while the permission request flow is done and then forwards them again with the request result. If this is a property on the event, then I would have to check for the presence of this property using reflection instead.

vinaysshenoy: One thing which I can think of for the future is still have the `RequiresPermission` interface:

```kotlin
fun interface RequiresPermission {
  var grant: Optional<PermissionResult> = None
}
```

and then change the definition of which permissions are required to an annotation instead:

```kotlin
@Permission(
  value = Manifest.permission.CALL_PHONE,
  requestCode = 100
)
object NormalCall: ContactPatientEvent, RequiresPermission
```

I am not sure if this would work out but I can try it out if you feel it looks better.

vinaysshenoy: > I understand now.
> Although the thing that I want to point here is that while we are moving runtime permissions from business logic, we are adding them to `Event` instead and eventually this looks like we are using `Event` to do more than what it's expected to. I don't think it is a problem if the business logic knows it needs permission to execute something. This could very well have been a side effect as well. Wdyt?
> Also, why did we choose to make `RequiresPermission` an `interface` and not a `data class` which can be passed around in an event? You can have default values there as well but they can be set in the `data class` itself instead of `Event`.

In response to your other concern, since permissions are something that are specific to the Android platform, I did not want this logic of how to request for permissions themselves to the Mobius loop. It is sufficient for our use case that when the mobius loop receives the event, it knows whether permission was granted or not for the event to proceed.

What do you think?

sanchita: Sure. This looks better. Let's discuss this then.

PR #1266, opened 2020-04-20
`PrescribedDrugsScreenController` test refactor

sanchita:
Pivotal story -> https://www.pivotaltracker.com/story/show/172145024

This PR covers Step 1 (Stabilize) of Mobius migration for `PrescribedDrugsScreenController`.

### DIFF
--- a/app/src/test/java/org/simple/clinic/drugs/selection/PrescribedDrugsScreenControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/drugs/selection/PrescribedDrugsScreenControllerTest.kt
@@ -1,6 +1,5 @@
 package org.simple.clinic.drugs.selection
 
-import com.nhaarman.mockitokotlin2.any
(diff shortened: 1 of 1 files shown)


vinaysshenoy:
One thing I noticed was that the controller uses a custom `PrescribedDrugsScreenCreated` event to get access to the `patientUuid`.

This is supposed to be provided through `AssistedInject` as part of the migration (See point 7 of the "Stabilize" section). Just wanted to confirm that this is being done in another PR.

sanchita:
> One thing I noticed was that the controller uses a custom `PrescribedDrugsScreenCreated` event to get access to the `patientUuid`.
> 
> This is supposed to be provided through `AssistedInject` as part of the migration (See point 7 of the "Stabilize" section). Just wanted to confirm that this is being done in another PR.

I missed this step. I will add a commit for this as well.

PR #1270, opened 2020-04-21
Trigger the contact patient bottom sheet when clicking on the phone number in the summary screen

vinaysshenoy:
https://www.pivotaltracker.com/story/show/170595045


app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -225,6 +227,13 @@ class PatientSummaryScreen(
         .map { PatientSummaryLinkIdCancelled }
   }
 
+  private fun phoneNumberClicks(): Observable<UiEvent> {
+    return if (config.callPatientFeatureEnabled)
```
sanchita, line 231: Is this a feature flag or developer flag?
If this flag will be removed once this feature is complete then its alright.
If this will go to production and can be toggled, then let's move this logic to `Update` class. Wdyt?

app/src/main/AndroidManifest.xml
```diff
@@ -116,7 +116,7 @@
       android:windowSoftInputMode="adjustResize" />
```
sanchita, line 116: You remembered :)
Thanks!

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -225,6 +227,13 @@ class PatientSummaryScreen(
...
+      contactTextView.clicks().map { ContactPatientClicked }
+    else
+      Observable.never()
```
sanchita, line 234: Instead of not emitting an event when click happens, we should disable click listeners. Wdyt?

vinaysshenoy:
> Overall PR looks good. Left minor comments.
> Do we need to update the Changelog?

This is still a WIP feature since the work to implement the "Remind Appointment In" and "Remove Appointment from overdue list" features are still pending. But I will add a WIP changelog item.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: This is technically a developer flag, but I added it as a remote flag since we can turn it on on the `SBX` environment and review the intermediate feature.

But the final PR of this feature will remove this flag and enable the feature permanently, so I decided to leave this in the screen level. Does this make sense?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: I think that makes sense, I will do this.

vinaysshenoy: Ok, I tried doing this, but it's a pain because the Rx stream will automatically make the view clickable when it sets up the stream (https://developer.android.com/reference/android/view/View.html#setOnClickListener%28android.view.View.OnClickListener%29).

Given that this is a developer flag and I am going to remove this at the end of the feature, I think I'd prefer to leave this as is for now since it's not worth the effort to fix.

sanchita: Yea. Not worth it.

PR #1272, opened 2020-04-21
Display the current app version in the settings screen

msasikanth:
https://www.pivotaltracker.com/story/show/168804289

### DIFF
--- a/app/src/main/java/org/simple/clinic/settings/SettingsEffect.kt
+++ b/app/src/main/java/org/simple/clinic/settings/SettingsEffect.kt
@@ -7,3 +7,5 @@ object LoadUserDetailsEffect : SettingsEffect()
 object LoadCurrentLanguageEffect : SettingsEffect()
 
 object OpenLanguageSelectionScreenEffect: SettingsEffect()
+
+data class LoadAppVersionEffect(val applicationId: String) : SettingsEffect()
--- a/app/src/main/java/org/simple/clinic/settings/SettingsEffect.kt
+++ b/app/src/main/java/org/simple/clinic/settings/SettingsEffect.kt
@@ -6,6 +6,8 @@ object LoadUserDetailsEffect : SettingsEffect()
 
 object LoadCurrentLanguageEffect : SettingsEffect()
 
-object OpenLanguageSelectionScreenEffect: SettingsEffect()
+object OpenLanguageSelectionScreenEffect : SettingsEffect()
 
 data class LoadAppVersionEffect(val applicationId: String) : SettingsEffect()
+
+object CheckAppUpdateAvailable : SettingsEffect()
--- a/app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
@@ -1,5 +1,6 @@
 package org.simple.clinic.settings
 
+import android.content.pm.PackageManager
 import com.spotify.mobius.rx2.RxMobius
 import com.squareup.inject.assisted.Assisted
 import com.squareup.inject.assisted.AssistedInject
@@ -13,6 +14,7 @@ class SettingsEffectHandler @AssistedInject constructor(
     private val userSession: UserSession,
     private val settingsRepository: SettingsRepository,
     private val schedulersProvider: SchedulersProvider,
+    private val appVersionFetcher: AppVersionFetcher,
     @Assisted private val uiActions: UiActions
 ) {
 
@@ -26,6 +28,7 @@ class SettingsEffectHandler @AssistedInject constructor(
       .addTransformer(LoadUserDetailsEffect::class.java, loadUserDetails())
       .addTransformer(LoadCurrentLanguageEffect::class.java, loadCurrentSelectedLanguage())
       .addAction(OpenLanguageSelectionScreenEffect::class.java, uiActions::openLanguageSelectionScreen, schedulersProvider.ui())
+      .addTransformer(LoadAppVersionEffect::class.java, loadAppVersion())
       .build()
 
   private fun loadUserDetails(): ObservableTransformer<LoadUserDetailsEffect, SettingsEvent> {
@@ -42,6 +45,16 @@ class SettingsEffectHandler @AssistedInject constructor(
       effectStream
           .flatMapSingle { settingsRepository.getCurrentLanguage().subscribeOn(schedulersProvider.io()) }
           .map(::CurrentLanguageLoaded)
+    }
+  }
+
+  private fun loadAppVersion(): ObservableTransformer<LoadAppVersionEffect, SettingsEvent> {
+    return ObservableTransformer { effectStream ->
+      effectStream
+          .map { appVersionEffect ->
+            val appVersionName = appVersionFetcher.appVersion(appVersionEffect.applicationId)
+            AppVersionLoaded(appVersionName)
+          }
     }
   }
 }
--- a/app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
@@ -1,47 +1,46 @@
 package org.simple.clinic.settings
 
 import com.spotify.mobius.rx2.RxMobius
+import com.squareup.inject.assisted.Assisted
+import com.squareup.inject.assisted.AssistedInject
 import io.reactivex.ObservableTransformer
 import io.reactivex.Scheduler
 import org.simple.clinic.user.UserSession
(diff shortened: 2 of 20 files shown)


app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
```diff
@@ -24,12 +27,14 @@ class SettingsEffectHandlerTest {
   private val userSession = mock<UserSession>()
   private val settingsRepository = mock<SettingsRepository>()
   private val uiActions = mock<UiActions>()
+  private val packageManager = mock<PackageManager>()
```
vinaysshenoy, line 33: We should not be doing this since we should not be mocking types that we do not own, i.e, `PackageManager` is not owned by us and we should not be mocking it.

Can we put this behind a type, say,

```kotlin
class AppVersionFetcher(packageManager) {
  fun appVersion(applicationId: String): String {
    // Do work here
  }
}
```

and provide this as a dependency to the effect handler instead?

app/src/main/java/org/simple/clinic/settings/SettingsInit.kt
```diff
@@ -8,8 +8,8 @@ class SettingsInit : Init<SettingsModel, SettingsEffect> {
 
   override fun init(model: SettingsModel): First<SettingsModel, SettingsEffect> {
     val effects = when {
-      model.userDetailsQueried -> setOf(LoadCurrentLanguageEffect)
-      else -> setOf(LoadUserDetailsEffect, LoadCurrentLanguageEffect)
+      model.userDetailsQueried -> setOf(LoadCurrentLanguageEffect, LoadAppVersionEffect(model.applicationId))
```
vinaysshenoy: The reason this is done is in order to prevent querying the app version again on state restoration unless we expect the data to change. Since the data won't change, should we add a check to prevent the app version from being fetched again as well?

app/src/main/java/org/simple/clinic/settings/SettingsEffect.kt
```diff
@@ -6,6 +6,8 @@ object LoadUserDetailsEffect : SettingsEffect()
 
 object LoadCurrentLanguageEffect : SettingsEffect()
 
-object OpenLanguageSelectionScreenEffect: SettingsEffect()
+object OpenLanguageSelectionScreenEffect : SettingsEffect()
 
 data class LoadAppVersionEffect(val applicationId: String) : SettingsEffect()
+
+object CheckAppUpdateAvailabilityEffect : SettingsEffect()
```
vinaysshenoy, line 13: Rename to `CheckAppUpdateAvailable`?

app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
```diff
@@ -65,4 +68,16 @@ class SettingsEffectHandler @AssistedInject constructor(
           }
     }
   }
+
+  private fun checkAppUpdateAvailability(
+      appUpdateAvailability: CheckAppUpdateAvailability
+  ): ObservableTransformer<CheckAppUpdateAvailabilityEffect, SettingsEvent> {
+    return ObservableTransformer { effectStream ->
+      effectStream
+          .flatMap {
```
vinaysshenoy, line 77: This should be a `.switchMap` instead since we would want to unsubscribe from the previous subscription if the effect is received.

app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
```diff
@@ -65,4 +68,16 @@ class SettingsEffectHandler @AssistedInject constructor(
...
+            appUpdateAvailability.listen()
```
vinaysshenoy, line 78: One thing to note is that the `CheckAppUpdateAvailability#listen` method will only emit if the version on the Play Store is ahead of the installed version by a certain value. This was implemented because we didn't want to show a pop-up to users every time they open the app.

However, for this specific screen, it might be better to show the button if there is an update available, regardless of how close to the current app version it is.

If this is the case, we might have to modify `CheckAppUpdateAvailability` to support this use case.

app/src/main/java/org/simple/clinic/settings/SettingsModel.kt
```diff
@@ -2,14 +2,16 @@ package org.simple.clinic.settings
...
 
 @Parcelize
 data class SettingsModel(
     val applicationId: String,
     val name: String?,
     val phoneNumber: String?,
     val currentLanguage: Language?,
-    val appVersion: String?
+    val appVersion: String?,
+    val appUpdateState: AppUpdateState?
```
vinaysshenoy, line 15: Since we don't care about the actual content of the `AppUpdateState` and only whether it is `ShowAppUpdate` or not, how about making this a nullable boolean, `isUpdateAvailable` instead?

app/src/test/java/org/simple/clinic/settings/SettingsInitTest.kt
```diff
@@ -13,17 +13,17 @@ class SettingsInitTest {
   private val spec = InitSpec<SettingsModel, SettingsEffect>(SettingsInit())
 
   @Test
-  fun `when screen is created, then user details, current language & app version should be loaded`() {
+  fun `when screen is created, then user details, current language & app version should be loaded & check for app update availability`() {
```
vinaysshenoy, line 16: I think we can leave the implementation details up to the method since the `hasEffects` matcher already tells us what is happening, I think we can just rename this to `when screen is created, then the initial data must be loaded`.

vinaysshenoy: Same comment for the other method where we test restoring of the model.

app/src/main/java/org/simple/clinic/settings/SettingsScreen.kt
```diff
@@ -41,14 +41,13 @@ class SettingsScreen(
...
   private val delegate: MobiusDelegate<SettingsModel, SettingsEvent, SettingsEffect> by unsafeLazy {
-    MobiusDelegate(
+    MobiusDelegate.forView(
         events = events,
         defaultModel = SettingsModel.FETCHING_USER_DETAILS,
         init = SettingsInit(),
         update = SettingsUpdate(),
         effectHandler = settingsEffectHandler.create(this).build(),
-        modelUpdateListener = uiRenderer::render,
-        crashReporter = crashReporter
```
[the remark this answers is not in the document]
msasikanth: Yes, I have noticed that after creating the PR. I have pushed that change already 👍

app/src/test/java/org/simple/clinic/settings/SettingsEffectHandlerTest.kt
msasikanth: Yes, this where I was confused about. Whether or not I can mock system classes in tests.

app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
msasikanth: Added a method `listenAllUpdates`

app/src/main/java/org/simple/clinic/settings/SettingsInit.kt
```diff
@@ -8,8 +8,10 @@ class SettingsInit : Init<SettingsModel, SettingsEffect> {
...
     val effects = when {
-      model.userDetailsQueried -> setOf(LoadCurrentLanguageEffect)
-      else -> setOf(LoadUserDetailsEffect, LoadCurrentLanguageEffect)
+      model.userDetailsQueried && model.appVersionQueried -> setOf(LoadCurrentLanguageEffect, CheckAppUpdateAvailable)
+      model.userDetailsQueried && model.appVersionQueried.not() -> setOf(LoadCurrentLanguageEffect, LoadAppVersionEffect(model.applicationId), CheckAppUpdateAvailable)
+      model.userDetailsQueried.not() && model.appVersionQueried -> setOf(LoadUserDetailsEffect, LoadCurrentLanguageEffect, CheckAppUpdateAvailable)
+      else -> setOf(LoadUserDetailsEffect, LoadCurrentLanguageEffect, LoadAppVersionEffect(model.applicationId), CheckAppUpdateAvailable)
     }
 
     return first(model, effects)
```
vinaysshenoy, line 17: I think we can simplify this by making the `effects` a `MutableSet` instead.

```kotlin
    val effects = mutableSetOf(LoadCurrentLanguageEffect, CheckAppUpdateAvailable)

    if (!model.userDetailsQueried) {
      effects.add(LoadUserDetailsEffect)
    }

    if (!model.appVersionQueried) {
      effects.add(LoadAppVersionEffect(model.applicationId))
    }

    return first(model, effects)
```

PR #1275, opened 2020-04-22
Verify PR description contains pivotal story

msasikanth:
https://www.pivotaltracker.com/story/show/172448351

### DIFF
--- a/.github/workflows/ci_checks.yml
+++ b/.github/workflows/ci_checks.yml
@@ -125,6 +125,22 @@ jobs:
       - name: Verify existing Room schemas are not modified
         run: current/.github/scripts/compare_directory_contents.sh master/app/schemas/org.simple.clinic.AppDatabase current/app/schemas/org.simple.clinic.AppDatabase
 
+  verify_pr_description:
+    runs-on: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_checks.yml
```diff
@@ -125,6 +125,14 @@ jobs:
       - name: Verify existing Room schemas are not modified
         run: current/.github/scripts/compare_directory_contents.sh master/app/schemas/org.simple.clinic.AppDatabase current/app/schemas/org.simple.clinic.AppDatabase
 
+  verify_pr_description:
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v2
```
vinaysshenoy, line 131: We won't need to checkout the code since this step only needs access to the PR meta-data.

msasikanth: I was using that to use the shell script in the scripts folder. But since this is just a simple regex check I can move the scrip to yaml file itself. wdyt?

vinaysshenoy: Ah, I see. Moving the regex to the YAML seems fine.

PR #1276, opened 2020-04-22
Add CI check for verifying that the CHANGELOG was updated

vinaysshenoy:
https://www.pivotaltracker.com/story/show/172448348

### DIFF
--- a/.github/workflows/ci_checks.yml
+++ b/.github/workflows/ci_checks.yml
@@ -127,9 +127,9 @@ jobs:
 
   verify_pr_description:
     runs-on: [ubuntu-latest]
+    if: github.event_name == 'pull_request'
     steps:
       - name: Verify PR description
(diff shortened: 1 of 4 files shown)


.github/workflows/ci_checks.yml
```diff
@@ -127,9 +127,9 @@ jobs:
 
   verify_pr_description:
     runs-on: [ubuntu-latest]
+    if: github.event_name == 'pull_request'
```
msasikanth, line 130: good catch, I thought I changed this 👍

vinaysshenoy: Actually both work, you can skip the job, or a step in the job as well.

msasikanth: Yes, but in the previous implementation, I only skipped a step. The empty job still gets triggered, so I wanted to change that.

PR #1281, opened 2020-04-23
Migrate `PrescribedDrugsScreenController` logic to Mobius

sanchita:
https://www.pivotaltracker.com/story/show/172145024

This PR completes the strangulation and migration of the controller to Mobius.


app/src/main/java/org/simple/clinic/drugs/EditMedicinesEffectHandler.kt
```diff
@@ -20,10 +29,29 @@ class EditMedicinesEffectHandler @AssistedInject constructor(
...
   }
+
+  private fun fetchDrugsList(io: Scheduler): ObservableTransformer<FetchPrescribedAndProtocolDrugs, EditMedicinesEvent> {
+    return ObservableTransformer { effects ->
+
+      val protocolDrugsStream = effects
+          .observeOn(io)
+          .flatMap { userSession.requireLoggedInUser() }
+          .switchMap { facilityRepository.currentFacility(it) }
+          .switchMap { protocolRepository.drugsForProtocolOrDefault(it.protocolUuid) }
```
vinaysshenoy, line 49: Since protocols are synced once a day, it is very unlikely that they would change while on this screen. Should we convert this to a non-reactive call and simplify the overall chain?

vinaysshenoy: This is technically not a part of this PR, so this could be something we also add to Pivotal as a task and pick up later.

app/src/main/java/org/simple/clinic/drugs/EditMedicinesEffectHandler.kt
```diff
@@ -20,10 +29,29 @@ class EditMedicinesEffectHandler @AssistedInject constructor(
...
+
+      val prescribedDrugsStream = effects
+          .flatMap { prescriptionRepository.newestPrescriptionsForPatient(it.patientUuid).subscribeOn(io) }
```
vinaysshenoy, line 54: Should this be a `switchMap` instead?

app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt
```diff
@@ -1,9 +1,42 @@
...
 import org.simple.clinic.mobius.ViewRenderer
+import org.simple.clinic.protocol.ProtocolDrugAndDosages
 
 class EditMedicinesUiRenderer(private val ui: PrescribedDrugUi) : ViewRenderer<EditMedicinesModel> {
   override fun render(model: EditMedicinesModel) {
+    if (model.prescribedDrugs == null || model.protocolDrugs == null)
+      return
+
+    if (model.prescribedDrugs != null && model.protocolDrugs != null)
+      renderPrescribedProtocolDrugs(model.prescribedDrugs, model.protocolDrugs)
```
vinaysshenoy, line 15: Suggestion: if we move the conditional for checking whether a model is ready to render to the model, we can simplify this a lot.

```kotlin
class EditMedicinesModel {
  val hasLoadedPrescriptionsAndProtocols: Boolean
    get() = prescribedDrugs != null && protocolDrugs != null
}

class EditMedicinesUiRenderer {
  fun render() {
    if(model.hasLoadedPrescriptionsAndProtocols) {
      // render UI here
    }
  }
}
```

app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt
```diff
@@ -1,9 +1,42 @@
...
+              id = index,
+              drugName = drugAndDosages.drugName,
+              prescribedDrug = matchingPrescribedDrug,
+              hideDivider = isAtLeastOneCustomDrugPrescribed.not() && index == protocolDrugs.lastIndex)
+        }
+
+    val customDrugs = prescribedDrugs
+        .filter { it.isProtocolDrug.not() }
+    val customPrescribedDrugItems = customDrugs
+        .sortedBy { it.updatedAt.toEpochMilli() }
```
vinaysshenoy, line 35: Since `Instant` already implements `Comparable`, I think we can skip the `.toEpochMilli` conversion.

app/src/main/java/org/simple/clinic/drugs/selection/PrescribedDrugsScreenController.kt
```diff
@@ -38,48 +33,6 @@ class PrescribedDrugsScreenController @AssistedInject constructor(
...
-
-          val customDrugs = prescribedDrugs
-              .filter { it.isProtocolDrug.not() }
-          val customPrescribedDrugItems = customDrugs
-              .sortedBy { it.updatedAt.toEpochMilli() }
-              .mapIndexed { index, prescribedDrug -> CustomPrescribedDrugListItem(prescribedDrug, index == customDrugs.lastIndex) }
-
-          protocolDrugSelectionItems + customPrescribedDrugItems
-        }
-        .map { { ui: Ui -> ui.populateDrugsList(it) } }
```
vinaysshenoy, line 87: 🚀

app/src/main/java/org/simple/clinic/drugs/EditMedicinesEffectHandler.kt
```diff
@@ -42,9 +42,13 @@ class EditMedicinesEffectHandler @AssistedInject constructor(
...
       val protocolDrugsStream = effects
           .observeOn(io)
-          .flatMap { userSession.requireLoggedInUser() }
-          .switchMap { facilityRepository.currentFacility(it) }
+          .map {
+            val loggedInUser = userSession.loggedInUserImmediate()!!
+            facilityRepository.currentFacilityImmediate(loggedInUser)!!
+          }
           .switchMap { protocolRepository.drugsForProtocolOrDefault(it.protocolUuid) }
+          .observeOn(schedulersProvider.ui())
```
vinaysshenoy, line 50: Is there a reason this is being observed on the UI thread?

app/src/main/java/org/simple/clinic/drugs/EditMedicinesEffectHandler.kt
```diff
@@ -42,9 +42,13 @@ class EditMedicinesEffectHandler @AssistedInject constructor(
 
       val protocolDrugsStream = effects
           .observeOn(io)
-          .flatMap { userSession.requireLoggedInUser() }
-          .switchMap { facilityRepository.currentFacility(it) }
+          .map {
+            val loggedInUser = userSession.loggedInUserImmediate()!!
+            facilityRepository.currentFacilityImmediate(loggedInUser)!!
```
vinaysshenoy, line 47: What we actually need here is the protocol UUID, so what you think about extracting this to a separate method that returns the protocol UUID instead?

app/src/main/java/org/simple/clinic/drugs/EditMedicinesEvent.kt
```diff
@@ -7,3 +7,10 @@ sealed class EditMedicinesEvent : UiEvent
 object AddNewPrescriptionClicked : EditMedicinesEvent() {
   override val analyticsName = "Drugs:Protocol:Add Custom Clicked"
 }
+
+data class ProtocolDrugClicked(
+    val drugName: String,
+    val prescriptionForProtocolDrug: PrescribedDrug?
```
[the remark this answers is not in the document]
sanchita: I can do this as part of refactor. Will make this change.

app/src/main/java/org/simple/clinic/drugs/EditMedicinesEffectHandler.kt
sanchita: I can create a non-reactive method in `ProtocolRepository`. 
Will do this.

app/src/main/java/org/simple/clinic/drugs/EditMedicinesEffectHandler.kt
sanchita: We do not need `facility` anywhere else. So I think its alright to keep the `protocolUuid` extraction here itself.

vinaysshenoy: Yeah, that's what I meant too. If we have a separate method in the effect handler

```kotlin
fun getCurrentProtocolUuid(): UUID {
  // get protocol UUID here
}
```

Then the implementation detail that the protocol ID is being fetched via the current facility is hidden from the Rx chain.

But this is just a suggestion. If you feel that the current implementation is fine, that's ok too.

app/src/main/java/org/simple/clinic/drugs/EditMedicinesEffectHandler.kt
sanchita: Where does the thread get switched? I wanted to ask this before as well.
Are the events already emitted on the `mainThread`?

app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt
sanchita: I just realized that I don't need the first `if` condition. I will remove that as well.

sanchita: Also, if I move the condition to the model, I will have to add null safety when accessing the two fields instead of being able to use Kotlin's smart cast.
I think it's better this way. Please suggest.

vinaysshenoy: I personally prefer extracting boolean logic out into a property or a local variable and give it a name that has some meaning, so I suggested this.

Even if we don't move this to the model, I think we can also remove that early return since it doesn't add any value since we have another check to render the UI in the first place.

app/src/main/java/org/simple/clinic/drugs/EditMedicinesEffectHandler.kt
vinaysshenoy: Since we have a `.observeOn(schedulersProvider.ui())` there, any event that gets emitted will be emitted on the UI thread.

I just wanted to know why this line was added.

app/src/main/java/org/simple/clinic/drugs/EditMedicinesEffectHandler.kt
```diff
@@ -42,16 +43,20 @@ class EditMedicinesEffectHandler @AssistedInject constructor(
...
-          .flatMap { prescriptionRepository.newestPrescriptionsForPatient(it.patientUuid).subscribeOn(io) }
+          .switchMap { prescriptionRepository.newestPrescriptionsForPatient(it.patientUuid).subscribeOn(io) }
 
       Observables
           .combineLatest(protocolDrugsStream, prescribedDrugsStream)
           .map { DrugsListFetched(it.first, it.second) }
     }
   }
+
+  private fun currentProtocolUuid(): UUID? {
```
vinaysshenoy, line 58: This need not be null since we'll generally have a protocol UUID. Otherwise, if we expect this to actually be nullable, we can return an `Optional` instead and filter the stream since Rx will crash if we return a `null`.

sanchita: This can be `null` because the `currentFacility` API returns a nullable. 
A better way to do this is to not propagate nullable facility post login because it's not going to happen. I can add null-safety in this method instead.

PR #1287, opened 2020-04-27
Implement the mobius loop for the set appointment reminder view

vinaysshenoy:

### DIFF
--- a/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUpdate.kt
@@ -26,13 +26,12 @@ class ContactPatientUpdate(
       is OverdueAppointmentLoaded -> next(model.overdueAppointmentLoaded(event.overdueAppointment))
       is NormalCallClicked -> directlyCallPatient(model, event)
       is SecureCallClicked -> maskedCallPatient(model, event)
-      is PatientMarkedAsAgreedToVisit -> dispatch(CloseScreen)
+      is PatientMarkedAsAgreedToVisit, ReminderSetForAppointment -> dispatch(CloseScreen)
       is PatientAgreedToVisitClicked -> dispatch(MarkPatientAsAgreedToVisit(model.appointment!!.get().appointment.uuid))
       is NextReminderDateClicked -> selectNextReminderDate(model)
       is PreviousReminderDateClicked -> selectPreviousReminderDate(model)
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUpdate.kt
```diff
@@ -28,9 +30,24 @@ class ContactPatientUpdate(
...
+  private fun updateWithManuallySelectedDate(
+      event: ManualDateSelected,
+      model: ContactPatientModel
+  ): Next<ContactPatientModel, ContactPatientEffect> {
+    val selectedDate = event.selectedDate
+    val currentDate = event.currentDate
+
+    val matchingPotentialAppointmentDate = model.potentialAppointments.firstOrNull { it.scheduledFor == selectedDate }
+
+    val reminderDate = matchingPotentialAppointmentDate ?: PotentialAppointmentDate(selectedDate, Days(currentDate daysTill selectedDate))
```
msasikanth, line 57: Selected date can't be less than current date right?

vinaysshenoy: It cannot. This is being controlled on the UI since the date pickers are disabled when the selected date is a day after the current date.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUpdate.kt
```diff
@@ -31,9 +31,16 @@ class ContactPatientUpdate(
       is NextReminderDateClicked -> selectNextReminderDate(model)
       is PreviousReminderDateClicked -> selectPreviousReminderDate(model)
       is ManualDateSelected -> updateWithManuallySelectedDate(event, model)
+      is AppointmentDateClicked -> showManualDatePicker(model)
     }
   }
 
+  private fun showManualDatePicker(model: ContactPatientModel): Next<ContactPatientModel, ContactPatientEffect> {
+    val datePickerBounds = model.potentialAppointments.first().scheduledFor..model.potentialAppointments.last().scheduledFor
```
msasikanth, line 43: should we separate the first and last `scheduledFor` to make it more readable? At first, it wasn't apparent it's a range between two dates. WDYT?

vinaysshenoy: Makes sense, I'll do this.

PR #1292, opened 2020-04-28
Implement the UI for the set appointment reminder view in the contact patient sheet

vinaysshenoy:


app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
```diff
@@ -3,6 +3,8 @@ package org.simple.clinic.contactpatient
 import android.content.Context
```
sanchita, line 3: I have a question about why did we choose to flip views in the `ContactPatientBottomSheet` rather than opening the sheet separately. Is it to avoid the navigation issues between multiple bottom sheets ?
And are we are planning to do the same for `RemoveAppointmentSheet` as well?

app/src/main/java/org/simple/clinic/contactpatient/views/SetAppointmentReminderView.kt
```diff
@@ -4,16 +4,35 @@ import android.content.Context
...
+private typealias DoneClicked = () -> Unit
+
 class SetAppointmentReminderView(
     context: Context,
     attributeSet: AttributeSet
 ) : ConstraintLayout(context, attributeSet) {
 
+  var previousStepperClicked: PreviousStepperClicked? = null
+
+  var nextStepperClicked: NextStepperClicked? = null
```
sanchita, line 34: Can we rename this to `incrementDateClicks` and `IncrementStepperClicked`?

app/src/main/java/org/simple/clinic/contactpatient/views/SetAppointmentReminderView.kt
```diff
@@ -4,16 +4,35 @@ import android.content.Context
...
+private typealias NextStepperClicked = () -> Unit
+private typealias AppointmentDateClicked = () -> Unit
+private typealias DoneClicked = () -> Unit
+
 class SetAppointmentReminderView(
     context: Context,
     attributeSet: AttributeSet
 ) : ConstraintLayout(context, attributeSet) {
 
+  var previousStepperClicked: PreviousStepperClicked? = null
```
sanchita, line 32: Can we rename this to `decrementDateClicks` and `DecrementStepperClicked`?

app/src/main/java/org/simple/clinic/contactpatient/UiMode.kt
```diff
@@ -0,0 +1,6 @@
+package org.simple.clinic.contactpatient
+
+enum class UiMode {
+  CallPatient,
+  SetAppointmentReminder
```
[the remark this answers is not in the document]
vinaysshenoy: Yup!

app/src/test/java/org/simple/clinic/contactpatient/ContactPatientCallPatientUiRendererTest.kt
```diff
@@ -21,7 +20,7 @@ import org.threeten.bp.LocalDate
 import org.threeten.bp.Period
 import java.util.UUID
 
-class ContactPatientUiRendererTest {
+class ContactPatientCallPatientUiRendererTest {
```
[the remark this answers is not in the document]
vinaysshenoy: I am following the convention of the rest of the codebase where the UI renderer is prefixed with the feature name. Maybe we should decide once and for all if we should have the prefix for the class name or nor?

My two cents is that we should because if we look at a test report, the names of the classes will all be `UiRendererTest`.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
vinaysshenoy: Hmm, this is weird. I remember adding the reasoning as a commit message, but it seems like I missed it.

The overall reasoning is correct, I did this to:

- Avoid navigation issues between multiple bottom sheets.
- Make migrating this set of screens to the navigation component easier. Once this is done, since nav component has support for bottom sheet fragments, we can split these screens out again. Mobius makes that pretty easy for us.
- In the final part of this PR, I change the overdue screen to open this screen itself and delete all the screens that are a part of the older call patient flow, so this screen is the only one necessary.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientModel.kt
```diff
@@ -17,6 +17,7 @@ import java.util.UUID
 @Parcelize
 data class ContactPatientModel(
     val patientUuid: UUID,
+    val mode: UiMode,
```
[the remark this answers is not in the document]
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
```diff
@@ -67,7 +70,10 @@ class ContactPatientUiRenderer(
       model: ContactPatientModel
   ) {
     val appointmentReminderPeriods = model.potentialAppointments.map { it.timeToAppointment }
-    val selectedReminderPeriod = model.potentialAppointments.first { it.scheduledFor == model.selectedAppointmentDate }.timeToAppointment
+    val selectedReminderPeriod = model
+        .potentialAppointments
+        .firstOrNull { it.scheduledFor == model.selectedAppointmentDate }?.timeToAppointment
+        ?: TimeToAppointment.Days(LocalDate.now(clock) daysTill model.selectedAppointmentDate)
```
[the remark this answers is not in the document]
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
vinaysshenoy: Added this reasoning to the kdoc of `UiMode`.

app/src/main/java/org/simple/clinic/contactpatient/views/SetAppointmentReminderView.kt
vinaysshenoy: Done

app/src/main/java/org/simple/clinic/contactpatient/views/SetAppointmentReminderView.kt
vinaysshenoy: Done.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
```diff
@@ -76,7 +76,7 @@ class ContactPatientBottomSheet : BottomSheetActivity(), ContactPatientUi, Conta
 
   private val permissionResults: Subject<ActivityPermissionResult> = PublishSubject.create()
 
-  private val dialogEvents: PublishSubject<ContactPatientEvent> = PublishSubject.create()
+  private val otherEvents: PublishSubject<ContactPatientEvent> = PublishSubject.create()
```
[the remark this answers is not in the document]
vinaysshenoy: Decided to name them `hotEvents` because this is essentially a "hot" observable (http://reactivex.io/documentation/observable.html). Accepting better naming suggestions 😓.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
sanchita: iirc, the issue with navigation occurs when we have to open a `Screen` from the `Sheet`. Since here we are opening different sheets from one sheet itself, I don't think it should be a problem. That will avoid the extra work of separating out the sheets later on as well.

app/src/test/java/org/simple/clinic/contactpatient/ContactPatientCallPatientUiRendererTest.kt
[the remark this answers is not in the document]
sanchita: I agree that we should have the feature names as prefix. I am just asking to remove the `ContactPatient` prefix since the package name and `CallPatient` is sufficient to point that this is a feature within `ContactPatient`.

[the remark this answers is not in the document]
vinaysshenoy: Ah, for this specific feature I think it makes sense. I will do this.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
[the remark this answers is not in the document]
sanchita: `hotEvents` is like saying "Today's hot news" 😄.
Let's just keep it as `otherEvents`. I couldn't think of anything else either.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
vinaysshenoy: The problem will be present here because the final view, the `RemoveAppointment` view is technically a "screen" and not a bottom sheet.

I think I'd like to get this in as it is now, and then separate them out once we migrate to the navigation component properly since that will resolve a lot of navigation problems that we have.

sanchita: Oh. I forgot we changed `RemoveAppointment` to a screen.
Cool. Let's keep it this way.

PR #1294, opened 2020-04-29
Extract blood pressure reading to a separate model

vinaysshenoy:
https://www.pivotaltracker.com/story/show/172552968


app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
```diff
@@ -377,7 +375,7 @@ class BloodPressureEntrySheetLogicTest {
     )
 
     val newInputDate = LocalDate.of(1991, 2, 14)
-    whenever(bloodPressureRepository.saveMeasurement(any(), any(), any(), any(), any(), any())).doReturn(Single.just(TestData.bloodPressureMeasurement()))
+    whenever(bloodPressureRepository.saveMeasurement(any(), any(), any(), any(), any())).doReturn(Single.just(TestData.bloodPressureMeasurement()))
```
sanchita, line 378: This method is using `any()`. Can you change it to use actual values instead?

app/src/main/java/org/simple/clinic/bp/Validation.kt
```diff
@@ -0,0 +1,15 @@
+package org.simple.clinic.bp
+
+// TODO: Rename to "Result".
+sealed class Validation {
+  // TODO: Rename to "Valid".
```
sanchita, line 5: Can we close these two `TODO`s in this PR itself?

app/src/main/java/org/simple/clinic/bp/Validation.kt
```diff
@@ -0,0 +1,15 @@
+package org.simple.clinic.bp
+
+// TODO: Rename to "Result".
```
sanchita, line 3: Rename to `ValidationResult`?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
```diff
@@ -125,6 +125,14 @@ class BloodPressureEntryUpdate(
     }
   }
 
+  private fun validateEnteredBp(model: BloodPressureEntryModel): Validation {
+    return when {
+      model.systolic.isBlank() -> Validation.ErrorSystolicEmpty
+      model.diastolic.isBlank() -> Validation.ErrorDiastolicEmpty
```
sanchita, line 131: Why do we need to extract the blank checks from the validation?

vinaysshenoy: Because `BloodPressureReading` stores the values as integers, while the model stores it as strings.

I think a long term solution would be to remove the blank validation errors and make them a part of the Mobius loop directly instead of validation errors, but I can do that later.

sanchita: Oh got it!

PR #1295, opened 2020-04-29
CI check to fix android ellipsis

msasikanth:
https://www.pivotaltracker.com/story/show/172579029

### DIFF
--- a/.github/scripts/strings_reader.py
+++ b/.github/scripts/strings_reader.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*- 
+
+import sys
+import re
+import os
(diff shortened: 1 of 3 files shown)


.github/scripts/strings_reader.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python
```
vinaysshenoy, line 1: The script looks good overall, can we also print the path of the file that is being processed?

msasikanth: We are already doing that

```python
print("Checking strings file at -> {}".format(strings_file_path))
```

vinaysshenoy: Ah, I think I missed that line.

PR #1296, opened 2020-04-29
Business logic for remove appointment view from the contact patient screen

vinaysshenoy:
https://www.pivotaltracker.com/story/show/172194352

### DIFF
--- a/app/src/test/java/org/simple/clinic/contactpatient/ContactPatientUpdateTest.kt
+++ b/app/src/test/java/org/simple/clinic/contactpatient/ContactPatientUpdateTest.kt
@@ -464,6 +464,21 @@ class ContactPatientUpdateTest {
         ))
   }
 
+  @Test
+  fun `when the appointment has been cancelled, the sheet must be closed`() {
+    val model = defaultModel()
+        .patientProfileLoaded(patientProfile)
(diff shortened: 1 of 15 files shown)


app/src/test/java/org/simple/clinic/contactpatient/ContactPatientUpdateTest.kt
```diff
@@ -479,6 +490,118 @@ class ContactPatientUpdateTest {
...
+        .patientProfileLoaded(patientProfile)
+        .overdueAppointmentLoaded(Just(overdueAppointment))
+        .removeAppointmentReasonSelected(MovedToPrivatePractitioner)
+
+    spec
+        .given(model)
+        .whenEvent(RemoveAppointmentDoneClicked)
+        .then(assertThatNext(
+            hasNoModel(),
+            hasEffects(CancelAppointment(appointmentUuid = appointmentUuid, reason = AppointmentCancelReason.MovedToPrivatePractitioner) as ContactPatientEffect)
```
msasikanth, line 585: Import the reason directly like the others?

vinaysshenoy: We cannot do this because we already have the `RemoveAppoinmentReason` sealed classes being imported which have a type named the same.

PR #1300, opened 2020-04-30
Fix `fix_strings_ellipsise` CI check

msasikanth:
https://www.pivotaltracker.com/story/show/172579029

### DIFF
--- a/.github/scripts/amend_and_push.sh
+++ b/.github/scripts/amend_and_push.sh
@@ -1,0 +1,20 @@
+readonly GIT_HAS_CHANGES=1 # 0 = clean, 1 = dirty
+
+git config user.name "$USER_NAME"
+git config user.email "$USER_EMAIL"
+
+# Following git command checks the git diff of staged files
+# and assigns the exit code to GIT_DIFF_STATUS variable.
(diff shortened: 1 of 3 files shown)


.github/scripts/commit_and_amend.sh
```diff
@@ -0,0 +1,22 @@
+readonly GIT_HAS_CHANGES=1 # 0 = clean, 1 = dirty
```
vinaysshenoy, line 1: Since this script is also pushing to the repo, should we rename this to `amend_and_push`?

.github/scripts/commit_and_amend.sh
```diff
@@ -0,0 +1,22 @@
...
+# them to new branch
+if [[ ${GIT_DIFF_STATUS} -eq ${GIT_HAS_CHANGES} ]]; then
+  git add -A
+  git commit --amend --no-edit
+
+  git push -f
+
+  exit 0
+else
+  exit 0
```
vinaysshenoy, line 21: Maybe we can pull the `exit 0` out of the if condition and skip the `else` block?

PR #1302, opened 2020-04-30
Migrate medical history summary view to Mobius

vinaysshenoy:
https://www.pivotaltracker.com/story/show/172573860


app/src/main/java/org/simple/clinic/summary/medicalhistory/MedicalHistorySummaryUiRenderer.kt
```diff
@@ -13,9 +13,21 @@ class MedicalHistorySummaryUiRenderer(
   // which will cause tests to break.
   private val medicalHistoryChangedCallback = ValueChangedCallback<MedicalHistory>()
 
+  private val diabetesManagementFlagChangedCallback = ValueChangedCallback<Boolean>()
```
sanchita, line 16: I had no idea this class `ValueChangedCallback` existed 😮

vinaysshenoy: Yeah, it was added specifically for migrating screens which were using `.distinctUntilChanged` in order to avoid having to update assertions. 😅

app/src/main/java/org/simple/clinic/summary/medicalhistory/MedicalHistorySummaryUiRenderer.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.simple.clinic.mobius.ViewRenderer
+import org.simple.clinic.util.ValueChangedCallback
+
+class MedicalHistorySummaryUiRenderer(
+    private val ui: MedicalHistorySummaryUi
+) : ViewRenderer<MedicalHistorySummaryModel> {
+
+  // Added to keep regression tests passing, otherwise whenever the
+  // model updates, the render method will call the UI methods again
+  // which will cause tests to break.
```
sanchita, line 13: I don't think we need this as a comment here. We can instead update the `migration-to-mobius` doc to add this use case. And maybe move this comment to the commit instead?

app/src/main/java/org/simple/clinic/summary/medicalhistory/MedicalHistorySummaryUi.kt
```diff
@@ -8,7 +8,4 @@ interface MedicalHistorySummaryUi: MedicalHistorySummaryUiActions {
   fun hideDiagnosisView()
```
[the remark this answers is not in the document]
vinaysshenoy: Yes, I did reorder some of them later on. If you check the branch out locally, you should see the actual order. GitHub shows them in the order of when they were created.

app/src/main/java/org/simple/clinic/summary/medicalhistory/MedicalHistorySummaryUiRenderer.kt
vinaysshenoy: Good point, I'll make this change.

app/src/main/java/org/simple/clinic/summary/medicalhistory/MedicalHistorySummaryUi.kt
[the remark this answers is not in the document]
vinaysshenoy: It was actually a WIP commit that I broke up later on. I guess GitHub treats it in the same way.

PR #1303, opened 2020-05-04
UI implementation for the "Remove from overdue list" part of the contact patient sheet

vinaysshenoy:
https://www.pivotaltracker.com/story/show/172194352

This PR completes the feature to contact the patient from the summary screen.

### DIFF
--- a/app/src/main/res/drawable/ic_summary_call_icon.xml
+++ b/app/src/main/res/drawable/ic_summary_call_icon.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+
+  <item>
+    <shape android:shape="oval">
+      <solid android:color="@color/green1" />
+    </shape>
+  </item>
+
+  <item>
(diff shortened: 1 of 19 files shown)


app/src/main/res/drawable/ic_summary_call_icon.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+  <item android:drawable="@drawable/summary_call_icon_background" />
+  <item android:drawable="@drawable/summary_call_icon_foreground" />
```
msasikanth, line 4: We can maybe use the insets inside the item instead of creating a new drawable, WDYT?
(for background drawable as well)

```suggestion
  <item>
    <inset
      android:drawable="@drawable/ic_call_9dp"
      android:inset="@dimen/spacing_4" />
  </item>
```

vinaysshenoy: I completely forgot that I could inline these. Done!

PR #1304, opened 2020-05-04
Replace flow with navigation component

msasikanth:
https://www.pivotaltracker.com/story/show/172448374

This PR will replace screen router usage with navigation component in `SetupActivity`

### DIFF
--- a/app/src/main/java/org/simple/clinic/navigation/ScreenHostView.kt
+++ b/app/src/main/java/org/simple/clinic/navigation/ScreenHostView.kt
@@ -1,0 +1,83 @@
+package org.simple.clinic.navigation
+
+import android.content.Context
+import android.os.Bundle
+import android.util.AttributeSet
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.annotation.NavigationRes
+import androidx.fragment.app.Fragment
+import androidx.navigation.NavController
+import androidx.navigation.NavHost
+import androidx.navigation.NavHostController
+import androidx.navigation.Navigation
+import org.simple.clinic.R
+
+class ScreenHostView : Fragment(), NavHost {
+
+  companion object {
+    private const val KEY_NAV_STATE = "org.simple.clinic.navigation:ScreenHostView:navState"
+    private const val KEY_GRAPH_ID = "org.simple.clinic.navigation:ScreenHostView:graphId"
+  }
+
+  @NavigationRes
+  private var graphId: Int = 0
+
+  private var navigationController: NavHostController? = null
+
+  override fun onInflate(context: Context, attrs: AttributeSet, savedInstanceState: Bundle?) {
+    super.onInflate(context, attrs, savedInstanceState)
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/simple/clinic/navigation/ScreenHostView.kt
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @NavigationRes
+  private var graphId: Int = 0
+
+  private var navigationController: NavHostController? = null
+
+  override fun onInflate(context: Context, attrs: AttributeSet, savedInstanceState: Bundle?) {
+    super.onInflate(context, attrs, savedInstanceState)
+    val ta = context.resources.obtainAttributes(
```
vinaysshenoy, line 31: Can we avoid using abbreviations here?

app/src/main/java/org/simple/clinic/setup/SetupActivity.kt
```diff
@@ -126,27 +112,19 @@ class SetupActivity : AppCompatActivity(), UiActions {
   }
 
   override fun showOnboardingScreen() {
-    screenRouter.popAndPush(OnboardingScreenKey(), RouterDirection.FORWARD)
+    // If onboarding screen is already being shown don't navigate again, it would cause
+    // duplicate destinations
+    if (navController.currentDestination?.id != R.id.onboardingScreen) {
+      navController.navigate(R.id.action_placeholderScreen_to_onboardingScreen)
+    }
```
vinaysshenoy, line 113: Do we need to do this check all the time? If so, we can move this check into `ScreenNavigator`, no?

msasikanth: Yeah, the reason why I didn't move it inside navigator is we may need to launch a destination even though it's already a current destination. So I didn't want to change the expected behaviour of the library to avoid issues when we are migrating to the fragment-based navigation in future.

vinaysshenoy: How does the default fragment based implementation handle this case?

msasikanth: We don't have to do it for every navigate call. This particular call was being made everytime when we are restoring state, so to avoid duplicates I placed this check.

msasikanth: I had to use the same check with when using fragments based navigation as well.

vinaysshenoy: > We don't have to do it for every navigate call. This particular call was being made everytime when we are restoring state, so to avoid duplicates I placed this check.

I see. Then I think the actual problem here is the way we are restoring state in `SetupActivity` since it is making a navigation call after state restoration, correct? If this is the case, then we can fix that problem later.

I think `Flow` internally checks if the same key is being pushed and handles this case itself, so moving to nav controller exposes a problem in our UI setup which we did not know about.

PR #1309, opened 2020-05-05
Add GitHub action for fixing unescaped characters in string resources

vinaysshenoy:
https://www.pivotaltracker.com/story/show/172666998


app/src/main/res/values-bn-rBD/strings.xml
```diff
@@ -505,7 +506,7 @@
   <string name="prescribeddrugsummary_item_name">%1$s %2$s</string>
 
   <!-- Contact patient -->
-  <string name="contactpatient_agreed_to_visit">\&quot;ভিজিটে আসতে সম্মত হয়েছেন&quot;\ বলে রোগীকে চিহ্নিত করুন</string>
```
vinaysshenoy, line 508: I am curious how the second `&quot;` did not break lint since it is not escaped. 😮

msasikanth: I don't think escape character is necessary because we are using `&quot;` instead of `"`?

msasikanth:
It's so much cleaner now 😃

app/src/main/res/values-bn-rBD/strings.xml
vinaysshenoy: Ah, right. I think the problem was happening when the translations had a apostrophe character.

Anyway, since the processor converts these to escaped quotes and apostrophes, I think it's ok.

PR #1311, opened 2020-05-06
Add teleconsultation fields to the facility model

msasikanth:
https://www.pivotaltracker.com/story/show/172667342

### DIFF
--- a/app/src/main/java/org/simple/clinic/facility/FacilityPayload.kt
+++ b/app/src/main/java/org/simple/clinic/facility/FacilityPayload.kt
@@ -58,9 +58,8 @@ data class FacilityPayload(
     @Json(name = "deleted_at")
     val deletedAt: Instant?,
 
-    // TODO(vs): 2020-01-22 Make this non-nullable once the feature is in PROD
     @Json(name = "config")
-    val config: FacilityConfig?
+    val config: FacilityConfig
 ) {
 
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/simple/clinic/facility/FacilityPayload.kt
```diff
@@ -89,7 +92,8 @@ data class FacilityPayload(
         updatedAt = updatedAt,
         syncStatus = syncStatus,
         deletedAt = deletedAt,
-        config = config ?: FacilityConfig(diabetesManagementEnabled = false)
+        teleconsultationPhoneNumber = "",
```
vinaysshenoy, line 95: This could be:

```kotlin
teleconsultationPhoneNumber = teleconsultationPhoneNumber ?: ""
```

so that we won't have to change this again when the API starts sending the field.

app/src/sharedTest/java/org/simple/clinic/TestData.kt
```diff
@@ -381,6 +381,7 @@ object TestData {
       createdAt: Instant = Instant.now(),
       updatedAt: Instant = Instant.now(),
       deletedAt: Instant? = null,
+      teleconsultationPhoneNumber: String? = null,
```
vinaysshenoy, line 384: This parameter is unused now, right?

PR #1313, opened 2020-05-06
Migrate `DosagePickerSheet` to Mobius - Strangulate controller

vinaysshenoy:
https://www.pivotaltracker.com/story/show/172669407


app/src/main/java/org/simple/clinic/drugs/selection/dosage/DosagePickerSheet.kt
```diff
@@ -28,7 +28,7 @@ import java.util.Locale
 import java.util.UUID
 import javax.inject.Inject
 
-class DosagePickerSheet : BottomSheetActivity() {
+class DosagePickerSheet : BottomSheetActivity(), DosagePickerUi {
```
msasikanth: We are implementing the `DosagePickerUi` in `DosagePickerSheet` right? why does commit message mention extract? 😅

vinaysshenoy: Because the interface was extracted from an already existing screen.

msasikanth: Ah, okay. 👍

PR #1325, opened 2020-05-15
Inject user and facility in `EffectHandler`s

sanchita:
https://www.pivotaltracker.com/story/show/172668123


app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandler.kt
```diff
@@ -146,14 +145,10 @@ class BloodSugarEntryEffectHandler @AssistedInject constructor(
...
-        .filterAndUnwrapJust()
-        .flatMap { user ->
-          facilityRepository
-              .currentFacility(user)
-              .map { facility -> user to facility }
-        }
+    val user = currentUser.get()
+    return facilityRepository
+        .currentFacility(user)
+        .map { facility -> user to facility }
```
vinaysshenoy, line 151: We can also inject the the current `Facility` directly, right?

sanchita: Yes we can. I didn't do this because I was only injecting `user` at first. Also this required changing the streams that were consuming this. I will make this change.

PR #1329, opened 2020-05-19
User should be able to consult a doctor remotely from the patient summary screen #3

msasikanth:
https://www.pivotaltracker.com/story/show/172667700

Addressing feedback that's left from this [PR](https://github.com/simpledotorg/simple-android/pull/1321).

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffect.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffect.kt
@@ -10,7 +10,7 @@ sealed class PatientSummaryEffect
 
 data class LoadPatientSummaryProfile(val patientUuid: UUID) : PatientSummaryEffect()
 
-object LoadCurrentFacility : PatientSummaryEffect()
+object LoadCurrentUserAndFacility : PatientSummaryEffect()
 
 data class HandleEditClick(
     val patientSummaryProfile: PatientSummaryProfile,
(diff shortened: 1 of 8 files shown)


msasikanth:
> PR looks good overall, but I think this necessity of managing multiple events is getting to be painful. Can we try something else?
> 
> ```kotlin
> data class LoadCurrentUserAndFaclity: PatientSummaryEffect()
> 
> data class CurrentUserAndFacilityLoaded(val user: User, val facility: Facility): PatientSummaryEvent()
> ```
> 
> By merging these two into a single event, things should become a lot simpler. We can kick off fetching the teleconsult info when this event is received instead of the facility event.

Initially, I was planning to do this but wasn't sure if it's the correct way to do this in Mobius 😅

vinaysshenoy:
> > PR looks good overall, but I think this necessity of managing multiple events is getting to be painful. Can we try something else?
> > ```kotlin
> > data class LoadCurrentUserAndFaclity: PatientSummaryEffect()
> > 
> > data class CurrentUserAndFacilityLoaded(val user: User, val facility: Facility): PatientSummaryEvent()
> > ```
> > 
> > 
> > By merging these two into a single event, things should become a lot simpler. We can kick off fetching the teleconsult info when this event is received instead of the facility event.
> 
> Initially, I was planning to do this but wasn't sure if it's the correct way to do this in Mobius 😅

It's perfectly fine. As long as the effect and event are named correctly and make sense from the screen perspective.

PR #1330, opened 2020-05-19
Fix PR description CI when using markdown syntax

msasikanth:
https://www.pivotaltracker.com/story/show/172909087

Because we were using a single quote when converting the PR body to a string, it started considering the URL markdown syntax `()` as "commands". In order to fix that, I have changed to to convert the received PR description to a proper double-quoted string so that it will ignore the "command" blocks inside the string.

### DIFF
--- a/.github/workflows/ci_checks.yml
+++ b/.github/workflows/ci_checks.yml
@@ -131,7 +131,8 @@ jobs:
     steps:
       - name: Verify PR description
         run: |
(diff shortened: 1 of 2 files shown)


msasikanth:
Marking it as WIP, it seems to be not working again in certain cases.

msasikanth:
Closing this PR and opening it as a local branch so that it will easier to rebase instead of rebasing a fork.

PR #1332, opened 2020-05-19
Fix crash when loading the facility in `ScheduleAppointmentSheet`

vinaysshenoy:
https://www.pivotaltracker.com/story/show/172909030

### DIFF
--- a/app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
+++ b/app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
@@ -22,6 +22,7 @@ import org.simple.clinic.patient.PatientRepository
 import org.simple.clinic.protocol.ProtocolRepository
 import org.simple.clinic.util.UserClock
 import org.simple.clinic.util.plus
+import org.simple.clinic.util.scheduler.SchedulersProvider
 import org.simple.clinic.util.unwrapJust
 import org.simple.clinic.widgets.ScreenCreated
 import org.simple.clinic.widgets.UiEvent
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -142,13 +144,15 @@ class ScheduleAppointmentSheetController @AssistedInject constructor(
   }
 
   private fun scheduleDefaultAppointmentDateForSheetCreates(events: Observable<UiEvent>): Observable<PotentialAppointmentDate> {
-    val protocolUuid = currentFacility.get().protocolUuid!!
-    val protocolStream = protocolRepository.protocol(protocolUuid)
+    val protocolUuid = Observable.fromCallable { currentFacility.get().protocolUuid!! }
```
msasikanth, line 147: Shouldn't we have to use a `Scheduler` here to switch to `IO` thread?

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentSheetController.kt
```diff
@@ -191,16 +195,17 @@ class ScheduleAppointmentSheetController @AssistedInject constructor(
         .replay()
         .refCount()
 
-    val appointmentStream = schedulingSkippedStream.map {
-      val currentFacilityUuid = currentFacility.get().uuid
+    val appointmentStream = schedulingSkippedStream
+        .map {
+          val currentFacilityUuid = currentFacility.get().uuid
```
msasikanth, line 200: We do not seem to be using a `Scheduler` here, shouldn't we need it here?

vinaysshenoy: Does not seem to be necessary, I tapped on the sheet over 20 times trying to trigger it. In any case, this will be fixed once the sheet is fully ported to Mobius, so 🤷

PR #1340, opened 2020-05-20
Prepare `SyncIndicatorController` for migration

sanchita:


app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorView.kt
```diff
@@ -60,37 +56,37 @@ class SyncIndicatorView(context: Context, attrs: AttributeSet) : LinearLayout(co
 
   private fun screenCreates() = Observable.just(ScreenCreated())
 
-  private fun viewClicks() = RxView.clicks(syncIndicatorLayout).map { SyncIndicatorViewClicked }
+  private fun viewClicks() = RxView.clicks(rootLayout).map { SyncIndicatorViewClicked }
```
vinaysshenoy, line 105: Nice! Can we also switch to `rxbinding3` while we are on this screen? Need not be done in this PR, we can do it in the final cleanup.

PR #1341, opened 2020-05-20
Migrate schedule appointment sheet to Mobius: Final cleanup

vinaysshenoy:
https://www.pivotaltracker.com/story/show/172714225

### DIFF
--- a/app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentEffectHandler.kt
@@ -90,16 +90,15 @@ class ScheduleAppointmentEffectHandler @AssistedInject constructor(
   private fun scheduleAppointmentForPatient(): ObservableTransformer<ScheduleAppointmentForPatient, ScheduleAppointmentEvent> {
     return ObservableTransformer { effects ->
       effects
-          .flatMapSingle { scheduleAppointment ->
-            appointmentRepository
-                .schedule(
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentEffectHandler.kt
```diff
@@ -107,12 +107,7 @@ class ScheduleAppointmentEffectHandler @AssistedInject constructor(
   private fun loadPatientDefaulterStatus(): ObservableTransformer<LoadPatientDefaulterStatus, ScheduleAppointmentEvent> {
     return ObservableTransformer { effects ->
       effects
-          // TODO (vs) 20/05/20: Make this a synchronous call
-          .switchMapSingle {
-            patientRepository
-                .isPatientDefaulter(it.patientUuid)
-                .firstOrError()
-          }
+          .map { patientRepository.isPatientDefaulter(it.patientUuid) }
```
msasikanth, line 110: Maybe we can use a single `map` function, wdyt?

```suggestion
effects
    .map {
      val isPatientDefaulter = patientRepository.isPatientDefaulter(it.patientUuid)
      PatientDefaulterStatusLoaded(isPatientDefaulter)
    }
```

vinaysshenoy: I could do this, will make the change.

vinaysshenoy: Done

PR #1342, opened 2020-05-20
Show patient address in overdue list

msasikanth:
https://www.pivotaltracker.com/story/show/172931419

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
@@ -36,7 +36,8 @@ data class OverdueAppointmentRow(
     val lastSeenDate: String,
     val diagnosedWithDiabetes: Answer?,
     val diagnosedWithHypertension: Answer?,
-    val showDiagnosisLabel: Boolean
+    val showDiagnosisLabel: Boolean,
+    val patientAddress: OverduePatientAddress
 ) : ItemAdapter.Item<UiEvent> {
 
   companion object {
@@ -68,7 +69,8 @@ data class OverdueAppointmentRow(
           lastSeenDate = dateFormatter.format(overdueAppointment.patientLastSeen.toLocalDateAtZone(clock.zone)),
           diagnosedWithDiabetes = overdueAppointment.diagnosedWithDiabetes,
           diagnosedWithHypertension = overdueAppointment.diagnosedWithHypertension,
-          showDiagnosisLabel = isDiabetesManagementEnabled
+          showDiagnosisLabel = isDiabetesManagementEnabled,
+          patientAddress = overdueAppointment.patientAddress
       )
     }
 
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
```diff
@@ -103,6 +106,12 @@ data class OverdueAppointmentRow(
 
     holder.patientNameTextView.text = context.getString(R.string.overdue_list_item_name_age, name, age.toString())
     holder.patientNameTextView.setCompoundDrawableStart(gender.displayIconRes)
+    holder.patientAddressTextView.text = if (patientAddress.streetAddress != null && patientAddress.colonyOrVillage != null) {
+      "${patientAddress.streetAddress}, ${patientAddress.colonyOrVillage}"
+    } else {
+      patientAddress.streetAddress ?: patientAddress.colonyOrVillage
+      ?: "${patientAddress.district}, ${patientAddress.state}"
+    }
```
vinaysshenoy, line 114: Can we replace this multiple if-else and elvis operators with a single level `when` clause?

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
```diff
@@ -48,13 +49,18 @@ import java.util.UUID
                 THEN 1 
               ELSE 0
             END
-          ) AS isAtHighRisk
+          ) AS isAtHighRisk,
+          
+          PA.uuid patient_address_uuid, PA.streetAddress patient_address_streetAddress, PA.colonyOrVillage patient_address_colonyOrVillage,
+          PA.zone patient_address_zone, PA.district patient_address_district, PA.state patient_address_state, PA.country patient_address_country,
+          PA.createdAt patient_address_createdAt, PA.updatedAt patient_address_updatedAt, PA.deletedAt patient_address_deletedAt
```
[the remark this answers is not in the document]
msasikanth: Yeah, we can probably create a new object. Only fields we will be avoiding are `uuid`, `country` and timestamps, so I wasn't sure if we should be creating a new object. We may have to show `zone` for other countries, so I think we should use the `PatientAddress`, WDYT? or should I go ahead and create an `OverduePatientAddress`?

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
msasikanth: Do you mean something like this?
```kotlin
holder.patientAddressTextView.text = when {
  patientAddress.streetAddress != null && patientAddress.colonyOrVillage != null -> {
    "${patientAddress.streetAddress}, ${patientAddress.colonyOrVillage}"
  }
  patientAddress.streetAddress != null -> {
    patientAddress.streetAddress
  }
  patientAddress.colonyOrVillage != null -> {
    patientAddress.colonyOrVillage
  }
  else -> {
    "${patientAddress.district}, ${patientAddress.state}"
  }
}
```

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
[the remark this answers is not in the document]
vinaysshenoy: We should try to only load the data that we need, so it's better to have a separate model. We can switch it to use the full model later on if required.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
vinaysshenoy: Yes, it seems more readable. Most of these can also have the braces removed.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentRow.kt
```diff
@@ -103,6 +105,14 @@ data class OverdueAppointmentRow(
...
     holder.patientNameTextView.text = context.getString(R.string.overdue_list_item_name_age, name, age.toString())
     holder.patientNameTextView.setCompoundDrawableStart(gender.displayIconRes)
+    holder.patientAddressTextView.text = when {
+      patientAddress.streetAddress != null && patientAddress.colonyOrVillage != null -> {
+        "${patientAddress.streetAddress}, ${patientAddress.colonyOrVillage}"
+      }
+      patientAddress.streetAddress != null -> patientAddress.streetAddress
+      patientAddress.colonyOrVillage != null -> patientAddress.colonyOrVillage
+      else -> "${patientAddress.district}, ${patientAddress.state}"
+    }
```
vinaysshenoy, line 115: One last thing, instead of `null` checks, should we use `!isNullOrBlank` instead?

PR #1343, opened 2020-05-20
Fix fetching teleconsultation info on patient summary screen restore

msasikanth:
We are now checking if teleconsultation is enabled for the facility and user is logged in before making this call.

https://www.pivotaltracker.com/story/show/172935677

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryInit.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryInit.kt
@@ -15,15 +15,8 @@ class PatientSummaryInit : Init<PatientSummaryModel, PatientSummaryEffect> {
       effects.add(LoadCurrentUserAndFacility)
     }
 
-    if (model.hasLoadedCurrentFacility) {
-      when (model.teleconsultInfo) {
-        null, is TeleconsultInfo.Fetching -> {
-          effects.add(FetchTeleconsultationInfo(model.currentFacility!!.uuid))
-        }
-        is TeleconsultInfo.NetworkError -> {
-          effects.add(ShowTeleconsultInfoError)
-        }
-      }
+    if (model.canCheckTeleconsultationInfo) {
+      effectForTeleconsultInfoState(model, effects)
     }
 
     if (!model.hasCheckedForInvalidPhone) {
@@ -35,5 +28,19 @@ class PatientSummaryInit : Init<PatientSummaryModel, PatientSummaryEffect> {
     }
 
     return first(model, effects)
+  }
+
+  private fun effectForTeleconsultInfoState(
+      model: PatientSummaryModel,
+      effects: MutableSet<PatientSummaryEffect>
+  ) {
+    when (model.teleconsultInfo) {
+      null, is TeleconsultInfo.Fetching -> {
+        effects.add(FetchTeleconsultationInfo(model.currentFacility!!.uuid))
+      }
+      is TeleconsultInfo.NetworkError -> {
+        effects.add(ShowTeleconsultInfoError)
+      }
+    }
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### Fixes
 - Fix camera crash when QR scanner view is paused/closed
+- Fix fetching teleconsultation error when `PatientSummaryScreen` is restored
 
 ## On Demo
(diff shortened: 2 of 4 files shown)


msasikanth:
Moving it to work-in-progress seems to be an issue

sanchita:
Removing reviewer for now.

app/src/main/java/org/simple/clinic/summary/PatientSummaryInit.kt
```diff
@@ -36,4 +30,18 @@ class PatientSummaryInit : Init<PatientSummaryModel, PatientSummaryEffect> {
...
+      model: PatientSummaryModel,
+      effects: MutableSet<PatientSummaryEffect>
+  ) {
+    when (model.teleconsultInfo) {
+      null, is TeleconsultInfo.Fetching -> {
+        effects.add(FetchTeleconsultationInfo(model.currentFacility!!.uuid))
+      }
+      is TeleconsultInfo.NetworkError -> {
+        effects.add(ShowTeleconsultInfoError)
+      }
```
sanchita, line 44: What happens in the remaining cases of `TeleconsultInfo`?
Can we add an `else` here?

app/src/main/java/org/simple/clinic/summary/PatientSummaryInit.kt
```diff
@@ -15,15 +15,9 @@ class PatientSummaryInit : Init<PatientSummaryModel, PatientSummaryEffect> {
...
-    if (model.hasLoadedCurrentFacility) {
-      when (model.teleconsultInfo) {
-        null, is TeleconsultInfo.Fetching -> {
-          effects.add(FetchTeleconsultationInfo(model.currentFacility!!.uuid))
-        }
-        is TeleconsultInfo.NetworkError -> {
-          effects.add(ShowTeleconsultInfoError)
-        }
-      }
+    val canCheckTeleconsultationInfo = model.hasLoadedCurrentFacility && model.isTeleconsultationEnabled && model.isUserLoggedIn
```
sanchita, line 18: Can we move this check to the `Model` class?

app/src/main/java/org/simple/clinic/summary/PatientSummaryInit.kt
```diff
@@ -15,15 +15,9 @@ class PatientSummaryInit : Init<PatientSummaryModel, PatientSummaryEffect> {
...
+    if (canCheckTeleconsultationInfo) {
+      checkTeleconsultationInfo(model, effects)
```
sanchita, line 20: Rename to `shouldFetchTeleconsultationInfo`?

app/src/main/java/org/simple/clinic/summary/PatientSummaryInit.kt
```diff
@@ -36,4 +30,18 @@ class PatientSummaryInit : Init<PatientSummaryModel, PatientSummaryEffect> {
 
     return first(model, effects)
   }
+
+  private fun checkTeleconsultationInfo(
+      model: PatientSummaryModel,
+      effects: MutableSet<PatientSummaryEffect>
+  ) {
+    when (model.teleconsultInfo) {
+      null, is TeleconsultInfo.Fetching -> {
```
sanchita, line 39: If we are in the `Fetching` state, why would we want to fetch again?

app/src/main/java/org/simple/clinic/summary/PatientSummaryInit.kt
msasikanth: We are not handling any other cases. So else will be empty, is that fine?

app/src/main/java/org/simple/clinic/summary/PatientSummaryInit.kt
msasikanth: Since we are checking during screen restore state, the model may not be updated from the initial request. So, just to be safe I also considering `Fetching` state.

app/src/main/java/org/simple/clinic/summary/PatientSummaryInit.kt
sanchita: Leave it be. We don't need to add `else`

msasikanth: Already added it with a comment `// Do nothing` 😅

PR #1347, opened 2020-05-22
Add `DeletedReason` for `Patient`

msasikanth:
https://www.pivotaltracker.com/story/show/172955500

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration67AndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration67AndroidTest.kt
@@ -1,0 +1,105 @@
+package org.simple.clinic.storage.migrations
+
+import org.junit.Test
+import org.simple.clinic.assertColumns
+import org.simple.clinic.assertValues
+import org.simple.clinic.insert
+import org.simple.clinic.patient.SyncStatus
+import org.threeten.bp.Instant
+import java.util.UUID
+
+class Migration67AndroidTest : BaseDatabaseMigrationTest(66, 67) {
+
+  @Test
+  fun deleted_reason_should_be_added_to_patient_table() {
+    before.assertColumns("Patient", setOf(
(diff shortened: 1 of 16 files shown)


app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration67AndroidTest.kt
```diff
@@ -0,0 +1,45 @@
+package org.simple.clinic.storage.migrations
+
+import org.junit.Test
+import org.simple.clinic.assertColumns
+
+class Migration67AndroidTest : BaseDatabaseMigrationTest(66, 67) {
+
+  @Test
+  fun deleted_reason_should_be_added_to_patient_table() {
```
vinaysshenoy, line 9: Can we also add another test to verify that the patient record is retained after the migration?

app/src/main/java/org/simple/clinic/patient/sync/PatientPayload.kt
```diff
@@ -65,7 +66,10 @@ data class PatientPayload(
 
     // TODO Make reminderConsent a non-nullable field once it is pushed to production
```
vinaysshenoy, line 67: Can we resolved this `TODO` while we are here?

app/src/sharedTest/java/org/simple/clinic/util/TestFunctions.kt
```diff
@@ -29,6 +30,10 @@ fun randomGender(): Gender {
   return Gender.TypeAdapter.knownMappings.keys.shuffled().first()
 }
 
+fun randomDeletedReason(): DeletedReason {
+  return DeletedReason.TypeAdapter.knownMappings.keys.shuffled().first()
```
vinaysshenoy, line 34: I think Kotlin has an extension function now to get a random element from a collection. So `.shuffled().first()` could be replaced with `.random()`.

PR #1351, opened 2020-05-26
Migrate `SyncIndicatorController.updateState` to Mobius

sanchita:

### DIFF
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffect.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffect.kt
@@ -3,3 +3,5 @@ package org.simple.clinic.sync.indicator
 sealed class SyncIndicatorEffect
 
 object FetchLastSyncedStatus : SyncIndicatorEffect()
+
+object FetchDataForSyncIndicatorState : SyncIndicatorEffect()
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffect.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffect.kt
@@ -1,7 +1,15 @@
 package org.simple.clinic.sync.indicator
 
+import java.util.concurrent.TimeUnit
+
 sealed class SyncIndicatorEffect
 
 object FetchLastSyncedStatus : SyncIndicatorEffect()
 
 object FetchDataForSyncIndicatorState : SyncIndicatorEffect()
+
+//TODO: Replace the two arguments with a Duration instead
+data class StartSyncedStateTimer(
+    val intervalAmount: Long,
+    val timeUnit: TimeUnit
+) : SyncIndicatorEffect()
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffect.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffect.kt
@@ -1,3 +1,5 @@
 package org.simple.clinic.sync.indicator
 
 sealed class SyncIndicatorEffect
+
+object FetchLastSyncedStatus : SyncIndicatorEffect()
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
@@ -1,11 +1,14 @@
 package org.simple.clinic.sync.indicator
 
+import com.f2prateek.rx.preferences2.Preference
 import com.spotify.mobius.rx2.RxMobius
 import com.squareup.inject.assisted.Assisted
 import com.squareup.inject.assisted.AssistedInject
 import io.reactivex.ObservableTransformer
+import org.simple.clinic.sync.LastSyncedState
 
 class SyncIndicatorEffectHandler @AssistedInject constructor(
+    private val lastSyncedState: Preference<LastSyncedState>,
     @Assisted private val uiActions: SyncIndicatorUiActions
 ) {
 
@@ -17,5 +20,14 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
   fun build(): ObservableTransformer<SyncIndicatorEffect, SyncIndicatorEvent> =
       RxMobius
           .subtypeEffectHandler<SyncIndicatorEffect, SyncIndicatorEvent>()
+          .addTransformer(FetchLastSyncedStatus::class.java, fetchLastSyncedState())
           .build()
+
+  private fun fetchLastSyncedState(): ObservableTransformer<FetchLastSyncedStatus, SyncIndicatorEvent> {
+    return ObservableTransformer { effect ->
+      effect
+          .switchMap { lastSyncedState.asObservable() }
+          .map(::LastSyncedStateFetched)
+    }
+  }
 }
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
@@ -4,11 +4,17 @@ import com.f2prateek.rx.preferences2.Preference
 import com.spotify.mobius.rx2.RxMobius
 import com.squareup.inject.assisted.Assisted
 import com.squareup.inject.assisted.AssistedInject
+import io.reactivex.Observable
 import io.reactivex.ObservableTransformer
 import org.simple.clinic.sync.LastSyncedState
+import org.simple.clinic.util.UtcClock
+import org.threeten.bp.Instant
 
 class SyncIndicatorEffectHandler @AssistedInject constructor(
     private val lastSyncedState: Preference<LastSyncedState>,
+    private val utcClock: UtcClock,
+    //TODO: Fetch the `SyncIndicatorConfig` instead of a stream
+    private val syncIndicatorConfig: Observable<SyncIndicatorConfig>,
     @Assisted private val uiActions: SyncIndicatorUiActions
 ) {
 
@@ -21,6 +27,7 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
       RxMobius
           .subtypeEffectHandler<SyncIndicatorEffect, SyncIndicatorEvent>()
           .addTransformer(FetchLastSyncedStatus::class.java, fetchLastSyncedState())
+          .addTransformer(FetchDataForSyncIndicatorState::class.java, fetchDataForSyncIndicatorState())
           .build()
 
   private fun fetchLastSyncedState(): ObservableTransformer<FetchLastSyncedStatus, SyncIndicatorEvent> {
@@ -28,6 +35,17 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
       effect
           .switchMap { lastSyncedState.asObservable() }
           .map(::LastSyncedStateFetched)
+    }
+  }
+
+  private fun fetchDataForSyncIndicatorState(): ObservableTransformer<FetchDataForSyncIndicatorState, SyncIndicatorEvent> {
+    return ObservableTransformer { effect ->
+      effect
+          .flatMap {
+            syncIndicatorConfig.map {
+              DataForSyncIndicatorStateFetched(Instant.now(utcClock), it.syncFailureThreshold)
+            }
+          }
     }
   }
 }
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
@@ -8,6 +8,7 @@ import io.reactivex.Observable
 import io.reactivex.ObservableTransformer
 import org.simple.clinic.sync.LastSyncedState
 import org.simple.clinic.util.UtcClock
+import org.simple.clinic.util.scheduler.SchedulersProvider
 import org.threeten.bp.Instant
 
 class SyncIndicatorEffectHandler @AssistedInject constructor(
@@ -15,6 +16,7 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
     private val utcClock: UtcClock,
     //TODO: Fetch the `SyncIndicatorConfig` instead of a stream
     private val syncIndicatorConfig: Observable<SyncIndicatorConfig>,
+    private val schedulersProvider: SchedulersProvider,
     @Assisted private val uiActions: SyncIndicatorUiActions
 ) {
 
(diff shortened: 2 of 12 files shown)


app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
```diff
@@ -28,8 +30,18 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
           .subtypeEffectHandler<SyncIndicatorEffect, SyncIndicatorEvent>()
           .addTransformer(FetchLastSyncedStatus::class.java, fetchLastSyncedState())
           .addTransformer(FetchDataForSyncIndicatorState::class.java, fetchDataForSyncIndicatorState())
+          .addTransformer(StartSyncedStateTimer::class.java, startTimer())
           .build()
 
+  private fun startTimer(): ObservableTransformer<StartSyncedStateTimer, SyncIndicatorEvent> {
+    return ObservableTransformer { effect ->
+      effect
+          .flatMap { Observable.interval(it.intervalAmount, it.timeUnit) }
```
vinaysshenoy, line 39: Should this be a `switchMap` instead? If we receive a new emission from upstream, we don't want two simultaneous ticks to be happening.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
```diff
@@ -28,8 +30,18 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
...
+          .observeOn(schedulersProvider.ui())
```
vinaysshenoy, line 40: Do we need to switch the thread here anymore since we are just emitting an event?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEvent.kt
```diff
@@ -13,3 +13,5 @@ data class DataForSyncIndicatorStateFetched(
     val currentTime: Instant,
     val syncIndicatorFailureThreshold: Duration
 ) : SyncIndicatorEvent()
+
+data class IncrementTimerTick(val minutes: Long) : SyncIndicatorEvent()
```
vinaysshenoy, line 17: Replace the `minutes` type with a `Duration` instead?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffect.kt
```diff
@@ -1,7 +1,14 @@
...
+
 sealed class SyncIndicatorEffect
 
 object FetchLastSyncedStatus : SyncIndicatorEffect()
 
 object FetchDataForSyncIndicatorState : SyncIndicatorEffect()
+
+data class StartSyncedStateTimer(
+    val intervalAmount: Long,
+    val timeUnit: TimeUnit
```
vinaysshenoy, line 13: What do you think of replacing these two with a `Duration` instead and converting to the `TimeUnit` at the usage site?

We could also add an extension function on `Observable` that lets us set up an `interval` with `Duration`. We have done this for [`Observable#timer`](https://github.com/simpledotorg/simple-android/blob/master/app/src/main/java/org/simple/clinic/util/RxFoo.kt#L16).

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorModel.kt
```diff
@@ -23,4 +24,10 @@ data class SyncIndicatorModel(
   fun syncIndicatorStateChanged(newState: SyncIndicatorState): SyncIndicatorModel {
     return copy(syncIndicatorState = newState)
   }
+
+  fun syncIndicatorSyncedTimerIncremented(newTime: Long): SyncIndicatorModel {
+    val syncedState = syncIndicatorState as Synced
```
vinaysshenoy, line 29: Is there a possibility that this cast can fail?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorModel.kt
```diff
@@ -23,4 +24,10 @@ data class SyncIndicatorModel(
...
+    val newDurationSince = syncedState.durationSince.plusMinutes(newTime)
```
vinaysshenoy, line 30: We could change this to a reducer method on `Synced`. 

```kotlin
val newSyncedState = syncedState.plusMinutes(newTime)
```

This way the implementation detail is hidden inside the `Synced` class.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorUiRenderer.kt
```diff
@@ -1,9 +1,19 @@
 package org.simple.clinic.sync.indicator
 
 import org.simple.clinic.mobius.ViewRenderer
+import org.simple.clinic.sync.indicator.SyncIndicatorState.SyncPending
 
-class SyncIndicatorUiRenderer : ViewRenderer<SyncIndicatorModel> {
+class SyncIndicatorUiRenderer(val ui: SyncIndicatorUi) : ViewRenderer<SyncIndicatorModel> {
```
vinaysshenoy, line 6: The `ui` property could be `private`.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorUiRenderer.kt
```diff
@@ -1,9 +1,19 @@
...
   override fun render(model: SyncIndicatorModel) {
-    //TODO: implement later
+    if (model.lastSyncedState == null)
+      return
```
vinaysshenoy, line 9: Can we either:

- Add braces here, or
- Move the `return` statement on the same line as the `if` clause

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorUpdate.kt
```diff
@@ -4,15 +4,78 @@ import com.spotify.mobius.Next
...
     }
   }
+
+  private fun updateSyncIndicatorState(
+      model: SyncIndicatorModel,
+      currentTime: Instant,
+      failureThreshold: Duration
+  ): Next<SyncIndicatorModel, SyncIndicatorEffect> {
+    if (model.lastSyncedState == null || model.lastSyncedState.isEmpty())
+      return noChange()
```
vinaysshenoy, line 37: Same comment as earlier about either wrapping this in braces or moving the `return` onto the same line.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorUpdate.kt
```diff
@@ -4,15 +4,78 @@ import com.spotify.mobius.Next
...
+  }
+
+  //TODO: Refactor this method after migration to Mobius is complete.
+  // This method is copied as is from `SyncIndicatorController` for migration purpose.
+  private fun syncIndicatorState(
+      syncState: LastSyncedState,
+      maxIntervalSinceLastSync: Duration,
+      currentTime: Instant
+  ): SyncIndicatorState {
+    val timestamp = syncState.lastSyncSucceededAt ?: return SyncPending
```
vinaysshenoy, line 60: I prefer using the elvis operator to provide an alternate value, but it seems like we're using it here to provide an alternate code path which is a little hard to understand at parse. What do you think of making this early return explicit by moving it to an `if` block?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorUpdate.kt
```diff
@@ -4,15 +4,78 @@ import com.spotify.mobius.Next
...
+
+    val timeSinceLastSync = Duration.between(timestamp, currentTime)
+    val mostFrequentSyncInterval = enumValues<SyncInterval>()
+        .map { it.frequency }
+        .min()!!
+
+    val syncHappenedInTheFuture = timeSinceLastSync.isNegative
```
vinaysshenoy, line 67: Might be useful to a add a comment that this situation is possible if the device time is changed.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorUpdate.kt
```diff
@@ -4,15 +4,78 @@ import com.spotify.mobius.Next
...
+
+    return when {
+      timeSinceLastSync > maxIntervalSinceLastSync -> ConnectToSync
+      timeSinceLastSync > mostFrequentSyncInterval -> SyncPending
```
vinaysshenoy, line 71: Can we extract these into separate variables and give them names, similar to what we did for `syncHappenedInTheFuture`?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -56,68 +45,10 @@ class SyncIndicatorViewController @Inject constructor(
...
-      timeSinceLastSync > mostFrequentSyncInterval -> SyncPending
-      syncHappenedInTheFuture -> SyncPending
-      else -> when (syncState.lastSyncProgress!!) {
-        SUCCESS -> Synced(timeSinceLastSync)
-        FAILURE -> SyncPending
-        SYNCING -> Syncing
-      }
-    }
-  }
-
```
vinaysshenoy, line 120: 🚀

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
```diff
@@ -4,11 +4,17 @@ import com.f2prateek.rx.preferences2.Preference
...
 import io.reactivex.ObservableTransformer
+import io.reactivex.rxkotlin.Observables
 import org.simple.clinic.sync.LastSyncedState
+import org.simple.clinic.util.UtcClock
+import org.threeten.bp.Instant
 
 class SyncIndicatorEffectHandler @AssistedInject constructor(
     private val lastSyncedState: Preference<LastSyncedState>,
+    private val utcClock: UtcClock,
+    private val syncIndicatorConfig: Observable<SyncIndicatorConfig>,
```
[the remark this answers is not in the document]
sanchita: This change will be made as part of the cleanup. This required changing tests considerably and it wasn't possible during the migration itself.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
vinaysshenoy: In addition, would we want to cancel this in some way if the last synced state preference changes? I have no idea what that would look like, though.

sanchita: This timer will only get started in `Synced` state. This state itself won't be emitted simultaneously. I don't understand how will this happen?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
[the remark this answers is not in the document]
vinaysshenoy: I personally prefer doing this as part of test stabilization since it simplifies the migration effort greatly. I guess it's something I forgot to add. 😓

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
vinaysshenoy: `.interval` will emit forever and since we are using a `.flatMap`, the previous emission won't be cancelled when upstream emits a new effect.

It is possible for the sync state to go from synced to pending and back to synced, which would cause a new effect to be emitted from upstream and a new `interval` subscription to be setup.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEvent.kt
sanchita: Do you want me to rename it or change the type itself?
Changing the type will require changing the implementation. I don't want to do that in this PR at all. This PR does not refactor much. We are just moving controller logic here.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorUiRenderer.kt
sanchita: This has been refactored and remove in later commits.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEvent.kt
vinaysshenoy: Yeah, I'd like us to not use primitives as much as possible and use types that are designed for this and expose this information.

We can change this in the final PR, no need to do it in this one.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorUpdate.kt
sanchita: I have added a comment here saying that this will be refactored later. I wanted to keep this method easy to review by not changing it at all from `Controller` class.
I will make this change in cleanup PR.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorUpdate.kt
sanchita: Same comment as above. Will make this change later.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorUpdate.kt
sanchita: Same comment as above.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
```diff
@@ -20,6 +20,16 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
...
       RxMobius
           .subtypeEffectHandler<SyncIndicatorEffect, SyncIndicatorEvent>()
-          .addTransformer(FetchLastSyncedStatus::class.java) { it.map { LastSyncedStateFetched(lastSyncedState.get()) } }
+          .addTransformer(FetchLastSyncedStatus::class.java, fetchLastSyncedState())
           .build()
+
+  private fun fetchLastSyncedState(): ObservableTransformer<FetchLastSyncedStatus, SyncIndicatorEvent> {
+    return ObservableTransformer { effect ->
+      effect
+          //TODO: Fetch the preference directly instead of turning this to a stream.
```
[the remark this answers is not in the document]
sanchita: Yes. You are right. I will remove this comment.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffect.kt
sanchita: I will pick this in cleanup as well.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
sanchita: That makes sense. I will make this change.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEvent.kt
sanchita: Added a `TODO` for this.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorModel.kt
sanchita: Yes. I will add a check to only trigger this in `Synced` state.

PR #1353, opened 2020-05-27
Users can soft delete patients #2

msasikanth:

I have rebased a couple of commits, so please check out the branch when reviewing.


app/src/main/java/org/simple/clinic/editpatient/deletepatient/DeletePatientScreen.kt
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  override fun showDeleteReasons(patientDeleteReasons: List<PatientDeleteReason>, selectedReason: PatientDeleteReason?) {
+  }
+
+  override fun showConfirmDeleteDialog(patientName: String, deletedReason: DeletedReason) {
+    val title = when (deletedReason) {
+      DeletedReason.Duplicate -> context.getString(R.string.deletereason_duplicate_patient_confirm_title)
+      DeletedReason.AccidentalRegistration -> context.getString(R.string.deletereason_accidental_registration_confirm_title)
+      is DeletedReason.Unknown -> deletedReason.actualValue
```
vinaysshenoy, line 102: Since we will never show the deleted reason in the UI for an already deleted patient, we might as well throw an `IllegalArgumentException` here instead of showing the non-human readable value. What do you think?

vinaysshenoy: We are throwing it when we read the message anyway.

app/src/main/java/org/simple/clinic/editpatient/deletepatient/DeletePatientScreen.kt
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    val message = when (deletedReason) {
+      DeletedReason.Duplicate -> context.getString(R.string.deletereason_duplicate_patient_confirm_message, patientName)
+      DeletedReason.AccidentalRegistration -> context.getString(R.string.deletereason_accidental_registration_confirm_message, patientName)
+      else -> throw IllegalArgumentException("Unknown deleted reason: $deletedReason")
+    }
+
+    AlertDialog.Builder(context, R.style.Clinic_V2_DialogStyle_Destructive)
```
vinaysshenoy, line 111: We have been using dialog fragments so that they are retained upon screen restoration. Can you switch to using them like we do elsewhere?

app/src/main/java/org/simple/clinic/main/TheActivityComponent.kt
```diff
@@ -132,6 +134,7 @@ interface TheActivityComponent :
   fun inject(target: EditPatientScreen)
   fun inject(target: SettingsScreen)
   fun inject(target: ChangeLanguageScreen)
+  fun inject(target: DeletePatientScreenInjector)
```
vinaysshenoy, line 137: Why is this needed? There's no point in injecting an interface, right?

app/src/main/java/org/simple/clinic/editpatient/EditPatientConfig.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.simple.clinic.remoteconfig.ConfigReader
+import javax.inject.Inject
+
+data class EditPatientConfig(
+    val canDeletePatient: Boolean
+) {
+
+  @Inject
+  constructor(configReader: ConfigReader) : this(
+      canDeletePatient = configReader.boolean("can_delete_patient", false)
```
vinaysshenoy, line 12: Rename this property to `deletePatientFeatureEnabled` to make it clear it is a feature flag?

app/src/main/java/org/simple/clinic/editpatient/deletepatient/DeletePatientScreen.kt
```diff
@@ -56,8 +60,16 @@ class DeletePatientScreen(context: Context, attrs: AttributeSet) : ConstraintLay
...
+      adapter = deleteReasonsAdapter
       addItemDecoration(DividerItemDecorator(context, marginStart = 56.dp, marginEnd = 16.dp))
     }
+
+    deleteReasonsAdapter
+        .itemEvents
+        .takeUntil(detaches())
+        .ofType<DeleteReasonItem.Event.Clicked>()
+        .map { PatientDeleteReasonClicked(it.reason) }
+        .subscribe { events.onNext(it) }
```
vinaysshenoy, line 72: 1. None of the events from `DeletePatientEvent` are getting pushed to analytics. Can we look at user generated events and ensure they are pushed correctly?
2. Can we follow our usual pattern of forwarding events?

```kotlin
val events: Observable<DeletePatientEvent> by unsafeLazy {
  Observable.mergeArray(
    dialogEvents(), // this can be the event subject that we have created
    adapterEvents() // set up another event stream here from the adapter
  )
  .compose(ReportAnalyticsEvents()) // This is missing
}
```

app/src/main/java/org/simple/clinic/editpatient/deletepatient/DeleteReasonItem.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    fun from(
+        deleteReasons: List<PatientDeleteReason>,
+        selectedReason: PatientDeleteReason?
+    ): List<DeleteReasonItem> {
+      return deleteReasons.map { deleteReason ->
+        val displayText = when (deleteReason) {
+          PatientDeleteReason.Duplicate -> R.string.deletereason_duplicate_patient
+          PatientDeleteReason.AccidentalRegistration -> R.string.deletereason_accidental_registration
+          PatientDeleteReason.Died -> R.string.deletereason_died
```
vinaysshenoy, line 27: Are we doing this in multiple places? Might be better to have a `displayResId` property on the `PatientDeleteReason` type itself.

app/src/main/java/org/simple/clinic/main/TheActivityComponent.kt
msasikanth: Oh wow, I have no idea why I added that 😅 . It's a mistake, removing it.

app/src/main/java/org/simple/clinic/editpatient/deletepatient/DeletePatientScreen.kt
msasikanth: Pragati mentioned about having a common title and message for the confirmation dialog. I will confirm that and make that change.

app/src/main/java/org/simple/clinic/editpatient/deletepatient/DeleteReasonItem.kt
msasikanth: We are using `PatientDeleteReason` in update and view renderer tests, so I wasn't sure having an Android string res there. WDYT?

vinaysshenoy: As long as the code path to query those are not executed in the tests, it should be fine.

vinaysshenoy: This will be similar to the labels we render for blood sugar types, where we do something similar, I think.

msasikanth: Yeah, that works. Changed it to an enum class (We can continue using the `sealed class` as well but readability was this was cleaner)

```kotlin
enum class PatientDeleteReason(@StringRes val displayText: Int) {
  Duplicate(R.string.deletereason_duplicate_patient),
  AccidentalRegistration(R.string.deletereason_accidental_registration),
  Died(R.string.deletereason_died)
}
```

app/src/main/java/org/simple/clinic/editpatient/deletepatient/DeletePatientScreen.kt
msasikanth: Seems like if I switch to dialog fragments, I have to either setup listeners for dialog actions using interfaces or Kotlin higher-order function or use its own Mobius loop _(I looked at other dialog fragment implementations in Simple and they are using a different a Mobius loop than the screen that opened the dialog)_. Should I use a different Mobius loop that handles the dialog actions? I am thinking of raising a different PR for it, WDYT?

vinaysshenoy: On second thought, I am thinking it's fine to leave it as is.

We have created Mobius loops for dialogs that are handling some business logic themselves. However in this case, it seems like the dialog just forwards events to the parent screen and logic is handled there.

Making this a dialog fragment would mean that we have to move the logic of deleting/marking patients as dead to the dialog class and creating a new mobius loop there since there is no easy way to reconnect events on state restoration and we'll have to check if the fragment exists or not before reattaching events.

So let's leave it as is. We might actually want to consider moving away from dialog fragments, looks like.

msasikanth:
One of the new change I have done during the rebase is using `margin_start` and `paddingStart` for the edit screen to avoid clipping of the button corners and shadows.

```xml
  <ScrollView
    android:id="@+id/formScrollView"
    android:layout_height="match_parent"
    android:layout_above="@+id/saveButton"
    android:layout_marginStart="@dimen/spacing_48"
    android:scrollbarStyle="insideOverlay"
    android:scrollbars="vertical">

    <LinearLayout
      android:clipToPadding="false"
      android:orientation="vertical"
      android:paddingStart="@dimen/spacing_8"
      android:paddingTop="@dimen/spacing_12"
      android:paddingEnd="@dimen/spacing_16"
      android:paddingBottom="64dp"
      tools:ignore="RtlSymmetry">
```

msasikanth:
Marking this as WIP, noticed some issues during the rebase.

PR #1354, opened 2020-05-27
Migrate Patients screen to Mobius: strangulate controller #2

vinaysshenoy:
https://www.pivotaltracker.com/story/show/172953157


app/src/main/java/org/simple/clinic/home/patients/PatientsEffect.kt
```diff
@@ -29,3 +29,7 @@ object LoadNumberOfPatientsRegistered: PatientsEffect()
 object OpenTrainingVideo: PatientsEffect()
 
 object LoadInfoForShowingAppUpdateMessage: PatientsEffect()
+
+object TouchAppUpdateShownAtTime: PatientsEffect()
```
msasikanth: Do we need `Touch` prefix here?

vinaysshenoy: I named this based on the Unix [`touch`](https://www.computerhope.com/unix/utouch.htm) command since that's essentially what we are doing. Couldn't think of a better name. 😅

msasikanth: TIL

PR #1357, opened 2020-05-27
Migrate Sync indicator controller to Mobius #2

sanchita:

This PR completes the controller migration. There will be one more PR that will cover clean-up.

### DIFF
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
@@ -47,7 +47,7 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
       effect
           .map {
             frequentlySyncingRepositories
-                .map { it.pendingSyncRecordCountImmediate() }
+                .map { it.pendingSyncRecordCount() }
                 .sum()
           }
           .map(::PendingSyncRecordCountFetched)
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
@@ -9,9 +9,11 @@ import io.reactivex.ObservableTransformer
 import org.simple.clinic.sync.DataSync
 import org.simple.clinic.sync.LastSyncedState
 import org.simple.clinic.sync.SyncGroup
+import org.simple.clinic.sync.SynceableRepository
 import org.simple.clinic.util.UtcClock
 import org.simple.clinic.util.scheduler.SchedulersProvider
 import org.threeten.bp.Instant
+import javax.inject.Named
 
 class SyncIndicatorEffectHandler @AssistedInject constructor(
     private val lastSyncedState: Preference<LastSyncedState>,
@@ -20,6 +22,7 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
     private val syncIndicatorConfig: Observable<SyncIndicatorConfig>,
     private val schedulersProvider: SchedulersProvider,
     private val dataSync: DataSync,
+    @Named("frequently_syncing_repositories") private val frequentlySyncingRepositories: ArrayList<SynceableRepository<*, *>>,
     @Assisted private val uiActions: SyncIndicatorUiActions
 ) {
 
@@ -35,8 +38,21 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
           .addTransformer(FetchDataForSyncIndicatorState::class.java, fetchDataForSyncIndicatorState())
           .addTransformer(StartSyncedStateTimer::class.java, startTimer())
           .addTransformer(InitiateDataSync::class.java, startDataSync())
+          .addTransformer(FetchPendingSyncRecordCount::class.java, fetchPendingSyncCount())
           .addConsumer(ShowDataSyncErrorDialog::class.java, { uiActions.showErrorDialog(it.errorType) }, schedulersProvider.ui())
           .build()
+
+  private fun fetchPendingSyncCount(): ObservableTransformer<FetchPendingSyncRecordCount, SyncIndicatorEvent> {
+    return ObservableTransformer { effect ->
+      effect
+          .map {
+            frequentlySyncingRepositories
+                .map { it.pendingSyncRecordCountImmediate() }
+                .sum()
+          }
+          .map(::PendingSyncRecordCountFetched)
+    }
+  }
 
   private fun startTimer(): ObservableTransformer<StartSyncedStateTimer, SyncIndicatorEvent> {
     return ObservableTransformer { effect ->
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
@@ -35,6 +35,7 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
           .addTransformer(FetchDataForSyncIndicatorState::class.java, fetchDataForSyncIndicatorState())
           .addTransformer(StartSyncedStateTimer::class.java, startTimer())
           .addTransformer(InitiateDataSync::class.java, startDataSync())
+          .addConsumer(ShowDataSyncErrorDialog::class.java, { uiActions.showErrorDialog(it.errorType) }, schedulersProvider.ui())
           .build()
 
   private fun startTimer(): ObservableTransformer<StartSyncedStateTimer, SyncIndicatorEvent> {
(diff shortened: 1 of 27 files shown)


app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
```diff
@@ -59,4 +63,13 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
           }
     }
   }
+
+  private fun startDataSync(): ObservableTransformer<InitiateDataSync, SyncIndicatorEvent> {
+    return ObservableTransformer { effect ->
+      effect
+          .doOnNext { dataSync.fireAndForgetSync(SyncGroup.FREQUENT) }
+          .switchMap { dataSync.streamSyncErrors() }
```
vinaysshenoy, line 88: Not necessary to change this now, but this need not be a part of this stream, right? What we want is to be notified whenever there is a sync error, in which case this can be set up as a separate effect which continuously observes the stream for errors.

vinaysshenoy: Never mind, this is intentional because we show a pop-up and we want it to be shown only if the user the manually triggered a sync.

Should we add a comment here explaining this?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorUpdate.kt
```diff
@@ -33,6 +33,18 @@ class SyncIndicatorUpdate : Update<SyncIndicatorModel, SyncIndicatorEvent, SyncI
       is DataForSyncIndicatorStateFetched -> updateSyncIndicatorState(model, event.currentTime, event.syncIndicatorFailureThreshold)
       is IncrementTimerTick -> incrementTimer(model, event)
       is DataSyncErrorReceived -> handleDataSyncErrors(event.errorType)
+      SyncIndicatorViewClicked -> syncIndicatorClicked(model.lastSyncedState)
+    }
+  }
+
+  private fun syncIndicatorClicked(lastSyncedState: LastSyncedState?): Next<SyncIndicatorModel, SyncIndicatorEffect> {
+    val isSyncing = lastSyncedState != null && lastSyncedState.lastSyncProgress != SYNCING
```
vinaysshenoy, line 53: I am a little confused why the property is named `isSyncing` and in the condition we check `lastSyncedState.lastSyncProgress != SYNCING`. Is this intentional?

vinaysshenoy: Looking at the usage, I am assuming it should be named `isNotSyncingCurrectly`. Correct if I am wrong.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
```diff
@@ -33,43 +27,12 @@ class SyncIndicatorViewController @Inject constructor(
...
-          lastSyncedState.lastSyncProgress == null || lastSyncedState.lastSyncProgress != SYNCING
-        }
-        .doOnNext { dataSync.fireAndForgetSync(FREQUENT) }
-        .switchMap { showErrorDialogOnSyncError() }
-  }
-
-  private fun showErrorDialogOnSyncError(): Observable<UiChange> {
-    return dataSync
-        .streamSyncErrors()
-        .take(1)
```
vinaysshenoy, line 70: This `.take(1)` ensures that the error message is only shown once and not multiple times if multiple sync errors occur (it is possible since there are multiple syncs that can fail). Can we investigate and ensure that this behaviour is not lost as part of the migration?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
```diff
@@ -35,9 +38,22 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
...
+          .addTransformer(FetchPendingSyncRecordCount::class.java, fetchPendingSyncCount())
           .addConsumer(ShowDataSyncErrorDialog::class.java, { uiActions.showErrorDialog(it.errorType) }, schedulersProvider.ui())
           .build()
 
+  private fun fetchPendingSyncCount(): ObservableTransformer<FetchPendingSyncRecordCount, SyncIndicatorEvent> {
+    return ObservableTransformer { effect ->
+      effect
+          .map {
+            frequentlySyncingRepositories
+                .map { it.pendingSyncRecordCountImmediate() }
```
vinaysshenoy: I think we had left it as an `Observable` here because we wanted the indicator to turn into the `Pending` state if any record on the database gets marked as `PENDING_SYNC`. In most cases, since we have to navigate away from this screen in order to modify a record and navigate back (in which case, the pending state will be queried again), it should be fine.

One case that I am concerned about though, is that this screen runs alongside the `OverdueScreen` in tabs, and that screen modifies `Appointment` records. Not implementing this in a reactive fashion might mean that this never changes to the `Pending` state even though there are pending records.

Please verify?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEvent.kt
```diff
@@ -22,3 +22,5 @@ data class DataSyncErrorReceived(val errorType: ResolvedError) : SyncIndicatorEv
 object SyncIndicatorViewClicked : SyncIndicatorEvent() {
   override val analyticsName = "Sync Indicator: View Clicked"
 }
+
+data class PendingSyncRecordCountFetched(val totalCount: Int) : SyncIndicatorEvent()
```
vinaysshenoy, line 26: Should we rename `totalCount` to `syncPendingRecordCount` to be more explicit?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
sanchita: I will add a comment. Also the naming below is misleading.  `dataSync.fireAndForgetSync(SyncGroup.FREQUENT)`

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorUpdate.kt
sanchita: Yes you are right. I will change the condition instead.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorViewController.kt
sanchita: This is behaving as expected. The dialog is not shown multiple times here.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
sanchita: Good point. I will verify this.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
vinaysshenoy: This was meant to expose the detail that if you call this method, the sync is going to get triggered and no more action is required. Would you mind explaining what you find misleading about this? We can think of alternate names as well.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
sanchita: This does not seem to be working before as well. I will fix this in the next PR.
Is that ok?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
sanchita: Sorry, I meant misleading because in this context it wasn't a `fireAndForgetSync` call. We are listening to errors here after we initiate sync. But I think when the sync call itself was moved to separate function, this became a fire and forget.

vinaysshenoy: Right, but the listening to sync errors is a separate call since you can invoke `fireAndForgetSync` and not have to subscribe to the errors at all.

But I see what you mean and I think a root cause here is that the api of the entire sync layer and how it exposes information about ongoing syncs and triggering them is poorly designed.

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
vinaysshenoy: Yes, let's fix this later then. If it's not an easy fix, can you raise a ticket for this instead? Someone else can pick this up since you'll be rolling off soon.

PR #1359, opened 2020-05-28
Users can soft delete patients - Final

msasikanth:


msasikanth:
> Overall PR looks good.
> Do we need to update changelog?

Since the delete feature got merged in today, this is still a part of that. So I think it's fine?

sanchita:
You need to re-roll your branch

PR #1360, opened 2020-05-28
Fix showing pending sync in indicator

sanchita:

### DIFF
--- a/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
@@ -47,7 +47,7 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
       effect
           .map {
             frequentlySyncingRepositories
-                .map { it.pendingSyncRecordCount() }
+                .map { it.pendingSyncRecordCountImmediate() }
                 .sum()
           }
           .map(::PendingSyncRecordCountFetched)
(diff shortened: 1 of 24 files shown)


app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
```diff
@@ -38,19 +39,22 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
...
 
-  private fun fetchPendingSyncCount(): ObservableTransformer<FetchPendingSyncRecordCount, SyncIndicatorEvent> {
+  private fun fetchPendingSync(): ObservableTransformer<FetchPendingSyncRecordsState, SyncIndicatorEvent> {
     return ObservableTransformer { effect ->
       effect
-          .map {
-            frequentlySyncingRepositories
-                .map { it.pendingSyncRecordCountImmediate() }
-                .sum()
+          .flatMap {
```
vinaysshenoy, line 49: This should be a `.switchMap`, right?

app/src/main/java/org/simple/clinic/sync/indicator/SyncIndicatorEffectHandler.kt
```diff
@@ -38,19 +39,22 @@ class SyncIndicatorEffectHandler @AssistedInject constructor(
...
+            val recordCounts = frequentlySyncingRepositories
+                .map { it.pendingSyncRecordCount() }
+
+            recordCounts
+                .combineLatest { counts -> counts.any { it > 0 } }
+                .filter { isSyncPending -> isSyncPending }
+                .map(::PendingSyncRecordsStateFetched)
```
vinaysshenoy, line 56: I was finding this a little hard to follow, so I tried rewriting it a little bit:

```kotlin
recordCounts
  .combineLatest { counts -> counts.sum() }
  .map { totalPendingSyncRecords -> totalPendingSyncRecords > 0 }
  .map(::PendingSyncRecordsStateFetched)
```

I felt we no longer needed a `.filter {}` here since the `Update` function is already checking the flag in the event that we emit. So the effect handler is only responsible for emitting the state of the pending sync records and the `Update` function is responsible for deciding what to do in the UI.

What do you think?

PR #1366, opened 2020-06-01
Use local flag for delete patient feature

msasikanth:


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Add support for building Android App Bundles (AABs)
 - Migrate `PatientsScreen` to Mobius
 - Migrated `SyncIndicatorController` to Mobius
+- Use local flag for patient delete feature
```
vinaysshenoy, line 20: Since this is the same release in which the "Delete patient" feature was added, I think we can skip this line.

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -108,8 +108,7 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
   @Inject
   lateinit var inputFields: InputFields
 
-  @Inject
-  lateinit var editPatientConfig: EditPatientConfig
+  private val editPatientConfig = EditPatientConfig()
```
vinaysshenoy, line 111: We should not be constructing this directly and they should be injected. Given that this is only a temporary thing, it's fine for now.

CHANGELOG.md
msasikanth: Should we even add the delete patients as part of the feature since the user cannot actually use this and we cannot enable it remotely? 🤔

vinaysshenoy: That's a good point. Let's remove this.

PR #1368, opened 2020-06-01
Fix delete patient sync

msasikanth:

Tested on `QA` build and sync is working correctly. This feature should be deployed `SBX` server by the end of today or tomorrow morning.

### DIFF
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientConfig.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientConfig.kt
@@ -1,3 +1,14 @@
 package org.simple.clinic.editpatient
 
-data class EditPatientConfig(val deletePatientFeatureEnabled: Boolean = false)
+import org.simple.clinic.remoteconfig.ConfigReader
+import javax.inject.Inject
+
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/editpatient/EditPatientConfig.kt
```diff
@@ -1,3 +1,14 @@
...
+import org.simple.clinic.remoteconfig.ConfigReader
+import javax.inject.Inject
+
+data class EditPatientConfig(
+    val deletePatientFeatureEnabled: Boolean
+) {
+
+  @Inject
+  constructor(configReader: ConfigReader) : this(
+      deletePatientFeatureEnabled = configReader.boolean("delete_patient_feature_enabled", false)
```
vinaysshenoy, line 12: Why is this disabled by default? Since the feature is complete, we can enable it by default, right?

msasikanth: Ah yes, I forgot to update it after changing it back to the remote flag 😅

PR #1371, opened 2020-06-02
New UI for on-boarding screen #1

Honey14:
https://www.pivotaltracker.com/story/show/173030066


app/src/main/res/drawable/ic_splash_simple_logo.xml
```diff
@@ -0,0 +1,16 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
vinaysshenoy, line 1: We seem to have a very similar resource already (`ic_logo_large`) that are used during the user registration flows. Could we use that existing resource instead of adding a new one?

Honey14: There is no drawable called `ic_logo_large` . There's one drawable called `ic_icons_logo` which has simple logo and not the text. Zeplin file contains both the logo and the simple text in a single SVG file. Do you want me to verify with Pragati if we should use the text explicitly instead of using it in the SVG file ?

vinaysshenoy: Whoops, I meant `logo_large`. There's no `ic_` prefix.

vinaysshenoy: Yeah, the logo includes the `Simple` text. That is not localised and we should include it in the image resource.

PR #1372, opened 2020-06-02
Fix a couple of flaky tests

vinaysshenoy:


app/src/test/java/org/simple/clinic/sync/DataSyncTest.kt
```diff
@@ -6,11 +6,16 @@ import io.reactivex.Completable
 import io.reactivex.Single
```
msasikanth, line 6: Please check the commit message description.
"This test was ~~not~~ failing earlier..."

app/src/test/java/org/simple/clinic/sync/DataSyncTest.kt
```diff
@@ -47,20 +52,22 @@ class DataSyncTest {
...
     val (sync1Completable, sync1Consumer) = Completable.complete().subscriptionTest()
     whenever(modelSync1.syncConfig()).thenReturn(createSyncConfig(SyncGroup.DAILY))
     whenever(modelSync1.sync()).thenReturn(sync1Completable)
+    whenever(modelSync1.name).thenReturn("sync1")
 
+    val modelSync2 = mock<ModelSync>()
     val sync2Completable = Completable.error(RuntimeException("TEST"))
     whenever(modelSync2.syncConfig()).thenReturn(createSyncConfig(SyncGroup.DAILY))
     whenever(modelSync2.sync()).thenReturn(sync2Completable)
+    whenever(modelSync2.name).thenReturn("sync2")
```
msasikanth, line 64: Is there a reason why we are returning the name for the test in this commit? Should we do that for other tests in this file as well? 🤔

app/src/main/java/org/simple/clinic/sync/DataSync.kt
```diff
@@ -74,6 +74,12 @@ class DataSync @Inject constructor(
         .subscribe()
   }
 
+  fun fireAndForgetSync() {
+    syncTheWorld()
+        .subscribeOn(schedulersProvider.io())
+        .subscribe()
```
msasikanth, line 80: Just a question, previously we had `onErrorComplete` for this call in the `LoginUserWithOtp` class. Why did we not use it over here? 🤔

app/src/test/java/org/simple/clinic/sync/DataSyncTest.kt
vinaysshenoy: This was done because in the event of a failure, we push the sync name and sync events to analytics. The test was throwing an error in the Rx chain because the sync name was returning `null`.

vinaysshenoy: The rest of the tests do not verify failures (iirc), so they do not attempt to read the `name` property.

app/src/main/java/org/simple/clinic/sync/DataSync.kt
vinaysshenoy: The individual syncs (See `sync(SyncGroup)`) themselves handle the errors internally, so this `onErrorComplete` was a redundant call.

app/src/test/java/org/simple/clinic/sync/DataSyncTest.kt
vinaysshenoy: Oh wait, the commit message is correct. We use `RxErrorsRule` because there are cases when an Rx chain can fail but the test continues to pass. This was one of those cases where the test was broken because a stub was missed, but it kept passing because we had missed adding the rule.

msasikanth: Ah, okay.

PR #1375, opened 2020-06-03
Change the button label from 'Add' to 'Next' when assigning a new BP passport to an existing patient.

msasikanth:
https://www.pivotaltracker.com/story/show/173129903


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Next Release
 ### Feature
 - User can soft delete patient
+- Rename the button label to 'Next' when assigning a new BP passport to an existing patient
```
vinaysshenoy, line 6: One thing I am still unsure of is naming this section 'Feature' since this is definitely *not* a feature, but it also is *neither* an internal change nor a bug fix.

Maybe this section should be named "User Facing". Something to think about.

app/src/main/res/layout/link_id_with_patient_view.xml
```diff
@@ -79,7 +79,7 @@
           android:layout_weight="1"
           android:backgroundTint="@color/blue1"
           android:gravity="center"
-          android:text="@string/linkidwithpatient_button_add"
```
vinaysshenoy, line 82: Can we not rename the resource ID? We can just change the string resource, Transifex handles this case. There does not need to be a 1:1 mapping between the name of the resource and the content.

msasikanth: If I don't remove the string from the translations, will the Transifex still handle that case? 🤔

CHANGELOG.md
msasikanth: Yeah, I was unsure about it as well. Although I am not entirely sure about "User Facing", perhaps we can name it as "Changes" and we can copy it to Play Store changelog as well. WDYT?

app/src/main/res/layout/link_id_with_patient_view.xml
vinaysshenoy: Yes, it will be marked as requiring update since the source has changed. The current translations will still be "Next", but that's perfectly fine.

CHANGELOG.md
vinaysshenoy: Yeah, let's do this from the next release onwards.

PR #1376, opened 2020-06-04
Add "Low" label for Blood Sugars

msasikanth:
https://www.pivotaltracker.com/story/show/173163349


vinaysshenoy:
This just fills the icon, right? What about changing the label? Will that be done in another PR?

msasikanth:
> This just fills the icon, right? What about changing the label? Will that be done in another PR?

We don't use labels for blood sugars, I talked to Pragati and she told me to use the filled icon for high and low range and outline icon for the normal value.

PR #1377, opened 2020-06-04
Add "Low" label for Blood Pressures

msasikanth:
https://www.pivotaltracker.com/story/show/173168583

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@
 
 ### Changes
 - Rename the button label to 'Next' when assigning a new BP passport to an existing patient
-- Indicate low blood sugars
+- Add visual indication for low blood sugars
+- Display the "Low" label when a blood pressure is low
 
 ### Internal
 - Change syncs on login to trigger at the usage site instead of in the `LoginUserWithOtp` class
--- a/app/src/main/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItem.kt
+++ b/app/src/main/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItem.kt
@@ -39,20 +39,9 @@ sealed class BloodPressureHistoryListItem : PagingItemAdapter.Item<Event> {
     @SuppressLint("SetTextI18n")
     override fun render(holder: ViewHolderX, subject: Subject<Event>) {
       val context = holder.itemView.context
-      val bpDateTime = if (bpTime != null) {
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItem.kt
```diff
@@ -38,29 +39,22 @@ sealed class BloodPressureHistoryListItem : PagingItemAdapter.Item<Event> {
...
-        context.getString(R.string.bloodpressurehistory_bp_time_date, bpDate, bpTime)
-      } else {
-        bpDate
-      }
-      val dateTimeTextAppearance = if (bpTime != null) {
-        TextAppearanceSpan(context, R.style.Clinic_V2_TextAppearance_Caption_Grey1)
-      } else {
-        TextAppearanceSpan(context, R.style.Clinic_V2_TextAppearance_Body2Left_Grey1)
-      }
-
```
vinaysshenoy, line 51: Something to keep a note of in future is that is ok (and preferable) to address Code Climate issues in a new commit since the previous commit is not actually broken, just could be improved.

Otherwise, this commit is doing two things:
- Refactoring some code
- Showing a different label for low BP

which should not be going together in a single commit.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 ### Changes
 - Rename the button label to 'Next' when assigning a new BP passport to an existing patient
 - Indicate low blood sugars
+- Indicate low blood pressures
```
vinaysshenoy, line 11: Can we reword these changelog items?

```md
- Add visual indication for low blood sugars
- Display the "Low" label when a blood sugar is low
```

app/src/main/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItem.kt
vinaysshenoy: This is not necessary to fix in this current PR.

msasikanth: Ah yes, I wanted to add it as a separate commit but wasn't sure about that. Will do that 👍

msasikanth: Done. Moved it a different commit

CHANGELOG.md
msasikanth: Done

PR #1379, opened 2020-06-05
Fix blood pressure "High" label displaying when blood pressure is low.

msasikanth:
https://www.pivotaltracker.com/story/show/173188293

### DIFF
--- a/app/src/main/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItem.kt
+++ b/app/src/main/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItem.kt
@@ -44,7 +44,6 @@ sealed class BloodPressureHistoryListItem : PagingItemAdapter.Item<Event> {
           .append(bloodPressureDateTime(context))
           .popSpan()
           .build()
-      val labelTextRes = measurement.level.displayTextRes.toNullable()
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/simple/clinic/bp/history/adapter/BloodPressureHistoryListItem.kt
```diff
@@ -44,17 +44,13 @@ sealed class BloodPressureHistoryListItem : PagingItemAdapter.Item<Event> {
...
           .popSpan()
           .build()
-      val labelTextRes = measurement.level.displayTextRes.toNullable()
 
       if (isBpHigh) {
         holder.heartImageView.setImageResource(R.drawable.bp_reading_high)
       } else {
         holder.heartImageView.setImageResource(R.drawable.bp_reading_normal)
       }
       holder.bpHighTextView.visibleOrGone(isBpHigh)
```
vinaysshenoy, line 53: I am assuming that the value set in the layout for this label is "High", correct?

msasikanth: Yep, I only used the `displayTextRes` in order to show the "Low" label text directly instead of checking it in the views.

PR #1380, opened 2020-06-05
Support receiving multiple teleconsultation phone numbers

msasikanth:
https://www.pivotaltracker.com/story/show/173188639

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/teleconsultation/api/TeleconsultationsResponse.kt
+++ b/app/src/main/java/org/simple/clinic/summary/teleconsultation/api/TeleconsultationsResponse.kt
@@ -5,6 +5,10 @@ import com.squareup.moshi.JsonClass
 
 @JsonClass(generateAdapter = true)
 data class TeleconsultationsResponse(
+    @Deprecated(message = "Use teleconsultationPhoneNumbers to get list of all MO phone numbers")
     @Json(name = "teleconsultation_phone_number")
-    val teleconsultationPhoneNumber: String?
+    val teleconsultationPhoneNumber: String?,
+
+    @Json(name = "teleconsultation_phone_numbers")
+    val teleconsultationPhoneNumbers: List<TeleconsultPhoneNumber>?
 )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Internal
 - Change syncs on login to trigger at the usage site instead of in the `LoginUserWithOtp` class
 - [In Progress: 01 Jun 2020] Introduce a `UUID` generator class to ease testing record creation
+- Support receiving multiple teleconsultation phone numbers
 
 ### Fixes
 - Fix blood pressure "High" label displaying when blood pressure is low.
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/simple/clinic/summary/teleconsultation/api/TeleconsultationsResponse.kt
```diff
@@ -5,6 +5,12 @@ import com.squareup.moshi.JsonClass
...
 @JsonClass(generateAdapter = true)
 data class TeleconsultationsResponse(
+    @Deprecated(
+        message = "Use teleconsultationPhoneNumbers to get list of all MO phone numbers",
+        replaceWith = ReplaceWith("teleconsultationPhoneNumbers", "org.simple.clinic.summary.teleconsultation.api")
+    )
     @Json(name = "teleconsultation_phone_number")
-    val teleconsultationPhoneNumber: String?
+    val teleconsultationPhoneNumber: String?,
+    @Json(name = "teleconsultation_phone_numbers")
```
vinaysshenoy, line 14: Can we add a line break here to separate the two properties and make it more readable?

app/src/main/java/org/simple/clinic/summary/teleconsultation/api/TeleconsultationsResponse.kt
```diff
@@ -5,6 +5,12 @@ import com.squareup.moshi.JsonClass
...
+    val teleconsultationPhoneNumbers: List<TeleconsultPhoneNumber>?
```
vinaysshenoy, line 15: Is the API spec shared somewhere? I could neither see it on the PRD nor on the API docs.

app/src/main/java/org/simple/clinic/summary/teleconsultation/api/TeleconsultationsResponse.kt
```diff
@@ -5,6 +5,12 @@ import com.squareup.moshi.JsonClass
 
 @JsonClass(generateAdapter = true)
 data class TeleconsultationsResponse(
+    @Deprecated(
+        message = "Use teleconsultationPhoneNumbers to get list of all MO phone numbers",
+        replaceWith = ReplaceWith("teleconsultationPhoneNumbers", "org.simple.clinic.summary.teleconsultation.api")
```
vinaysshenoy, line 10: This is incorrect usage of `ReplaceWith` since it should not cause compilation errors/failing tests after running. If we don't have a replacement that does not cause a compilation error, we should avoid using it and let the developer decide how to migrate a deprecated usage.

app/src/main/java/org/simple/clinic/summary/teleconsultation/api/TeleconsultationsResponse.kt
msasikanth: Yeah, it's been shared on slack. 
https://simpledotorg.slack.com/archives/C012M349Q4X/p1591261415030900

app/src/main/java/org/simple/clinic/summary/teleconsultation/api/TeleconsultationsResponse.kt
msasikanth: Ah, okay I used it to make easier to replace at call sites. I will remove it.

app/src/main/java/org/simple/clinic/summary/teleconsultation/api/TeleconsultationsResponse.kt
msasikanth: Done

PR #1388, opened 2020-06-08
Add Diagnosis and Blood Sugar readings to the WhatsApp message

msasikanth:
https://www.pivotaltracker.com/story/show/173195647


msasikanth:
Pushing another commit to fix the code climate issue

app/src/test/java/org/simple/clinic/summary/PatientSummaryEffectHandlerTest.kt
```diff
@@ -327,18 +328,28 @@ class PatientSummaryEffectHandlerTest {
...
         bloodSugars = bloodSugars,
-        prescriptions = prescriptions
+        prescriptions = prescriptions,
+        medicalHistory = medicalHistory
     )
 
     whenever(bloodPressureRepository.newestMeasurementsForPatientImmediate(patientUuid, patientSummaryConfig.numberOfMeasurementsForTeleconsultation)) doReturn bloodPressures
     whenever(prescriptionRepository.newestPrescriptionsForPatientImmediate(patientUuid)) doReturn prescriptions
     whenever(bloodSugarRepository.latestMeasurementsImmediate(patientUuid, patientSummaryConfig.numberOfMeasurementsForTeleconsultation)) doReturn bloodSugars
+    whenever(medicalHistoryRepository.historyForPatientOrDefaultImmediate(uuidGenerator.v4(), patientUuid)) doReturn medicalHistory
```
vinaysshenoy, line 352: The first parameter here should be the `medicalHistoryUuid`, right? It is an implementation detail that it will be the same in the test, but the method stub should be explicit.

app/src/main/res/values/strings.xml
```diff
@@ -206,6 +206,11 @@
   <string name="patientsummary_bangladesh_national_id">ID</string>
 
   <string name="patientsummary_contact_doctor">Doctor</string>
+  <string name="patientsummary_contact_doctor_diagnosis" translatable="false">*Diagnosis*</string>
+  <string name="patientsummary_contact_doctor_diagnosis_hypertension" translatable="false">Hypertension: </string>
+  <string name="patientsummary_contact_doctor_diagnosis_diabetes" translatable="false">Diabetes: </string>
```
vinaysshenoy, line 211: iirc, the Android build tools will remove blank spaces at the end of a message. Can we verify this? Might want to use the escape character instead.

app/src/main/java/org/simple/clinic/summary/teleconsultation/TeleconsultationMessageBuilder.kt
```diff
@@ -34,37 +34,36 @@ class TeleconsultationMessageBuilder @Inject constructor(
         .appendln("https://app.simple.org/patient/${patientTeleconsultationInfo.patientUuid}")
         .appendln("")
 
-    message.appendln(resources.getString(R.string.patientsummary_contact_doctor_diagnosis))
+    diagnosis(patientTeleconsultationInfo, message)
+    bloodPressures(patientTeleconsultationInfo, message)
+    bloodSugars(patientTeleconsultationInfo, message)
+    prescriptions(patientTeleconsultationInfo, message)
```
vinaysshenoy, line 40: The names of these methods are not too descriptive since they do not expose the detail that the `message` parameter is going to be modified by them. Can we rename the methods so this detail is exposed?

Ex: `addDiagnosisSectionToMessage(patientTeleconsultationInfo, message)`.

app/src/main/java/org/simple/clinic/bloodsugar/BloodSugarRepository.kt
```diff
@@ -60,6 +60,10 @@ class BloodSugarRepository @Inject constructor(
     return dao.latestMeasurements(patientUuid, limit)
   }
 
+  fun latestMeasurementsImmediate(patientUuid: UUID, limit: Int): List<BloodSugarMeasurement> {
```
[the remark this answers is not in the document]
msasikanth: The implementation is the same as `latestMeasurements`. Didn't talk about not having separate tests for blocking calls with same implementation? 🤔

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
```diff
@@ -256,13 +256,15 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
           .map {
             val bloodPressures = bloodPressureRepository.newestMeasurementsForPatientImmediate(it.patientUuid, patientSummaryConfig.numberOfMeasurementsForTeleconsultation)
             val prescriptions = prescriptionRepository.newestPrescriptionsForPatientImmediate(it.patientUuid)
+            val bloodSugars = bloodSugarRepository.latestMeasurementsImmediate(it.patientUuid, patientSummaryConfig.numberOfMeasurementsForTeleconsultation)
```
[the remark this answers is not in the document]
msasikanth: Done

app/src/main/java/org/simple/clinic/summary/teleconsultation/TeleconsultationMessageBuilder.kt
```diff
@@ -53,4 +82,22 @@ class TeleconsultationMessageBuilder @Inject constructor(
...
+    }
+  }
+
+  private fun unitForBloodSugarType(type: BloodSugarMeasurementType): String {
+    return when (type) {
+      Random, PostPrandial, Fasting -> resources.getString(R.string.patientsummary_contact_doctor_unit_type_mg_dl)
+      HbA1c -> resources.getString(R.string.patientsummary_contact_doctor_unit_type_percentage)
+      is Unknown -> ""
+    }
+  }
```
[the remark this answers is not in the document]
msasikanth: I thought about it but was unsure about moving them into the `BloodSugarReading` class since we have to pass in context or resources. WDYT?

app/src/test/java/org/simple/clinic/summary/PatientSummaryEffectHandlerTest.kt
msasikanth: Ah right, forgot about that. 😅

app/src/main/res/values/strings.xml
msasikanth: I think I can remove those spaces since I am adding a new line when using `StringBuilder`, so there is no point having those spaces will remove those.

msasikanth: I am adding the space when building the message, so I can remove the spaces from the strings. Since we are planning on having hardcoded strings, either way, it makes sense I guess.

app/src/main/java/org/simple/clinic/summary/teleconsultation/TeleconsultationMessageBuilder.kt
msasikanth: Done

app/src/main/java/org/simple/clinic/summary/teleconsultation/TeleconsultationMessageBuilder.kt
```diff
@@ -33,6 +34,22 @@ class TeleconsultationMessageBuilder @Inject constructor(
...
+    message.appendln(resources.getString(R.string.patientsummary_contact_doctor_diagnosis))
+
+    val hyperTensionTitle = resources.getString(R.string.patientsummary_contact_doctor_diagnosis_hypertension)
+    val diagnosedWithHypertension = patientTeleconsultationInfo.medicalHistory.diagnosedWithHypertension
+    if (diagnosedWithHypertension.isAnswered) {
+      message.appendln("$hyperTensionTitle ${textForDiagnosisAnswer(diagnosedWithHypertension)}")
+    }
+
+    val diabetesTitle = resources.getString(R.string.patientsummary_contact_doctor_diagnosis_diabetes)
+    val diagnosedWithDiabetes = patientTeleconsultationInfo.medicalHistory.diagnosedWithHypertension
```
vinaysshenoy: A bug we found while reviewing: this should be using `diagnosedWithDiabetes` and not `diagnosedWithHypertension`.

msasikanth: Oh yeah, my bad. Fixed it 👍

app/src/main/java/org/simple/clinic/bloodsugar/BloodSugarRepository.kt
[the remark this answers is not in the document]
vinaysshenoy: Ah, if there already is a test for that, let's skip it for now.

app/src/main/java/org/simple/clinic/summary/teleconsultation/TeleconsultationMessageBuilder.kt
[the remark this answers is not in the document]
vinaysshenoy: We don't need to do that and can just return the resource ID from the type.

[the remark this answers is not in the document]
vinaysshenoy: Anyway, this is not a change that needs to be done in this PR.

vinaysshenoy:
> PR looks good overall. Left a few comments.
> 
> One thing which is a little annoying now is that `TelconsultationMessageBuilder` has a mixture of hardcoded strings and string resources. We should probably fix this to be consistent.
> 
> Since we do not want these messages to be localised, we could probably just hardcode the strings in the class and leave a comment in the class documentation that these are intentionally not part of string resources.
> 
> It is not necessary to make this change in this PR, we can create a ticket and get to it later.

Can we create a ticket for this? PR looks good, I am merging it in.

app/src/main/java/org/simple/clinic/summary/teleconsultation/TeleconsultationMessageBuilder.kt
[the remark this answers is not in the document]
msasikanth: I will make this change in the hardcoded strings for teleconsultation message PR.

msasikanth:
> > PR looks good overall. Left a few comments.
> > One thing which is a little annoying now is that `TelconsultationMessageBuilder` has a mixture of hardcoded strings and string resources. We should probably fix this to be consistent.
> > Since we do not want these messages to be localised, we could probably just hardcode the strings in the class and leave a comment in the class documentation that these are intentionally not part of string resources.
> > It is not necessary to make this change in this PR, we can create a ticket and get to it later.
> 
> Can we create a ticket for this? PR looks good, I am merging it in.

Done, tracking [here](https://www.pivotaltracker.com/story/show/173242299)

PR #1396, opened 2020-06-10
Show `ContactDoctorSheet` when there are multiple teleconsultation phone numbers #2

msasikanth:
https://www.pivotaltracker.com/story/show/173240199


app/src/test/java/org/simple/clinic/summary/PatientSummaryUpdateTest.kt
```diff
@@ -631,7 +631,7 @@ class PatientSummaryUpdateTest {
   }
 
   @Test
-  fun `when contact doctor button is clicked, then load patient teleconsultation information`() {
+  fun `when contact doctor button is clicked and only one phone number is present, then load patient teleconsultation information`() {
```
vinaysshenoy, line 634: Instead of describing the implementation detail only one phone number is present, can we reword this to reflect the real world usage? Ex: `only one doctor is available for teleconsultation`.

app/src/test/java/org/simple/clinic/summary/PatientSummaryUpdateTest.kt
```diff
@@ -655,6 +655,27 @@ class PatientSummaryUpdateTest {
         )
   }
 
+  @Test
+  fun `when contact doctor button is clicked and more than one phone number is present, then open contact doctor sheet`() {
```
vinaysshenoy, line 659: Instead of describing the implementation detail only one phone number is present, can we reword this to reflect the real world usage? Ex: `multiple doctors are available for teleconsultation`.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -135,6 +137,7 @@ class PatientSummaryScreen(
         .setActionTextColor(ContextCompat.getColor(context, R.color.green2))
         .setAnchorView(R.id.doneButtonFrame)
   }
+  private val detaches = detaches()
```
vinaysshenoy, line 140: Why was this extracted? It seems like it is not being used anywhere else.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -172,6 +175,20 @@ class PatientSummaryScreen(
     super.onDetachedFromWindow()
   }
 
+  private fun teleconsultPhoneNumberSelected(): Observable<PatientSummaryEvent> {
+    return screenRouter
+        .streamScreenResults()
+        .ofType<ActivityResult>()
+        .filter { it.requestCode == CONTACT_DOCTOR_SHEET && it.succeeded() && it.data != null }
+        .map { it.data!! }
```
vinaysshenoy, line 183: We have this exact same codeblock everywhere where the only thing that changes is the type of the result data and the result code. Maybe we should think of moving this to an extension function.

```kotlin
inline fun <reified T> Observable<ActivityResult>.extractSuccessful(
    requestCode: Int,
    crossinline resultMapper: (Intent) -> T
): Observable<T> {
  return filter { it.requestCode == requestCode && it.succeeded() && it.data != null }
      .map { resultMapper(it.data!!) }
}
```

The usage then becomes a lot easier to read and developers can focus on the business logic instead of this boilerplate.

```kotlin
screenRouter
        .streamScreenResults()
        .ofType<ActivityResult>()
        .extractSuccessful(CONTACT_DOCTOR_SHEET, ContactDoctorSheet.Companion::readPhoneNumberExtra)
        .map { teleconsultPhoneNumberString ->
          val teleconsultPhoneNumber = TeleconsultPhoneNumber(teleconsultPhoneNumberString)

          ContactDoctorPhoneNumberSelected(teleconsultPhoneNumber)
        }
```

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -425,7 +442,11 @@ class PatientSummaryScreen(
       facility: Facility,
       phoneNumbers: List<TeleconsultPhoneNumber>
   ) {
-    // TODO (SM): Open contact doctor sheet
+    val phoneNumbersArrayList = arrayListOf<TeleconsultPhoneNumber>().apply {
+      addAll(phoneNumbers)
+    }
```
vinaysshenoy, line 447: Why are we creating a new list here of the exact same thing? We can pass the `phoneNumbers` list directly to the intent factory method, right?

CHANGELOG.md
```diff
@@ -4,8 +4,8 @@
 ### Feature
 - User can soft delete patient
 - Add diagnosis and blood sugar readings to the teleconsultation message
+- Support multiple teleconsultation phone numbers
```
vinaysshenoy, line 7: Can we reword this to better describe a user facing feature?

Ex: "User can now select from multiple doctors when requesting a teleconsultation".

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
```diff
@@ -99,20 +99,32 @@ class PatientSummaryUpdate : Update<PatientSummaryModel, PatientSummaryEvent, Pa
...
+          }
+          else -> {
+            LoadPatientTeleconsultationInfo(
+                model.patientUuid,
+                model.patientSummaryProfile?.bpPassport,
+                model.currentFacility,
+                teleconsultInfo.doctorsPhoneNumbers.first()
+            )
+          }
+        }
```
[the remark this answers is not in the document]
msasikanth: I mainly did that for readability, but once I extract the boolean logic I can switch to an if/else condition.

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
```diff
@@ -99,20 +99,32 @@ class PatientSummaryUpdate : Update<PatientSummaryModel, PatientSummaryEvent, Pa
...
-      is TeleconsultInfo.Fetched -> dispatch(LoadPatientTeleconsultationInfo(
-          model.patientUuid,
-          model.patientSummaryProfile?.bpPassport,
-          model.currentFacility,
-          model.teleconsultInfo.doctorsPhoneNumbers.first()
-      ) as PatientSummaryEffect)
+    return when (val teleconsultInfo = model.teleconsultInfo) {
+      is TeleconsultInfo.Fetched -> {
+        val effect = when {
+          teleconsultInfo.doctorsPhoneNumbers.size > 1 -> {
```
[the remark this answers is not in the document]
msasikanth: Yeah, since we know this won't be triggered when there are no phone numbers we will have either 1 or more numbers states.

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
```diff
@@ -99,20 +99,32 @@ class PatientSummaryUpdate : Update<PatientSummaryModel, PatientSummaryEvent, Pa
...
+            OpenContactDoctorSheet(teleconsultInfo.doctorsPhoneNumbers)
```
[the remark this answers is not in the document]
msasikanth: Yeah, makes sense 👍

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: No, intent takes in an `ArrayList` and doesn't accept the Kotlin `List`.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: Should this be part of this PR?

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
vinaysshenoy, line 106: One more thing I forgot to add: this commit breaks the app. The reason being that the method that opens the contact patient sheet is a no-op on the UI so clicking the button when there are no doctors available will result in a button-click that does nothing.

We should squash the first three commits together so that this change goes in as a single commit.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: Oh, I forgot to add `takeUntil` for the `streamScreenResults` 😅

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: Then this is a leaky abstraction since the factory method is exposing this detail. We should change the factory method to expect a `List<>` and then convert it to an array list inside that method.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: Not necessary.

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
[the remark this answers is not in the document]
msasikanth: Done

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: Yeah, changed it

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
msasikanth: I have squashed the commits, but when no doctors are available the button should go to disable state. So the user won't be able to press the button.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: Can we create a ticket for this?

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
vinaysshenoy: Sorry, I meant when more than one doctor is available.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: Done: https://www.pivotaltracker.com/story/show/173282881

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: This is not necessary since Mobius disposes the streams when the screen gets detached.

vinaysshenoy: We need to manually dispose streams only when they are not part of a Mobius loop.

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
```diff
@@ -99,20 +99,39 @@ class PatientSummaryUpdate : Update<PatientSummaryModel, PatientSummaryEvent, Pa
...
+      }
+      else -> noChange()
+    }
+  }
+
+  private fun eventForContactDoctorClicked(
+      teleconsultInfo: TeleconsultInfo.Fetched,
+      model: PatientSummaryModel
+  ): PatientSummaryEffect {
+    return if (areMultipleDoctorsAvailable(teleconsultInfo)) {
```
vinaysshenoy, line 115: I don't really think this needs to be a function since it is only used in one place. Maybe we can convert it to a computed property on the `TeleconsultInfo.Fetched` class instead?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: Ah okay 👍

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
msasikanth: Ah right, forgot about that. Done

PR #1397, opened 2020-06-11
Create `ProgressMaterialButton`

### LINKED ISSUE #1374: `ProgressMaterialButton`
# Problem

At times we need to show the progress bar for the `MaterialButton` before enable/disable it. In order to do that right now, we are manually triggering the state changes in the screen we need this behaviour. 

It would be better we if can extract these things into its own `View`, which will improve the reusability of these kinds of behaviours. The reason why I didn't went with a layout file and compound view approach is that we have to handle theming all the attrs for `MaterialButton`.

# Proposal

We have to use a custom view, We cannot directly place the `ProgressBar` inside the `MaterialButton` because it's just a `TextView`. So we create custom `FrameLayout` that will add `MaterialButton` & `ProgressBar` internally.

```kotlin
class ProgressMaterialButton @JvmOverloads constructor(
  context: Context,
  attrs: AttributeSet? = null,
  defStyleAttr: Int = 0
) : FrameLayout(context, attrs, defStyleAttr) {

  enum class ButtonState {
    InProgress, Enabled, Disabled
  }

  companion object {
    private val BUTTON_STYLES = arrayOf(
      R.attr.materialButtonStyle,
      R.attr.materialButtonOutlinedStyle
    )
  }

  private var buttonState = ButtonState.InProgress
  private var buttonText: String? = null
  private var buttonIcon: Drawable? = null

  private val materialButton: MaterialButton
  private val progress: ProgressBar

  init {
    val typedArray = context.obtainStyledAttributes(attrs, R.styleable.ProgressMaterialButton)

    buttonText = typedArray.getString(R.styleable.ProgressMaterialButton_android_text)
    buttonIcon = typedArray.getDrawable(R.styleable.ProgressMaterialButton_icon)
    val materialButtonStyleAttr = BUTTON_STYLES[typedArray.getInt(R.styleable.ProgressMaterialButton_progressButtonStyle, 0)]
    val progressColor = typedArray.getColor(R.styleable.ProgressMaterialButton_progressColor, 0)

    materialButton = createButton(context, attrs, materialButtonStyleAttr)
    progress = createProgress(context, attrs).apply {
      isIndeterminate = true

      // We are setting the translationZ to set the view on top of the material button,
      // since material button has elevation with out the translationZ of progress bar it
      // will be behind it.
      translationZ = 8.dp.toFloat()

      indeterminateTintList = ColorStateList.valueOf(progressColor)
    }

    addView(materialButton)
    addView(progress)

    setButtonState(buttonState)

    progress.updateLayoutParams<LayoutParams> {
      height = 32.dp
      gravity = Gravity.CENTER
    }

    typedArray.recycle()

    clipToPadding = false
  }

  override fun setOnClickListener(l: OnClickListener?) {
    @Suppress("LABEL_NAME_CLASH")
    materialButton.setOnClickListener {
      if (buttonState != ButtonState.Enabled) {
        return@setOnClickListener
      }
      l?.onClick(this)
    }
  }

  private fun createButton(context: Context, attrs: AttributeSet?, @AttrRes materialButtonStyleAttr: Int): MaterialButton {
    return MaterialButton(context, attrs, materialButtonStyleAttr)
  }

  private fun createProgress(context: Context, attrs: AttributeSet?): ProgressBar {
    return ProgressBar(context, attrs)
  }

  fun setButtonState(buttonState: ButtonState) {
    when (buttonState) {
      ButtonState.InProgress -> {
        progress.visibility = View.VISIBLE
        materialButton.icon = null
        materialButton.text = null
        materialButton.isEnabled = true
      }
      ButtonState.Enabled -> {
        progress.visibility = View.GONE
        materialButton.icon = buttonIcon
        materialButton.text = buttonText
        materialButton.isEnabled = true
      }
      ButtonState.Disabled -> {
        progress.visibility = View.GONE
        materialButton.icon = buttonIcon
        materialButton.text = buttonText
        materialButton.isEnabled = false
      }
    }
    this.buttonState = buttonState
  }
}
```

## Problems with this approach
- We have to copy some/all attrs from the `MaterialButton` to our `attrs.xml` file in order to let the auto-complete work. _(We have to copy attrs for `ProgressBar` if we need the auto-complete for it as well)_
- Applying style to `MaterialButton` seems to be problematic, we cannot directly apply the style to the custom `View` because it would apply it for the underlying `FrameLayout` & `ProgressBar` as well. Adding a custom attr to get the theme for `MaterialButton` and using `ContextThemeWrapper` when creating the `MaterialButton` doesn't seem to work as well. So it's something we need to figure out or keep in mind.

msasikanth:
https://www.pivotaltracker.com/story/show/172854764

closes #1374

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenUi.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenUi.kt
@@ -5,10 +5,9 @@ interface PatientSummaryScreenUi {
   fun showEditButton()
   fun showDiabetesView()
   fun hideDiabetesView()
-  fun showContactDoctorButtonTextAndIcon()
-  fun showContactButtonProgress()
-  fun enableContactDoctorButton()
-  fun disableContactDoctorButton()
   fun showContactDoctorButton()
   fun hideContactDoctorButton()
+  fun enableContactDoctorButton()
+  fun disableContactDoctorButton()
+  fun fetchingTeleconsultInfo()
 }
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryViewRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryViewRenderer.kt
@@ -23,24 +23,18 @@ class PatientSummaryViewRenderer(
 
   private fun setupUiForTeleconsult(model: PatientSummaryModel) {
     if (model.isTeleconsultationEnabled && model.isUserLoggedIn) {
-      ui.showContactDoctorButton()
-
-      when (model.teleconsultInfo) {
-        is TeleconsultInfo.Fetched -> {
-          ui.showContactDoctorButtonTextAndIcon()
-          ui.enableContactDoctorButton()
-        }
-        is TeleconsultInfo.MissingPhoneNumber, is TeleconsultInfo.NetworkError -> {
-          ui.showContactDoctorButtonTextAndIcon()
-          ui.disableContactDoctorButton()
-        }
-        is TeleconsultInfo.Fetching -> {
-          ui.showContactButtonProgress()
-          ui.enableContactDoctorButton()
-        }
-      }
+      renderContactDoctorButton(model)
     } else {
       ui.hideContactDoctorButton()
+    }
+  }
+
+  private fun renderContactDoctorButton(model: PatientSummaryModel) {
+    ui.showContactDoctorButton()
+    when (model.teleconsultInfo) {
+      is TeleconsultInfo.Fetched -> ui.enableContactDoctorButton()
+      is TeleconsultInfo.MissingPhoneNumber, is TeleconsultInfo.NetworkError -> ui.disableContactDoctorButton()
+      is TeleconsultInfo.Fetching -> ui.fetchingTeleconsultInfo()
     }
   }
 
--- a/app/src/main/java/org/simple/clinic/widgets/ProgressMaterialButton.kt
+++ b/app/src/main/java/org/simple/clinic/widgets/ProgressMaterialButton.kt
@@ -1,0 +1,85 @@
+package org.simple.clinic.widgets
+
+import android.content.Context
+import android.graphics.drawable.Drawable
+import android.util.AttributeSet
+import androidx.vectordrawable.graphics.drawable.Animatable2Compat
+import androidx.vectordrawable.graphics.drawable.AnimatedVectorDrawableCompat
+import com.google.android.material.button.MaterialButton
+import org.simple.clinic.R
+
+class ProgressMaterialButton(
+    context: Context,
+    attrs: AttributeSet? = null
+) : MaterialButton(context, attrs) {
+
+  enum class ButtonState {
+    InProgress, Enabled, Disabled
+  }
+
+  private var buttonState: ButtonState
+  private var buttonText: String? = null
+  private var buttonIcon: Drawable? = null
+  private var buttonIconGravity: Int
+
+  private var progressDrawable = AnimatedVectorDrawableCompat.create(context, R.drawable.avd_progress)
+
+  private val progressAvdCallback = object : Animatable2Compat.AnimationCallback() {
+    override fun onAnimationEnd(drawable: Drawable?) {
+      super.onAnimationEnd(drawable)
+      progressDrawable?.start()
+    }
+  }
+
+  init {
+    buttonText = text.toString()
+    buttonIcon = icon
+    buttonIconGravity = iconGravity
+
+    val typeArray = context.obtainStyledAttributes(attrs, R.styleable.ProgressMaterialButton)
+    buttonState = ButtonState.values()[typeArray.getInt(R.styleable.ProgressMaterialButton_buttonState, 0)]
+
+    setButtonState(buttonState)
+
+    typeArray.recycle()
+  }
+
+  override fun onAttachedToWindow() {
+    super.onAttachedToWindow()
+    progressDrawable?.run {
+      registerAnimationCallback(progressAvdCallback)
+      start()
+    }
+  }
+
+  override fun onDetachedFromWindow() {
+    super.onDetachedFromWindow()
+    progressDrawable?.unregisterAnimationCallback(progressAvdCallback)
+  }
+
+  fun setButtonState(buttonState: ButtonState) {
+    when (buttonState) {
+      ButtonState.InProgress -> {
+        isEnabled = true
+        isClickable = false
+        icon = progressDrawable
+        text = null
+        iconGravity = ICON_GRAVITY_TEXT_START
+      }
+      ButtonState.Enabled -> {
+        isEnabled = true
+        isClickable = true
+        icon = buttonIcon
+        text = buttonText
+        iconGravity = buttonIconGravity
+      }
+      ButtonState.Disabled -> {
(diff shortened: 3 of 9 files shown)


msasikanth:
I have tested it in the app and the behaviour is the same as the old one. We can probably ignore the code climate issues WDYT?

vinaysshenoy:
> I have tested it in the app and the behaviour is the same as the old one. We can probably ignore the code climate issues WDYT?

The only reasons I'd be ok with skipping fixing code climate issues are the following:
- We are going to fix them in an immediate upcoming PR which is scoped and scheduled to be picked up/worked on.
- The code is not ours, and is something we copied and dropped into the codebase (like `OmegaCenterIconButton` was).
- The code is going to be removed/changed as part of a future PR.
- Fixing them will take an inordinate amount of time (in which case, we should probably create a ticket and track it).

If it's not falling into any of these cases, we *should* fix Code Climate issues.

Are any other cases where you feel we can skip fixing these issues? We can add them to the list.

msasikanth:
> > I have tested it in the app and the behaviour is the same as the old one. We can probably ignore the code climate issues WDYT?
> 
> The only reasons I'd be ok with skipping fixing code climate issues are the following:
> 
> * We are going to fix them in an immediate upcoming PR which is scoped and scheduled to be picked up/worked on.
> * The code is not ours, and is something we copied and dropped into the codebase (like `OmegaCenterIconButton` was).
> * The code is going to be removed/changed as part of a future PR.
> * Fixing them will take an inordinate amount of time (in which case, we should probably create a ticket and track it).
> 
> If it's not falling into any of these cases, we _should_ fix Code Climate issues.
> 
> Are any other cases where you feel we can skip fixing these issues? We can add them to the list.

The reason why I suggest ignoring code climate in this PR was because of deep nesting. Code climate was complaining about null check inside a `when` block, but I can replace it with a different implementation. 👍

vinaysshenoy:
> > > I have tested it in the app and the behaviour is the same as the old one. We can probably ignore the code climate issues WDYT?
> > 
> > 
> > The only reasons I'd be ok with skipping fixing code climate issues are the following:
> > 
> > * We are going to fix them in an immediate upcoming PR which is scoped and scheduled to be picked up/worked on.
> > * The code is not ours, and is something we copied and dropped into the codebase (like `OmegaCenterIconButton` was).
> > * The code is going to be removed/changed as part of a future PR.
> > * Fixing them will take an inordinate amount of time (in which case, we should probably create a ticket and track it).
> > 
> > If it's not falling into any of these cases, we _should_ fix Code Climate issues.
> > Are any other cases where you feel we can skip fixing these issues? We can add them to the list.
> 
> The reason why I suggest ignoring code climate in this PR was because of deep nesting. Code climate was complaining about null check inside a `when` block, but I can replace it with a different implementation. 👍

Yes please, the point of adding these checks is to force us to make smaller functions and separate related bits of code out. This will help have a more maintainable codebase in the long run.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenUi.kt
```diff
@@ -5,10 +5,9 @@ interface PatientSummaryScreenUi {
...
   fun hideDiabetesView()
-  fun showContactDoctorButtonTextAndIcon()
-  fun showContactButtonProgress()
-  fun enableContactDoctorButton()
-  fun disableContactDoctorButton()
   fun showContactDoctorButton()
   fun hideContactDoctorButton()
+  fun contactDoctorButtonEnabled()
+  fun contactDoctorButtonDisabled()
+  fun contactDoctorButtonInProgress()
```
vinaysshenoy, line 12: Is this name change necessary? The older names seem more appropriate to me since it is still up to the UI to change the button state.

app/src/main/java/org/simple/clinic/summary/PatientSummaryViewRenderer.kt
```diff
@@ -23,27 +23,27 @@ class PatientSummaryViewRenderer(
...
+    when (model.teleconsultInfo) {
+      is TeleconsultInfo.Fetched -> {
+        ui.contactDoctorButtonEnabled()
+      }
+      is TeleconsultInfo.MissingPhoneNumber, is TeleconsultInfo.NetworkError -> {
+        ui.contactDoctorButtonDisabled()
+      }
+      is TeleconsultInfo.Fetching -> {
+        ui.contactDoctorButtonInProgress()
+      }
```
vinaysshenoy, line 43: We can remove the braces on the `when` clause patterns now.

app/src/main/java/org/simple/clinic/widgets/ProgressMaterialButton.kt
```diff
@@ -0,0 +1,80 @@
...
+class ProgressMaterialButton(
+    context: Context,
+    attrs: AttributeSet? = null
+) : MaterialButton(context, attrs) {
+
+  enum class ButtonState {
+    InProgress, Enabled, Disabled
+  }
+
+  private var buttonState = ButtonState.InProgress
```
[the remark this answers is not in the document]
msasikanth: Yeah, that's a good idea. I will add an attr for setting state from xml

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenUi.kt
msasikanth: Previously we had multiple methods from showing/hiding button, showing/hiding text and icon or progress bar and enabling/disabling the button itself. Since we no longer have to handle all of those cases separately I changed the method names to represent the `ProgressMaterialButton` states.

app/src/main/java/org/simple/clinic/summary/PatientSummaryViewRenderer.kt
msasikanth: Done

app/src/main/java/org/simple/clinic/widgets/ProgressMaterialButton.kt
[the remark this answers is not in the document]
msasikanth: Done

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreenUi.kt
vinaysshenoy: Right, I think my thoughts were more along the lines of the phrasing of the method names.

Most methods fall into three categories:
1. Querying something for information
2. Transforming data from one form to another
3. Changing the state in some way

Which category a method falls into should be evident from reading the name. Here, the methods seem to fall into category 3 since they are changing the state of the UI, but the name does not make it clear. In fact, from reading the name, you cannot tell what category the method falls into at all.

msasikanth: Ah right. I will rename the methods to better reflect that.

msasikanth: Renamed, hopefully these makes sense. wdyt?
```kt
fun enableContactDoctorButton()
fun disableContactDoctorButton()
fun fetchingTeleconsultInfo()
```

vinaysshenoy: These look good, let's go ahead.

PR #1402, opened 2020-06-12
Migrate registration phone entry to Mobius: sterilize screen

vinaysshenoy:
https://www.pivotaltracker.com/story/show/173144564


app/src/test/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinScreenControllerTest.kt
```diff
@@ -47,7 +46,6 @@ class RegistrationConfirmPinScreenControllerTest {
   @Test
   fun `when 4 digits are entered then the PIN should be submitted automatically`() {
     whenever(userSession.ongoingRegistrationEntry()).thenReturn(Single.just(OngoingRegistrationEntry(pin = "1234")))
-    whenever(userSession.saveOngoingRegistrationEntry(any())).thenReturn(Completable.complete())
```
msasikanth, line 50: Should we add `verify` statements to verify that method is being called instead?

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenController.kt
```diff
@@ -213,11 +214,14 @@ class RegistrationFacilitySelectionScreenController @Inject constructor(
...
         .ofType<RegistrationFacilityConfirmed>()
         .map { it.facilityUuid }
-        .flatMap { facilityUuid ->
-          userSession.ongoingRegistrationEntry()
-              .map { it.copy(facilityId = facilityUuid) }
-              .doOnSuccess(userSession::saveOngoingRegistrationEntry)
-              .flatMapCompletable { userSession.saveOngoingRegistrationEntryAsUser() }
+        .map { facilityUuid ->
+          val entry = (userSession.ongoingRegistrationEntry() as Just).value
+          entry.copy(facilityId = facilityUuid)
```
msasikanth, line 219: Should we provide a method to change this from `OngoingRegistrationEntry` instead of using `copy` directly? WDYT?

app/src/test/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinScreenControllerTest.kt
vinaysshenoy: Good catch, I'll add it!

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreenController.kt
vinaysshenoy: It's a good idea, I'll do this when I migrate `RegistrationFacilitySelectionScreenController` to Mobius.

app/src/test/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinScreenControllerTest.kt
vinaysshenoy: I just realised that most of the tests already had the assertion, it just was not visible in the diff on GitHub. One test was missing it, so I added it.

PR #1403, opened 2020-06-15
Fix bug where country ISD code would not be displayed when signing in

vinaysshenoy:
https://www.pivotaltracker.com/story/show/173326632

### DIFF
--- a/app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreen.kt
+++ b/app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreen.kt
@@ -81,6 +81,7 @@ class RegistrationPhoneScreen(
     }
     context.injector<Injector>().inject(this)
 
+    isdCodeEditText.setText(resources.getString(R.string.registrationphone_country_code, country.isdCode))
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreen.kt
```diff
@@ -81,6 +81,7 @@ class RegistrationPhoneScreen(
     }
     context.injector<Injector>().inject(this)
 
+    isdCodeEditText.setText(resources.getString(R.string.registrationphone_country_code, country.isdCode))
```
msasikanth, line 84: I have a doubt regarding this, aren't we sending pre-fill fields effect in `Init` won't that trigger the `preFillUserDetails`, why did it not work?

vinaysshenoy: The issue was that in the implementation, the `preFillUserDetails` method would be called only if a registration entry existed, not when it was being created for the first time. On the ISD code would get pre-filled on screen restoration, but not creation.

Since the code is part of the country and will not change with the model, I moved it out.

PR #1404, opened 2020-06-15
Redesign patient card in home screen

msasikanth:
https://www.pivotaltracker.com/story/show/172931883

~~I have used `Patient#createdAt` to check if the patient registration is a new registration, I have avoided using `recordedAt` since it is updated every time a new BP, blood sugar is added, so wasn't sure if that is correct behaviour.~~ https://github.com/simpledotorg/simple-android/pull/1404#discussion_r439976772


msasikanth:
Marking as WIP to fix the broken test

app/src/main/java/org/simple/clinic/patient/RecentPatient.kt
```diff
@@ -32,7 +34,7 @@ data class RecentPatient(
     companion object {
 
       const val RECENT_PATIENT_QUERY = """
-        SELECT P.uuid, P.fullName, P.gender, P.dateOfBirth, P.age_value, P.age_updatedAt,
+        SELECT P.uuid, P.fullName, P.gender, P.dateOfBirth, P.age_value, P.age_updatedAt, P.createdAt patientCreatedAt,
```
vinaysshenoy, line 37: What we want is the day was registered in the system, not the day the database row was created. For this, we have another property, `recordedAt`, which is what we should be using instead.

app/src/sharedTest/java/org/simple/clinic/TestData.kt
```diff
@@ -987,13 +987,15 @@ object TestData {
       gender: Gender = randomGender(),
       dateOfBirth: LocalDate? = null,
       age: Age? = null,
+      patientCreatedAt: Instant = Instant.now(),
```
vinaysshenoy, line 990: Can we use a hardcoded `Instant` instead of the current timestamp? There should be a live template in the Simple tooling plugin that adds this.

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
```diff
@@ -89,16 +91,23 @@ class RecentPatientsViewController @Inject constructor(
...
-          uuid = recentPatient.uuid,
-          name = recentPatient.fullName,
-          age = age(recentPatient),
-          gender = recentPatient.gender,
-          updatedAt = recentPatient.updatedAt,
-          dateFormatter = dateFormatter,
-          clock = userClock
-      )
+  private fun recentPatientItem(recentPatient: RecentPatient): RecentPatientItem {
+    val today = Instant.now(userClock).toLocalDateAtZone(userClock.zone)
```
vinaysshenoy, line 95: What do you think of passing the current date to this method as a parameter since it should be the same for all the recent patients? We can compute this once and pass it.

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
```diff
@@ -89,16 +91,23 @@ class RecentPatientsViewController @Inject constructor(
...
+    val patientCreatedAt = recentPatient.patientCreatedAt.toLocalDateAtZone(userClock.zone)
+
+    val isNewRegistration = today.isEqual(patientCreatedAt)
```
vinaysshenoy, line 98: We could also use `today == patientCreatedAt` to be more idiomatic.

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
```diff
@@ -89,16 +91,23 @@ class RecentPatientsViewController @Inject constructor(
...
-          name = recentPatient.fullName,
-          age = age(recentPatient),
-          gender = recentPatient.gender,
-          updatedAt = recentPatient.updatedAt,
-          dateFormatter = dateFormatter,
-          clock = userClock
-      )
+  private fun recentPatientItem(recentPatient: RecentPatient): RecentPatientItem {
+    val today = Instant.now(userClock).toLocalDateAtZone(userClock.zone)
+    val patientCreatedAt = recentPatient.patientCreatedAt.toLocalDateAtZone(userClock.zone)
```
vinaysshenoy, line 96: Rename this to `patientRegisteredOnDate`?

app/src/main/java/org/simple/clinic/patient/RecentPatient.kt
msasikanth: Ah okay, I misunderstood the implementation of it. I thought `recordedAt` was being updated every time BP or blood sugar is added, I forgot it only updates if the record date was older than the patient recorded date. Will update the implementation 👍

app/src/sharedTest/java/org/simple/clinic/TestData.kt
msasikanth: Going forward to all implementation in `TestData` will use `Instant.parse` right?

app/src/main/java/org/simple/clinic/patient/RecentPatient.kt
msasikanth: Done

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
msasikanth: Done

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
msasikanth: Done

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
msasikanth: Done

app/src/sharedTest/java/org/simple/clinic/TestData.kt
vinaysshenoy: Yes. Same for `LocalDate` as well.

msasikanth: Done

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
```diff
@@ -32,6 +35,8 @@ class RecentPatientsViewController @Inject constructor(
     @Named("full_date") private val dateFormatter: DateTimeFormatter
 ) : ObservableTransformer<UiEvent, UiChange> {
 
+  private val today = Instant.now(userClock).toLocalDateAtZone(userClock.zone)
```
vinaysshenoy, line 38: This need not be a class property, right? We could instantiate it inside the operator where it is being used.

```kotlin
.map { 
  it.map { recentPatient -> recentPatientItem(recentPatient, today) } 
}
```

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientItem.kt
```diff
@@ -43,22 +45,25 @@ data class RecentPatientItem(
...
 
-    viewHolder.render(name, age, gender, updatedAt, dateFormatter, clock)
+    viewHolder.render(name, age, gender, updatedAt, dateFormatter, clock, isNewRegistration)
   }
 
   class RecentPatientViewHolder(rootView: View) : ViewHolder(rootView) {
     private val nameAgeTextView by bindView<TextView>(R.id.recentpatient_item_title)
     private val lastSeenTextView by bindView<TextView>(R.id.recentpatient_item_last_seen)
     private val genderImageView by bindView<ImageView>(R.id.recentpatient_item_gender)
+    private val newRegistrationTextView by bindView<TextView>(R.id.recentpatient_new_registration)
```
vinaysshenoy, line 55: While we are here, should we switch to synthetic view accessors?

msasikanth: Yeah let's do it

msasikanth: I am going to open a different PR for this change since I will be converting the groupie usage with our regular adapter usage. Some of the changes may affect `RecentPatientsScreen` and it's items so I want to avoid doing that in this PR.

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsViewController.kt
```diff
@@ -64,7 +67,12 @@ class RecentPatientsViewController @Inject constructor(
         .refCount()
 
     val updateRecentPatients = recentPatientsStream
-        .map { it.map(::recentPatientItem) }
+        .map {
+          it.map { recentPatient ->
+            val today = Instant.now(userClock).toLocalDateAtZone(userClock.zone)
```
vinaysshenoy, line 72: This will still end up creating a new instance for every patient item. We should leave it inside the Rx `.map` operator block, but outside the list `.map` block.

vinaysshenoy:
```kotlin
.map { 
  it.map { recentPatient -> recentPatientItem(recentPatient, today) } 
}
```

msasikanth: Ah right, sorry 😓

vinaysshenoy: Also, we don't need to create an `Instant` to get a `LocalDate`. We can just create a `LocalDate` directly using `LocalDate.now(Clock)`.

msasikanth: Yes, forgot about it. Changing it to `LocalDate`

PR #1405, opened 2020-06-15
Update CHANGELOG

Honey14:


CHANGELOG.md
```diff
@@ -1,11 +1,14 @@
 # CHANGELOG
 
 ## Next Release
+### Internal
```
msasikanth, line 4: This should be `Feature` right?

vinaysshenoy: It should be in "Changes", I think. A feature is something that the user can use, but this is not something that the nurse can take action on.

Honey14: Can you tell me the difference between Internal | Changes ?

vinaysshenoy: This was my bad, I should have caught this when we paired on this PR.

Internal -> "Not a user facing change"
"Feature" -> "A new tool which the user can use"
"Changes" -> "A user facing change, but which is not in control of the user"
"Fixes" -> "Bug fixes"

msasikanth:
re-rolling the PR on to master branch

PR #1406, opened 2020-06-15
Abstract away the source of local HTML text

vinaysshenoy:
https://www.pivotaltracker.com/story/show/173302730


app/src/test/java/org/simple/clinic/help/HelpRepositoryTest.kt
```diff
@@ -1,128 +0,0 @@
-package org.simple.clinic.help
```
msasikanth, line 1: Do we no longer need `HelpRepositoryTest`?

vinaysshenoy: In the next PR, this will be replaced by a Room database backed implementation which is very simple and does not need to be tested.

vinaysshenoy:
> PR looks good to me. Maybe update the CHANGELOG as well to mark this change internally?

The next PR in the series will replace it with a room database. This is where I will add it to the CHANGELOG.

PR #1407, opened 2020-06-15
Migrate `RecentPatientItemType` to use `ItemAdapter`

msasikanth:


app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsView.kt
```diff
@@ -55,23 +52,26 @@ class RecentPatientsView(context: Context, attrs: AttributeSet) : FrameLayout(co
...
   }
 
   private fun screenCreates() = Observable.just(ScreenCreated())
 
-  fun updateRecentPatients(recentPatients: List<RecentPatientItemType<out ViewHolder>>) {
-    recentPatients.forEach { it.uiEvents = adapterEvents }
-    groupAdapter.update(recentPatients)
+  private fun adapterEvents() = recentAdapter.itemEvents
+      .ofType<UiEvent>()
+      .takeUntil(detaches)
```
vinaysshenoy, line 66: This is not necessary since `adapterEvents` is already being passed to the controller, which will dispose it when the screen is detached.

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientItem.kt
```diff
@@ -1,27 +1,23 @@
...
+import org.simple.clinic.widgets.recyclerview.ViewHolderX
 import org.simple.clinic.widgets.visibleOrGone
 import org.threeten.bp.Instant
 import org.threeten.bp.format.DateTimeFormatter
 import java.util.UUID
 
-sealed class RecentPatientItemType<VH : ViewHolder>(adapterId: Long) : GroupieItemWithUiEvents<VH>(adapterId) {
-  override lateinit var uiEvents: Subject<UiEvent>
-}
+sealed class RecentPatientItemType : ItemAdapter.Item<UiEvent>
```
vinaysshenoy, line 20: 👏

app/src/main/java/org/simple/clinic/recentpatientsview/RecentPatientsView.kt
msasikanth: fixed it. So if I got it right, all the observable events that are passed to `MobiusDelegate` and controller will automatically get disposed when screen is detached?

vinaysshenoy: Yes.

PR #1410, opened 2020-06-16
Redesign patient card in search results

msasikanth:
https://www.pivotaltracker.com/story/show/172931660


app/src/main/res/layout/view_patient_search_result.xml
```diff
@@ -35,65 +34,83 @@
...
+          style="@style/Clinic.V2.TextAppearance.Body2Bold.Grey1"
           android:layout_width="wrap_content"
           android:layout_height="wrap_content"
-          android:layout_marginTop="6dp"
-          android:layout_marginEnd="@dimen/spacing_12"
-          android:textAppearance="@style/Clinic.V2.TextAppearance.Body2Left.Numeric.Grey1"
-          tools:text="24-Mar-1975" />
+          android:text="@string/patientsearchresults_item_phone_number" />
 
         <org.simple.clinic.widgets.TintableCompoundDrawableTextView
```
vinaysshenoy, line 63: Does the new material components library let us get rid of this widget? We wanted it so we could tint the compound drawables?

Not necessary to fix this in this PR, just whether it's something we can get rid off now.

msasikanth: I think we no longer need to use this widget over here since we don't show any icon or handle the clicks for it. Since this is merged in, I will do the clean up during recent patient screen item view redesign. As for tinting drawables, we can still do it with `TextView`s using `drawableTint`.

vinaysshenoy: Sounds good. Can we also remove the widget from the codebase if `drawableTint` is available now?

msasikanth: Yes, I will verify if both of those have the same behaviour before deleting it.

PR #1413, opened 2020-06-16
Bump CameraX dependencies

msasikanth:


app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -75,11 +75,11 @@ constructor(
...
           it.setAnalyzer(cameraExecutor, ZxingQrCodeAnalyzer(scans::onNext))
         }
-
     cameraProvider.unbindAll()
 
-    val camera = cameraProvider.bindToLifecycle(qrCodeScannerLifecycle, cameraSelector, preview, analyzer)
-    preview.setSurfaceProvider(previewView.createSurfaceProvider(camera.cameraInfo))
+    preview.setSurfaceProvider(previewView.createSurfaceProvider())
+
+    cameraProvider.bindToLifecycle(qrCodeScannerLifecycle, cameraSelector, preview, analyzer)
```
vinaysshenoy, line 82: I'm curious, why was this change in ordering of statements done? Was this something we were doing incorrectly earlier?

msasikanth: Oh weren't doing anything wrong earlier, previously we set after `bindToLifecycle` call since we required the `Camera` instance, but we no longer require it. So they updated the documentation to call it before the `bindToLifecycle` call.

msasikanth: Okay, seems like even if I don't change the position it works fine. Reverting back the order change. (It gets the Camera from `SurfaceRequest`)

PR #1414, opened 2020-06-16
Retrying on a failed teleconsultation fetch only works once

msasikanth:
https://www.pivotaltracker.com/story/show/172910991

This was happening because of a race condition of dismissing and showing snackbar again when in airplane mode.


app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -134,7 +134,9 @@ class PatientSummaryScreen(
   private val teleconsultationErrorSnackbar by unsafeLazy {
     Snackbar.make(rootLayout, R.string.patientsummary_teleconsult_network_error, Snackbar.LENGTH_INDEFINITE)
         .setAction(R.string.patientsummary_teleconsult_network_error_retry) {
-          snackbarActionClicks.onNext(RetryFetchTeleconsultInfo)
+          postDelayed({
+            snackbarActionClicks.onNext(RetryFetchTeleconsultInfo)
+          }, 100)
```
vinaysshenoy, line 139: The solution to all problems. 👏

PR #1419, opened 2020-06-19
New illustrations and design updates

msasikanth:
This PR will add new illustrations for 
- `RegistrationLocationPermissionScreen`.
- `OverdueScreen` when the overdue list is empty.
- `ReportsScreen` when there are no reports.

I have also updated the design for `RegistrationLocationPermissionScreen` to match the new design


msasikanth:
Here are the screens preview

app/src/main/res/layout/screen_registration_location_permission.xml
```diff
@@ -39,10 +39,9 @@
   <ImageView
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:layout_above="@+id/registrationlocation_rationale_card"
-    android:layout_below="@+id/toolbar"
-    android:layout_marginBottom="-8dp"
-    android:scaleType="centerCrop"
+    android:layout_centerInParent="true"
+    android:adjustViewBounds="true"
+    android:scaleType="fitXY"
```
vinaysshenoy, line 44: `fitXY` can cause aspect ratio issues, correct?

msasikanth: Oh I forgot to remove it, `android:adjustViewBounds` should take care of that so we don't actually need a `scaleType`.  I will remove it. But yeah `fitXY`, when used alone, can cause aspect ratio issues.

msasikanth:
> PR looks, good. Left one comment.
> 
> One more question: Do we have vector drawables we can use instead of PNGs?

I have tried importing SVGs as vector drawables but some elements are getting messed up.

vinaysshenoy:
> > PR looks, good. Left one comment.
> > One more question: Do we have vector drawables we can use instead of PNGs?
> 
> I have tried importing SVGs as vector drawables but some elements are getting messed up. Only certain images are getting messed up
> 
> Screenshot 2020-06-19 at 10 43 14 AM

Before using PNGs, can we check with design to see if the SVGs can be corrected?

msasikanth:
Yep, will notify the design team about the SVG issues and see if they can be resolved.

msasikanth:
@vinaysshenoy Switched to vector drawables

PR #1420, opened 2020-06-19
Migrate registration name screen to Mobius: Strangulate controller

vinaysshenoy:
https://www.pivotaltracker.com/story/show/173216658

### DIFF
--- a/app/src/main/java/org/simple/clinic/registration/name/RegistrationNameEffect.kt
+++ b/app/src/main/java/org/simple/clinic/registration/name/RegistrationNameEffect.kt
@@ -5,3 +5,5 @@ import org.simple.clinic.user.OngoingRegistrationEntry
 sealed class RegistrationNameEffect
 
 data class PrefillFields(val entry: OngoingRegistrationEntry): RegistrationNameEffect()
+
+data class ValidateEnteredName(val name: String) : RegistrationNameEffect()
--- a/app/src/main/java/org/simple/clinic/registration/name/RegistrationNameEffect.kt
+++ b/app/src/main/java/org/simple/clinic/registration/name/RegistrationNameEffect.kt
@@ -4,8 +4,10 @@ import org.simple.clinic.user.OngoingRegistrationEntry
 
 sealed class RegistrationNameEffect
 
-data class PrefillFields(val entry: OngoingRegistrationEntry): RegistrationNameEffect()
+data class PrefillFields(val entry: OngoingRegistrationEntry) : RegistrationNameEffect()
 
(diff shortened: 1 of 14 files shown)


vinaysshenoy:
Marked as WIP to fix code climate issues.

app/src/main/java/org/simple/clinic/registration/name/RegistrationNameEffect.kt
```diff
@@ -5,3 +5,5 @@ import org.simple.clinic.user.OngoingRegistrationEntry
 sealed class RegistrationNameEffect
 
 data class PrefillFields(val entry: OngoingRegistrationEntry): RegistrationNameEffect()
+
+data class ValidatedEnteredName(val name: String): RegistrationNameEffect()
```
msasikanth: Can we rename this class name please?
```suggestion
data class ValidateEnteredName(val name: String): RegistrationNameEffect()
```

app/src/main/java/org/simple/clinic/registration/name/RegistrationNameEffectHandler.kt
```diff
@@ -19,7 +21,16 @@ class RegistrationNameEffectHandler @AssistedInject constructor(
...
-        .addConsumer(PrefillFields::class.java, { uiActions.preFillUserDetails(it.entry)}, schedulers.ui())
+        .addConsumer(PrefillFields::class.java, { uiActions.preFillUserDetails(it.entry) }, schedulers.ui())
+        .addTransformer(ValidatedEnteredName::class.java, validateNameEntry())
         .build()
   }
+
+  private fun validateNameEntry(): ObservableTransformer<ValidatedEnteredName, RegistrationNameEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .map { if (it.name.isBlank()) Blank else Valid }
```
msasikanth, line 36: One question I have is when do we consider doing validations in effect handler vs update function. Because I remember previously we have done some validations in update functions. Since this is a simple validation should we have done this in update function or am I missing something? 🤔

vinaysshenoy: You're right. I am planning to do an entire sweep of the entire registration flow once I migrate all the screens to Mobius. Unifying the way we do validations here is one of them.

I will make this change then.

app/src/main/java/org/simple/clinic/registration/name/RegistrationNameEffect.kt
vinaysshenoy: Good catch, done!

PR #1421, opened 2020-06-19
Show total overdue patients count in tabs

msasikanth:
https://www.pivotaltracker.com/story/show/172931754

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/HomeScreen.kt
+++ b/app/src/main/java/org/simple/clinic/home/HomeScreen.kt
@@ -10,6 +10,7 @@ import kotlinx.android.synthetic.main.screen_home.view.*
 import org.simple.clinic.R
 import org.simple.clinic.bindUiToController
 import org.simple.clinic.facility.change.FacilityChangeActivity
+import org.simple.clinic.home.HomeTab.OVERDUE
 import org.simple.clinic.home.help.HelpScreenKey
 import org.simple.clinic.main.TheActivity
 import org.simple.clinic.router.screen.ScreenRouter
@@ -95,10 +96,23 @@ class HomeScreen(context: Context, attrs: AttributeSet) : RelativeLayout(context
   }
 
   fun showOverdueAppointmentCount(count: Int) {
-    // TODO (SM): Update overdue tab badge to display the overdue count
+    val overdueTabIndex = HomeTab.values().indexOf(OVERDUE)
+    val overdueTab = homeTabLayout.getTabAt(overdueTabIndex)
+
+    overdueTab?.run {
+      @Suppress("UsePropertyAccessSyntax")
+      getOrCreateBadge().apply {
+        isVisible = true
+        maxCharacterCount = 3
+        number = count
+      }
+    }
   }
 
   fun removeOverdueAppointmentCount() {
-    // TODO (SM): Remove overdue appointment count badge from overdue tab
+    val overdueTabIndex = HomeTab.values().indexOf(OVERDUE)
+    val overdueTab = homeTabLayout.getTabAt(overdueTabIndex)
+
+    overdueTab?.removeBadge()
   }
 }
--- a/app/src/main/java/org/simple/clinic/home/HomeScreen.kt
+++ b/app/src/main/java/org/simple/clinic/home/HomeScreen.kt
@@ -52,6 +52,7 @@ class HomeScreen(context: Context, attrs: AttributeSet) : RelativeLayout(context
     rootLayout.hideKeyboard()
 
     viewPager.adapter = HomePagerAdapter(context)
+    homeTabLayout.setupWithViewPager(viewPager)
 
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
```diff
@@ -129,6 +129,17 @@ class AppointmentRepository @Inject constructor(
     )
   }
 
+  fun overdueAppointmentsCount(
+      since: LocalDate,
+      facility: Facility
+  ): Observable<List<Int>> {
+    return overdueDao.overdueAtFacilityCount(
```
vinaysshenoy, line 136: What do you think of changing the repository method to return `Observable<Int>` instead and map the list by summing the values within the repository method? This way, the abstraction no longer leaks and we can change the query later without affecting consumers of this API.

app/src/main/java/org/simple/clinic/home/HomeScreenController.kt
```diff
@@ -26,11 +32,49 @@ class HomeScreenController @Inject constructor(
...
+  private fun overdueAppointmentCountStream(facilityStream: Observable<Facility>): Observable<List<Int>> {
+    return facilityStream
+        .flatMap { currentFacility ->
+          appointmentRepository.overdueAppointmentsCount(
+              since = LocalDate.now(userClock),
+              facility = currentFacility
+          )
+        }
+        .replay()
+        .refCount()
```
vinaysshenoy, line 74: This stream does not need to be shared since it has only one consumer.

app/src/main/java/org/simple/clinic/home/HomeScreenController.kt
```diff
@@ -39,7 +83,8 @@ class HomeScreenController @Inject constructor(
                 facilityRepository.currentFacility(it)
               }
         }
-        .map { { ui: Ui -> ui.setFacility(it.name) } }
+        .replay()
+        .refCount()
```
vinaysshenoy, line 87: The sharing of this stream should not be done here, but only in the call site where it is needed.

app/src/main/java/org/simple/clinic/home/HomeScreen.kt
```diff
@@ -99,10 +100,21 @@ class HomeScreen(context: Context, attrs: AttributeSet) : RelativeLayout(context
   }
 
   fun showOverdueAppointmentCount(count: Int) {
-    // TODO (SM): Update overdue tab badge to display the overdue count
+    val overdueTabIndex = HomeTab.values().indexOf(OVERDUE)
+    val overdueTab = homeTabLayout.getTabAt(overdueTabIndex)
+
+    overdueTab?.run {
+      val badge = orCreateBadge
```
vinaysshenoy, line 107: Ugh, it took me a while to realise that this `orCreateBadge` is actually a method, `getOrCreateBadge` which Kotlin is treating as a getter method when it is not.

Can we use the complete method call instead?

vinaysshenoy: I tried this, and added a `.apply {}` block to the badge.

```kotlin
overdueTab?.run {
  @Suppress("UsePropertyAccessSyntax")
  getOrCreateBadge().apply {
      isVisible = true
      maxCharacterCount = 3
      number = count 
  }
}
```

app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
msasikanth: Does this makes sense?
```kt
    return overdueDao.overdueAtFacilityCount(
        facilityUuid = facility.uuid,
        scheduledBefore = since,
        scheduledAfter = since.minus(appointmentConfig.periodForIncludingOverdueAppointments)
    ).map { it.size }
```

app/src/main/java/org/simple/clinic/home/HomeScreenController.kt
msasikanth: done

app/src/main/java/org/simple/clinic/home/HomeScreenController.kt
msasikanth: Done

app/src/main/java/org/simple/clinic/home/HomeScreen.kt
msasikanth: Done

app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
vinaysshenoy: Yup, looks good.

msasikanth:
> PR looks good, leaving it unapproved for now until we figure out what's going wrong with the gesture navigation.

Removed the migration to ViewPager 2 which was causing the issues. I have rebased the PR, so this will be the first commit 9acf5e0

PR #1422, opened 2020-06-19
Ethiopia phone number validation

Honey14:
https://www.pivotaltracker.com/story/show/173219527

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Show overdue appointment count badge in overdue tab
 - Changed normal call buttons in the app from green to blue
 - Updated translations: `bn_BD`, `kn_IN`, `mr_IN`, `am_ET`
+- Ethiopia phone number validation
 
 ### Internal
 - Migrated `RecentPatientItemType` to use `ItemAdapter`
--- a/app/src/main/java/org/simple/clinic/editpatient/EditablePatientEntry.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditablePatientEntry.kt
@@ -2,20 +2,20 @@ package org.simple.clinic.editpatient
 
 import android.os.Parcelable
 import kotlinx.android.parcel.Parcelize
-import org.simple.clinic.editpatient.EditPatientValidationError.AGE_EXCEEDS_MAX_LIMIT
-import org.simple.clinic.editpatient.EditPatientValidationError.AGE_EXCEEDS_MIN_LIMIT
-import org.simple.clinic.editpatient.EditPatientValidationError.BOTH_DATEOFBIRTH_AND_AGE_ABSENT
-import org.simple.clinic.editpatient.EditPatientValidationError.COLONY_OR_VILLAGE_EMPTY
-import org.simple.clinic.editpatient.EditPatientValidationError.DATE_OF_BIRTH_EXCEEDS_MAX_LIMIT
-import org.simple.clinic.editpatient.EditPatientValidationError.DATE_OF_BIRTH_EXCEEDS_MIN_LIMIT
-import org.simple.clinic.editpatient.EditPatientValidationError.DATE_OF_BIRTH_IN_FUTURE
-import org.simple.clinic.editpatient.EditPatientValidationError.DATE_OF_BIRTH_PARSE_ERROR
-import org.simple.clinic.editpatient.EditPatientValidationError.DISTRICT_EMPTY
-import org.simple.clinic.editpatient.EditPatientValidationError.FULL_NAME_EMPTY
-import org.simple.clinic.editpatient.EditPatientValidationError.PHONE_NUMBER_EMPTY
-import org.simple.clinic.editpatient.EditPatientValidationError.PHONE_NUMBER_LENGTH_TOO_LONG
-import org.simple.clinic.editpatient.EditPatientValidationError.PHONE_NUMBER_LENGTH_TOO_SHORT
-import org.simple.clinic.editpatient.EditPatientValidationError.STATE_EMPTY
+import org.simple.clinic.editpatient.EditPatientValidationError.AgeExceedsMaxLimit
+import org.simple.clinic.editpatient.EditPatientValidationError.AgeExceedsMinLimit
+import org.simple.clinic.editpatient.EditPatientValidationError.BothDateOfBirthAndAgeAdsent
+import org.simple.clinic.editpatient.EditPatientValidationError.ColonyOrVillageEmpty
+import org.simple.clinic.editpatient.EditPatientValidationError.DateOfBirthExceedsMaxLimit
+import org.simple.clinic.editpatient.EditPatientValidationError.DateOfBirthExceedsMinLimit
+import org.simple.clinic.editpatient.EditPatientValidationError.DateOfBirthInFuture
+import org.simple.clinic.editpatient.EditPatientValidationError.DateOfBirthParseError
+import org.simple.clinic.editpatient.EditPatientValidationError.DistrictEmpty
+import org.simple.clinic.editpatient.EditPatientValidationError.FullNameEmpty
+import org.simple.clinic.editpatient.EditPatientValidationError.PhoneNumberEmpty
+import org.simple.clinic.editpatient.EditPatientValidationError.PhoneNumberLengthTooLong
+import org.simple.clinic.editpatient.EditPatientValidationError.PhoneNumberLengthTooShort
+import org.simple.clinic.editpatient.EditPatientValidationError.StateEmpty
 import org.simple.clinic.editpatient.EditablePatientEntry.EitherAgeOrDateOfBirth.EntryWithAge
 import org.simple.clinic.editpatient.EditablePatientEntry.EitherAgeOrDateOfBirth.EntryWithDateOfBirth
 import org.simple.clinic.patient.Age
@@ -31,7 +31,6 @@ import org.simple.clinic.registration.phone.PhoneNumberValidator.Result.LengthTo
 import org.simple.clinic.registration.phone.PhoneNumberValidator.Type
 import org.simple.clinic.util.valueOrEmpty
 import org.simple.clinic.widgets.ageanddateofbirth.UserInputAgeValidator
-import org.simple.clinic.widgets.ageanddateofbirth.UserInputAgeValidator.*
 import org.simple.clinic.widgets.ageanddateofbirth.UserInputAgeValidator.Result.Invalid.ExceedsMaxAgeLimit
 import org.simple.clinic.widgets.ageanddateofbirth.UserInputAgeValidator.Result.Invalid.ExceedsMinAgeLimit
 import org.simple.clinic.widgets.ageanddateofbirth.UserInputDateValidator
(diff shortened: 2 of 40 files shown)


app/src/main/java/org/simple/clinic/registration/phone/PhoneNumberValidator.kt
```diff
@@ -45,3 +45,26 @@ class IndianPhoneNumberValidator : PhoneNumberValidator {
...
+      }
+      MOBILE -> {
+        when{
+          number.isBlank() -> Result.BLANK
+          number.length < 9 -> LENGTH_TOO_SHORT
+          number.length > 10 -> LENGTH_TOO_LONG
+          else -> VALID
+        }
+      }
+    }
```
vinaysshenoy: Since these two code blocks are identical, we can skip the check of the `type` and just add the validations directly, right?

app/src/main/java/org/simple/clinic/registration/phone/PhoneNumberValidator.kt
```diff
@@ -13,11 +14,11 @@ interface PhoneNumberValidator {
...
 
-  enum class Result {
-    VALID,
-    LENGTH_TOO_SHORT,
-    LENGTH_TOO_LONG,
-    BLANK
+  sealed class Result {
+    object ValidNumber : Result()
+    data class LengthTooShort(val limit: Int) : Result()
+    data class LengthTooLong(val limit: Int) : Result()
```
vinaysshenoy, line 20: Instead of `limit`, can we use more appropriate names which indicate what they are?

Ex: `mininumRequiredNumberLength` and `maximumAllowedNumberLength`.

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
```diff
@@ -123,8 +123,8 @@ class PatientEntryEffectHandler(
 
   private fun hidePhoneLengthErrors() {
     with(validationActions) {
-      showLengthTooLongPhoneNumberError(false)
-      showLengthTooShortPhoneNumberError(false)
+      showLengthTooLongPhoneNumberError(false, 0)
+      showLengthTooShortPhoneNumberError(false, 0)
```
vinaysshenoy, line 127: Ah, looks like this method is responsible for both hiding *and* showing the error, which now requires us to pass in a parameter which is useless when hiding the error message. A classic violation of single responsibility principle.

We'll have to schedule some work to fix this at some point.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
```diff
@@ -432,17 +432,17 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
     }
   }
 
-  override fun showLengthTooShortPhoneNumberError(show: Boolean) {
+  override fun showLengthTooShortPhoneNumberError(show: Boolean, requiredNumberLength: Int) {
     if (show) {
-      phoneNumberInputLayout.error = context.getString(R.string.patiententry_error_phonenumber_length_less)
+      phoneNumberInputLayout.error = String.format(context.resources.getString(R.string.patiententry_error_phonenumber_length_less), requiredNumberLength)
```
vinaysshenoy, line 437: `String.format` is the wrong method to use here since it's a JVM method that is unaware of the Android resources framework. We should be using the version of [`Context#getString`](https://developer.android.com/reference/android/content/Context#getString(int,%20java.lang.Object...)) that accepts vararg format arguments instead.

app/src/test/java/org/simple/clinic/newentry/PatientEntryScreenLogicTest.kt
```diff
@@ -367,8 +368,8 @@ class PatientEntryScreenLogicTest {
     verify(validationActions, atLeastOnce()).showEmptyColonyOrVillageError(true)
     verify(validationActions, atLeastOnce()).showEmptyDistrictError(true)
     verify(validationActions, atLeastOnce()).showEmptyStateError(true)
-    verify(validationActions, atLeastOnce()).showLengthTooShortPhoneNumberError(true)
-    verify(validationActions, atLeastOnce()).showLengthTooLongPhoneNumberError(true)
+    verify(validationActions, atLeastOnce()). showLengthTooShortPhoneNumberError(true, 10)
```
vinaysshenoy, line 371: Looks like an unnecessary space was added?

CHANGELOG.md
```diff
@@ -5,8 +5,9 @@
 - New Onboarding UI
 - Show new registration label for recent patients in home screen
 - Redesigned patient card in search results
-- Redesigned patient card in recent patients screen
 - New illustrations for location permission screen, overdue screen, reports screen
+- Redesigned patient card in recent patients screen
```
vinaysshenoy, line 9: This (reordering of the line items) seems to been a rebase conflict issue. Not necessary to fix, something to watch out for in future.

app/src/main/res/values-am-rET/strings.xml
```diff
@@ -76,8 +76,8 @@
   <string name="patiententry_error_both_dateofbirth_and_age_empty">የታካሚውን ዕድሜ ያስገቡ</string>
```
vinaysshenoy, line 76: Can we squash this commit with the earlier one where we made the changes to the source strings?

app/src/main/java/org/simple/clinic/registration/phone/PhoneNumberValidator.kt
```diff
@@ -28,43 +28,41 @@ class IndianPhoneNumberValidator : PhoneNumberValidator {
   override fun validate(number: String, type: PhoneNumberValidator.Type): Result {
```
vinaysshenoy, line 28: Can we fix the commit messages?

1. The tone is incorrect, it should be written in a way that it completes the sentence: "If applied, this commit will..."
2. We need to be more detailed in the commit messages. Instead of something like "Refactor `PhoneNumberValidator`", we should be saying exactly what is being done. In this case, it will be something like, "Extract validation logic into discrete methods in `PhoneNumberValidator`".

app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
```diff
@@ -135,11 +135,7 @@ data class OngoingNewPatientEntry(
         errors += dobValidationCheck(dobValidator, dateOfBirth, ageValidator)
```
vinaysshenoy, line 135: We need to be more detailed in the commit messages. Instead of something like "Refactor `OngoingNewPatientEntry`", we should be saying exactly what is being done. In this case, it will be something like, "Extract validation logic into discrete methods in `OngoingNewPatientEntry`".

app/src/main/java/org/simple/clinic/editpatient/EditablePatientEntry.kt
```diff
@@ -180,11 +180,14 @@ data class EditablePatientEntry @Deprecated("Use the `from` factory function ins
     when (numberValidator.validate(phoneNumber, Type.LANDLINE_OR_MOBILE)) {
```
vinaysshenoy, line 180: We need to be more detailed in the commit messages. Instead of something like "Refactor `EditablePatientEntry`", we should be saying exactly what is being done. In this case, it will be something like, "Extract validation logic into discrete methods in `EditablePatientEntry`".

app/src/main/res/values-am-rET/strings.xml
```diff
@@ -561,4 +561,4 @@
```
vinaysshenoy: Making this change here means that we have broken the tests in earlier commits since they will have compilation issues. We will need to make these changes a part of the earlier commit where we changed the `PatientEntryValidationError` type to a sealed class.

CHANGELOG.md
Honey14: noted

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
Honey14: Agreed. Hiding errors should have had a different method. I can work on it whenever we do schedule.

app/src/main/java/org/simple/clinic/registration/phone/PhoneNumberValidator.kt
Honey14: Made this change in this commit `Extract validation logic into discrete methods in `PhoneNumberValidator``. Since we are already separating it into different functions.

app/src/main/java/org/simple/clinic/registration/phone/PhoneNumberValidator.kt
```diff
@@ -28,44 +28,35 @@ class IndianPhoneNumberValidator : PhoneNumberValidator {
...
+  fun validateLandlineOrMobile(number: String, startLimit: Int, endLimit: Int): Result {
+    return when {
+      number.isBlank() -> Blank
+      number.length < startLimit -> LengthTooShort(startLimit)
+      number.length > endLimit -> LengthTooLong(endLimit)
+      else -> ValidNumber
+    }
+  }
+
+  fun validateMobile(number: String, limit: Int, smallerThanNumber: Int, greaterThanNumber: Int): Result {
```
vinaysshenoy, line 48: Is the `limit` parameter necessary here? We can use `smallerThanNumber` and `greaterThanNumber` right? They are the same thing, but we are adding an additional parameter that adds no value.

vinaysshenoy, line 48: Can we rename the parameters which represent the bounds in both these methods?

Ex: `minimumRequiredLength` and `maximumAllowedLength`

app/src/main/java/org/simple/clinic/registration/phone/PhoneNumberValidator.kt
```diff
@@ -28,44 +28,35 @@ class IndianPhoneNumberValidator : PhoneNumberValidator {
...
-      MOBILE -> {
-        when {
-          number.isBlank() -> Blank
-          number.length < 9 -> LengthTooShort(9)
-          number.length > 10 -> LengthTooLong(10)
-          else -> ValidNumber
-        }
-      }
-    }
+    return IndianPhoneNumberValidator().validateLandlineOrMobile(number, 9, 10)
```
vinaysshenoy, line 60: With this, we are introduce coupling between the Ethiopia number validator and the Indian number validator. We should keep the two classes separate since there really is no reason for the Ethiopia number validator to depend on the Indian number validator.

However, the fact that you did this leads me to believe that we can merge the two classes into one, say a `LengthBasedNumberValidator` that can take in two `IntRange` (one for `MOBILE` and another for `LANDLINE_OR_MOBILE`) and those ranges can change based on the current `Country`.

app/src/main/res/values/strings.xml
```diff
@@ -83,8 +83,8 @@
   <string name="patiententry_error_both_dateofbirth_and_age_empty">Enter patient’s age</string>
   <string name="patiententry_error_invalid_dateofbirth">Enter patient’s date of birth in DD/MM/YYYY</string>
   <string name="patiententry_error_dateofbirth_is_in_future">Fix patient’s date of birth</string>
-  <string name="patiententry_error_phonenumber_length_less">Phone number cannot be less than 6 digits</string>
-  <string name="patiententry_error_phonenumber_length_more">Phone number cannot be more than 12 digits</string>
+  <string name="patiententry_error_phonenumber_length_less">Phone number cannot be less than %1$s digits</string>
+  <string name="patiententry_error_phonenumber_length_more">Phone number cannot be more than %1$s digits</string>
```
vinaysshenoy, line 87: Changing these strings will also break `PatientEditScreen` since it depends on these resources as well. Can we update that screen to account for the differences in number limits in countries as well?

app/src/main/java/org/simple/clinic/registration/RegistrationModule.kt
```diff
@@ -21,13 +22,19 @@ class RegistrationModule {
...
+      Country.ETHIOPIA -> LengthBasedNumberValidator(
+          minimumRequiredLengthMobile = 9,
+          maximumAllowedLengthMobile = 10,
+          minimumRequiredLengthLandlinesOrMobile = 9,
+          maximumAllowedLengthLandlinesOrMobile = 10)
+      else -> LengthBasedNumberValidator(
+          minimumRequiredLengthMobile = 10,
+          maximumAllowedLengthMobile = 10,
+          minimumRequiredLengthLandlinesOrMobile = 9,
+          maximumAllowedLengthLandlinesOrMobile = 10)
```
vinaysshenoy, line 37: These should be 6 and 12, right?

Honey14: Yes, that was a mistake

PR #1427, opened 2020-06-23
Migrate registration PIN entry to Mobius: Sterilize tests

vinaysshenoy:

### DIFF
--- a/app/src/test/java/org/simple/clinic/registration/pin/RegistrationPinScreenControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/registration/pin/RegistrationPinScreenControllerTest.kt
@@ -42,12 +42,14 @@ class RegistrationPinScreenControllerTest {
 
   @Test
   fun `when 4 digits are entered then the PIN should be submitted automatically`() {
+    // when
     setupController()
     uiEvents.onNext(RegistrationPinTextChanged("1"))
     uiEvents.onNext(RegistrationPinTextChanged("12"))
     uiEvents.onNext(RegistrationPinTextChanged("123"))
     uiEvents.onNext(RegistrationPinTextChanged("1234"))
 
+    // then
     verify(userSession).ongoingRegistrationEntry()
     verify(userSession).saveOngoingRegistrationEntry(ongoingRegistrationEntry.withPin("1234"))
     verify(screen).hideIncompletePinError()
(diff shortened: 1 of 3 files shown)


app/src/test/java/org/simple/clinic/registration/pin/RegistrationPinScreenControllerTest.kt
```diff
@@ -81,16 +88,19 @@ class RegistrationPinScreenControllerTest {
     uiEvents.onNext(RegistrationPinTextChanged("123"))
     uiEvents.onNext(RegistrationPinDoneClicked())
 
-    verify(screen).showIncompletePinError()
     verify(userSession, never()).saveOngoingRegistrationEntry(any())
+    verify(screen).showIncompletePinError()
     verify(screen, never()).openRegistrationConfirmPinScreen()
```
msasikanth, line 103: Since we have `verifyNoMoreInteractions` for the screen, we can perhaps remove this?

app/src/test/java/org/simple/clinic/registration/pin/RegistrationPinScreenControllerTest.kt
```diff
@@ -49,6 +49,9 @@ class RegistrationPinScreenControllerTest {
     uiEvents.onNext(RegistrationPinTextChanged("1234"))
 
     verify(userSession).saveOngoingRegistrationEntry(ongoingRegistrationEntry.withPin("1234"))
```
msasikanth, line 53: Should we also add `verifyNoMoreInteractions` for `UserSession`?

app/src/test/java/org/simple/clinic/registration/pin/RegistrationPinScreenControllerTest.kt
vinaysshenoy: We should keep this explicitly since this is something we do not want to happen. If we remove the `verifyNoMoreInteractions` line in future for whatever reason, we will have to remember to add this back and I'm sure this will be missed.

vinaysshenoy: Basically, anything we are asserting as part of the test itself should be explicit even if it's covered by a `verifyNoMoreInteractions` call.

app/src/test/java/org/simple/clinic/registration/pin/RegistrationPinScreenControllerTest.kt
vinaysshenoy: We can, I'll do this.

app/src/test/java/org/simple/clinic/registration/pin/RegistrationPinScreenControllerTest.kt
msasikanth: Got it 👍

app/src/test/java/org/simple/clinic/registration/pin/RegistrationPinScreenControllerTest.kt
vinaysshenoy: Done

PR #1429, opened 2020-06-23
Migrate registration PIN screen to Mobius: Prepare controller for migration

vinaysshenoy:


app/src/main/java/org/simple/clinic/registration/pin/RegistrationPinScreen.kt
```diff
@@ -28,6 +30,17 @@ class RegistrationPinScreen(
...
 
+  private val events by unsafeLazy {
+    Observable
+        .merge(
+            screenCreates(),
+            pinTextChanges(),
+            doneClicks()
+        )
+        .compose(ReportAnalyticsEvents())
+        .share()
```
Honey14, line 47: ~Can you help me with understanding why are we using `share()` and what it does?~ Are we sharing this because we need to share the events in Controller and Mobius loop as well?

vinaysshenoy: Yup, that is right. In the final parts of the refactoring after we remove the controller, we will also stop sharing the stream.

PR #1431, opened 2020-06-24
Prescription created from a protocol drug is not shown in the "Edit Medicines" screen if the drug is no longer in the protocol

msasikanth:
https://www.pivotaltracker.com/story/show/172974575

### DIFF
--- a/app/src/main/java/org/simple/clinic/drugs/EditMedicinesModel.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/EditMedicinesModel.kt
@@ -18,6 +18,10 @@ data class EditMedicinesModel(
     }
   }
 
+  fun isProtocolDrug(prescribedDrug: PrescribedDrug): Boolean {
+    return protocolDrugs!!.any { it.matches(prescribedDrug) }
+  }
+
   fun prescribedDrugsFetched(listFetched: List<PrescribedDrug>) =
       copy(prescribedDrugs = listFetched)
 
--- a/app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt
@@ -10,14 +10,39 @@ class EditMedicinesUiRenderer(private val ui: EditMedicinesUi) : ViewRenderer<Ed
 
   override fun render(model: EditMedicinesModel) {
     if (model.prescribedDrugs != null && model.protocolDrugs != null)
-      renderPrescribedProtocolDrugs(model.prescribedDrugs, model.protocolDrugs)
+      renderPrescribedProtocolDrugs(model, model.prescribedDrugs, model.protocolDrugs)
   }
 
-  private fun renderPrescribedProtocolDrugs(prescribedDrugs: List<PrescribedDrug>, protocolDrugs: List<ProtocolDrugAndDosages>) {
-    val prescribedProtocolDrugs = prescribedDrugs.filter { it.isProtocolDrug }
-    val isAtLeastOneCustomDrugPrescribed = prescribedDrugs.any { it.isProtocolDrug.not() }
+  private fun renderPrescribedProtocolDrugs(
+      model: EditMedicinesModel,
+      prescribedDrugs: List<PrescribedDrug>,
+      protocolDrugs: List<ProtocolDrugAndDosages>
+  ) {
+    val (prescribedProtocolDrugs, prescribedCustomDrugs) = prescribedDrugs.partition(model::isProtocolDrug)
+    val isAtLeastOneCustomDrugPrescribed = prescribedCustomDrugs.isNotEmpty()
+
+    val protocolDrugSelectionItems = protocolDrugSelectionItems(protocolDrugs, prescribedProtocolDrugs, isAtLeastOneCustomDrugPrescribed)
+    val customPrescribedDrugItems = customPrescribedDrugItems(prescribedCustomDrugs)
+    val drugsList = protocolDrugSelectionItems + customPrescribedDrugItems
+
+    ui.populateDrugsList(drugsList)
+  }
+
+  private fun customPrescribedDrugItems(
+      prescribedCustomDrugs: List<PrescribedDrug>
+  ): List<CustomPrescribedDrugListItem> {
+    return prescribedCustomDrugs
+        .sortedBy { it.updatedAt }
+        .mapIndexed { index, prescribedDrug -> CustomPrescribedDrugListItem(prescribedDrug, index == prescribedCustomDrugs.lastIndex) }
+  }
+
+  private fun protocolDrugSelectionItems(
+      protocolDrugs: List<ProtocolDrugAndDosages>,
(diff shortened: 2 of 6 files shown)


msasikanth:
Marked as WIP to fix code climate issue

msasikanth:
Assigned to @vinaysshenoy since @Honey14 is on leave

app/src/main/java/org/simple/clinic/drugs/EditMedicinesModel.kt
```diff
@@ -18,6 +18,14 @@ data class EditMedicinesModel(
     }
   }
 
+  fun isProtocolDrug(prescribedDrug: PrescribedDrug): Boolean {
+    return protocolDrugs!!.any {
+      it.drugs.any { protocolDrug ->
+        protocolDrug.name == prescribedDrug.name && protocolDrug.dosage == prescribedDrug.dosage
+      }
```
vinaysshenoy, line 25: What do you think of making this block a separate method on the `ProtocolDrugAndDosages` class?

```kotlin
fun matches(prescribedDrug: PrescribedDrug): Boolean {
    return drugs.any { protocolDrug ->
      protocolDrug.name == prescribedDrug.name && protocolDrug.dosage == prescribedDrug.dosage
    }
}
```

Then the `isProtocolDrug` method only operates at its own level of knowledge.

```kotlin
fun isProtocolDrug(prescribedDrug: PrescribedDrug): Boolean {
    return protocolDrugs!!.any { it.matches(prescribedDrug) }
}
```
```

File: app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt [hunk 2]
```diff
@@ -10,14 +10,43 @@ class EditMedicinesUiRenderer(private val ui: EditMedicinesUi) : ViewRenderer<Ed
...
+  ) {
     val prescribedProtocolDrugs = prescribedDrugs.filter { it.isProtocolDrug }
-    val isAtLeastOneCustomDrugPrescribed = prescribedDrugs.any { it.isProtocolDrug.not() }
+    val isAtLeastOneCustomDrugPrescribed = prescribedDrugs.any { prescribedDrug -> model.isProtocolDrug(prescribedDrug).not() }
+
+    val protocolDrugSelectionItems = protocolDrugSelectionItems(protocolDrugs, prescribedProtocolDrugs, isAtLeastOneCustomDrugPrescribed)
+    val customPrescribedDrugItems = customPrescribedDrugItems(model, prescribedDrugs)
+    val drugsList = protocolDrugSelectionItems + customPrescribedDrugItems
+
+    ui.populateDrugsList(drugsList)
```
vinaysshenoy, line 28: Can we add a UI renderer test to verify that a `PrescribedDrug` which was added as a `ProtocolDrug` initially shows up as a non-protocol drug if the protocol drug was removed?

app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt
```diff
@@ -10,14 +10,43 @@ class EditMedicinesUiRenderer(private val ui: EditMedicinesUi) : ViewRenderer<Ed
...
+      renderPrescribedProtocolDrugs(model, model.prescribedDrugs, model.protocolDrugs)
   }
 
-  private fun renderPrescribedProtocolDrugs(prescribedDrugs: List<PrescribedDrug>, protocolDrugs: List<ProtocolDrugAndDosages>) {
+  private fun renderPrescribedProtocolDrugs(
+      model: EditMedicinesModel,
+      prescribedDrugs: List<PrescribedDrug>,
+      protocolDrugs: List<ProtocolDrugAndDosages>
+  ) {
     val prescribedProtocolDrugs = prescribedDrugs.filter { it.isProtocolDrug }
```
vinaysshenoy, line 21: We should not be using `isProtocolDrug` property on `PrescribedDrug` anymore since it is unreliable. The only thing we'll use it for is for reporting on the dashboard. The app should check using the `ProtocolDrugAndDosage` check we introduced earlier.

With that in mind, I think we can use `partition` to split the list into protocol and custom drugs easily:

```kotlin
val (prescribedProtocolDrugs, prescribedCustomDrugs) = prescribedDrugs.partition(model::isProtocolDrug)
```

msasikanth: I had no idea Kotlin has a method called `partition`, will use that.

app/src/main/java/org/simple/clinic/drugs/EditMedicinesModel.kt
msasikanth: Yeah, that makes sense. Will change it

msasikanth: Done

app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt
msasikanth: Done

app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt
msasikanth: Done

app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt
```diff
@@ -10,14 +10,39 @@ class EditMedicinesUiRenderer(private val ui: EditMedicinesUi) : ViewRenderer<Ed
...
-  private fun renderPrescribedProtocolDrugs(prescribedDrugs: List<PrescribedDrug>, protocolDrugs: List<ProtocolDrugAndDosages>) {
-    val prescribedProtocolDrugs = prescribedDrugs.filter { it.isProtocolDrug }
-    val isAtLeastOneCustomDrugPrescribed = prescribedDrugs.any { it.isProtocolDrug.not() }
+  private fun renderPrescribedProtocolDrugs(
+      model: EditMedicinesModel,
+      prescribedDrugs: List<PrescribedDrug>,
+      protocolDrugs: List<ProtocolDrugAndDosages>
+  ) {
+    val (prescribedProtocolDrugs, prescribedCustomDrugs) = prescribedDrugs.partition(model::isProtocolDrug)
+    val isAtLeastOneCustomDrugPrescribed = prescribedDrugs.any { prescribedDrug -> model.isProtocolDrug(prescribedDrug).not() }
```
vinaysshenoy, line 22: This check can be simplified:

```kotlin
val isAtLeastOneCustomDrugPrescribed = !prescribedCustomDrugs.isEmpty()
```

msasikanth: Yep, changed to `val isAtLeastOneCustomDrugPrescribed = prescribedCustomDrugs.isNotEmpty()`

PR #1432, opened 2020-06-24
Implement a new way to manage feature flags

### LINKED ISSUE #1364: A better way to add feature flags
# The Problem
Currently, we declare and inject a configuration class via the dependency injection (DI) layer whenever we need to define a feature flag.

```kotlin
data class PhoneNumberMaskerConfig(
    val proxyPhoneNumber: String,
    val phoneMaskingFeatureEnabled: Boolean
) {

  companion object {

    fun read(reader: ConfigReader): PhoneNumberMaskerConfig {
      return PhoneNumberMaskerConfig(
          proxyPhoneNumber = reader.string("phonenumbermasker_proxy_phone_number", default = ""),
          phoneMaskingFeatureEnabled = reader.boolean("phonenumbermasker_masking_enabled", default = false)
      )
    }
  }
}
```

This is problematic for a few reasons:

- It is a lot of overhead to create a new config class if one does not already exist for a feature.
- The configuration objects are scoped to a particular package, but sometimes a feature flag cuts across multiple packages. This leads to leakage of information across boundaries.
- It is very cumbersome to inject an entire class just to get access to a feature flag.
- There is no way for us to find all the feature flags being used across the codebase easily.

We need a better solution, one that can solve these issues while retaining the benefits of the existing solution:

- Be able to configure these flags in tests
- Be able to switch between local and remote flags
- Override these flags for local development

## Proposal

We define features using enums, and provide a default value.

```kotlin
enum class Feature(
    val enabled: Boolean,
    val remoteConfigKey: String = ""
) {
  SecureCall(false),
  Telemedicine(true, "telemedicine_enabled")
}
```

Once this is done, we can define a singleton which can be used throughout the app to get access to the feature flag values:

```kotlin
object Flags {

  // Set this during app/test initialization
  lateinit var remoteConfig: ConfigReader

  fun isEnabled(feature: Feature): Boolean {
    return when {
      feature.remoteConfigKey.isBlank() -> feature.enabled
      else -> remoteConfig.boolean(feature.remoteConfigKey, feature.enabled)
    }
  }
}
```

With this, we can avoid all the boilerplate around creating a class and injecting it and simply call `Flags.isEnabled(SecureCallEnabled)` to check whether a feature is enabled or not. All our feature flags will also be defined in a single place.

The only thing missing from the list of desired benefits is a way to override the values in tests, which can be solved by maintaining a hash map inside the `Flags` object.

```kotlin
var overrides = mutableMapOf<Feature, Boolean>()

fun isEnabled(feature: Feature): Boolean {
  return when {
    feature in overrides -> overrides.getValue(feature)
    feature.remoteConfigKey.isBlank() -> feature.enabled
    else -> remoteConfig.boolean(feature.remoteConfigKey, feature.enabled)
  }
}
```

Overriding this value in tests is very simple.

```kotlin
@Test
fun `the feature is overriden in this test`() {
  Flags.overrides[Feature.SecureCall] = true
}
```

What does everyone else think? Is this something that appeals to us as a team? Any other solutions people can think of?

vinaysshenoy:
https://www.pivotaltracker.com/story/show/173485480

This PR resolves #1364.


app/src/main/java/org/simple/clinic/summary/bloodsugar/view/BloodSugarSummaryView.kt
```diff
@@ -292,7 +298,7 @@ class BloodSugarSummaryView(
   }
 
   private fun isBloodSugarEditable(measurement: BloodSugarMeasurement): Boolean {
-    if (bloodSugarSummaryConfig.bloodSugarEditFeatureEnabled.not() || measurement.reading.type is Unknown) {
+    if (features.isEnabled(EditBloodSugar).not() || measurement.reading.type is Unknown) {
```
msasikanth, line 301: Since we have `Features#isDisabled` should we use that? WDYT?

msasikanth:
Also, we have broken test references that need to be updated with new feature toggles, PTAL

msasikanth:
> ~Also, it seems like we have broken test references that need to be updated with new feature toggles, PTAL~ NVM, I mistook the error log, we have a separate `BloodSugarSummaryConfigModule` for release type. You haven't removed the `bloodSugarEditFeatureEnabled` param from there.

We have previously added it because we wanted to enable the editing by default for debug builds and enable remote config for release builds. But we have started enabling the edit feature for both builds, so maybe can move `BloodSugarSummaryConfigModule` to `main`

app/src/main/java/org/simple/clinic/summary/bloodsugar/view/BloodSugarSummaryView.kt
vinaysshenoy: Good catch, I introduced that method to handle this case, but it looks like I forgot to use it 😓

vinaysshenoy:
> ~Also, it seems like we have broken test references that need to be updated with new feature toggles, PTAL~ NVM, I mistook the error log, we have a separate `BloodSugarSummaryConfigModule` for release type. You haven't removed the `bloodSugarEditFeatureEnabled` param from there.

I am so happy we decided to build the release APKs, otherwise this would not have been caught until it got merged it and then a `SBX` release was attempted. Will fix.

vinaysshenoy:
> > ~Also, it seems like we have broken test references that need to be updated with new feature toggles, PTAL~ NVM, I mistook the error log, we have a separate `BloodSugarSummaryConfigModule` for release type. You haven't removed the `bloodSugarEditFeatureEnabled` param from there.
> 
> We have previously added it because we wanted to enable the editing by default for debug builds and enable remote config for release builds. But we have started enabling the edit feature for both builds, so maybe can move `BloodSugarSummaryConfigModule` to `main`

Good point, I will merge the modules.

app/src/main/java/org/simple/clinic/summary/bloodsugar/view/BloodSugarSummaryView.kt
```diff
@@ -292,7 +297,7 @@ class BloodSugarSummaryView(
   }
 
   private fun isBloodSugarEditable(measurement: BloodSugarMeasurement): Boolean {
-    if (bloodSugarSummaryConfig.bloodSugarEditFeatureEnabled.not() || measurement.reading.type is Unknown) {
+    if (features.isDisabled(EditBloodSugar) || measurement.reading.type is Unknown) {
```
msasikanth, line 300: Do we even need a feature flag for `EditBloodSugar` anymore since this is enabled locally?

vinaysshenoy: We don't. Let's remove this later.

PR #1442, opened 2020-06-26
Migrate registration PIN screen to Mobius: Sterilize the screen

vinaysshenoy:


vinaysshenoy:
Assigning to @Honey14 since Sasikanth is on leave.

app/src/main/java/org/simple/clinic/registration/pin/RegistrationPinScreen.kt
```diff
@@ -92,7 +88,9 @@ class RegistrationPinScreen(
   }
 
   private fun pinTextChanges() =
-      pinEditText.textChanges()
+      pinEditText
+          .textChanges()
+          .skip(1)
```
Honey14, line 93: Can you tell me why are using `skip()` here ?

vinaysshenoy: `rxbinding` uses an `InitialValueObservable` which emits the initial state of the view when it is subscribed to. It was needed in the older architecture, but is no longer required in Mobius screens, so we skip the first value which gets emitted.

PR #1443, opened 2020-06-26
Migrate confirm remove prescription dialog to Mobius : Prepare controller

msasikanth:
https://www.pivotaltracker.com/story/show/173527048

### DIFF
--- a/app/src/main/java/org/simple/clinic/drugs/selection/entry/confirmremovedialog/ConfirmRemovePrescriptionDialog.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/selection/entry/confirmremovedialog/ConfirmRemovePrescriptionDialog.kt
@@ -14,6 +14,7 @@ import io.reactivex.android.schedulers.AndroidSchedulers.mainThread
 import io.reactivex.schedulers.Schedulers.io
 import io.reactivex.subjects.PublishSubject
 import org.simple.clinic.R
+import org.simple.clinic.ReportAnalyticsEvents
 import org.simple.clinic.bp.entry.confirmremovebloodpressure.RemovePrescriptionClicked
 import org.simple.clinic.di.injector
 import org.simple.clinic.drugs.PrescriptionRepository
@@ -38,6 +39,16 @@ class ConfirmRemovePrescriptionDialog : AppCompatDialogFragment(), UiActions {
     requireArguments().getSerializable(KEY_PRESCRIPTION_UUID) as UUID
   }
 
+  private val events by unsafeLazy {
+    Observable
+        .merge(
+            screenDestroys,
+            removeClicks()
+        )
+        .compose(ReportAnalyticsEvents())
+        .share()
+  }
+
   @SuppressLint("CheckResult")
   override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
     val dialog = AlertDialog.Builder(requireContext(), R.style.Clinic_V2_DialogStyle_Destructive)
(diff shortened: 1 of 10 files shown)


vinaysshenoy:
Assigning to myself because Sasikanth is on leave and we need to get his PRs merged in.

app/src/main/java/org/simple/clinic/drugs/selection/entry/confirmremovedialog/ConfirmRemovePrescriptionDialog.kt
```diff
@@ -38,6 +39,12 @@ class ConfirmRemovePrescriptionDialog : AppCompatDialogFragment(), UiActions {
     requireArguments().getSerializable(KEY_PRESCRIPTION_UUID) as UUID
   }
 
+  private val events by unsafeLazy {
+    removeClicks()
+        .compose(ReportAnalyticsEvents())
```
vinaysshenoy, line 49: When we add it here, we will also have to remove it from the corresponding controller, otherwise the analytics events will get reported twice.

Since you are on leave, I will make this correction.

app/src/main/java/org/simple/clinic/drugs/selection/entry/confirmremovedialog/ConfirmRemovePrescriptionDialog.kt
```diff
@@ -57,7 +64,7 @@ class ConfirmRemovePrescriptionDialog : AppCompatDialogFragment(), UiActions {
   }
 
   private fun setupDialog(): Observable<UiChange> {
-    return Observable.merge(screenDestroys, removeClicks())
```
vinaysshenoy, line 60: We cannot remove `screenDestroys` here because it is used in the controller (see [source](https://github.com/simpledotorg/simple-android/blob/032ccc500c98295b1a67ad641c9fd976be4f8692/app/src/main/java/org/simple/clinic/ReplayUntilScreenIsDestroyed.kt) of `ReplayUntilScreenIsDestroyed`) to dispose the stream automatically.

Since you are on leave, I will make this correction.

app/src/main/java/org/simple/clinic/drugs/selection/entry/confirmremovedialog/ConfirmRemovePrescriptionDialog.kt
```diff
@@ -45,6 +50,20 @@ class ConfirmRemovePrescriptionDialog : AppCompatDialogFragment(), UiActions {
         .share()
   }
 
+  private val delegate: MobiusDelegate<ConfirmRemovePrescriptionDialogModel, ConfirmRemovePrescriptionDialogEvent, ConfirmRemovePrescriptionDialogEffect> by unsafeLazy {
+    MobiusDelegate.forView(
```
vinaysshenoy, line 54: Since this is a `Fragment`, we should be using a different saved state handle than the views since state saving and restoration will not work as expected.

We can use the `.forActivity` handle here since their mechanism and fragment's are similar compared to the view's. We should probably rename that to `.forActivityOrFragment` as well, although this can be done later.

I will make this correction since you are on leave.

vinaysshenoy:
I realized when reviewing this PR that we have not yet used `MobiusDelegate` with a `DialogFragment`. Running this confirmed this because the app crashed in a couple of scenarios.

I will mark this as WIP for now and will look at it tomorrow once I am done migrating the current screen.

PR #1450, opened 2020-06-30
Add support for Medical Record Number in Ethiopia

Honey14:
https://www.pivotaltracker.com/story/show/173333535


app/src/main/res/values/strings.xml
```diff
@@ -55,7 +55,7 @@
   <string name="patiententry_full_name">Patient\'s name</string>
   <string name="patiententry_phone_number">Phone number</string>
   <string name="patiententry_bangladesh_national_id">National ID number</string>
-  <string name="patiententry_ethiopia_national_id">National ID</string>
+  <string name="patiententry_ethiopia_medical_record_number">Medical Record Number</string>
```
vinaysshenoy, line 59: If we are changing the key here, we should be deleting the strings in the resource files for other languages since Transifex will not pick it up as a new string and mark them as translated, which will be incorrect since the source string has changed.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -548,6 +549,7 @@ class PatientRepository @Inject constructor(
     return when (identifierType) {
       BpPassport -> createBpPassportMetaData(assigningUser)
       BangladeshNationalId -> createBangladeshNationalIdMetadata(assigningUser)
+      EthiopiaMedicalRecordNumber -> TODO()
```
vinaysshenoy: This commit is now broken since `TODO()` will cause a crash since the screens will end up creating a record of type `EthiopiaMedicalRecordNumber` which will be passed here.

Suggestion: We can move the enabling of the identifier for the `Country` to a separate, final commit so that we can build this independently without needing to worry about this.

app/src/main/java/org/simple/clinic/patient/businessid/Identifier.kt
```diff
@@ -38,6 +39,7 @@ data class Identifier(
       }
       BangladeshNationalId -> value
       is Unknown -> value
+      EthiopiaMedicalRecordNumber -> value
```
vinaysshenoy, line 42: Can we move this up and keep the `Unknown` clause at the end?

app/src/main/java/org/simple/clinic/patient/businessid/Identifier.kt
```diff
@@ -46,6 +48,7 @@ data class Identifier(
       BpPassport -> resources.getString(R.string.identifiertype_bp_passport)
       BangladeshNationalId -> resources.getString(R.string.identifiertype_bangladesh_national_id)
       is Unknown -> resources.getString(R.string.identifiertype_unknown)
+      EthiopiaMedicalRecordNumber -> resources.getString(R.string.identifiertype_ethiopia_medical_record_number)
```
vinaysshenoy, line 51: Same comment as earlier about keeping the `Unknown` clause at the end.

app/src/androidTest/java/org/simple/clinic/patient/businessid/BusinessIdMetaDataAdapterAndroidTest.kt
```diff
@@ -41,6 +41,10 @@ class BusinessIdMetaDataAdapterAndroidTest {
                 assigningUserUuid = UUID.randomUUID(),
```
vinaysshenoy, line 41: We are adding the meta data type to the `BusinessId` type and not to the table, so the commit message should reflect this. Can we reword the commit, please?

app/src/androidTest/java/org/simple/clinic/patient/businessid/BusinessIdMetaDataAdapterAndroidTest.kt
```diff
@@ -41,6 +41,10 @@ class BusinessIdMetaDataAdapterAndroidTest {
...
                 assigningFacilityUuid = UUID.randomUUID()
             )
+            BusinessId.MetaDataVersion.MedicalRecordNumberMetaDataV1 -> BusinessIdMetaData.MedicalRecordNumberMetaDataV1(
+                assigningUserUuid = UUID.randomUUID(),
+                assigningFacilityUuid = UUID.randomUUID()
```
vinaysshenoy, line 46: Can we change this (and the rest of the class) to use fixed UUID instances instead of random ones, please? This is part of an ongoing effort to reduce the amount of non-determinism in our tests.

I also do not know if you are aware of this, but we have a tooling plugin which adds live templates to make this easier: https://github.com/simpledotorg/simple-android/tree/master/tooling/intellij-plugin

app/src/main/java/org/simple/clinic/patient/PatientModule.kt
```diff
@@ -27,7 +27,9 @@ class PatientModule {
         BusinessId.MetaDataVersion.BpPassportMetaDataV1 to
             moshi.adapter(BusinessIdMetaData.BpPassportMetaDataV1::class.java) as JsonAdapter<BusinessIdMetaData>,
         BusinessId.MetaDataVersion.BangladeshNationalIdMetaDataV1 to
-            moshi.adapter(BusinessIdMetaData.BangladeshNationalIdMetaDataV1::class.java) as JsonAdapter<BusinessIdMetaData>
+            moshi.adapter(BusinessIdMetaData.BangladeshNationalIdMetaDataV1::class.java) as JsonAdapter<BusinessIdMetaData>,
+        BusinessId.MetaDataVersion.MedicalRecordNumberMetaDataV1 to
+            moshi.adapter(BusinessIdMetaData.MedicalRecordNumberMetaDataV1::class.java) as JsonAdapter<BusinessIdMetaData>
```
vinaysshenoy, line 32: Can we squash this change with the previous commit? Otherwise, that commit is broken since the metadata-adapter does not include this.

app/src/androidTest/java/org/simple/clinic/patient/businessid/BusinessIdMetaDataAdapterAndroidTest.kt
Honey14: Can this be done by using `FakeUuidGenerator` class ?

vinaysshenoy: That's not meant for this use case. That class is meant for use cases where we want to use a fixed UUID wherever we need to generate a random UUID in production code.

I'd suggest installing the plugin I linked since that has a live template which generates random UUIDs to use in tests.

vinaysshenoy:
LGTM!

PR #1452, opened 2020-07-02
Migrate registration confirm PIN screen to Mobius: Prepare the controller

vinaysshenoy:
https://www.pivotaltracker.com/story/show/173391366

### DIFF
--- a/app/src/test/java/org/simple/clinic/registration/pin/RegistrationPinScreenLogicTest.kt
+++ b/app/src/test/java/org/simple/clinic/registration/pin/RegistrationPinScreenLogicTest.kt
@@ -53,8 +53,9 @@ class RegistrationPinScreenLogicTest {
     uiEvents.onNext(RegistrationPinDoneClicked())
 
     // then
-    verify(userSession).saveOngoingRegistrationEntry(ongoingRegistrationEntry.withPin(input))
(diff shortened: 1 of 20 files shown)


app/src/test/java/org/simple/clinic/registration/pin/RegistrationPinScreenLogicTest.kt
```diff
@@ -53,8 +53,9 @@ class RegistrationPinScreenLogicTest {
     uiEvents.onNext(RegistrationPinDoneClicked())
 
     // then
-    verify(userSession).saveOngoingRegistrationEntry(ongoingRegistrationEntry.withPin(input))
-    verify(uiActions).openRegistrationConfirmPinScreen()
+    val entryWithPin = ongoingRegistrationEntry.withPin(input)
```
Honey14, line 56: Is this the common test for both screens `RegistrationPinScreen` and `RegistrationConfirmPinScreen` ?

vinaysshenoy: No, it's not a common test. But since we are passing it to the screen key of `RegistrationConfirmPinScreen`, we have to modify the UI action of `RegistrationPinScreen` to accept `OngoingRegistrationEntry` as a parameter, so the test has to account for this as well.

PR #1454, opened 2020-07-02
Migrate registration confirm PIN to Mobius: Strangulate controller

vinaysshenoy:
https://www.pivotaltracker.com/story/show/173391366

### DIFF
--- a/app/src/main/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinEffectHandler.kt
@@ -8,11 +8,14 @@ import org.simple.clinic.registration.confirmpin.RegistrationConfirmPinValidatio
 import org.simple.clinic.registration.confirmpin.RegistrationConfirmPinValidationResult.Valid
 import org.simple.clinic.user.OngoingRegistrationEntry
 import org.simple.clinic.user.UserSession
+import org.simple.clinic.util.UtcClock
 import org.simple.clinic.util.scheduler.SchedulersProvider
+import org.threeten.bp.Instant
 
 class RegistrationConfirmPinEffectHandler @AssistedInject constructor(
     private val schedulers: SchedulersProvider,
     private val userSession: UserSession,
+    private val utcClock: UtcClock,
     @Assisted private val uiActions: RegistrationConfirmPinUiActions
 ) {
 
@@ -35,7 +38,7 @@ class RegistrationConfirmPinEffectHandler @AssistedInject constructor(
     return ObservableTransformer { effects ->
       effects
           .map { (pinConfirmation, entry) -> checkIfConfirmedPinIsValid(pinConfirmation, entry) }
-          .map(::PinConfirmationValidated)
+          .map { PinConfirmationValidated(it, Instant.now(utcClock)) }
     }
   }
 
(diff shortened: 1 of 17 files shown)


app/src/main/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinEffectHandler.kt
```diff
@@ -35,7 +38,7 @@ class RegistrationConfirmPinEffectHandler @AssistedInject constructor(
     return ObservableTransformer { effects ->
       effects
           .map { (pinConfirmation, entry) -> checkIfConfirmedPinIsValid(pinConfirmation, entry) }
-          .map(::PinConfirmationValidated)
+          .map { PinConfirmationValidated(it, Instant.now(utcClock)) }
```
Honey14, line 42: Can you explain how is adding timestamp is making the `Update` function pure?

app/src/main/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinUpdate.kt
```diff
@@ -16,20 +15,22 @@ class RegistrationConfirmPinUpdate : Update<RegistrationConfirmPinModel, Registr
...
       model: RegistrationConfirmPinModel,
       event: PinConfirmationValidated
   ): Next<RegistrationConfirmPinModel, RegistrationConfirmPinEffect> {
-    val updatedModel = model.validatedPinConfirmation(event.result)
+    var updatedModel = model.validatedPinConfirmation(event.result)
 
-    return if (updatedModel.pinConfirmationDoesNotMatchEnteredPin) {
-      next(updatedModel, ClearPin as RegistrationConfirmPinEffect)
+    return if (updatedModel.pinConfirmationMatchesEnteredPin) {
+      updatedModel = updatedModel.updateRegistrationEntryWithPinConfirmation(updatedModel.enteredPinConfirmation, event.timestamp)
```
Honey14, line 30: Is it fine to mutate the model like this in the `Update` function?

app/src/main/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinEvent.kt
```diff
@@ -3,3 +3,7 @@ package org.simple.clinic.registration.confirmpin
 import org.simple.clinic.widgets.UiEvent
```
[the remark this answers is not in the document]
vinaysshenoy: Whoops, good catch! Will fix!

app/src/main/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinEffectHandler.kt
vinaysshenoy: Good question! The `Instant` itself is fine to track in `Update` functions, because it is immutable. What we should not be doing is fetching the current timestamp in `Update` functions, which is why we read it in the effect handler, and pass it to `Update` as an event.

app/src/main/java/org/simple/clinic/registration/confirmpin/RegistrationConfirmPinUpdate.kt
vinaysshenoy: We are not actually mutating the model here, the model is still immutable. What are we doing is reassigning the reference of the variable `updatedModel` to a new instance of the model.

This is fine because the reference is local to this function. In addition, this is reassigned only in a single thread and will never involve concurrent access, so it's safe to do so.

PR #1460, opened 2020-07-06
Bump navigation component

msasikanth:
- Added checks to verify if the current screen is `placeHolderScreen` in `SetupActivity`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -75,7 +75,7 @@ buildscript {
       zxing               : '3.3.3',
       paging              : '2.1.2',
       appcompat           : '1.2.0-beta01',
-      navigation          : '2.2.2',
+      navigation          : '2.3.0',
       lottie              : '3.4.0',
       desugarJdk          : '1.0.5'
   ]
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -75,7 +75,7 @@ buildscript {
       zxing               : '3.3.3',
       paging              : '2.1.2',
       appcompat           : '1.2.0-beta01',
-      navigation          : '2.2.2',
+      navigation          : '2.3.0',
```
vinaysshenoy, line 78: Did we check the release notes to see if there are any breaking changes? Since this is going to be cherry-picked onto a release branch, I'm a little wary of introducing possible behavioural changes.

msasikanth: Yes, I have tested it after updating the version as well. There are a bunch of bug fixes and some new features, but if you aren't comfortable with making this change right now, I can drop this for now and bump it for next release. WDYT?

https://developer.android.com/jetpack/androidx/releases/navigation

vinaysshenoy: If you've tested it, it's fine. Just wanted to confirm.

msasikanth: Can you verify it on your device as well, please?

PR #1461, opened 2020-07-07
Clean up the user registration flow

vinaysshenoy:


app/src/androidTest/java/org/simple/clinic/user/UserSessionAndroidTest.kt
```diff
@@ -15,12 +15,11 @@ import org.simple.clinic.facility.FacilityRepository
 import org.simple.clinic.rules.LocalAuthenticationRule
 import org.simple.clinic.user.User.LoggedInStatus.LOGGED_IN
 import org.simple.clinic.user.User.LoggedInStatus.UNAUTHORIZED
-import org.simple.clinic.util.Just
-import org.simple.clinic.util.None
 import org.simple.clinic.util.Optional
 import org.simple.clinic.util.Rules
 import org.simple.clinic.util.RxErrorsRule
 import org.simple.clinic.util.toOptional
+import org.threeten.bp.Instant
```
Honey14, line 22: Can you tell me why do we need the current timestamp when creating user, please?

vinaysshenoy: All records on the system have three timestamps. There are all in UTC with millisecond precision.

`createdAt`: When this record was created at
`updatedAt`: When this record was updated
`deletedAt`: When this record was deleted

It is up to us to set and change this values correctly. So that the time of record creation, we should set both `createdAt` and `updatedAt` to the appropriate values.

PR #1463, opened 2020-07-07
Migrate registration location permission screen to Mobius: Prepare controller for migration

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/231/migrate-registration-location-permission-screen-to-mobius

### DIFF
--- a/app/src/test/java/org/simple/clinic/registration/location/RegistrationLocationPermissionScreenControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/registration/location/RegistrationLocationPermissionScreenControllerTest.kt
@@ -9,7 +9,6 @@ import org.junit.After
 import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
-import org.simple.clinic.util.Just
 import org.simple.clinic.platform.util.RuntimePermissionResult.GRANTED
 import org.simple.clinic.util.Optional
 import org.simple.clinic.util.RxErrorsRule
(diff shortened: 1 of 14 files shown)


app/src/test/java/org/simple/clinic/registration/location/RegistrationLocationPermissionScreenControllerTest.kt
```diff
@@ -34,12 +36,23 @@ class RegistrationLocationPermissionScreenControllerTest {
...
     uiEvents.onNext(RequestLocationPermission(permission = Optional.of(GRANTED)))
-
+    
     // then
     verify(screen).openFacilitySelectionScreen()
     verifyNoMoreInteractions(screen)
   }
 
+  @Test
+  fun `when location permission is denied then facility selection screen should not be opened`() {
```
Honey14, line 50: Question :
Steps: While user registration, I clicked deny then clicked on skip and I got the facility selection screen when location denied? this is the right behaviour, right? 
Also when checked the same from the master branch, I could not click the `Allow Location access` button, was it disabled before?

vinaysshenoy: > While user registration, I clicked deny then clicked on skip and I got the facility selection screen when location denied? this is the right behaviour, right?

Yup, this is the correct behaviour. If the user has allowed location access, we show the facilities that are within the user's vicinity together at the top and the rest of the facilities at the bottom. If they have denied it, we show all the facilities in one list.

> Also when checked the same from the master branch, I could not click the Allow Location access button, was it disabled before?

It wasn't. Maybe it was a one-off thing? I haven't seen this issue. 🤔

Honey14: > Yup, this is the correct behaviour. If the user has allowed location access, we show the facilities that are within the user's vicinity together at the top and the rest of the facilities at the bottom. If they have denied it, we show all the facilities in one list.

Got it

> It wasn't. Maybe it was a one-off thing? I haven't seen this issue. 🤔

Not sure about this

vinaysshenoy: > Not sure why. I tried twice just to re-check, it's was still happening

Weird. Given that the current branch is fine and it was working, I think we can ignore this for now and keep a watch for anything else.

PR #1479, opened 2020-07-10
Remove ThreeTenBp in favour of desugared java.time classes

vinaysshenoy:
Ticket: https://app.clubhouse.io/simpledotorg/story/338/remove-threetenbp-in-favour-of-java-time-apis

Ran the app on an API 21 emulator and did some basic flow testing. Things looked fine without any issues.


vinaysshenoy:
@Honey14 this is a large-ish PR, but almost all of it is just a replacing of import statements.

msasikanth:
Just a quick question, how does the timezone management work with `java.time`? Because I remember threeten has its own separate library.

vinaysshenoy:
> Just a quick question, how does the timezone management work with `java.time`? Because I remember threeten has its own separate library.

That's for `Chronology`. Timezones for the Gregorian calendar are a part of the platform.

PR #1492, opened 2020-07-13
Migrate reports screen to Mobius - Prepare controller

msasikanth:
https://app.clubhouse.io/simpledotorg/story/342/migrate-reports-screen-to-mobius

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/report/ReportsEvent.kt
+++ b/app/src/main/java/org/simple/clinic/home/report/ReportsEvent.kt
@@ -1,0 +1,3 @@
+package org.simple.clinic.home.report
+
+sealed class ReportsEvent
--- a/app/src/main/java/org/simple/clinic/home/report/ReportsModel.kt
+++ b/app/src/main/java/org/simple/clinic/home/report/ReportsModel.kt
@@ -1,0 +1,7 @@
+package org.simple.clinic.home.report
+
+import android.os.Parcelable
+import kotlinx.android.parcel.Parcelize
+
+@Parcelize
+class ReportsModel : Parcelable
--- a/app/src/main/java/org/simple/clinic/home/report/ReportsModel.kt
+++ b/app/src/main/java/org/simple/clinic/home/report/ReportsModel.kt
@@ -4,4 +4,9 @@ import android.os.Parcelable
 import kotlinx.android.parcel.Parcelize
 
 @Parcelize
-class ReportsModel : Parcelable
+class ReportsModel : Parcelable {
+
+  companion object {
+    fun create() = ReportsModel()
+  }
+}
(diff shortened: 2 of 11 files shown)


app/src/main/java/org/simple/clinic/home/report/ReportsModel.kt
```diff
@@ -4,4 +4,9 @@ import android.os.Parcelable
 import kotlinx.android.parcel.Parcelize
 
 @Parcelize
-class ReportsModel : Parcelable
+class ReportsModel : Parcelable {
+
+  companion object {
```
Honey14, line 9: Can you tell me why do we need this here?

app/src/main/java/org/simple/clinic/home/report/ReportsScreen.kt
```diff
@@ -34,7 +42,7 @@ class ReportsScreen(context: Context, attrs: AttributeSet) : FrameLayout(context
 
     bindUiToController(
         ui = this,
-        events = screenCreates(),
+        events = events,
         controller = controller,
         screenDestroys = detaches().map { ScreenDestroyed() }
```
Honey14, line 66: Can we extract `screenDestroys` to a shared stream, just like we did for events, please?

app/src/main/java/org/simple/clinic/home/report/ReportsEvent.kt
```diff
@@ -0,0 +1,3 @@
+package org.simple.clinic.home.report
```
Honey14, line 1: Can you create `ReportsUiActions` interface and extend in `ReportsUi` in no-op implementation ?

msasikanth:
> Looks good to me. Can you update the CHANGELOG file, please?

There is already a CHANGELOG entry for the same day for the migration.

app/src/main/java/org/simple/clinic/home/report/ReportsModel.kt
msasikanth: We are using it when creating `MobiusTestFixture` in `ReportsScreenControllerTest `

app/src/main/java/org/simple/clinic/home/report/ReportsScreen.kt
msasikanth: We don't need `screenDestroys` as part of the shared stream. It's only required when binding the controller here.

app/src/main/java/org/simple/clinic/home/report/ReportsEvent.kt
msasikanth: There are no `UiActions ` in this screen, so we don't have to create it.

app/src/main/java/org/simple/clinic/home/report/ReportsScreen.kt
msasikanth: @vinaysshenoy can you confirm this once? If I am not using `ScreenDestroyed` in the controller, and it's only being used in `ReplayUntilScreenIsDestroyed`. Do I still need to make the `screenDestroys` as part of the shared stream? 🤔

vinaysshenoy: No longer necessary. We will only need it if screen destroyed events are specifically used for something else.

PR #1493, opened 2020-07-13
Migrate reports screen to Mobius - Strangulate & Cleanup

msasikanth:
https://app.clubhouse.io/simpledotorg/story/342/migrate-reports-screen-to-mobius


app/src/main/java/org/simple/clinic/home/report/ReportsUiRenderer.kt
```diff
@@ -1,12 +1,28 @@
 package org.simple.clinic.home.report
 
 import org.simple.clinic.mobius.ViewRenderer
+import org.simple.clinic.util.ValueChangedCallback
 
 class ReportsUiRenderer(
     private val ui: ReportsUi
 ) : ViewRenderer<ReportsModel> {
 
+  private val valueChangedCallback = ValueChangedCallback<ReportsModel>()
```
vinaysshenoy, line 10: Adding a changed callback on the model is generally not a good idea since we only care about a certain property in it has changed or not.

In this case, since it's a nullable string, it might be a little tricky to track in the value changed callback since it does not support null values. I think it's ok to modify the tests in these scenarios since we are adding a lot of additional complexity here.

app/src/main/java/org/simple/clinic/home/report/ReportsUiRenderer.kt
```diff
@@ -1,12 +1,28 @@
...
+
   override fun render(model: ReportsModel) {
+    valueChangedCallback.pass(model) {
+      val reportsContent = it.reportsContent
+      if (reportsContent != null) {
```
vinaysshenoy, line 15: We could extract this to a computed property, say `hasLoadedReports` on `ReportsModel`.

app/src/main/java/org/simple/clinic/home/report/ReportsScreen.kt
```diff
@@ -5,26 +5,20 @@ import android.content.Context
 import android.os.Parcelable
```
vinaysshenoy, line 5: Shouldn't the commit message be "Remove `ReportsScreenController`"?

app/src/main/java/org/simple/clinic/home/report/ReportsScreen.kt
```diff
@@ -58,13 +52,6 @@ class ReportsScreen(context: Context, attrs: AttributeSet) : FrameLayout(context
     webView.settings.javaScriptEnabled = true
 
     context.injector<Injector>().inject(this)
-
-    bindUiToController(
-        ui = this,
-        events = events,
```
vinaysshenoy, line 64: We could also remove emission of the `ScreenCreated` event now that it is no longer in use.

app/src/main/java/org/simple/clinic/home/report/ReportsUiRenderer.kt
msasikanth: Removed it. We don't have to modify tests as well, I forgot I added a null check before rendering the reports. So we don't have to use `ValueChangedCallback` or modify any tests.

vinaysshenoy: This was initially added to handle cases where we were using `.distinctUntilChanged` to prevent the UI from rendering again and there's no standard way to do this in the Mobius style. Not doing it would cause tests to break since something would get rendered twice instead of once.

Since rendering is mostly cheap and using this adds additional complexity, I have been tending towards modifying the tests instead of using this since those tests will eventually get removed when we add new features to the screens.

app/src/main/java/org/simple/clinic/home/report/ReportsScreen.kt
```diff
@@ -5,40 +5,24 @@ import android.content.Context
...
-        .compose(ReportAnalyticsEvents())
-        .share()
-  }
-
   private val delegate by unsafeLazy {
     val uiRenderer = ReportsUiRenderer(this)
 
     MobiusDelegate.forView(
-        events = events.ofType(),
+        events = Observable.empty(),
```
vinaysshenoy, line 25: Can we use `Observable.never` instead? `Observable.empty()` immediately signals the completion of the stream, which is not what an actual event stream behaves like (never completes).

msasikanth: Okay, I wasn't sure what was the difference between these 2 in this context. So went with `empty`

PR #1494, opened 2020-07-13
Migrate add phone number dialog to Mobius - Stabilize

msasikanth:
https://app.clubhouse.io/simpledotorg/story/347/migrate-add-phone-number-dialog-to-mobius

### DIFF
--- a/app/src/test/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialogControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialogControllerTest.kt
@@ -133,12 +133,15 @@ class AddPhoneNumberDialogControllerTest {
   }
 
   private fun setupController() {
-    val controller = AddPhoneNumberDialogController(repository, validator, FakeUuidGenerator.fixed(generatedPhoneUuid))
+    val controller = AddPhoneNumberDialogController(
+        repository,
+        validator,
+        FakeUuidGenerator.fixed(generatedPhoneUuid),
+        patientUuid
+    )
 
     controllerSubscription = uiEvents
         .compose(controller)
         .subscribe { uiChange -> uiChange(dialog) }
-
-    uiEvents.onNext(AddPhoneNumberDialogCreated(patientUuid))
   }
 }
--- a/app/src/test/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialogControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialogControllerTest.kt
@@ -9,28 +9,22 @@ import com.nhaarman.mockitokotlin2.whenever
 import io.reactivex.Completable
 import io.reactivex.disposables.Disposable
 import io.reactivex.subjects.PublishSubject
-import junitparams.JUnitParamsRunner
-import junitparams.Parameters
 import org.junit.After
 import org.junit.Rule
 import org.junit.Test
-import org.junit.runner.RunWith
 import org.simple.clinic.patient.PatientRepository
 import org.simple.clinic.patient.PhoneNumberDetails
 import org.simple.clinic.registration.phone.PhoneNumberValidator
-import org.simple.clinic.registration.phone.PhoneNumberValidator.Result
 import org.simple.clinic.registration.phone.PhoneNumberValidator.Result.Blank
 import org.simple.clinic.registration.phone.PhoneNumberValidator.Result.LengthTooLong
 import org.simple.clinic.registration.phone.PhoneNumberValidator.Result.LengthTooShort
 import org.simple.clinic.registration.phone.PhoneNumberValidator.Result.ValidNumber
 import org.simple.clinic.registration.phone.PhoneNumberValidator.Type.LANDLINE_OR_MOBILE
 import org.simple.clinic.util.RxErrorsRule
-import org.simple.clinic.util.exhaustive
 import org.simple.clinic.uuid.FakeUuidGenerator
 import org.simple.clinic.widgets.UiEvent
 import java.util.UUID
 
-@RunWith(JUnitParamsRunner::class)
 class AddPhoneNumberDialogControllerTest {
 
   @get:Rule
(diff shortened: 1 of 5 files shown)


msasikanth:
Marking it as `WIP` in order to fix parameter tests.

msasikanth:
I have rebased commits, please check out the branch when reviewing

app/src/test/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialogControllerTest.kt
```diff
@@ -118,11 +112,14 @@ class AddPhoneNumberDialogControllerTest {
...
-  private fun setupController() {
+  private fun setupController(newNumber: String) {
     val controller = AddPhoneNumberDialogController(repository, validator, FakeUuidGenerator.fixed(generatedPhoneUuid))
 
     controllerSubscription = uiEvents
         .compose(controller)
         .subscribe { uiChange -> uiChange(dialog) }
+
+    uiEvents.onNext(AddPhoneNumberDialogCreated(patientUuid))
+    uiEvents.onNext(AddPhoneNumberSaveClicked(newNumber))
```
vinaysshenoy, line 129: Can we have this particular event done separately in each test? We should only move emission of the `ScreenCreated` event since that is a part of screen initialization and something that is supposed to be triggered for every test.

app/src/test/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialogControllerTest.kt
```diff
@@ -71,6 +72,7 @@ class AddPhoneNumberDialogControllerTest {
         numberDetails = numberDetails,
         active = true
     )
+    verifyNoMoreInteractions(repository)
```
vinaysshenoy, line 79: Can we also verify no interactions on the `dialog` here?

app/src/test/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialogControllerTest.kt
```diff
@@ -90,6 +92,7 @@ class AddPhoneNumberDialogControllerTest {
     setupController(newNumber)
 
     verify(repository, never()).createPhoneNumberForPatient(any(), any(), any(), any())
+    verifyNoMoreInteractions(repository)
```
vinaysshenoy, line 99: Can we also verify no interactions on the `dialog` here?

app/src/test/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialogControllerTest.kt
```diff
@@ -79,13 +80,10 @@ class AddPhoneNumberDialogControllerTest {
...
   @Test
-  @Parameters(method = "validation errors")
-  fun `when save is clicked, the number should not be saved if it's invalid`(
-      validationError: Result
-  ) {
+  fun `when save is clicked, the number should not be saved if it's invalid`() {
     // given
     val newNumber = "123"
-    whenever(validator.validate(newNumber, type = LANDLINE_OR_MOBILE)).thenReturn(validationError)
+    whenever(validator.validate(newNumber, type = LANDLINE_OR_MOBILE)).thenReturn(validationErrors.random())
```
vinaysshenoy, line 86: When we remove parameterized tests, we should be adding a separate test for each one of the removed parameter combinations instead of using a random value. Can we do that in this commit?

vinaysshenoy, line 86: Can we use a real instance of `PhoneNumberValidator` here? See step 3 in [`Stabilize`](https://github.com/simpledotorg/simple-android/blob/master/doc/mobius/migrating-to-mobius.md#steps).

app/src/test/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialogControllerTest.kt
```diff
@@ -118,11 +112,14 @@ class AddPhoneNumberDialogControllerTest {
...
 
-  private fun setupController() {
+  private fun setupController(newNumber: String) {
     val controller = AddPhoneNumberDialogController(repository, validator, FakeUuidGenerator.fixed(generatedPhoneUuid))
 
     controllerSubscription = uiEvents
         .compose(controller)
         .subscribe { uiChange -> uiChange(dialog) }
+
+    uiEvents.onNext(AddPhoneNumberDialogCreated(patientUuid))
```
vinaysshenoy, line 128: One more thing I see that is missing in this PR is changing the controller to accept the parameters via assisted injection instead of a custom `ScreenCreated` event (See step 7 of [`Stabilize`](https://github.com/simpledotorg/simple-android/blob/master/doc/mobius/migrating-to-mobius.md#steps)).

Was this missed or we going to do it in another PR?

msasikanth: Ah right, I missed it. Will fix it

app/src/test/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialogControllerTest.kt
msasikanth: I initially went with that approach and it seemed like a lot of code duplication. So using random felt like a good option, but will change back to having different tests.

app/src/test/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialogControllerTest.kt
msasikanth: You're talking about `AddPhoneNumberSaveClicked` event right?

vinaysshenoy: Yes.

app/src/test/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialogControllerTest.kt
vinaysshenoy: The problem with this is that it adds non-determinism to the tests, which makes the test suite unreliable. Code duplication is encouraged in tests to keep them isolated and focused.

PR #1503, opened 2020-07-14
Migrate add phone number dialog to Mobius - Prepare controller

msasikanth:
https://app.clubhouse.io/simpledotorg/story/347/migrate-add-phone-number-dialog-to-mobius

I have rebased commits, please check out the branch when reviewing.


app/src/main/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialog.kt
```diff
@@ -73,9 +77,22 @@ class AddPhoneNumberDialog : AppCompatDialogFragment(), AddPhoneNumberUi {
...
+        update = AddPhoneNumberUpdate(),
+        effectHandler = effectHandlerFactory.create(this).build(),
+        modelUpdateListener = uiRenderer::render
+    )
+  }
+
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     TheActivity.component.inject(this)
+    delegate.onRestoreInstanceState(savedInstanceState)
```
Honey14, line 94: Are we doing this here because we shouldn't expect dialog to work like activities and did not use onRestoreInstanceState?

app/src/main/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialog.kt
```diff
@@ -89,15 +86,22 @@ class AddPhoneNumberDialog : AppCompatDialogFragment(), AddPhoneNumberUi {
...
+  private var layout: View? = null
+
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     requireContext().injector<Injector>().inject(this)
     delegate.onRestoreInstanceState(savedInstanceState)
   }
 
+  override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+    return layout
```
Honey14, line 98: Can you tell me why are we creating the global state for `layout`?

app/src/main/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialog.kt
msasikanth: Yeah, the way we restore is different in fragments and activities compared to views.

app/src/main/java/org/simple/clinic/summary/addphone/AddPhoneNumberDialog.kt
msasikanth: In order to use Kotlin synthetic accessors, we need to override the `onCreateView` and provide the layout. Since we are inflating the layout and setting it as a custom view for the dialog in `onCreate`. I have made it a variable in the class.

Honey14: Why do we need to override `onCreateView` for using kotlin extensions?

msasikanth: Kotlin extensions try to get the view from the `Activity` or `Fragment`. If we don't assign a layout in `Activity#setContentView` or `Fragment#onCreateView` it will fail.

vinaysshenoy: Ok, it is clear that the current mobius setup does not work as well for dialogs as it does for the other UI components. I see two approaches we can take here:

- See what needs to change in `MobiusDelegate` to support dialogs, or
- Make dialogs really dumb and pure presentation only. They can emit events which can be subscribed to by the parent screen/activity/fragment and then fed into their own mobius loop.

This PR is fine for now and we can merge it in, maybe this change is something we should decide on before migrating any other dialogs.

vinaysshenoy:
@msasikanth Can you reroll this and fix the conflicts? I can approve since Honey is unavailable now and she already has approved it.

PR #1508, opened 2020-07-15
Migrate registration facility screen to Mobius: strangulate controller

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/237/migrate-registration-facility-selection-screen-to-mobius

### DIFF
--- a/app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionEffectHandler.kt
@@ -25,6 +25,7 @@ class RegistrationFacilitySelectionEffectHandler @AssistedInject constructor(
         .subtypeEffectHandler<RegistrationFacilitySelectionEffect, RegistrationFacilitySelectionEvent>()
         .addTransformer(FetchCurrentLocation::class.java, fetchLocation())
         .addTransformer(LoadFacilitiesWithQuery::class.java, loadFacilitiesWithQuery())
+        .addTransformer(LoadTotalFacilityCount::class.java, loadTotalCountOfFacilities())
         .build()
   }
 
@@ -52,6 +53,19 @@ class RegistrationFacilitySelectionEffectHandler @AssistedInject constructor(
                 .facilities(searchQuery = effect.query)
                 .subscribeOn(schedulersProvider.io())
                 .map { FacilitiesFetched(query = effect.query, facilities = it) }
+          }
+    }
+  }
+
+  private fun loadTotalCountOfFacilities(): ObservableTransformer<LoadTotalFacilityCount, RegistrationFacilitySelectionEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .switchMap {
+            facilityRepository
+                .recordCount()
+                .subscribeOn(schedulersProvider.io())
+                .take(1)
+                .map(::TotalFacilityCountLoaded)
           }
     }
   }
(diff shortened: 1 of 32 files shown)


app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionInit.kt
```diff
@@ -1,12 +1,35 @@
...
 import com.spotify.mobius.First
-import com.spotify.mobius.First.first
 import com.spotify.mobius.Init
+import org.simple.clinic.mobius.first
+import org.simple.clinic.registration.RegistrationConfig
+import java.time.Duration
 
-class RegistrationFacilitySelectionInit: Init<RegistrationFacilitySelectionModel, RegistrationFacilitySelectionEffect> {
+class RegistrationFacilitySelectionInit(
+    private val locationUpdateInterval: Duration,
```
msasikanth, line 10: If these parameters are part of `RegistrationConfig`, couldn't we have injected it directly in effect handler? 🤔

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionEffectHandler.kt
```diff
@@ -40,4 +43,16 @@ class RegistrationFacilitySelectionEffectHandler @AssistedInject constructor(
           .map(::LocationFetched)
     }
   }
+
+  private fun loadFacilitiesWithQuery(): ObservableTransformer<LoadFacilitiesWithQuery, RegistrationFacilitySelectionEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .switchMap { effect ->
```
msasikanth, line 60: Should we observe on `io` here or subscribing on `io` for the nested chain is fine?

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionEffectHandler.kt
```diff
@@ -70,4 +77,16 @@ class RegistrationFacilitySelectionEffectHandler @AssistedInject constructor(
           }
     }
   }
+
+  private fun saveCurrentRegistrationEntry(): ObservableTransformer<SaveRegistrationEntryAsUser, RegistrationFacilitySelectionEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .doOnNext { userSession.saveOngoingRegistrationEntry(it.entry) }
```
msasikanth, line 85: Should we these operations on `io` thread?

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionInit.kt
vinaysshenoy: We could have, except I felt in this case that the effect handler should not be in charge of making the decisions around what the location update interval could be and it should be provided that via the effect.

I think the actual problem here is that we do not inject the delegate which requires us to manually construct all these things in the screen. Something to consider in the back of our mind to see if we _can_ inject delegates.

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionEffectHandler.kt
vinaysshenoy: If we are invoking a reactive method, I generally prefer to do it in the nested chain.

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionEffectHandler.kt
vinaysshenoy: We don't need to since they do not do any I/O at all and are just objects in memory. As a conclusion of this migration, I have completely removed user session for being responsible for storing this ongoing entry and made it an object that is passed via the screen keys. This will be seen in later PRs.

PR #1509, opened 2020-07-15
Migrate recent patients view to Mobius: Strangulate controller

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/311/migrate-recent-patients-view-to-mobius

### DIFF
--- a/app/src/main/java/org/simple/clinic/recentpatientsview/LatestRecentPatientsEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/recentpatientsview/LatestRecentPatientsEffectHandler.kt
@@ -27,6 +27,7 @@ class LatestRecentPatientsEffectHandler @AssistedInject constructor(
     return RxMobius
         .subtypeEffectHandler<LatestRecentPatientsEffect, LatestRecentPatientsEvent>()
         .addTransformer(LoadRecentPatients::class.java, loadRecentPatients())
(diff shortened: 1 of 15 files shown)


msasikanth:
Assigning it to my self since @Honey14 is no leave

app/src/main/java/org/simple/clinic/recentpatientsview/LatestRecentPatientsEffectHandler.kt
```diff
@@ -19,6 +26,23 @@ class LatestRecentPatientsEffectHandler @AssistedInject constructor(
...
+  private fun loadRecentPatients(): ObservableTransformer<LoadRecentPatients, LatestRecentPatientsEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .switchMap { effect ->
+            userSession
+                .loggedInUser()
+                .subscribeOn(schedulers.io())
+                .filterAndUnwrapJust()
+                .take(1)
+                .switchMap(facilityRepository::currentFacility)
```
msasikanth, line 44: Aren't we using `private val currentFacility: Lazy<Facility>` to get current facility or will that be done in later commits?

vinaysshenoy: We will not change this during strangulation and retain whatever method we were using in the controller. We'll switch to using the injected providers in the final clean up PRs.

PR #1510, opened 2020-07-16
Migrate login pin screen to Mobius - Stabilise tests

msasikanth:

### DIFF
--- a/app/src/test/java/org/simple/clinic/login/pin/LoginPinScreenControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/login/pin/LoginPinScreenControllerTest.kt
@@ -54,12 +54,14 @@ class LoginPinScreenControllerTest {
 
   @Test
   fun `when screen starts, show phone number`() {
-    whenever(userSession.ongoingLoginEntry())
-        .thenReturn(Single.just(ongoingLoginEntry))
+    // given
+    whenever(userSession.ongoingLoginEntry()).thenReturn(Single.just(ongoingLoginEntry))
 
+    // when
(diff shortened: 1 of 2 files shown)


app/src/test/java/org/simple/clinic/login/pin/LoginPinScreenControllerTest.kt
```diff
@@ -59,7 +60,7 @@ class LoginPinScreenControllerTest {
 
     // when
     setupController()
-    uiEvents.onNext(PinScreenCreated())
+    uiEvents.onNext(ScreenCreated())
```
vinaysshenoy, line 63: Can we move this to `setupController` so that it is initialized in every test?

There is also no need to replace a custom event with `ScreenCreated`. The point of doing this is to remove the need for a `ScreenCreated` event of any type since that role is being played by the `Init` class once we move to Mobius.

msasikanth: Yeah, so the reason why I still used the `ScreenCreated` is it has to deal with UI. So I am planning on remove it during strangulate. WDYT?

vinaysshenoy: What I mean is that there is no need to use `ScreenCreated` instead of `PinScreenCreated` since they are functionally the same. What we should do, however, is move the emission of this event into the `setupController` method because we want the tests to be as close to production as possible and this particular event should be emitted in each test.

PR #1512, opened 2020-07-16
Migrate registration facility selection screen to Mobius: Sterilize screen

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/237/migrate-registration-facility-selection-screen-to-mobius


app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionUiRenderer.kt
```diff
@@ -8,12 +10,17 @@ import org.simple.clinic.location.LocationUpdate.Unavailable
...
 
-class RegistrationFacilitySelectionUiRenderer(
-    private val ui: RegistrationFacilitySelectionUi,
+class RegistrationFacilitySelectionUiRenderer @AssistedInject constructor(
+    @Assisted private val ui: RegistrationFacilitySelectionUi,
     private val facilityListItemBuilder: FacilityListItemBuilder,
     private val registrationConfig: RegistrationConfig
 ) : ViewRenderer<RegistrationFacilitySelectionModel> {
 
+  @AssistedInject.Factory
```
msasikanth, line 19: Is it a new approach we will be implementing for UiRenderer?

vinaysshenoy: We will only use this in places where the UI renderer needs access to dependencies that are injectable.

PR #1515, opened 2020-07-20
Clean up facility selection screens

vinaysshenoy:


app/src/main/java/org/simple/clinic/facility/change/FacilityChangeActivity.kt
```diff
@@ -158,15 +156,6 @@ class FacilityChangeActivity : AppCompatActivity() {
   }
 
   fun updateFacilities(facilityItems: List<FacilityListItem>, updateType: FacilitiesUpdateType) {
-    // Avoid animating the items on their first entry.
```
msasikanth, line 161: Any particular reason for removing this?

app/src/main/java/org/simple/clinic/facility/change/FacilityListItem.kt
```diff
@@ -27,6 +61,42 @@ sealed class FacilityListItem {
...
+        }
+        is Address.WithoutStreet -> {
+          holder.itemView.resources.getString(
+              R.string.registrationfacilities_facility_address_without_street,
+              address.district,
+              address.state)
+        }
+      }
+
+      holder.itemView.setOnClickListener { subject.onNext(FacilityItemClicked(facility)) }
```
msasikanth, line 98: It's not related to this PR, but one thing I have been thinking about is. Isn't it not recommended to have click listeners when we are binding the view to recycler view since it's called for every item? 🤔

app/src/main/java/org/simple/clinic/facility/change/FacilityChangeActivity.kt
vinaysshenoy: Because we are using `ItemAdapter` which will ship with default animations already. With the current approach, we were adding additional complexity to track whether the animations were already shown once or not for little benefit.

app/src/main/java/org/simple/clinic/facility/change/FacilityListItem.kt
vinaysshenoy: The alternative is to track the current item in a mutable variable and update it when binding, which adds additional complexity. We made this call to make the views as stateless as possible. Eventually, when Jetpack Compose becomes production ready, this will make it easier for us to migrate if we would like to.

msasikanth: or can't we get the views in `onCreateView`? and get the item for the position when the item is clicked. WDYT?

vinaysshenoy: I didn't understand this. `onCreateView` is used where?

msasikanth: Sorry, `onCreateViewHolder`.  Since that is only called when we are creating the `ViewHolder`, that seems like a good option.

vinaysshenoy: Ah. But that would mean then that the item view needs to maintain a reference to the entire list in some way so it can look up the item when a click happens.

msasikanth: When using `ListAdapter` we can just call `getItem()`, and `ViewHolder` should have the adapter position available in click listeners.

vinaysshenoy: That's not possible here since we have a generic view holder that wraps an `Item` generic type. These types do not have access to the parent adapter so they cannot invoke that method.

PR #1519, opened 2020-07-21
Appointment should be scheduled by default at the patient's assigned facility

msasikanth:
https://app.clubhouse.io/simpledotorg/story/217/appointment-should-be-scheduled-by-default-at-the-patient-s-assigned-facility

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Next Release
 ### Features
-- [In Progress: 20th Jul 2020] Show assigned facility in patient summary screen
+- Show assigned facility in the patient summary screen
+- [In Progress] Use patient's assigned facility to schedule an appointment
 
 ### Internal
 - Use `app:srcCompat` for `ImageView`
--- a/app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentEffectHandler.kt
@@ -84,7 +84,7 @@ class ScheduleAppointmentEffectHandler @AssistedInject constructor(
 
   private fun loadAppointmentFacility(): ObservableTransformer<LoadAppointmentFacility, ScheduleAppointmentEvent> {
     return ObservableTransformer { effects ->
-      effects.map { CurrentFacilityLoaded(currentFacility.get()) }
+      effects.map { AppointmentFacilityLoaded(currentFacility.get()) }
     }
   }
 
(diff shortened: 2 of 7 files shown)


app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentEffectHandler.kt
```diff
@@ -84,10 +89,20 @@ class ScheduleAppointmentEffectHandler @AssistedInject constructor(
 
   private fun loadAppointmentFacility(): ObservableTransformer<LoadAppointmentFacility, ScheduleAppointmentEvent> {
     return ObservableTransformer { effects ->
-      effects.map { AppointmentFacilityLoaded(currentFacility.get()) }
+      effects
+          .observeOn(schedulers.io())
+          .map { patientRepository.patientImmediate(it.patientUuid) }
+          .map { getAssignedFacility(it).toNullable() ?: currentFacility.get() }
```
vinaysshenoy, line 95: While we know that this is temporary, we have no idea how long it might be temporary for. In reality, it might be months or even more than a year till we reach the point where we can remove this.

I think we should load both the current facility and the patient's assigned facility in this screen and decide in `Update` which one to use.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Next Release
 ### Features
-- [In Progress: 20th Jul 2020] Show assigned facility in patient summary screen
+- [In Progress: 21th Jul 2020] Patient assigned facility
```
vinaysshenoy, line 5: Is there something wrong with the CHANGELOG here?

We can have two different lines, right?

- Show assigned facility in the patient summary screen
- [In Progress] Use patient's assigned facility to schedule an appointment

msasikanth: I feel like these changes all fall under patient assigned facility feature, should we track them separately?

vinaysshenoy: Yeah, I think we should. These are all small features that can go out separately.

msasikanth: Done

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentEffectHandler.kt
msasikanth: Done

app/src/main/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentUpdate.kt
```diff
@@ -36,6 +36,14 @@ class ScheduleAppointmentUpdate(
     }
   }
 
+  private fun appointmentFacilityLoaded(
+      model: ScheduleAppointmentModel,
+      event: AppointmentFacilitiesLoaded
+  ): Next<ScheduleAppointmentModel, ScheduleAppointmentEffect> {
+    val appointmentFacility = event.assignedFacility ?: event.currentFacility
```
vinaysshenoy, line 43: I am merging this for now, but can we add additional tests in the Mobius style to verify this behaviour? We should add similar tests to the effect handler as well.

msasikanth: Will do

PR #1520, opened 2020-07-21
Migrate registration loading screen to Mobius: prepare screen for migration

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/246/migrate-complete-registration-screen-to-mobius

### DIFF
--- a/app/src/main/java/org/simple/clinic/registration/register/RegistrationLoadingScreen.kt
+++ b/app/src/main/java/org/simple/clinic/registration/register/RegistrationLoadingScreen.kt
@@ -9,8 +9,8 @@ import io.reactivex.Observable
 import kotlinx.android.synthetic.main.screen_registration_loading.view.*
 import org.simple.clinic.R
 import org.simple.clinic.bindUiToController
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/simple/clinic/registration/register/RegistrationLoadingScreen.kt
```diff
@@ -30,6 +32,16 @@ class RegistrationLoadingScreen(
   @Inject
   lateinit var screenRouter: ScreenRouter
 
+  private val events by unsafeLazy {
+    Observable
+        .merge(
+            screenCreates(),
+            retryClicks()
+        )
+        .compose(ReportAnalyticsEvents())
```
msasikanth, line 47: If the controller didn't have `ReportAnalyticsEvents`, should we add it to new events stream during migration? 🤔

vinaysshenoy: Yes, this was missed in the initial screen implementation.

PR #1533, opened 2020-07-24
Extract facility picker UI into a separate component: Extract view from registration facility selection

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/253/extract-facility-picking-into-a-separate-view


app/src/main/res/layout/view_facilitypicker.xml
```diff
@@ -0,0 +1,76 @@
...
+
+    <LinearLayout
+      android:id="@+id/progressView"
+      android:layout_width="match_parent"
+      android:layout_height="match_parent"
+      android:layout_below="@+id/toolbarViewFlipper"
+      android:orientation="vertical"
+      android:visibility="gone">
+
+      <include layout="@layout/list_facility_selection_progress" />
```
msasikanth, line 46: Are we doing this to simulate the list shimmering effect?

app/src/main/java/org/simple/clinic/facilitypicker/FacilityPickerUiRenderer.kt
```diff
@@ -1,11 +1,53 @@
...
   override fun render(model: FacilityPickerModel) {
+    if (model.hasLoadedFacilities) {
+      ui.hideProgressIndicator()
+      renderFacilities(model)
+    } else {
+      ui.showProgressIndicator()
+    }
+
+    if (model.hasLoadedTotalFacilityCount) {
+      if (model.totalFacilityCount!! > 0) {
```
msasikanth: Couldn't we have checked for facilities list size instead? 🤔. That way we could avoid doing another DB query to get the count. Are we doing this to be more explicit?

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreen.kt
```diff
@@ -68,16 +47,15 @@ class RegistrationFacilitySelectionScreen(
...
 
     MobiusDelegate.forView(
         events = events.ofType(),
         defaultModel = RegistrationFacilitySelectionModel.create(screenKey.ongoingRegistrationEntry),
         update = RegistrationFacilitySelectionUpdate(),
         effectHandler = effectHandlerFactory.create(this).build(),
-        init = RegistrationFacilitySelectionInit.create(config),
-        modelUpdateListener = uiRenderer::render
+        init = RegistrationFacilitySelectionInit(),
+        modelUpdateListener = { /* Nothing to do here */ }
```
msasikanth, line 58: We can remove the `modelUpdateListener` since it already has a default implementation in `MobiusDelegate`

app/src/main/java/org/simple/clinic/facilitypicker/FacilityPickerView.kt
```diff
@@ -1,47 +1,93 @@
...
+    toolbarViewWithSearch.setNavigationOnClickListener { backClicked?.invoke() }
+    toolbarViewWithoutSearch.setNavigationOnClickListener { backClicked?.invoke() }
+
+    facilityRecyclerView.layoutManager = LinearLayoutManager(context)
+    facilityRecyclerView.adapter = recyclerViewAdapter
+
+    searchEditText.requestFocus()
+
+    // Hiding the keyboard without adding a post{} block doesn't seem to work.
+    post { hideKeyboard() }
```
msasikanth, line 68: Shouldn't we be opening keyboard when this screen is opened since the user can search for a facility?

vinaysshenoy: We intentionally do not show it because we expect the location permission to be given so that we can show nearby facilities at the top of the list and the nurse does not need to search most of the time.

app/src/main/java/org/simple/clinic/facilitypicker/FacilityPickerUiRenderer.kt
vinaysshenoy: Yes, they are two different things. In addition, we do not want this functionality to break if we move the screen to paging library later.

app/src/main/res/layout/view_facilitypicker.xml
vinaysshenoy: Yes. Although I am planning to remove this in order for a more consistent loading animation of the large indeterminate progress which we use everywhere else.

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreen.kt
vinaysshenoy: Good point, I'll make this change.

PR #1537, opened 2020-07-27
Migrate login pin screen to Mobius - Migrate

msasikanth:


app/src/main/java/org/simple/clinic/login/pin/LoginPinUpdate.kt
```diff
@@ -1,11 +1,25 @@
...
+    return when (event) {
+      is OngoingLoginEntryLoaded -> next(
+          model.ongoingLoginEntryUpdated(event.ongoingLoginEntry)
+      )
+      is LoginPinAuthenticated -> next(
+          model.ongoingLoginEntryUpdated(event.newLoginEntry),
+          SaveOngoingLoginEntry(event.newLoginEntry)
+      )
+      is LoginPinScreenUpdatedLoginEntry -> next(
+          model.ongoingLoginEntryUpdated(event.ongoingLoginEntry),
```
vinaysshenoy, line 19: Is it necessary to emit a model change here since the entry is not really changing at all and it is already present in the model?

msasikanth: Yeah, I wasn't sure about it either but kept it to make sure I haven't missed anything during migration. Will change it to just dispatch the effect.

msasikanth: Done

PR #1538, opened 2020-07-27
Update `materialButton` to `ProgressMaterialButton` in `PatientEntryScreen`

Honey14:
https://app.clubhouse.io/simpledotorg/story/498/add-progress-ui-for-registering-a-patient

### DIFF
--- a/app/src/main/java/org/simple/clinic/newentry/ButtonState.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/ButtonState.kt
@@ -1,0 +1,7 @@
+package org.simple.clinic.newentry
+
+enum class ButtonState {
+  SAVING,
+  SAVED
+}
+
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
@@ -41,6 +41,7 @@ import org.simple.clinic.patient.PatientRepository
 import org.simple.clinic.user.UserSession
 import org.simple.clinic.util.ValueChangedCallback
 import org.simple.clinic.util.scheduler.SchedulersProvider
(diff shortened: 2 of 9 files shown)


app/src/main/java/org/simple/clinic/newentry/ButtonState.kt
```diff
@@ -0,0 +1,6 @@
+package org.simple.clinic.newentry
+
+enum class ButtonState {
+  CLICKED,
```
msasikanth, line 4: Can we rename the button states to be more domain-specific like `SAVING`, `SAVED`? Currently, these are named to be more screen-specific and we usually avoid that particular kind of naming when placing inside the Mobius model.

app/src/main/java/org/simple/clinic/newentry/ButtonState.kt
```diff
@@ -0,0 +1,6 @@
...
+  ENABLED
+}
```
msasikanth, line 6: Please add a line feed here.

You can enable the setting in Android Studi / Idea to automatically add a line feed on save.
`Preferences > Editor > General > Ensure line feed at file end on Save`

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
```diff
@@ -147,7 +148,7 @@ class PatientEntryEffectHandler(
           .subscribeOn(scheduler)
           .flatMapSingle { savePatientEntry(it) }
           .doOnNext { patientRegisteredCount.set(patientRegisteredCount.get().plus(1)) }
-          .map { PatientEntrySaved }
+          .map { PatientEntrySaved(ButtonState.ENABLED) }
```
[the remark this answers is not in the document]
Honey14: I think that sounds right. We don't need to pass it from the effect handler.

PR #1541, opened 2020-07-27
Migrate login pin screen to Mobius - Cleanup & Sterlize

msasikanth:


app/src/main/java/org/simple/clinic/login/pin/LoginPinScreen.kt
```diff
@@ -142,11 +122,6 @@ class LoginPinScreen(context: Context, attrs: AttributeSet) : RelativeLayout(con
     return backClicksFromView.mergeWith(backClicksFromSystem)
   }
 
-  private fun otpReceived(): Observable<LoginPinOtpReceived>? {
```
vinaysshenoy, line 145: Was this event not being used anywhere?

vinaysshenoy: Saw the deleted event and the related comment.

PR #1543, opened 2020-07-28
Use facility picker view in all facility selection screens

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/253/extract-facility-picking-into-a-separate-view


app/src/main/java/org/simple/clinic/facility/change/confirm/FacilityChangeComponent.kt
```diff
@@ -2,10 +2,14 @@ package org.simple.clinic.facility.change.confirm
...
 import dagger.Subcomponent
 import org.simple.clinic.activity.BindsActivity
+import org.simple.clinic.di.AssistedInjectModule
 import org.simple.clinic.facility.change.FacilityChangeActivity
+import org.simple.clinic.facilitypicker.FacilityPickerView
 
-@Subcomponent
-interface FacilityChangeComponent {
+@Subcomponent(
+    modules = [AssistedInjectModule::class]
```
msasikanth, line 10: I am guessing we are doing this because `AssistedInjectModule` that is used at `AppComponent` level scope wont work with subcomponents?

vinaysshenoy: We cannot use `AssistedInjectedModule` at the app level because there are dependencies that require access to the `Activity` it runs in. So we need to add it separately onto every subcomponent that uses it.

PR #1546, opened 2020-07-28
Unify search screens in preparation for Mobius migration

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/580/unity-search-screens


vinaysshenoy:
@Honey14 you might to pair with @msasikanth on this PR since it's pretty large and is mostly in the legacy architecture. Most of the change is deletions, though, so shouldn't be too much of a pain to review hopefully.

vinaysshenoy:
> Looks good. Should we add the internal changeLog for this?

Good idea, I will add it in the next PR.

PR #1552, opened 2020-07-30
Migrate forgot confirm pin screen to Mobius - Prepare

msasikanth:
https://app.clubhouse.io/simpledotorg/story/608/migrate-forgot-confirm-pin-screen-to-mobius

### DIFF
--- a/app/src/main/java/org/simple/clinic/forgotpin/confirmpin/ForgotPinConfirmPinModel.kt
+++ b/app/src/main/java/org/simple/clinic/forgotpin/confirmpin/ForgotPinConfirmPinModel.kt
@@ -1,0 +1,7 @@
+package org.simple.clinic.forgotpin.confirmpin
+
+import android.os.Parcelable
+import kotlinx.android.parcel.Parcelize
+
+@Parcelize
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/simple/clinic/forgotpin/confirmpin/ForgotPinConfirmPinModel.kt
```diff
@@ -4,4 +4,9 @@ import android.os.Parcelable
 import kotlinx.android.parcel.Parcelize
 
 @Parcelize
-class ForgotPinConfirmPinModel : Parcelable
+class ForgotPinConfirmPinModel : Parcelable {
+
+  companion object {
```
Honey14, line 9: Should the companion object be added in `Add no-op Mobius classes` commit?

msasikanth: Since we will only start using it once we set up the test fixture it doesn't matter. I personally preferer adding during this commit but we can also add it in no-op if we want.

PR #1553, opened 2020-07-30
Update `ProgressMaterialButton` in `EditPatientScreen`

Honey14:
https://app.clubhouse.io/simpledotorg/story/499/add-progress-ui-for-editing-a-patient

### DIFF
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientModel.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientModel.kt
@@ -19,7 +19,8 @@ data class EditPatientModel(
     val savedPatient: Patient,
     val savedAddress: PatientAddress,
     val savedPhoneNumber: PatientPhoneNumber?,
-    val savedBangladeshNationalId: BusinessId?
+    val savedBangladeshNationalId: BusinessId?,
+    val saveButtonState: EditPatientState?
 ) : Parcelable {
   companion object {
     fun from(
@@ -27,17 +28,19 @@ data class EditPatientModel(
         address: PatientAddress,
         phoneNumber: PatientPhoneNumber?,
         dateOfBirthFormatter: DateTimeFormatter,
-        bangladeshNationalId: BusinessId?
+        bangladeshNationalId: BusinessId?,
+        saveButtonState: EditPatientState?
     ): EditPatientModel {
       val savedEntry = EditablePatientEntry.from(
           patient,
           address,
           phoneNumber,
           dateOfBirthFormatter,
-          bangladeshNationalId
+          bangladeshNationalId,
+          saveButtonState
       )
       val ongoingEntry = savedEntry.copy()
-      return EditPatientModel(savedEntry, ongoingEntry, patient, address, phoneNumber, bangladeshNationalId)
+      return EditPatientModel(savedEntry, ongoingEntry, patient, address, phoneNumber, bangladeshNationalId, saveButtonState)
     }
   }
 
@@ -73,4 +76,7 @@ data class EditPatientModel(
 
   fun updateStreetAddress(streetAddress: String): EditPatientModel =
       copy(ongoingEntry = ongoingEntry.updateStreetAddress(streetAddress))
+
+  fun buttonStateChanged(buttonState: EditPatientState?): EditPatientModel =
+      copy(saveButtonState = buttonState)
 }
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/simple/clinic/editpatient/EditPatientViewRenderer.kt
```diff
@@ -19,5 +20,15 @@ class EditPatientViewRenderer(private val ui: EditPatientUi) : ViewRenderer<Edit
...
     ui.setDateOfBirthAndAgeVisibility(fieldToShow)
+
+    manageButtonState(model)
+  }
+
+  private fun manageButtonState(model: EditPatientModel) {
+    if (model.saveButtonState == ButtonState.SAVING) {
+      ui.showProgressUntilPatientUpdated()
+    } else {
+      ui.patientUpdated()
```
vinaysshenoy, line 31: This state is not exactly `patientUpdated` but more like `patient update is not happening`. Can we rename the methods to better reflect this?

In addition, the UI need not care about the conditions for showing the progress since that responsibility is a part of the UI renderer and we can drop the `UntilPatientUpdated` from the method name.

Suggestions:
- `showProgress`
- `hideProgress`

app/src/main/java/org/simple/clinic/editpatient/EditPatientUpdate.kt
```diff
@@ -51,7 +52,7 @@ class EditPatientUpdate(
       is ZoneChanged -> next(model.updateZone(event.zone))
       is StreetAddressChanged -> next(model.updateStreetAddress(event.streetAddress))
       is BackClicked -> onBackClicked(model)
-      is PatientSaved -> dispatch(GoBackEffect)
+      is PatientSaved -> next(model.buttonStateChanged(ButtonState.SAVED), GoBackEffect)
```
vinaysshenoy, line 55: This screen does not have a valid "Saved" state since we close the screen when that happens. The model should reflect that.

I think we can introduce a new enum here with two states:

- `SavingPatient`
- `NotSavingPatient`

And use these to represent the state.

app/src/main/java/org/simple/clinic/editpatient/EditPatientModel.kt
```diff
@@ -19,7 +20,8 @@ data class EditPatientModel(
     val savedPatient: Patient,
     val savedAddress: PatientAddress,
     val savedPhoneNumber: PatientPhoneNumber?,
-    val savedBangladeshNationalId: BusinessId?
+    val savedBangladeshNationalId: BusinessId?,
+    val saveButtonState: ButtonState? = null
```
vinaysshenoy, line 24: Can we avoid using default values in the constructors? This should be set in the factory function.

vinaysshenoy, line 24: Does this need to be nullable? This reflects the state of the button and it does not need any async work and can be defaulted to the non-progress state, right?

app/src/main/java/org/simple/clinic/editpatient/EditPatientUpdate.kt
Honey14: `NotSavingPatient` doesn't make sense to me, what do you think? 

States 
* `SavingPatient` concludes button state in-progress
* `SavedPatient` concludes success and button state can now be enabled

vinaysshenoy: The problem is that `SavedPatient` is not a good name for a state, but a name for an event. We need a name that indicates this.

app/src/main/java/org/simple/clinic/editpatient/EditPatientModel.kt
Honey14: Yes it can be, keeping it enabled as the default value

vinaysshenoy: Was this done? The type itself is still nullable (`EditPatientStates?`).

app/src/main/java/org/simple/clinic/editpatient/EditPatientStates.kt
```diff
@@ -0,0 +1,6 @@
+package org.simple.clinic.editpatient
+
+enum class EditPatientStates {
```
vinaysshenoy, line 3: Can we rename this enum type to be in the singular (`EditPatientState`)?

PR #1554, opened 2020-07-31
Migrate facility selection activity to Mobius

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/286/migrate-facility-selection-activity-to-mobius


app/src/main/java/org/simple/clinic/scheduleappointment/facilityselection/FacilitySelectionUi.kt
```diff
@@ -0,0 +1,7 @@
+package org.simple.clinic.scheduleappointment.facilityselection
+
+import org.simple.clinic.facility.Facility
+
+interface FacilitySelectionUi {
+  fun sendSelectedFacility(selectedFacility: Facility)
```
msasikanth: I have a question, we know this call is a UI action call and this screen doesn't have any UI to render on-screen (at least for this loop). So should we have gone with extracting out `UiActions` and not implemented UI renderer? 🤔

vinaysshenoy: We could have, yes. I just ended up doing it this way out of habit.

PR #1580, opened 2020-08-03
Migrate forgot confirm pin screen to Mobius - Cleanup & Sterilise

msasikanth:
https://app.clubhouse.io/simpledotorg/story/608/migrate-forgot-confirm-pin-screen-to-mobius


app/src/main/java/org/simple/clinic/forgotpin/confirmpin/ForgotPinConfirmPinUi.kt
```diff
@@ -1,6 +1,6 @@
 package org.simple.clinic.forgotpin.confirmpin
 
-interface ForgotPinConfirmPinUi : UiActions {
```
vinaysshenoy, line 3: Can we rename the `UiActions` interface to be prefixed with the `ForgotPinConfirmPin` name?

msasikanth: Will do 👍🏾

PR #1581, opened 2020-08-03
Migrate overdue screen to paging library

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/373/migrate-overdue-screen-to-mobius


app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
```diff
@@ -135,6 +136,21 @@ data class OverdueAppointment(
         scheduledAfter: LocalDate
     ): Observable<List<OverdueAppointment>>
 
+    @Query("""
+      SELECT * FROM OverdueAppointment
+      WHERE 
+        appt_facilityUuid = :facilityUuid
```
msasikanth, line 142: Weren't we supposed to load overdue list based on patient assigned facility id?

msasikanth: For `overdueAtFacility` we have used this `IFNULL(patientAssignedFacilityUuid, appt_facilityUuid) = :facilityUuid`

app/src/main/java/org/simple/clinic/home/overdue/OverdueEffectHandler.kt
```diff
@@ -25,8 +26,8 @@ class OverdueEffectHandler @AssistedInject constructor(
     return RxMobius
         .subtypeEffectHandler<OverdueEffect, OverdueEvent>()
         .addTransformer(LoadCurrentFacility::class.java, loadCurrentFacility())
-        .addTransformer(LoadOverdueAppointments::class.java, loadOverdueAppointments())
-        .addConsumer(OpenContactPatientScreen::class.java, { uiActions.openPhoneMaskBottomSheet(it.patientUuid)}, schedulers.ui())
+        .addConsumer(LoadOverdueAppointments::class.java, ::loadOverdueAppointments, schedulers.ui())
```
msasikanth, line 29: Doesn't data source factory change the thread when we attach to paging adapter? Do we have manually change the thread here?

app/src/main/java/org/simple/clinic/di/PagingModule.kt
```diff
@@ -0,0 +1,33 @@
...
+import androidx.paging.PagedList
+import dagger.Module
+import dagger.Provides
+import javax.inject.Named
+
+@Module
+class PagingModule {
+
+  @Provides
+  @Named("for_measurement_history")
```
msasikanth, line 13: We should perhaps start using `Qualifiers` later on to be more type-safe, WDYT?

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
vinaysshenoy: Oh yes, good catch! This screen was done before the assigned facility work was done, so I think I missed updating it while I rerolled. Will fix.

app/src/main/java/org/simple/clinic/home/overdue/OverdueEffectHandler.kt
vinaysshenoy: Hmm, I am not sure about this. I added the thread switching out of habit.

app/src/main/java/org/simple/clinic/di/PagingModule.kt
vinaysshenoy: Good idea, I haven't worked with them before. Maybe you could create an issue on GitHub which describes what this would look like and we can decide on an approach?

app/src/main/java/org/simple/clinic/home/overdue/OverdueEffectHandler.kt
vinaysshenoy: Yeah, we will need to call it on the UI thread because the data source internally subscribes to the screen detaches which needs to be invoked on the UI thread.

app/src/main/java/org/simple/clinic/di/PagingModule.kt
msasikanth: Will do that 👍🏾

PR #1582, opened 2020-08-04
Add progress UI when saving blood pressure

Honey14:
https://app.clubhouse.io/simpledotorg/story/496/add-progress-ui-for-saving-blood-pressure-measurements

### DIFF
--- a/app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
+++ b/app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
@@ -598,6 +598,7 @@ class BloodPressureEntrySheetLogicTest {
     )
     verify(ui).hideRemoveBpButton()
     verify(ui).showEnterNewBloodPressureTitle()
+    verify(ui, times(3)).hideProgress()
     verifyNoMoreInteractions(ui)
   }
 
@@ -621,6 +622,7 @@ class BloodPressureEntrySheetLogicTest {
     verify(ui).setDiastolic(bp.reading.diastolic.toString())
     verify(ui).showRemoveBpButton()
     verify(ui).showEditBloodPressureTitle()
+    verify(ui, times(4)).hideProgress()
     verifyNoMoreInteractions(ui)
   }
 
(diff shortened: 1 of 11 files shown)


app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryLogicTest.kt
```diff
@@ -598,6 +598,7 @@ class BloodPressureEntrySheetLogicTest {
     )
```
vinaysshenoy, line 598: When we add new functionality to a screen migrated from the old legacy architecture, we should add relevant tests in the Mobius style. Can we do this for the `Update` and `ViewRenderer` changes we have introduced in this PR?

Honey14: Added tests for `ViewRenderer` and I think we already have tests for saving BP's. Can you check once?

vinaysshenoy: I meant adding `UpdateSpec` tests for verifying that the state flags are being set correctly.

Honey14: Got it

app/src/test/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdateTest.kt
```diff
@@ -60,4 +65,113 @@ class BloodPressureEntryUpdateTest {
...
+        .then(
+            assertThatNext(
+                hasModel(defaultModel.bloodPressureStateChanged(BloodPressureSaveState.NOT_SAVING_BLOOD_PRESSURE)),
+                hasEffects(SetBpSavedResultAndFinish as BloodPressureEntryEffect)
+            )
+        )
+  }
+
+  @Test
+  fun `when the save button is clicked, then show no change`() {
```
vinaysshenoy, line 138: The name of this test does not indicate that the fact that it is testing the condition where a measurement save operation is ongoing. Please update?

PR #1585, opened 2020-08-04
Migrate Enter OTP screen to Mobius: Stabilize the tests

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/502/migrate-enter-otp-screen-to-mobius

### DIFF
--- a/app/src/test/java/org/simple/clinic/enterotp/EnterOtpScreenControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/enterotp/EnterOtpScreenControllerTest.kt
@@ -55,7 +55,12 @@ class EnterOtpScreenControllerTest {
   private val phoneNumber = "1234567890"
   private val loggedInUserUuid = UUID.fromString("13e563ba-a148-4b5d-838d-e38d42dca72c")
   private val registrationFacilityUuid = UUID.fromString("f33bfc01-f595-42ce-8ad8-b70150ccbde2")
-  private val user = TestData.loggedInUser(uuid = loggedInUserUuid, phone = phoneNumber)
+  private val user = TestData.loggedInUser(
+      uuid = loggedInUserUuid,
+      phone = phoneNumber,
+      status = ApprovedForSyncing,
+      loggedInStatus = OTP_REQUESTED
+  )
   private val ongoingLoginEntry = OngoingLoginEntry(
       uuid = loggedInUserUuid,
       phoneNumber = phoneNumber,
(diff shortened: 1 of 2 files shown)


app/src/test/java/org/simple/clinic/enterotp/EnterOtpScreenControllerTest.kt
```diff
@@ -113,57 +111,45 @@ class EnterOtpScreenControllerTest {
...
 
   @Test
-  @Parameters(
-      "|false",
-      "1|false",
-      "11|false",
-      "111|false",
-      "1111|false",
-      "11111|false",
-      "111111|true"
```
Honey14, line 151: Removing parameters will also remove multiple test cases, right?

app/src/test/java/org/simple/clinic/enterotp/EnterOtpScreenControllerTest.kt
```diff
@@ -387,100 +390,166 @@ class EnterOtpScreenControllerTest {
...
-    )
+    uiEvents.onNext(EnterOtpResendSmsClicked())
+
+    verify(screen, never()).showSmsSentMessage()
+  }
+
+  @Test
+  fun `when the resend sms call fails with a server error, the sms sent message must not be shown`() {
+    whenever(activateUser.activate(loggedInUserUuid, pin))
+        .doReturn(ActivateUser.Result.ServerError(400))
```
Honey14, line 731: Should the `ServerError` code be 500 instead or 400 is fine?

app/src/test/java/org/simple/clinic/enterotp/EnterOtpScreenControllerTest.kt
vinaysshenoy: Yeah, this was intentional. I felt the overhead of having parameterized tests was not worth the effort of testing each edge case here.

app/src/test/java/org/simple/clinic/enterotp/EnterOtpScreenControllerTest.kt
vinaysshenoy: Good point; Although it makes no difference to the tests, we should strive to match real world scenarios. I'll make this change in a later PR.

app/src/test/java/org/simple/clinic/enterotp/EnterOtpScreenControllerTest.kt
Honey14: So we test the success case and failed case only, right?

PR #1592, opened 2020-08-06
Migrate patient search screen to Mobius: Strangulate and remove controller

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/639/migrate-patient-search-screen-to-mobius


app/src/main/java/org/simple/clinic/search/PatientSearchUpdate.kt
```diff
@@ -1,14 +1,33 @@
...
 import org.simple.clinic.mobius.next
 
 class PatientSearchUpdate : Update<PatientSearchModel, PatientSearchEvent, PatientSearchEffect> {
 
   override fun update(model: PatientSearchModel, event: PatientSearchEvent): Next<PatientSearchModel, PatientSearchEffect> {
-    return when(event) {
+    return when (event) {
       is SearchQueryTextChanged -> next(model.queryChanged(event.text))
+      is SearchClicked -> {
+        val validationErrors = validateInput(model.enteredQuery)
```
Honey14: Can we extract this in a private function?

Honey14:
Tests were running fine in the build. Can you check this once?

app/src/main/java/org/simple/clinic/search/PatientSearchUpdate.kt
vinaysshenoy: It was done in a later commit in the same PR.

vinaysshenoy:
> Tests were running fine in the build. Can you check this once?

I looked at the logs and it seems like it wasn't able to fetch a particular dependency. Running again.

PR #1595, opened 2020-08-06
Migrate scan Simple ID screen to Mobius - Migrate

msasikanth:
https://app.clubhouse.io/simpledotorg/story/688/migrate-scan-simple-id-screen-to-mobius

### DIFF
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
@@ -23,6 +23,7 @@ class ScanSimpleIdEffectHandler @AssistedInject constructor(
       .addAction(HideShortCodeValidationError::class.java, uiActions::hideShortCodeValidationError, schedulersProvider.ui())
       .addConsumer(ShowShortCodeValidationError::class.java, { uiActions.showShortCodeValidationError(it.failure) }, schedulersProvider.ui())
       .addTransformer(ValidateShortCode::class.java, validateShortCode())
+      .addConsumer(OpenPatientShortCodeSearch::class.java, { uiActions.openPatientShortCodeSearch(it.shortCode) }, schedulersProvider.ui())
       .build()
 
   private fun validateShortCode(): ObservableTransformer<ValidateShortCode, ScanSimpleIdEvent> {
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
@@ -25,6 +25,7 @@ class ScanSimpleIdEffectHandler @AssistedInject constructor(
       .addTransformer(ValidateShortCode::class.java, validateShortCode())
       .addConsumer(OpenPatientShortCodeSearch::class.java, { uiActions.openPatientShortCodeSearch(it.shortCode) }, schedulersProvider.ui())
       .addConsumer(OpenPatientSummary::class.java, { uiActions.openPatientSummary(it.patientUuid) }, schedulersProvider.ui())
+      .addConsumer(OpenAddIdToPatientScreen::class.java, { uiActions.openAddIdToPatientScreen(it.identifier) }, schedulersProvider.ui())
       .build()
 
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -1,17 +1,29 @@
...
       ShortCodeChanged -> dispatch(HideShortCodeValidationError)
-      is ShortCodeValidated -> noChange()
+      is ShortCodeValidated -> shortCodeValidated(model, event)
+      is ShortCodeSearched -> next(model.shortCodeChanged(event.shortCode), ValidateShortCode(event.shortCode))
     }
   }
+
+  private fun shortCodeValidated(model: ScanSimpleIdModel, event: ShortCodeValidated): Next<ScanSimpleIdModel, ScanSimpleIdEffect> {
+    val effect = when (event.result) {
+      Success -> OpenPatientShortCodeSearch(model.shortCode?.shortCodeText.orEmpty())
```
vinaysshenoy, line 50: Since if it's a `Success`, we know for certain that the `shortCode` property in the model will not be `null`, we could use the `!!` and skip the null handling.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
```diff
@@ -26,8 +29,23 @@ class ScanSimpleIdEffectHandler @AssistedInject constructor(
...
+    return ObservableTransformer { effects ->
+      val scannedUuid = effects
+          .map { it.scannedUuid }
+          .share()
+
+      effects
+          .observeOn(schedulersProvider.io())
+          .flatMap { patientRepository.findPatientWithBusinessId(it.scannedUuid.toString()) }
+          .withLatestFrom(scannedUuid)
+          .map { (patient, scannedUuid) -> PatientSearchCompleted(patient, scannedUuid) }
```
vinaysshenoy, line 45: I tried to get rid of the `.share()` operator usage here and make this into a single chain.

```kotlin
effects
          .switchMap { effect ->
            val scannedUuid = effect.scannedUuid

            patientRepository
                .findPatientWithBusinessId(scannedUuid.toString())
                .subscribeOn(schedulersProvider.io())
                .map { PatientSearchCompleted(it, scannedUuid) }
          }
```

What do you think about using this instead?

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenController.kt
```diff
@@ -26,54 +17,8 @@ class ScanSimpleIdScreenController @Inject constructor(
...
-  private fun mergeWithScannedBpPassportCodes() = ObservableTransformer<UiEvent, UiEvent> { events ->
-    val scannedBpPassportCodes = events
-        .ofType<ScanSimpleIdScreenQrCodeScanned>()
-        .map { it.text }
-        .map { scannedQrCode ->
-          try {
-            val bpPassportCode = UUID.fromString(scannedQrCode)
-            ValidPassportCode(bpPassportCode)
-          } catch (e: IllegalArgumentException) {
-            InvalidPassportCode
```
vinaysshenoy, line 51: I see this is not being handled in the update function. IIRC, we just ignore invalid QR codes. Just confirming that ignoring it is the expected behaviour.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
msasikanth: Done

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
msasikanth: Yeah, this works 👍🏾

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreenController.kt
msasikanth: Yes, ignoring invalid QR codes is the expected behaviour

PR #1596, opened 2020-08-06
Migrate Enter OTP screen to Mobius: Prepare screen

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/502/migrate-enter-otp-screen-to-mobius

### DIFF
--- a/app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
+++ b/app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
@@ -11,6 +11,7 @@ import com.jakewharton.rxbinding3.view.detaches
 import com.jakewharton.rxbinding3.widget.editorActions
 import io.reactivex.Observable
 import kotlinx.android.synthetic.main.screen_enterotp.view.*
+import org.simple.clinic.LOGIN_OTP_LENGTH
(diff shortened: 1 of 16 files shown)


app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
```diff
@@ -85,25 +88,25 @@ class EnterOtpScreen(context: Context, attributeSet: AttributeSet) : RelativeLay
     userPhoneNumberTextView.text = phoneNumberWithCountryCode
   }
 
-  fun goBack() {
+  override fun goBack() {
```
msasikanth, line 137: Since we are no longer using `goBack` in the controller, should we have no extracted it in the UI interface?

vinaysshenoy: Good question, let me check and see if we can remove this.

vinaysshenoy: I checked and it's still used in the controller. We only moved handling of back clicks, but the method itself is still invoked from the controller if the login OTP arrives and the user gets logged in in the background.

PR #1598, opened 2020-08-06
Fix issue where recent and overdue patients on home screen would not update after changing facilities

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/742/home-screen-does-not-update-when-changing-facilities

## Reason
Both `HomeScreen` and `OverdueScreen` are in the screen where the changing of current facility happens. Since `FacilityChangeActivity` is a new activity and not a new screen in `TheActivity`, returning to `HomeScreen` after switching the current facility does not cause the screen to be reinitialized and thus, does not notify a change in the current facility.

## Fix
This PR fixes the issue by changing the two screen to observe the current facility for changes instead of just loading it once.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@
 - Migrated `HelpScreen` to Mobius
 - [In Progress: 05 Aug 2020] Migrate `HelpScreen` to Mobius
 
+### Fixes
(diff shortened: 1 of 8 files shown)


Honey14:
Might be the downloading lib issue. Can you try pushing this again? @vinaysshenoy

vinaysshenoy:
> Might be the downloading lib issue. Can you try pushing this again? @vinaysshenoy

Yup, this was the issue. Reran the tests and they pass.

If you see this happening, you can rerun the tests yourself from the GitHub Actions menu.
Screenshot 2020-08-07 at 11 05 10 AM

PR #1599, opened 2020-08-06
Migrate home screen to Mobius - Migrate

msasikanth:
https://app.clubhouse.io/simpledotorg/story/726/migrate-home-screen-to-mobius

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/HomeScreenEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/home/HomeScreenEffectHandler.kt
@@ -18,5 +18,6 @@ class HomeScreenEffectHandler @AssistedInject constructor(
 
   fun build(): ObservableTransformer<HomeScreenEffect, HomeScreenEvent> = RxMobius
       .subtypeEffectHandler<HomeScreenEffect, HomeScreenEvent>()
+      .addAction(OpenFacilitySelection::class.java, uiActions::openFacilitySelection, schedulersProvider.ui())
       .build()
 }
--- a/app/src/main/java/org/simple/clinic/home/HomeScreenEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/home/HomeScreenEffectHandler.kt
@@ -5,12 +5,17 @@ import com.squareup.inject.assisted.Assisted
 import com.squareup.inject.assisted.AssistedInject
 import io.reactivex.ObservableTransformer
 import org.simple.clinic.facility.FacilityRepository
+import org.simple.clinic.overdue.AppointmentRepository
 import org.simple.clinic.user.UserSession
+import org.simple.clinic.util.UserClock
 import org.simple.clinic.util.scheduler.SchedulersProvider
+import java.time.LocalDate
 
 class HomeScreenEffectHandler @AssistedInject constructor(
     private val userSession: UserSession,
     private val facilityRepository: FacilityRepository,
+    private val appointmentRepository: AppointmentRepository,
+    private val userClock: UserClock,
     private val schedulersProvider: SchedulersProvider,
     @Assisted private val uiActions: HomeScreenUiActions
 ) {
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/simple/clinic/home/HomeScreenEffectHandler.kt
```diff
@@ -19,5 +23,16 @@ class HomeScreenEffectHandler @AssistedInject constructor(
...
       .addAction(OpenFacilitySelection::class.java, uiActions::openFacilitySelection, schedulersProvider.ui())
+      .addTransformer(LoadCurrentFacility::class.java, loadCurrentFacility())
       .build()
+
+  private fun loadCurrentFacility(): ObservableTransformer<LoadCurrentFacility, HomeScreenEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .flatMap { userSession.requireLoggedInUser() }
+          .flatMap { facilityRepository.currentFacility(it) }
```
vinaysshenoy, line 52: In a recent PR (#1598), we added a way to inject a stream of the current facility as `Observable<Facility>`. In the final cleanup of this screen, we might want to switch to using this.

app/src/main/java/org/simple/clinic/home/HomeScreenEffectHandler.kt
```diff
@@ -24,8 +29,21 @@ class HomeScreenEffectHandler @AssistedInject constructor(
...
       .addAction(OpenFacilitySelection::class.java, uiActions::openFacilitySelection, schedulersProvider.ui())
       .addTransformer(LoadCurrentFacility::class.java, loadCurrentFacility())
+      .addTransformer(LoadOverdueAppointmentCount::class.java, loadOverdueAppointmentCount())
       .build()
 
+  private fun loadOverdueAppointmentCount(): ObservableTransformer<LoadOverdueAppointmentCount, HomeScreenEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .flatMap { (facility) ->
```
vinaysshenoy, line 39: We should ideally be using `.switchMap` by default and only using `.flatMap` where it is required.

app/src/main/java/org/simple/clinic/home/HomeScreenUiRenderer.kt
```diff
@@ -8,6 +8,18 @@ class HomeScreenUiRenderer(private val ui: HomeScreenUi) : ViewRenderer<HomeScre
...
     }
+
+    if (model.hasOverdueAppointmentCount) {
+      renderOverdueAppointmentCount(model)
+    }
+  }
+
+  private fun renderOverdueAppointmentCount(model: HomeScreenModel) {
+    val overdueAppointmentCount = model.overdueAppointmentCount!!
+    if (overdueAppointmentCount > 0)
```
vinaysshenoy, line 19: What do you think about making this check a part of `HomeScreenModel` as a computed property, say, `hasAtLeastOneOverdueAppointment`?

app/src/main/java/org/simple/clinic/home/HomeScreenEffectHandler.kt
msasikanth: Correct me if I am wrong, we need to use `switchMap` to get the latest observable stream right? Does that mean `flatMap` will retain older streams?

app/src/main/java/org/simple/clinic/home/HomeScreenUiRenderer.kt
msasikanth: Oh yeah, I should have done that. Completely missed it, will do it in final PR 👍🏾

app/src/main/java/org/simple/clinic/home/HomeScreenEffectHandler.kt
vinaysshenoy: Yes, that's exactly what `flatMap` does.

vinaysshenoy: If you look at the source code for [`ObservableFlatMap`](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/observable/ObservableFlatMap.java) and [`ObservableSwitchMap`](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/observable/ObservableSwitchMap.java), you'll see the different between the two.

PR #1604, opened 2020-08-07
UI updates when creating blood pressure

Honey14:
https://app.clubhouse.io/simpledotorg/story/496/add-progress-ui-for-saving-blood-pressure-measurements


app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -392,15 +392,13 @@ class BloodPressureEntrySheet : BottomSheetActivity(), BloodPressureEntryUi, Rem
```
msasikanth, line 392: Can you rename the commit message to "Move `ProgressBar` inside the parent layout" please?

PR #1605, opened 2020-08-10
Bump Dagger to 2.28.3

msasikanth:
Since Dagger 2.25 they have added the support to add annotation like `Named`, `Qualifier`s without using Kotlin `field`. I have migrated that usage in this commit as well.

So when injecting a dependency with `Named` or `Qualifier` we can do this
```kt
@Inject
@Named("sample_dependency")
```
instead of these
```kt
// Approach 1
@field:[Inject Named("sample_dependency")]

// Approach 2
@Inject
@field:Named("sample_dependency")
```

This dependency update will also allow us to use new Dagger changes like
- Dagger lint
- Improved error messaging
- Hilt (if interested)


vinaysshenoy:
This looks good to me.

PR #1609, opened 2020-08-10
Migrate logged out of device dialog to Mobius - Stabilise & Prepare

msasikanth:
https://app.clubhouse.io/simpledotorg/story/784/migrate-logged-out-of-device-dialog-to-mobius

### DIFF
--- a/app/src/test/java/org/simple/clinic/registration/phone/loggedout/LoggedOutOfDeviceDialogControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/registration/phone/loggedout/LoggedOutOfDeviceDialogControllerTest.kt
@@ -9,6 +9,7 @@ import com.nhaarman.mockitokotlin2.whenever
 import io.reactivex.Single
 import io.reactivex.disposables.Disposable
 import io.reactivex.plugins.RxJavaPlugins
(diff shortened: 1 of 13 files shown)


app/src/test/java/org/simple/clinic/registration/phone/loggedout/LoggedOutOfDeviceDialogControllerTest.kt
```diff
@@ -32,23 +33,21 @@ class LoggedOutOfDeviceDialogControllerTest {
...
-        .subscribe({ uiChange -> uiChange(dialog) }, { throw it })
+  @After
+  fun tearDown() {
+    controllerSubscription.dispose()
   }
 
   @Test
   fun `when the dialog is created, the okay button must be disabled`() {
+    // given
+    RxJavaPlugins.setErrorHandler(null)
```
vinaysshenoy, line 48: We can pass the error handler as a parameter to `setupController` and avoid setting this up in each test, right?

msasikanth: Ah right, will do that in next PR.

PR #1613, opened 2020-08-11
Migrate patient search results screen to Mobius: Strangulate controller

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/708/migrate-search-results-view-to-mobius

### DIFF
--- a/app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsEffectHandler.kt
@@ -5,12 +5,17 @@ import com.squareup.inject.assisted.Assisted
 import com.squareup.inject.assisted.AssistedInject
 import io.reactivex.Observable
 import io.reactivex.ObservableTransformer
+import org.simple.clinic.facility.FacilityRepository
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsEffectHandler.kt
```diff
@@ -43,4 +49,21 @@ class PatientSearchResultsEffectHandler @AssistedInject constructor(
...
     }
   }
+
+  private fun openPatientEntryScreen(): ObservableTransformer<OpenPatientEntryScreen, PatientSearchResultsEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulers.io())
+          .switchMap {
+            userSession
+                .loggedInUser()
```
Honey14, line 59: We only need `Facility` in `openPatientEntryScreen`, right? Why do we need to add loggedInUser here?

vinaysshenoy: `FacilityRepository#currentFacility` requires the current user a parameter to fetch the current facility.

PR #1614, opened 2020-08-11
Accept and save user signature

Honey14:
https://app.clubhouse.io/simpledotorg/story/735/implement-a-screen-to-accept-and-save-the-user-signature


app/src/main/java/org/simple/clinic/signature/SignatureActivity.kt
```diff
@@ -0,0 +1,12 @@
...
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatActivity
+import org.simple.clinic.R
+
+class SignatureActivity : AppCompatActivity() {
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.activity_signature)
+  }
+}
```
msasikanth, line 92: Please add line break at the end of the file. You can enable this setting in Android Studio

app/src/main/res/drawable/ic_accept.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
msasikanth, line 1: Please rename this file to `ic_signature_accept` to be more accurate.

app/src/main/res/drawable/ic_accept.xml
```diff
@@ -0,0 +1,10 @@
...
+    android:width="64dp"
```
msasikanth, line 2: Can you run code format on this file, please?

app/src/main/res/drawable/ic_undo.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
msasikanth, line 1: Please rename this file to `ic_signature_undo`

app/src/main/res/drawable/ic_undo.xml
```diff
@@ -0,0 +1,10 @@
...
+    android:width="64dp"
```
msasikanth, line 2: Can you run code format on this file, please?

app/src/main/res/layout/activity_signature.xml
```diff
@@ -0,0 +1,76 @@
...
+
+    <LinearLayout
+      android:layout_width="match_parent"
+      android:layout_height="wrap_content"
+      android:layout_below="@+id/linearLayout"
+      android:layout_marginTop="@dimen/spacing_16"
+      android:layout_marginBottom="@dimen/spacing_16"
+      android:orientation="horizontal">
+
+      <ImageView
```
msasikanth, line 59: Looks like you forgot to suppress `ContentDescription` warning here.

`tools:ignore="ContentDescription" `

app/src/main/res/layout/activity_signature.xml
```diff
@@ -0,0 +1,76 @@
...
+        android:id="@+id/clear_signature"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_weight="1"
+        android:src="@drawable/ic_undo" />
```
msasikanth, line 64: We should not be using this to set image resource, we have started using `app:srcCompat` since `ImageView` will be inflated as `AppCompatImageView` by AppCompat and we want to avoid any unnecessary issues with the resource.

app/src/main/java/org/simple/clinic/signature/SignatureActivity.kt
```diff
@@ -0,0 +1,12 @@
+package org.simple.clinic.signature
+
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatActivity
+import org.simple.clinic.R
+
+class SignatureActivity : AppCompatActivity() {
```
msasikanth: Instead of `Activity`, why aren't we just using a Dialog? 🤔

app/src/main/java/org/simple/clinic/signature/SignatureEffect.kt
```diff
@@ -0,0 +1,3 @@
+package org.simple.clinic.signature
```
msasikanth, line 1: Just an FYI. When we are working on new features, we should not implement no-op Mobius classes. We should be implementing code as needed.

app/src/main/java/org/simple/clinic/signature/SignatureActivity.kt
```diff
@@ -2,11 +2,19 @@ package org.simple.clinic.signature
...
 import androidx.appcompat.app.AppCompatActivity
+import kotlinx.android.synthetic.main.activity_signature.*
 import org.simple.clinic.R
 
-class SignatureActivity : AppCompatActivity() {
+class SignatureActivity : AppCompatActivity(), SignatureUi {
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     setContentView(R.layout.activity_signature)
+    clear_signature.setOnClickListener {
```
msasikanth, line 48: Can we rename the view ids to be camel case instead of snake case, please?

app/src/main/java/org/simple/clinic/signature/SignatureEffectHandler.kt
```diff
@@ -1,12 +1,24 @@
...
 
 import com.spotify.mobius.rx2.RxMobius
+import com.squareup.inject.assisted.Assisted
+import com.squareup.inject.assisted.AssistedInject
 import io.reactivex.ObservableTransformer
+import org.simple.clinic.util.scheduler.SchedulersProvider
 
-class SignatureEffectHandler {
+class SignatureEffectHandler @AssistedInject constructor(
+    @Assisted private val ui: SignatureUi,
```
msasikanth, line 12: We should be passing `SignatureUiActions` over here. We use `SignatureUi` in UI renderer, any one of actions should be present in `SignatureUiActions` interface.

app/src/main/java/org/simple/clinic/signature/SignatureUi.kt
```diff
@@ -0,0 +1,5 @@
+package org.simple.clinic.signature
+
+interface SignatureUi {
+  fun signatureCleared()
```
msasikanth, line 4: Can we rename this method to `clearSignature`, please?

app/src/main/java/org/simple/clinic/signature/SignatureActivity.kt
```diff
@@ -1,20 +1,92 @@
...
 import org.simple.clinic.R
+import org.simple.clinic.mobius.MobiusDelegate
+import org.simple.clinic.util.unsafeLazy
+import javax.inject.Inject
 
 class SignatureActivity : AppCompatActivity(), SignatureUi {
+
+  private lateinit var component: SignatureComponent
+
+  private val context: Context = this
```
msasikanth, line 20: We don't have to do this, since we are building the `MobiusDelegate` lazily we can directly get the context there

```kt
    MobiusDelegate.forActivity(
         events = events.ofType(),
         defaultModel = SignatureModel.create(filesDir),
         update = SignatureUpdate(),
         effectHandler = effectHandlerFactory.create(this).build()
     )
```

app/src/test/java/org/simple/clinic/signature/SignatureUpdateTest.kt
```diff
@@ -1,17 +1,24 @@
...
 import com.spotify.mobius.test.UpdateSpec
 import com.spotify.mobius.test.UpdateSpec.assertThatNext
 import org.junit.Test
+import java.io.File
 
 class SignatureUpdateTest {
 
+  private val bitmap = mock<Bitmap>()
+  private val filePath = mock<File>()
+  val model = SignatureModel.create(filePath)
```
msasikanth, line 16: These can be private to the class

app/src/main/AndroidManifest.xml
```diff
@@ -128,6 +128,22 @@
...
+      android:screenOrientation="portrait">
+      <intent-filter>
+        <action android:name="android.intent.action.VIEW" />
+
+        <category android:name="android.intent.category.BROWSABLE" />
+        <category android:name="android.intent.category.DEFAULT" />
+
+        <data
+          android:host="app.simple.org"
+          android:pathPrefix="/hidden/teleconsult/signature"
```
msasikanth, line 142: Have we confirmed this deep link schema or is it just temporary?

Honey14: Vinay gave this as a temporary link I suppose.

vinaysshenoy: This is temporary and is only meant to be used for testing this screen. In reality, a different deeplink will open another screen which will then lead to this screen.

app/src/main/java/org/simple/clinic/signature/SignatureEffect.kt
Honey14: We'll make sure the next time. Thanks

app/src/main/java/org/simple/clinic/signature/SignatureActivity.kt
Honey14: @vinaysshenoy can you help me here? I asked the same question and remember vaguely, not sure about it.

vinaysshenoy: The reason is that I am not sure if a dialog will support landscape mode if the parent activity it is running in does not.

Honey14: Now that we're keeping the screen orientation as portrait, this could have worked, right?

vinaysshenoy: We should support landscape mode for signature signing since smaller width screens might not give enough real estate for a complete signature.

app/src/main/AndroidManifest.xml
```diff
@@ -128,6 +128,22 @@
       </intent-filter>
     </activity>
 
+    <activity
+      android:name=".signature.SignatureActivity"
+      android:screenOrientation="portrait">
```
vinaysshenoy, line 133: I am curious, why did we lock this screen to portrait?

Honey14: > We should support landscape mode for signature signing since smaller width screens might not give enough real estate for a complete signature.

This seemed like a better user behaviour since we weren't handling the states but I get the reason now. Updating it to support both.

vinaysshenoy: We can support auto rotation while also disabling state saving and restoration using [this property](https://developer.android.com/reference/android/view/View#attr_android:saveEnabled) on the view.

app/src/main/java/org/simple/clinic/signature/SignatureActivity.kt
msasikanth: Just a thought for later, instead of just moving the signature screen to a separate activity. Should we consider moving the entire doctor deep link flow to a different activity? 🤔 

One downside for this might be breaking the single activity approach we have right now.

app/src/main/java/org/simple/clinic/signature/SignatureActivity.kt
```diff
@@ -1,18 +1,82 @@
...
+        effectHandler = effectHandlerFactory.create(this).build()
+    )
+  }
+
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     setContentView(R.layout.activity_signature)
   }
 
+  private fun savingSignature() = acceptSignature
```
msasikanth, line 48: You can do it in next PR, but we usually name it as `acceptSignatureClicks` or `undoClicks` to convey these are view click streams.

app/src/main/java/org/simple/clinic/signature/SignatureActivity.kt
```diff
@@ -46,10 +52,15 @@ class SignatureActivity : AppCompatActivity(), SignatureUiActions {
         AcceptClicked(bitmap)
       }
 
-  override fun clearSignature() {
+  private fun clearSignature() = clearSignature
+      .clicks()
+      .map { UndoClicked }
+
+  override fun signatureCleared() {
```
msasikanth, line 59: You can do this in another PR, rename this to `clearSignature`.

PR #1615, opened 2020-08-11
Migrate the activity to Mobius: Stabilize tests

vinaysshenoy:

### DIFF
--- a/app/src/test/java/org/simple/clinic/activity/TheActivityControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/activity/TheActivityControllerTest.kt
@@ -54,9 +54,6 @@ class TheActivityControllerTest {
   private val patientRepository = mock<PatientRepository>()
   private val lockAfterTimestamp = mock<Preference<Instant>>()
   private val uiEvents = PublishSubject.create<UiEvent>()
-  private val userSubject = PublishSubject.create<Optional<User>>()
-  private val userUnauthorizedSubject = PublishSubject.create<Boolean>()
-  private val userDisapprovedSubject = PublishSubject.create<Boolean>()
 
   private val currentTimestamp = Instant.parse("2018-01-01T00:00:00Z")
   private val clock = TestUtcClock(currentTimestamp)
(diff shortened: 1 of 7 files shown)


app/src/test/java/org/simple/clinic/activity/TheActivityControllerTest.kt
```diff
@@ -99,31 +99,73 @@ class TheActivityControllerTest {
...
+    whenever(lockAfterTimestamp.get()).thenReturn(lockAfterTime)
+
+    uiEvents.onNext(Started(null))
+
+    verify(activity).showAppLockScreen()
+  }
+
+  @Test
+  fun `when activity is started, user is not logged in and user was inactive then app lock should not be shown`() {
+    whenever(userSession.isUserLoggedIn()).thenReturn(true)
```
msasikanth, line 146: Shouldn't this return `false`?

app/src/test/java/org/simple/clinic/activity/TheActivityControllerTest.kt
```diff
@@ -299,6 +311,7 @@ class TheActivityControllerTest {
 
     //then
     verify(activity, never()).showAccessDeniedScreen(fullName)
+    verifyNoMoreInteractions(activity)
     verify(patientRepository, never()).clearPatientData()
```
msasikanth, line 351: Should we also add verify no interactions with repository?

app/src/test/java/org/simple/clinic/activity/TheActivityControllerTest.kt
vinaysshenoy: Good catch. Also will need to check why the test did not break.

app/src/test/java/org/simple/clinic/activity/TheActivityControllerTest.kt
vinaysshenoy: I am only adding it for outgoing changes caused by the controller since it's a part of the public API. What the controller uses internally is an implementation detail and it makes the tests and production code covariant so I intentionally don't add these assertions for these kind of dependencies.

app/src/test/java/org/simple/clinic/activity/TheActivityControllerTest.kt
vinaysshenoy: Ok, I just checked and the test is correct. The problem is that `UserSession#isUserLoggedIn` method is badly named.

What it actually does is check if there is a user present locally or not, but not if they are _actually_ logged in. This is handled when creating the user for the test with a `loggedInStatus` of `NOT_LOGGED_IN`.

Created a [ticket](https://app.clubhouse.io/simpledotorg/story/835/rename-badly-named-method-in-usersession) to track renaming this.

PR #1616, opened 2020-08-11
Setup the UI via effects in patient entry screen

vinaysshenoy:


app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
```diff
@@ -180,4 +186,13 @@ class PatientEntryEffectHandler @AssistedInject constructor(
       ui.scrollToFirstFieldWithError()
     }
   }
+
+  private fun loadInputFields(): ObservableTransformer<LoadInputFields, PatientEntryEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .map { inputFieldsFactory.fieldsFor(country) }
+          .map(::InputFields)
```
Honey14, line 195: ~I get that we're passing the country and we get the respective fields to load in `InputFieldsLoaded`, but can you tell me why are we injecting `InputFields` here?~ Saw the test for it. It returns the list of country fields

PR #1617, opened 2020-08-11
Migrate update phone number dialog to Mobius - Stabilise & Prepare

msasikanth:
https://app.clubhouse.io/simpledotorg/story/793/migrate-update-phone-number-dialog-to-mobius

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/updatephone/UpdatePhoneNumberDialog.kt
+++ b/app/src/main/java/org/simple/clinic/summary/updatephone/UpdatePhoneNumberDialog.kt
@@ -16,9 +16,10 @@ import io.reactivex.Observable
 import io.reactivex.subjects.PublishSubject
 import kotterknife.bindView
 import org.simple.clinic.R
-import org.simple.clinic.main.TheActivity
 import org.simple.clinic.bindUiToController
+import org.simple.clinic.main.TheActivity
 import org.simple.clinic.patient.PatientUuid
+import org.simple.clinic.widgets.ScreenCreated
 import org.simple.clinic.widgets.ScreenDestroyed
 import org.simple.clinic.widgets.UiEvent
 import org.simple.clinic.widgets.setTextAndCursor
@@ -56,7 +57,7 @@ class UpdatePhoneNumberDialog : AppCompatDialogFragment() {
   }
 
   @Inject
-  lateinit var controller: UpdatePhoneNumberDialogController
+  lateinit var controller: UpdatePhoneNumberDialogController.Factory
 
   private val phoneInputLayout by bindView<TextInputLayout>(R.id.updatephone_phone_inputlayout)
   private val numberEditText by bindView<EditText>(R.id.updatephone_phone)
@@ -96,6 +97,7 @@ class UpdatePhoneNumberDialog : AppCompatDialogFragment() {
   private fun setupDialog(screenDestroys: Observable<ScreenDestroyed>) {
     val cancelButton = (dialog as AlertDialog).getButton(DialogInterface.BUTTON_NEGATIVE)
     val saveButton = (dialog as AlertDialog).getButton(DialogInterface.BUTTON_POSITIVE)
+    val patientUuid = arguments!!.getSerializable(KEY_PATIENT_UUID) as PatientUuid
 
     bindUiToController(
         ui = this,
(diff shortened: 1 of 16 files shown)


app/src/test/java/org/simple/clinic/summary/updatephone/UpdatePhoneNumberDialogControllerTest.kt
```diff
@@ -85,16 +90,20 @@ class UpdatePhoneNumberDialogControllerTest {
       validationError: Result
   ) {
     val newNumber = "123"
-    val existingPhoneNumber = TestData.patientPhoneNumber(patientUuid = patientUuid, number = "old-number")
+    val existingPhoneNumber = TestData.patientPhoneNumber(
+        uuid = UUID.fromString("ab3f84b5-683f-49ae-987a-e319fd1db7d2"),
+        patientUuid = patientUuid,
+        number = "old-number"
```
vinaysshenoy, line 133: This is a nitpick, but while we are here, can we also update the data to be more "realistic" like using something like "1234567890" instead of an alphabetical string?

app/src/test/java/org/simple/clinic/summary/updatephone/UpdatePhoneNumberDialogControllerTest.kt
```diff
@@ -67,11 +68,15 @@ class UpdatePhoneNumberDialogControllerTest {
...
+    val existingPhoneNumber = TestData.patientPhoneNumber(
+        uuid = UUID.fromString("104e74c7-381a-4c07-8728-d6db77086dd3"),
+        patientUuid = patientUuid,
+        number = "0987654321"
+    )
 
     whenever(validator.validate(newNumber, type = LANDLINE_OR_MOBILE)).thenReturn(ValidNumber)
     whenever(repository.phoneNumber(patientUuid)).thenReturn(Observable.just(Just(existingPhoneNumber)))
-    whenever(repository.updatePhoneNumberForPatient(eq(patientUuid), any())).thenReturn(Completable.complete())
+    whenever(repository.updatePhoneNumberForPatient(patientUuid, existingPhoneNumber.copy(number = newNumber))).thenReturn(Completable.complete())
```
vinaysshenoy, line 84: We are using `existingPhoneNumber.copy(number = newNumber)` quite often in this class. What do you think of creating an extension function local to this test class that does this?

app/src/test/java/org/simple/clinic/summary/updatephone/UpdatePhoneNumberDialogControllerTest.kt
```diff
@@ -84,51 +78,69 @@ class UpdatePhoneNumberDialogControllerTest {
...
-  ) {
+  fun `when save is clicked, the number should not be saved if it's blank and an error should be shown`() {
+    val newNumber = ""
+    val existingPhoneNumber = TestData.patientPhoneNumber(
+        uuid = UUID.fromString("0e4bf753-009b-4cd6-ae30-aa9935bf2ea6"),
+        patientUuid = patientUuid,
+        number = "old-number"
+    )
+
+    whenever(validator.validate(newNumber, type = LANDLINE_OR_MOBILE)).thenReturn(Blank)
```
vinaysshenoy, line 110: One more thing which we are supposed to do as part of the Mobius migration is replace these mocks with actual objects as much as possible. Can we do this for the validator?

app/src/main/java/org/simple/clinic/summary/updatephone/UpdatePhoneNumberDialog.kt
```diff
@@ -94,18 +112,25 @@ class UpdatePhoneNumberDialog : AppCompatDialogFragment(), UpdatePhoneNumberDial
     numberEditText.showKeyboard()
   }
 
+  @SuppressLint("CheckResult")
+  override fun onResume() {
+    super.onResume()
+    events
+        .takeUntil(screenDestroys)
+        .subscribe(dialogEvents::onNext)
```
vinaysshenoy, line 151: Alright, once we finish the migration, let's take a look at how dialogs work with our Mobius delegate and come up with something that works for all our cases.

app/src/main/java/org/simple/clinic/summary/updatephone/UpdatePhoneNumberDialog.kt
```diff
@@ -139,7 +134,7 @@ class UpdatePhoneNumberDialog : AppCompatDialogFragment(), UpdatePhoneNumberDial
   override fun onStart() {
     super.onStart()
     onStarts.onNext(Any())
-    numberEditText.showKeyboard()
+    dialog?.numberEditText?.showKeyboard()
```
vinaysshenoy, line 137: Is adding null safety here the correct thing to do? We expect the dialog to be available, correct? In which case we should be using `!!` so we are notified of issues correctly.

app/src/test/java/org/simple/clinic/summary/updatephone/UpdatePhoneNumberDialogControllerTest.kt
msasikanth: Ah yes, forgot about that.

PR #1621, opened 2020-08-12
Migrate link id with patient view to Mobius - Stabilise & Prepare

msasikanth:
https://app.clubhouse.io/simpledotorg/story/836/migrate-link-id-with-patient-view-to-mobius

### DIFF
--- a/app/src/test/java/org/simple/clinic/summary/linkId/LinkIdWithPatientViewControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/summary/linkId/LinkIdWithPatientViewControllerTest.kt
@@ -27,7 +27,7 @@ class LinkIdWithPatientViewControllerTest {
   @get:Rule
   val rxErrorsRule = RxErrorsRule()
 
-  private val view = mock<LinkIdWithPatientView>()
+  private val ui = mock<LinkIdWithPatientViewUi>()
   private val patientRepository = mock<PatientRepository>()
   private val userSession = mock<UserSession>()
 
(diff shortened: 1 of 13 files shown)


app/src/test/java/org/simple/clinic/summary/linkId/LinkIdWithPatientViewControllerTest.kt
```diff
@@ -25,24 +25,22 @@ class LinkIdWithPatientViewControllerTest {
...
   private val userSession = mock<UserSession>()
 
   private val uiEvents = PublishSubject.create<UiEvent>()
 
   private val patientUuid = UUID.fromString("755bfa1a-afa5-4c80-9ec7-57d81dff2ca1")
-
+  private val identifierUuid = UUID.fromString("097a39e5-945f-44de-8293-f75960c0a54e")
   private val identifier = Identifier(
-      value = "40269f4d-f177-44a5-9db7-3cb8a7a53b33",
+      value = identifierUuid.toString(),
```
vinaysshenoy, line 44: The value of an `Identifier` is different from the `UUID` it generates. `identifierUuid` is meant to be that UUID and it should not be used for the value of this `Identifier`.

msasikanth: Ah okay, this always confuses me 🤔

vinaysshenoy: Basically, if the type was not a `BpPassport`, the value would not look like a `UUID` at all. But the ID of the generated `BusinessId` will be one.

PR #1623, opened 2020-08-13
Migrate app lock screen to Mobius - Stabilise

msasikanth:
https://app.clubhouse.io/simpledotorg/story/856/migrate-app-lock-screen-to-mobius

### DIFF
--- a/app/src/test/java/org/simple/clinic/login/applock/AppLockScreenControllerTest.kt
+++ b/app/src/test/java/org/simple/clinic/login/applock/AppLockScreenControllerTest.kt
@@ -45,15 +45,19 @@ class AppLockScreenControllerTest {
 
   @Test
   fun `when PIN is authenticated, the last-unlock-timestamp should be updated and then the app should be unlocked`() {
+    // given
     val facility = TestData.facility(
         uuid = UUID.fromString("f0e27682-796c-47c9-8187-dbcca66c4273"),
         name = "PHC Obvious"
     )
+
     whenever(facilityRepository.currentFacility(loggedInUser)).thenReturn(Observable.just(facility))
 
+    // when
     setupController()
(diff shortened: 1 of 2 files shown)


app/src/test/java/org/simple/clinic/login/applock/AppLockScreenControllerTest.kt
```diff
@@ -46,22 +45,35 @@ class AppLockScreenControllerTest {
 
   @Test
   fun `when PIN is authenticated, the last-unlock-timestamp should be updated and then the app should be unlocked`() {
+    val facility = TestData.facility(
+        uuid = UUID.fromString("f0e27682-796c-47c9-8187-dbcca66c4273"),
+        name = "PHC Obvious"
+    )
+    whenever(facilityRepository.currentFacility(loggedInUser)).thenReturn(Observable.just(facility))
```
vinaysshenoy: Should we make these class properties and set up these stubs in the `setupController` method?

msasikanth: I thought about it, but there is a test which returns 2 facilities and verifies if the facility name is changed. I didn't want to change that behaviour so I went with it. During sterilise I replace it with current facility stream instead of just injecting current facility as well.

vinaysshenoy: I think in the app lock screen we need not handle this case of the facility changing since it cannot change while this screen is visible. We could remove that test and just use the current facility.

msasikanth: Will do that in next PR 👍🏾

PR #1624, opened 2020-08-13
Migrate patient search results screen to Mobius: Sterilize screen

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/676/migrate-patient-search-results-screen-to-mobius


app/src/main/java/org/simple/clinic/search/results/PatientSearchResultsEffectHandler.kt
```diff
@@ -54,16 +52,10 @@ class PatientSearchResultsEffectHandler @AssistedInject constructor(
...
-                .loggedInUser()
-                .extractIfPresent()
-                .switchMap(facilityRepository::currentFacility)
-                .take(1)
-                .observeOn(schedulers.ui())
-                .doOnNext(uiActions::openPatientEntryScreen)
-                .switchMap { Observable.empty<PatientSearchResultsEvent>() }
-          }
+          .map { currentFacility.get() }
+          .observeOn(schedulers.ui())
```
msasikanth, line 56: Should we send a event to open patient entry screen instead of doing this?

vinaysshenoy: Not really, this is only been done this way because we need to switch threads separately, otherwise it would be a consumer.

An improvement I have planned is moving the thread switching for ui actions outside the effect handler, so we will not need to manually switch threads anymore.

vinaysshenoy:
@msasikanth DId you forget to approve this after rerolling?

msasikanth:
> @msasikanth DId you forget to approve this after rerolling?

OMG yeah, I am sorry. re-rolling and approving it.

PR #1626, opened 2020-08-13
Separate out Input field providers on a per-country bases via Dagger

vinaysshenoy:


app/src/test/java/org/simple/clinic/editpatient/EditPatientEffectHandlerTest.kt
```diff
@@ -64,14 +64,15 @@ class EditPatientEffectHandlerTest {
       identifier = Identifier(value = "1234567890abcd", type = BangladeshNationalId)
   )
 
+  private val india = TestData.country(isoCountryCode = Country.INDIA)
+  private val bangladesh = TestData.country(isoCountryCode = Country.BANGLADESH)
```
msasikanth, line 69: We have a effect handler with country set to Bangladesh in this test, shouldn't we construct a new `InputFieldsFactory` for that?

This is the test
`adding an id to an empty alternative id should create a new Business id if country has alternative id`

app/src/main/java/org/simple/clinic/newentry/country/di/InputFieldsFactoryModule.kt
```diff
@@ -0,0 +1,33 @@
...
+import java.time.format.DateTimeFormatter
+import javax.inject.Named
+
+@Module
+class InputFieldsFactoryModule {
+
+  @Provides
+  fun provideInputFieldsProvider(
+      country: Country,
+      @Named("date_for_user_input") dateTimeFormatter: DateTimeFormatter,
```
msasikanth, line 21: Since we are touching this `Named` annotation should we consider changing it to `Qualifier`?

vinaysshenoy: In the interest of time, I did not do this as part of this feature. Will do it once the PRs are merged in.

app/src/test/java/org/simple/clinic/editpatient/EditPatientEffectHandlerTest.kt
vinaysshenoy: Good catch, will add this.

vinaysshenoy: Actually, this is not necessary because in a later commit in this feature, I have changed all the input field factories in tests to use the one for Bangladesh because the Indian one has become more complex.

PR #1630, opened 2020-08-13
Add different address field labels for Chennai for patient entry and edit screens

vinaysshenoy:


app/src/test/java/org/simple/clinic/editpatient/EditPatientEffectHandlerTest.kt
```diff
@@ -68,7 +68,7 @@ class EditPatientEffectHandlerTest {
   private val india = TestData.country(isoCountryCode = Country.INDIA)
   private val bangladesh = TestData.country(isoCountryCode = Country.BANGLADESH)
 
-  private val inputFieldsFactory = InputFieldsFactory(IndiaInputFieldsProvider(
+  private val inputFieldsFactory = InputFieldsFactory(BangladeshInputFieldsProvider(
```
msasikanth, line 71: Not important, but should we just create a `TestInputFieldsProvider` to avoid any confusion in future?

vinaysshenoy: Yeah, that's a good idea. I am planning to take a fresh look at how the patient entry and edit screens are represented and setup. I will add this when doing that.

PR #1631, opened 2020-08-13
Migrate app lock screen to Mobius - Migrate

msasikanth:
https://app.clubhouse.io/simpledotorg/story/856/migrate-app-lock-screen-to-mobius

### DIFF
--- a/app/src/main/java/org/simple/clinic/login/applock/AppLockUiRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/login/applock/AppLockUiRenderer.kt
@@ -1,5 +1,6 @@
 package org.simple.clinic.login.applock
 
+import org.simple.clinic.facility.Facility
 import org.simple.clinic.mobius.ViewRenderer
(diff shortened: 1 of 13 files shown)


app/src/main/java/org/simple/clinic/login/applock/AppLockUiRenderer.kt
```diff
@@ -1,10 +1,17 @@
 package org.simple.clinic.login.applock
 
 import org.simple.clinic.mobius.ViewRenderer
+import org.simple.clinic.user.User
 
 class AppLockUiRenderer(private val ui: AppLockScreenUi) : ViewRenderer<AppLockModel> {
 
   override fun render(model: AppLockModel) {
+    if (model.hasUser) {
+      renderUsername(model.user!!)
```
vinaysshenoy, line 11: This is a nitpick, but this name should be `renderUserName` and not `renderUsername` since the name of the user and the user's username are two different things.

PR #1633, opened 2020-08-14
Migrate `TheActivity` to Mobius: Strangulate controller

vinaysshenoy:

### DIFF
--- a/app/src/main/java/org/simple/clinic/main/TheActivityEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/main/TheActivityEffectHandler.kt
@@ -42,6 +42,7 @@ class TheActivityEffectHandler @AssistedInject constructor(
         .addAction(RedirectToLoginScreen::class.java, uiActions::redirectToLogin, schedulers.ui())
         .addTransformer(ListenForUserDisapprovals::class.java, listenForUserDisapprovals())
         .addTransformer(ClearPatientData::class.java, clearPatientData())
+        .addTransformer(ShowAccessDeniedScreen::class.java, openAccessDeniedScreen())
         .build()
   }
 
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/simple/clinic/main/TheActivityEffectHandler.kt
```diff
@@ -129,4 +130,15 @@ class TheActivityEffectHandler @AssistedInject constructor(
...
     }
   }
+
+  private fun openAccessDeniedScreen(): ObservableTransformer<ShowAccessDeniedScreen, TheActivityEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulers.io())
+          .map { userSession.loggedInUserImmediate()!!.fullName }
+          .observeOn(schedulers.ui())
+          .doOnNext(uiActions::showAccessDeniedScreen)
```
msasikanth, line 140: I just want to make sure regarding this approach. If we need to trigger a UI action right after a small task are we allowed to follow this approach or stick with sending out event and dispatching effect? 🤔

vinaysshenoy: I don't understand the question? Here, the effect is to show the access denied screen and to do it we need the current user. It's perfectly fine to query it in the effect handler.

What should be pushed into the update loop is conditional logic, so if we have branching paths based on the result of a DB or network query, the branching decision should be pushed into the update function.

msasikanth: Ah okay, I wanted to know when to send a event to update function. So just get this correct, if we require a object/data that is not present in model and is only required for that particular UiAction, it's perfectly fine to do this right?

PR #1638, opened 2020-08-17
Migrate `TheActivity` to Mobius: Sterilize screen

vinaysshenoy:


app/src/main/java/org/simple/clinic/main/TypedPreference.kt
```diff
@@ -0,0 +1,11 @@
+package org.simple.clinic.main
+
+import javax.inject.Qualifier
+
+@Qualifier
+annotation class TypedPreference(val value: Type) {
+
+  enum class Type {
```
Honey14, line 8: Why do we have enum type as a parameter?

vinaysshenoy: Because with this, we will add a new enum type for each `Preference` that we use to give it a statically typed value. This is essentially a replacement for the `@Named` parameter annotations that we use otherwise.

PR #1640, opened 2020-08-17
Add screen for adjusting medicine duration

msasikanth:
https://app.clubhouse.io/simpledotorg/story/891/add-screen-for-adjusting-medicine-duration

### DIFF
--- a/app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationInit.kt
+++ b/app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationInit.kt
@@ -1,0 +1,17 @@
+package org.simple.clinic.teleconsultlog.drugduration
+
+import com.spotify.mobius.First
+import com.spotify.mobius.First.first
+import com.spotify.mobius.Init
+
+class DrugDurationInit : Init<DrugDurationModel, DrugDurationEffect> {
+
+  override fun init(model: DrugDurationModel): First<DrugDurationModel, DrugDurationEffect> {
+    val effects = mutableSetOf<DrugDurationEffect>()
+    if (model.hasDuration) {
+      effects.add(SetDrugDuration(model.duration))
+    }
+
+    return first(model, effects)
+  }
+}
--- a/app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationSheet.kt
+++ b/app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationSheet.kt
@@ -1,0 +1,163 @@
+package org.simple.clinic.teleconsultlog.drugduration
+
+import android.app.Activity
+import android.content.Context
+import android.content.Intent
+import android.os.Bundle
+import android.view.View
+import android.view.inputmethod.EditorInfo
+import com.jakewharton.rxbinding3.widget.editorActions
+import io.github.inflationx.viewpump.ViewPumpContextWrapper
+import io.reactivex.Observable
+import io.reactivex.rxkotlin.ofType
+import kotlinx.android.synthetic.main.sheet_drug_duration.*
+import org.simple.clinic.ClinicApp
+import org.simple.clinic.R
+import org.simple.clinic.ReportAnalyticsEvents
+import org.simple.clinic.di.InjectorProviderContextWrapper
+import org.simple.clinic.drugs.PrescribedDrug
+import org.simple.clinic.mobius.MobiusDelegate
+import org.simple.clinic.teleconsultlog.drugduration.di.DrugDurationComponent
+import org.simple.clinic.util.LocaleOverrideContextWrapper
+import org.simple.clinic.util.unsafeLazy
+import org.simple.clinic.util.wrap
+import org.simple.clinic.widgets.BottomSheetActivity
+import org.simple.clinic.widgets.setTextAndCursor
+import org.simple.clinic.widgets.textChanges
+import java.util.Locale
+import javax.inject.Inject
+
+class DrugDurationSheet : BottomSheetActivity(), DrugDurationUiActions {
+
+  @Inject
+  lateinit var locale: Locale
+
+  @Inject
+  lateinit var effectHandlerFactory: DrugDurationEffectHandler.Factory
+
+  companion object {
+    private const val EXTRA_DRUG_UUID = "prescribedDrugUuid"
+    private const val EXTRA_DRUG_NAME = "prescribedDrugTitle"
+    private const val EXTRA_DRUG_DOSAGE = "prescribedDrugDosage"
+    private const val EXTRA_DURATION = "drugDuration"
+    private const val EXTRA_SAVED_DRUG_UUID = "savedDrugUuid"
+    private const val EXTRA_SAVED_DURATION = "savedDrugDuration"
+
+    fun intent(
+        context: Context,
+        drug: PrescribedDrug,
+        drugDuration: String
+    ): Intent {
+      return Intent(context, DrugDurationSheet::class.java).apply {
+        putExtra(EXTRA_DRUG_UUID, drug.uuid)
+        putExtra(EXTRA_DRUG_NAME, drug.name)
+        putExtra(EXTRA_DRUG_DOSAGE, drug.dosage)
+        putExtra(EXTRA_DURATION, drugDuration)
+      }
+    }
+  }
+
+  private lateinit var component: DrugDurationComponent
+
+  private val events by unsafeLazy {
+    Observable
+        .merge(
+            imeClicks(),
+            durationChanges()
+        )
+        .compose(ReportAnalyticsEvents())
+  }
+
+  private val delegate by unsafeLazy {
+    val duration = intent.getStringExtra(EXTRA_DURATION)!!
+
+    MobiusDelegate.forActivity(
+        events = events.ofType(),
+        defaultModel = DrugDurationModel.create(duration),
+        init = DrugDurationInit(),
+        update = DrugDurationUpdate(),
+        effectHandler = effectHandlerFactory.create(this).build()
+    )
+  }
+
+  override fun onStart() {
+    super.onStart()
+    delegate.start()
+  }
+
+  override fun onStop() {
+    delegate.stop()
+    super.onStop()
+  }
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.sheet_drug_duration)
+    delegate.onRestoreInstanceState(savedInstanceState)
+
+    val drugName = intent.getStringExtra(EXTRA_DRUG_NAME)!!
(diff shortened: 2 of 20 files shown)


app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationUiActions.kt
```diff
@@ -0,0 +1,5 @@
+package org.simple.clinic.teleconsultlog.drugduration
+
+interface DrugDurationUiActions {
+  fun showBlankDurationError()
```
vinaysshenoy, line 4: This should be done in the UI renderer, right?

We have kept these in the UI actions as effects when migrating from Mobius because it's simpler that way. But for new screens, we should be tracking any error states to display in the model and then render them in the UI renderer.

An exception is when error messages are shown as modals.

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationUiActions.kt
```diff
@@ -2,4 +2,5 @@ package org.simple.clinic.teleconsultlog.drugduration
 
 interface DrugDurationUiActions {
   fun showBlankDurationError()
+  fun hideDurationError()
```
vinaysshenoy, line 5: Same as earlier comment about these being part of the UI renderer flow.

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationInit.kt
```diff
@@ -0,0 +1,17 @@
...
+import com.spotify.mobius.First
+import com.spotify.mobius.First.first
+import com.spotify.mobius.Init
+
+class DrugDurationInit : Init<DrugDurationModel, DrugDurationEffect> {
+
+  override fun init(model: DrugDurationModel): First<DrugDurationModel, DrugDurationEffect> {
+    val effects = mutableSetOf<DrugDurationEffect>()
+    if (model.hasDuration) {
+      effects.add(SetDrugDuration(model.duration))
```
vinaysshenoy, line 12: Since the view is an edit text, do we need to do this since it will persist its state correctly?

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationUpdate.kt
```diff
@@ -3,12 +3,13 @@ package org.simple.clinic.teleconsultlog.drugduration
...
 import org.simple.clinic.mobius.dispatch
+import org.simple.clinic.mobius.next
 
 class DrugDurationUpdate : Update<DrugDurationModel, DrugDurationEvent, DrugDurationEffect> {
 
   override fun update(model: DrugDurationModel, event: DrugDurationEvent): Next<DrugDurationModel, DrugDurationEffect> {
     return when (event) {
-      DurationChanged -> dispatch(HideDurationError)
+      is DurationChanged -> next(model.durationChanged(event.duration), HideDurationError)
       is DrugDurationSaveClicked -> drugDurationSaveClicked(event)
```
vinaysshenoy, line 13: In the body of `drugDurationSaveClicked`, we can use the `duration` from the model instead of reading it from the `DrugDurationSaveClicked` event now.

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationSheet.kt
```diff
@@ -0,0 +1,156 @@
...
+    intent.getParcelableExtra<PrescribedDrug>(EXTRA_DRUG)!!
+  }
+
+  private val events by unsafeLazy {
+    Observable
+        .merge(
+            imeClicks(),
+            durationChanges()
+        )
+        .compose(ReportAnalyticsEvents())
```
vinaysshenoy, line 67: Can we ensure the `analyticsProperty` value is overriden on all the events that we are creating?

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationSheet.kt
```diff
@@ -0,0 +1,156 @@
...
+    private const val EXTRA_DURATION = "drugDuration"
+    private const val EXTRA_SAVED_DURATION = "savedDrugDuration"
+
+    fun intent(
+        context: Context,
+        drug: PrescribedDrug,
+        drugDuration: String
+    ): Intent {
+      return Intent(context, DrugDurationSheet::class.java).apply {
+        putExtra(EXTRA_DRUG, drug)
```
vinaysshenoy, line 49: Do we need the entire drug object in the extras? We can just use the `UUID`, correct?

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationSheet.kt
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  override fun hideDurationError() {
+    drugDurationErrorTextView.text = null
+    drugDurationErrorTextView.visibility = View.GONE
+  }
+
+  override fun saveDrugDuration(duration: Int) {
+    val intent = Intent()
+    intent.putExtra(EXTRA_SAVED_DURATION, duration)
```
vinaysshenoy, line 148: Won't we also need to add the `UUID` of the `PrescribedDrug` so that the caller knows which one needs to be updated?

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationInit.kt
msasikanth: It was mainly for setting the default duration that is passed from the screen arguments. Should I have used a different approach for that?

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationUpdate.kt
msasikanth: Ah yes, good catch. I have added the duration to model later on & forgot to change it here as well.

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationSheet.kt
msasikanth: Since this mainly for setting the title and dosage I wasn't sure if we should do another DB query in this screen and save it to model. WDYT?

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationUiActions.kt
msasikanth: I have question regarding this approach. In this case doesn't UI actions makes more sense? because the error message visibility not only is dependent on the duration validation result but also on the duration text changes. How does that translate to the model 🤔

vinaysshenoy: Basically anything that is renderered as part of the screen should be done through the model. What this means is that we have a nullable property in the model, say a `ValidationResult` enum with `[NOT_VALIDATED, BLANK]` states and then change the state in the model based on what we want to render.

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationInit.kt
vinaysshenoy: Ah, I see. This makes sense.

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationSheet.kt
vinaysshenoy: Ah, we can pass the entire model in the factory function and then put only the values we need as part of the arguments.

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationUiActions.kt
msasikanth: Ah okay. Just to confirm that means whenever the duration text changes we need to update the model twice, once for updating the duration in the model and second one is to update the validation result to set it to `NOT_VALIDATED` in order to remove the error whenever the duration text changes. WDYT?

msasikanth: Something like this
```kt
      is DurationChanged -> next(
          model
              .durationChanged(event.duration)
              .validationResultChanged(NOT_VALIDATED)
      )
```

vinaysshenoy: This works, but I prefer to make it so that `durationChanged` could internally set the validation result to `NOT_VALIDATED`. We could then add another method that is `.invalid(ValidationResult)` which can be used when the model is invalid.

msasikanth: Yeah, I was thinking about doing that but felt like it's a hidden behaviour 😅

vinaysshenoy: It's not really hidden, the business logic is part of the model.

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationValidationResult.kt
```diff
@@ -0,0 +1,10 @@
+package org.simple.clinic.teleconsultlog.drugduration
+
+import android.os.Parcelable
+import kotlinx.android.parcel.Parcelize
+
+@Parcelize
```
vinaysshenoy, line 6: We don't need to explicit parcel enums because they are `Serializable` by default.

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationSheet.kt
```diff
@@ -0,0 +1,165 @@
...
+    fun intent(
+        context: Context,
+        drug: PrescribedDrug,
+        drugDuration: String
+    ): Intent {
+      return Intent(context, DrugDurationSheet::class.java).apply {
+        putExtra(EXTRA_DRUG_UUID, drug.uuid)
+        putExtra(EXTRA_DRUG_NAME, drug.name)
+        putExtra(EXTRA_DRUG_DOSAGE, drug.dosage)
+        putExtra(EXTRA_DURATION, drugDuration)
```
vinaysshenoy, line 55: We don't need to do this in this PR, but I can suggest that we extract the arguments required for this screen into their own class and save that as a single `Parcelable` argument.

PR #1642, opened 2020-08-18
Migrate patient search view to Mobius: Strangulate controller and sterilize the screen

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/708/migrate-search-results-view-to-mobius


app/src/main/java/org/simple/clinic/searchresultsview/SearchResultsEffectHandler.kt
```diff
@@ -25,10 +25,9 @@ class SearchResultsEffectHandler @Inject constructor(
   }
 
   private fun searchForPatients(): ObservableTransformer<SearchWithCriteria, SearchResultsEvent> {
-    val currentFacilityStream = userSession
-        .requireLoggedInUser()
+    val currentFacilityStream = Observable
+        .fromCallable { currentFacility.get() }
         .subscribeOn(schedulers.io())
-        .switchMap(facilityRepository::currentFacility)
         .take(1)
```
msasikanth: Do we still need to use a separate facility stream with `take(1)`, since `Lazy<Facility>` is not a observable stream we can directly use that right?

vinaysshenoy: Good catch, I think I missed that. Will fix.

vinaysshenoy: Or wait, this was fixed in the last commit. Please take a look?

msasikanth: Yep, I completely missed it 🤦🏾

PR #1643, opened 2020-08-18
Add screen to notify a successful Teleconsultation 

Honey14:
https://app.clubhouse.io/simpledotorg/story/887/add-screen-for-notifying-a-successful-teleconsultation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 - Migrated `TheActivity` to Mobius
 - Migrated `AppLockScreen` to Mobius
 - Migrated `PatientSearchView` to Mobius
+- Added `TeleConsultSuccessScreen` to notify teleconsultation recorded successfully
+
 
 ## On Demo
 ### Features
--- a/app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffect.kt
+++ b/app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffect.kt
@@ -1,5 +1,6 @@
 package org.simple.clinic.teleconsultlog.success
 
+import org.simple.clinic.patient.Patient
 import java.util.UUID
 
 sealed class TeleConsultSuccessEffect {
@@ -7,4 +8,6 @@ sealed class TeleConsultSuccessEffect {
   data class LoadPatientDetails(val patientUuid: UUID) : TeleConsultSuccessEffect()
 
   object GoToHomeScreen : TeleConsultSuccessEffect()
+
+  data class GoToPrescriptionScreen(val patient: Patient) : TeleConsultSuccessEffect()
 }
--- a/app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffect.kt
+++ b/app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffect.kt
@@ -1,0 +1,8 @@
+package org.simple.clinic.teleconsultlog.success
+
+import java.util.UUID
+
+sealed class TeleConsultSuccessEffect {
+
+  data class LoadPatientDetails(val patientUuid: UUID) : TeleConsultSuccessEffect()
+}
--- a/app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffect.kt
+++ b/app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffect.kt
@@ -5,4 +5,6 @@ import java.util.UUID
 sealed class TeleConsultSuccessEffect {
 
   data class LoadPatientDetails(val patientUuid: UUID) : TeleConsultSuccessEffect()
+
+  object GoToHomeScreen : TeleConsultSuccessEffect()
 }
--- a/app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffectHandler.kt
@@ -1,27 +1,31 @@
 package org.simple.clinic.teleconsultlog.success
 
 import com.spotify.mobius.rx2.RxMobius
+import com.squareup.inject.assisted.Assisted
 import com.squareup.inject.assisted.AssistedInject
 import io.reactivex.ObservableTransformer
 import io.reactivex.Scheduler
 import org.simple.clinic.patient.PatientRepository
+import org.simple.clinic.teleconsultlog.success.TeleConsultSuccessEffect.GoToHomeScreen
 import org.simple.clinic.teleconsultlog.success.TeleConsultSuccessEffect.LoadPatientDetails
 import org.simple.clinic.util.scheduler.SchedulersProvider
 
 class TeleConsultSuccessEffectHandler @AssistedInject constructor(
     private val schedulersProvider: SchedulersProvider,
-    private val patientRepository: PatientRepository
+    private val patientRepository: PatientRepository,
+    @Assisted private val uiActions: TeleConsultSuccessScreenUiActions
 ) {
 
   @AssistedInject.Factory
   interface Factory {
-    fun create(): TeleConsultSuccessEffectHandler
+    fun create(uiActions: TeleConsultSuccessScreenUiActions): TeleConsultSuccessEffectHandler
   }
 
   fun build(): ObservableTransformer<TeleConsultSuccessEffect, TeleConsultSuccessEvent> {
     return RxMobius
         .subtypeEffectHandler<TeleConsultSuccessEffect, TeleConsultSuccessEvent>()
         .addTransformer(LoadPatientDetails::class.java, loadPatientDetails(schedulersProvider.io()))
+        .addAction(GoToHomeScreen::class.java, { uiActions.goToHomeScreen() }, schedulersProvider.ui())
         .build()
   }
 
--- a/app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffectHandler.kt
@@ -7,6 +7,7 @@ import io.reactivex.ObservableTransformer
 import io.reactivex.Scheduler
 import org.simple.clinic.patient.PatientRepository
 import org.simple.clinic.teleconsultlog.success.TeleConsultSuccessEffect.GoToHomeScreen
+import org.simple.clinic.teleconsultlog.success.TeleConsultSuccessEffect.GoToPrescriptionScreen
 import org.simple.clinic.teleconsultlog.success.TeleConsultSuccessEffect.LoadPatientDetails
 import org.simple.clinic.util.scheduler.SchedulersProvider
 
@@ -26,6 +27,7 @@ class TeleConsultSuccessEffectHandler @AssistedInject constructor(
         .subtypeEffectHandler<TeleConsultSuccessEffect, TeleConsultSuccessEvent>()
         .addTransformer(LoadPatientDetails::class.java, loadPatientDetails(schedulersProvider.io()))
         .addAction(GoToHomeScreen::class.java, { uiActions.goToHomeScreen() }, schedulersProvider.ui())
+        .addConsumer(GoToPrescriptionScreen::class.java, { uiActions.goToPrescriptionScreen(it.patient) }, schedulersProvider.ui())
         .build()
   }
 
--- a/app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffectHandler.kt
@@ -31,7 +31,7 @@ class TeleConsultSuccessEffectHandler @AssistedInject constructor(
         .build()
   }
 
-  private fun loadPatientDetails(scheduler: Scheduler): ObservableTransformer<TeleConsultSuccessEffect.LoadPatientDetails, TeleConsultSuccessEvent> {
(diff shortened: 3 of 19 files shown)


app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessInit.kt
```diff
@@ -0,0 +1,13 @@
+package org.simple.clinic.teleconsultlog.success
+
+import com.spotify.mobius.First
+import com.spotify.mobius.Init
+import org.simple.clinic.mobius.first
+import org.simple.clinic.teleconsultlog.success.TeleConsultSuccessEffect.LoadPatientDetails
+
+class TeleConsultSuccessInit : Init<TeleConsultSuccessModel, TeleConsultSuccessEffect> {
+  override fun init(model: TeleConsultSuccessModel): First<TeleConsultSuccessModel, TeleConsultSuccessEffect> {
+    return first(model, LoadPatientDetails(model.patientUuid))
```
vinaysshenoy, line 10: Not necessary to do it in this PR, but we should split this into two commits:

- One where we add the effect and its handler
- The next where we trigger the effect from the init function

app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffectHandler.kt
```diff
@@ -17,6 +22,21 @@ class TeleConsultSuccessEffectHandler @AssistedInject constructor(
...
+        .addTransformer(LoadPatientDetails::class.java, loadPatientDetails(schedulersProvider.io()))
         .build()
   }
+
+  private fun loadPatientDetails(scheduler: Scheduler): ObservableTransformer<TeleConsultSuccessEffect.LoadPatientDetails, TeleConsultSuccessEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .switchMap {
+            patientRepository
+                .patient(it.patientUuid)
```
vinaysshenoy, line 42: Can we avoid using reactive calls unless we explicitly require the change notifications please? We have a synchronous method call already which we should prefer to use here.

app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessUpdate.kt
```diff
@@ -2,9 +2,12 @@ package org.simple.clinic.teleconsultlog.success
 
 import com.spotify.mobius.Next
 import com.spotify.mobius.Update
+import org.simple.clinic.mobius.next
 
 class TeleConsultSuccessUpdate : Update<TeleConsultSuccessModel, TeleConsultSuccessEvent, TeleConsultSuccessEffect> {
   override fun update(model: TeleConsultSuccessModel, event: TeleConsultSuccessEvent): Next<TeleConsultSuccessModel, TeleConsultSuccessEffect> {
-    return Next.noChange()
+    return when (event) {
+      is PatientDetailsLoaded -> next(model.patientDetailLoaded(event.patient))
```
vinaysshenoy, line 14: In a previous commit, we load the patient details whenever the screen is created. In the screen restoration scenario, if the model already has the patient details loaded we do not need to load it again, right?

app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffectHandler.kt
```diff
@@ -1,28 +1,32 @@
...
   interface Factory {
-    fun create(): TeleConsultSuccessEffectHandler
+    fun create(uiActions: TeleConsultSuccessScreenUiActions): TeleConsultSuccessEffectHandler
   }
 
   fun build(): ObservableTransformer<TeleConsultSuccessEffect, TeleConsultSuccessEvent> {
     return RxMobius
         .subtypeEffectHandler<TeleConsultSuccessEffect, TeleConsultSuccessEvent>()
         .addTransformer(LoadPatientDetails::class.java, loadPatientDetails(schedulersProvider.io()))
+        .addConsumer(GoToHomeScreen::class.java, { uiActions.goToHomeScreen() }, schedulersProvider.ui())
```
vinaysshenoy, line 31: We are not using anything from the effect so this could be an action, right?

app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffectHandler.kt
```diff
@@ -27,6 +28,7 @@ class TeleConsultSuccessEffectHandler @AssistedInject constructor(
         .subtypeEffectHandler<TeleConsultSuccessEffect, TeleConsultSuccessEvent>()
         .addTransformer(LoadPatientDetails::class.java, loadPatientDetails(schedulersProvider.io()))
         .addConsumer(GoToHomeScreen::class.java, { uiActions.goToHomeScreen() }, schedulersProvider.ui())
+        .addConsumer(GoToPrescriptionScreen::class.java, { uiActions.goToPrescriptionScreen() }, schedulersProvider.ui())
```
vinaysshenoy, line 32: Same comment as earlier about this being an action instead of a consumer.

app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffect.kt
```diff
@@ -7,4 +7,6 @@ sealed class TeleConsultSuccessEffect {
   data class LoadPatientDetails(val patientUuid: UUID) : TeleConsultSuccessEffect()
 
   object GoToHomeScreen : TeleConsultSuccessEffect()
+
+  object GoToPrescriptionScreen : TeleConsultSuccessEffect()
```
vinaysshenoy, line 11: I am not entirely sure an `object` is the right kind to use here since the prescription screen will need to know which patient to load prescriptions for.

app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffectHandler.kt
```diff
@@ -29,6 +30,7 @@ class TeleConsultSuccessEffectHandler @AssistedInject constructor(
         .addTransformer(LoadPatientDetails::class.java, loadPatientDetails(schedulersProvider.io()))
         .addConsumer(GoToHomeScreen::class.java, { uiActions.goToHomeScreen() }, schedulersProvider.ui())
         .addConsumer(GoToPrescriptionScreen::class.java, { uiActions.goToPrescriptionScreen() }, schedulersProvider.ui())
+        .addConsumer(GoToTeleConsultScreen::class.java, { uiActions.goToTeleConsultScreen() }, schedulersProvider.ui())
```
vinaysshenoy, line 33: Same comment as earlier about this being an action instead of a consumer.

app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessScreen.kt
```diff
@@ -0,0 +1,104 @@
...
+  lateinit var screenRouter: ScreenRouter
+
+  private val events: Observable<TeleConsultSuccessEvent> by unsafeLazy {
+    Observable
+        .merge(
+            backClicks(),
+            yesClicks(),
+            noClicks()
+        )
+        .cast<TeleConsultSuccessEvent>()
```
vinaysshenoy, line 44: I think we can skip adding the explicit `<TeleConsultSuccessEvent>` to the `.cast` operator since Kotlin's type inference is smart enough to figure it out.

app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessScreen.kt
```diff
@@ -0,0 +1,104 @@
...
+  override fun goToHomeScreen() {
+    screenRouter.clearHistoryAndPush(HomeScreenKey(), direction = BACKWARD)
+  }
+
+  override fun goToPrescriptionScreen() {
+    TODO("Not yet implemented")
+  }
+
+  override fun goToTeleConsultScreen() {
+    TODO("Not yet implemented")
```
vinaysshenoy, line 110: Can we leave these implementations empty instead of adding `TODO` calls? These will crash the app if invoked.

app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessScreenInjector.kt
```diff
@@ -0,0 +1,5 @@
+package org.simple.clinic.teleconsultlog.success
+
+interface TeleConsultSuccessScreenInjector {
```
vinaysshenoy, line 3: We have lately been following the convention of making this interface a nested interface of the screen that needs injection and skipping the screen name prefix. Can we follow that here please?

app/src/main/res/layout/screen_teleconsult_success.xml
```diff
@@ -0,0 +1,112 @@
+<?xml version="1.0" encoding="utf-8"?>
+<org.simple.clinic.teleconsultlog.success.TeleConsultSuccessScreen xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:app="http://schemas.android.com/apk/res-auto"
+  xmlns:tools="http://schemas.android.com/tools"
+  android:layout_width="match_parent"
+  android:layout_height="match_parent"
+  android:background="@color/grey4">
+
+  <RelativeLayout
```
vinaysshenoy, line 9: Is there a reason we are creating a separate layout here instead of using the `Toolbar` component?

app/src/main/res/layout/screen_teleconsult_success.xml
```diff
@@ -0,0 +1,112 @@
...
+    style="@style/Clinic.LoginPinLabel"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:layout_marginTop="@dimen/spacing_16"
+    android:text="@string/screen_teleconsult_success_teleconsult_recorded"
+    app:layout_constraintEnd_toEndOf="parent"
+    app:layout_constraintStart_toStartOf="parent"
+    app:layout_constraintTop_toBottomOf="@+id/imageView" />
+
+  <LinearLayout
```
vinaysshenoy, line 69: Why do we need this specific `LinearLayout` here if the parent is already a `ConstraintLayout`? We _should_ be able to achieve the same layout with the constraints available to us.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 ### Changes
 - Added `SignatureActivity` to accept and save user signature
 - Increase error text sizes
+- Added `TeleConsultSuccessScreen` to notify teleconsultation recorded successfully
```
vinaysshenoy, line 7: This could be an "Internal" change since the screen itself is not accessible at this point.

app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessUiRenderer.kt
```diff
@@ -0,0 +1,11 @@
+package org.simple.clinic.teleconsultlog.success
+
+import org.simple.clinic.mobius.ViewRenderer
+
+class TeleConsultSuccessUiRenderer(
+    private val ui: TeleConsultSuccessUi
+) : ViewRenderer<TeleConsultSuccessModel> {
+  override fun render(model: TeleConsultSuccessModel) {
+    ui.showPatientInfo(model.patient!!)
```
vinaysshenoy, line 9: We'll need to add a check here if the patient has not loaded, otherwise the app will crash on initial load.

vinaysshenoy (changes requested):
Overall approach looks good, left some comments. Please take a look?

app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessInit.kt
Honey14: Question on 1. 
So the event translated as a response from the effect handler should also be in the same commit?

vinaysshenoy: Yes, that should be in the same commit. This is more about making commits atomic and easier to review than anything else since we squash and merge PRs anyway.

app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessEffectHandler.kt
Honey14: This would be `addConsumer` since we're sending `patientUuid` to Prescription screen

app/src/main/res/layout/screen_teleconsult_success.xml
Honey14: Removed `Recorded` text from the layout, but we need a have a parent layout which has white background only for the two material buttons. So keeping `LinearLayout` for that. PTAL?

Honey14:
Removed `GoToTeleConsultScreen` effect, since that would be the screen before this after implementation. We do not need an explicit effect for that.

app/src/main/java/org/simple/clinic/teleconsultlog/success/TeleConsultSuccessUpdate.kt
vinaysshenoy: The current change makes it so that the patient details are never loaded since there is no effect being emitted at all. We should emit the effect conditionally based on whether the model has the details or not.

vinaysshenoy: Please ignore the earlier comment, I was only looking a single commit instead of the full set of changes.

Honey14:
Moving it to WIP, including tests to check if patient detail is empty.

vinaysshenoy:
> Moving it to WIP, including tests to check if patient detail is empty.

Cool, let me know once its done. Overall PR looks good so once the tests are fixed we can merge this in.

vinaysshenoy (approved):

PR #1645, opened 2020-08-18
Update the `User` model to include the teleconsultation number

msasikanth:
https://app.clubhouse.io/simpledotorg/story/877/update-the-user-model-to-include-the-teleconsultation-number

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration71AndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration71AndroidTest.kt
@@ -1,0 +1,96 @@
+package org.simple.clinic.storage.migrations
+
+import org.junit.Test
+import org.simple.clinic.assertValues
+import org.simple.clinic.insert
+import java.time.Instant
+import java.util.UUID
+
+class Migration71AndroidTest : BaseDatabaseMigrationTest(70, 71) {
+
+  @Test
+  fun user_information_should_be_retained_after_the_migration() {
+    val facilityUuid = UUID.fromString("eac9f788-1a8c-4f29-87ff-5d16d5a4eeba")
+    before.insert("Facility", mapOf(
+        "uuid" to facilityUuid,
+        "name" to "PHC Obvious",
+        "facilityType" to null,
+        "streetAddress" to null,
+        "villageOrColony" to null,
+        "district" to "Bathinda",
+        "state" to "Punjab",
+        "country" to "India",
+        "pinCode" to null,
+        "protocolUuid" to null,
+        "groupUuid" to null,
+        "location_latitude" to null,
+        "location_longitude" to null,
+        "createdAt" to Instant.parse("2018-01-01T00:00:00Z"),
+        "updatedAt" to Instant.parse("2018-01-01T00:00:01Z"),
+        "deletedAt" to null,
+        "config_diabetesManagementEnabled" to true,
+        "config_teleconsultationEnabled" to true,
+        "syncStatus" to "DONE"
+    ))
+    before.insert("LoggedInUser", mapOf(
+        "uuid" to UUID.fromString("dfe2653f-62d7-4b05-b2c0-ae596361de51"),
+        "fullName" to "John Doe",
+        "phoneNumber" to "1111111111",
+        "pinDigest" to "1111",
+        "status" to "allowed",
+        "createdAt" to Instant.parse("2018-01-01T00:00:00Z"),
+        "updatedAt" to Instant.parse("2018-01-01T00:00:00Z"),
+        "loggedInStatus" to "LOGGED_IN",
+        "registrationFacilityUuid" to facilityUuid,
+        "currentFacilityUuid" to facilityUuid
+    ))
+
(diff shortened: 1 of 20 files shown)


msasikanth:
Moving to WIP to fix a failing test

app/src/main/java/org/simple/clinic/user/LoggedInUserPayload.kt
```diff
@@ -30,5 +30,8 @@ data class LoggedInUserPayload(
     val createdAt: Instant,
 
     @Json(name = "updated_at")
-    val updatedAt: Instant
+    val updatedAt: Instant,
+
+    @Json(name = "teleconsult_phone_number")
+    val teleconsultPhoneNumber: String?
```
vinaysshenoy, line 36: IIRC, we are also going to make other changes to the user model and add the "capabilities" property. Can we try to make this part of the same migration?

app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration71AndroidTest.kt
```diff
@@ -0,0 +1,37 @@
+package org.simple.clinic.storage.migrations
+
+import org.junit.Test
+import org.simple.clinic.assertColumns
+
+class Migration71AndroidTest : BaseDatabaseMigrationTest(70, 71) {
+
+  @Test
+  fun registered_facility_id_and_assigned_facility_id_should_be_added_to_patient_table() {
+    before.assertColumns("LoggedInUser", setOf(
```
vinaysshenoy, line 10: Can we also create a default entry and run the migration to ensure that the values are retained after migration correctly?

app/src/main/java/org/simple/clinic/login/pin/LoginPinEffectHandler.kt
```diff
@@ -88,7 +88,8 @@ class LoginPinEffectHandler @AssistedInject constructor(
         updatedAt = entry.updatedAt!!,
         loggedInStatus = User.LoggedInStatus.OTP_REQUESTED,
         registrationFacilityUuid = entry.registrationFacilityUuid!!,
-        currentFacilityUuid = entry.registrationFacilityUuid
+        currentFacilityUuid = entry.registrationFacilityUuid,
+        teleconsultPhoneNumber = null
```
vinaysshenoy, line 92: `OngoingLoginEntry` is created from the `UsersApi#activate` API endpoint which returns a complete `LoggedInUserPayload` which will include the teleconsultation number.

So setting it to `null` is the incorrect thing to do here since we should also update `OngoingLoginEntry` to save the teleconsult number.

app/src/main/java/org/simple/clinic/storage/migrations/Migration_71.kt
```diff
@@ -0,0 +1,19 @@
+package org.simple.clinic.storage.migrations
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+import org.simple.clinic.storage.inTransaction
+import javax.inject.Inject
+
+@Suppress("Classname")
+class Migration_71 @Inject constructor() : Migration(70, 71) {
```
vinaysshenoy, line 9: Just FYI, since we squash and merge now, we can move the migration to a different commit for easier review.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/user/LoggedInUserPayload.kt
msasikanth: Yes it has `capabilities` as well. But I wasn't sure it was confirmed and it was listed as a separate story. Should I go ahead with that or wait until all the API contracts are confirmed? I picked this up because it's the most simple API contract.

app/src/main/java/org/simple/clinic/storage/migrations/Migration_71.kt
msasikanth: Will do that from next migrations 👍🏾

app/src/main/java/org/simple/clinic/login/pin/LoginPinEffectHandler.kt
msasikanth: Ah okay, my bad. I was under the impression that `OngoingLoginEntry` is only local to device for registration purpose.

app/src/main/java/org/simple/clinic/user/LoggedInUserPayload.kt
vinaysshenoy: Hmm, alright. Let's add another migration for that once it's finalized.

vinaysshenoy:
@msasikanth Do we need to update the CHANGELOG here? We have updated the database schema, so I think adding a line is warranted.

msasikanth:
> @msasikanth Do we need to update the CHANGELOG here? We have updated the database schema, so I think adding a line is warranted.

Yeah, I was thinking about having a in-progress feature were we track all these changes. But I guess I can place this change in internal.

vinaysshenoy (approved):

vinaysshenoy (approved):

PR #1649, opened 2020-08-24
Fix crash on saving the state during registration facility selection

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1030/app-crashes-during-registration-when-it-is-backgrounded


msasikanth (approved):
LGTM

PR #1652, opened 2020-08-24
Add support for sending teleconsult request via SMS #1

msasikanth:
https://app.clubhouse.io/simpledotorg/story/885/add-support-for-sending-teleconsult-request-via-sms

### DIFF
--- a/app/src/main/java/org/simple/clinic/util/messagesender/SmsMessageSender.kt
+++ b/app/src/main/java/org/simple/clinic/util/messagesender/SmsMessageSender.kt
@@ -1,0 +1,19 @@
+package org.simple.clinic.util.messagesender
+
+import android.content.Intent
+import android.net.Uri
+import androidx.appcompat.app.AppCompatActivity
+import javax.inject.Inject
+
+class SmsMessageSender @Inject constructor(
+    private val activity: AppCompatActivity
+) : MessageSender {
+
+  override fun send(phoneNumber: String, message: String) {
+    val uri = Uri.parse("smsto:$phoneNumber")
+    val intent = Intent(Intent.ACTION_SENDTO, uri).apply {
+      putExtra("sms_body", message)
+    }
+    activity.startActivity(intent)
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Migrated `PatientSearchView` to Mobius
 - Added `TeleConsultSuccessScreen` to notify teleconsultation recorded successfully
 - Add `teleconsultPhoneNumber` in `User`, `LoggedInUserPayload` & `OngoingLoginEntry`
+- [In Progress: 24 Aug 2020] Add support for sending teleconsult request via SMS
 
 ## 2020-08-24-7397
 ### Fixes
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
@@ -54,8 +54,8 @@ import org.simple.clinic.summary.updatephone.UpdatePhoneNumberDialog
 import org.simple.clinic.util.Truss
 import org.simple.clinic.util.Unicode
 import org.simple.clinic.util.UserClock
-import org.simple.clinic.util.WhatsAppMessageSender
 import org.simple.clinic.util.extractSuccessful
+import org.simple.clinic.util.messagesender.WhatsAppMessageSender
 import org.simple.clinic.util.toLocalDateAtZone
 import org.simple.clinic.util.unsafeLazy
 import org.simple.clinic.widgets.ProgressMaterialButton.ButtonState
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
@@ -47,9 +47,9 @@ import org.simple.clinic.summary.addphone.AddPhoneNumberDialog
 import org.simple.clinic.summary.linkId.LinkIdWithPatientCancelled
 import org.simple.clinic.summary.linkId.LinkIdWithPatientLinked
 import org.simple.clinic.summary.linkId.LinkIdWithPatientViewShown
-import org.simple.clinic.summary.teleconsultation.TeleconsultationMessageBuilder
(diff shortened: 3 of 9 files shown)


app/src/main/java/org/simple/clinic/util/messagesender/SmsMessageSender.kt
```diff
@@ -0,0 +1,19 @@
...
+import android.net.Uri
+import androidx.appcompat.app.AppCompatActivity
+import javax.inject.Inject
+
+class SmsMessageSender @Inject constructor(
+    private val activity: AppCompatActivity
+) : MessageSender {
+
+  override fun send(phoneNumber: String, message: String) {
+    val uri = Uri.parse("smsto:$phoneNumber")
```
vinaysshenoy, line 13: Is this the official way to do this? AFAIK, there is a dedicated Intent provided by the platform instead of a URI scheme, right?

app/src/main/java/org/simple/clinic/summary/teleconsultation/messagebuilder/TeleconsultationSmsMessageBuilder.kt
```diff
@@ -0,0 +1,19 @@
+package org.simple.clinic.summary.teleconsultation.messagebuilder
+
+import org.simple.clinic.summary.PatientTeleconsultationInfo
+import javax.inject.Inject
+
+class TeleconsultationSmsMessageBuilder @Inject constructor() : TeleconsultationMessageBuilder {
```
vinaysshenoy, line 6: We can still send the long version of the message over SMS; the only reason we are doing it in this way is because we want to fit into the character limit for a single SMS.

What do you think of naming these message builders, `LongMessageBuilder` and `ShortMessageBuilder` instead? We can separate the length of the message from the medium used as the communication channel.

app/src/main/java/org/simple/clinic/summary/teleconsultation/messagebuilder/TeleconsultationSmsMessageBuilder.kt
```diff
@@ -0,0 +1,19 @@
...
+
+  // TODO (SM): 2020-08-24 Change the deep link
+  override fun message(patientTeleconsultationInfo: PatientTeleconsultationInfo): String {
+    val message = StringBuilder("Teleconsult request")
+        .appendLine()
+        .appendLine()
+        .appendLine("Review patient & log a record:")
+        .appendLine("https://app.simple.org/patient/${patientTeleconsultationInfo.patientUuid}")
```
vinaysshenoy, line 14: I am assuming that the change to the teleconsult URL format will happen later.

app/src/main/java/org/simple/clinic/summary/teleconsultation/messagebuilder/TeleconsultationSmsMessageBuilder.kt
```diff
@@ -0,0 +1,19 @@
...
+import javax.inject.Inject
+
+class TeleconsultationSmsMessageBuilder @Inject constructor() : TeleconsultationMessageBuilder {
+
+  // TODO (SM): 2020-08-24 Change the deep link
+  override fun message(patientTeleconsultationInfo: PatientTeleconsultationInfo): String {
+    val message = StringBuilder("Teleconsult request")
+        .appendLine()
+        .appendLine()
+        .appendLine("Review patient & log a record:")
```
vinaysshenoy, line 13: I wonder if we should add comments in this class and in the other message builder that we are explicitly not localizing these strings because we do not know the preferred language of the recipient. Otherwise, someone who is well meaning might end up extracting these to resources in the future without knowing.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/util/messagesender/SmsMessageSender.kt
msasikanth: Yes, this seems to be [official](https://developer.android.com/guide/components/intents-common#Messaging) way.

app/src/main/java/org/simple/clinic/summary/teleconsultation/messagebuilder/TeleconsultationSmsMessageBuilder.kt
msasikanth: Yes, that makes sense but the reason I went with this approach is, it seems more common to have different messages depending on the client. For example, we might have a completely different long message if it is an email. WDYT? or should I go ahead and change it to `LongMessageBuilder` & `ShortMessageBuilder`?

app/src/main/java/org/simple/clinic/summary/teleconsultation/messagebuilder/TeleconsultationSmsMessageBuilder.kt
msasikanth: Yep, I am trying to figure out how to handle both the old and new deep links URLs.

app/src/main/java/org/simple/clinic/summary/teleconsultation/messagebuilder/TeleconsultationSmsMessageBuilder.kt
msasikanth: Yep, will do that in later commits 👍🏾

msasikanth: NVM, ended up changing it in this PR.

app/src/main/java/org/simple/clinic/summary/teleconsultation/messagebuilder/TeleconsultationSmsMessageBuilder.kt
vinaysshenoy: What I mean is that a class should not have any idea about how it is going to be used. For the message builder itself, whether the message that it constructs is sent via Email, SMS, Whatsapp, or anything else is irrelevant since that decision is taken elsewhere.

So adding the communication channel to the class name adds this knowledge to the class while the class itself doesn't do anything with that knowledge internally.

msasikanth: Okay 👍🏾

vinaysshenoy (approved):
LGTM!

PR #1654, opened 2020-08-24
Simplify queries to fetch the current facility

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1054/remove-unnecessary-user-parameter-when-loading-current-facility


Honey14 (approved):
Looks good

PR #1655, opened 2020-08-24
Use UiRenderer for setting drug duration in `DrugDurationSheet`

msasikanth:


app/src/test/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationInitTest.kt
```diff
@@ -18,7 +18,7 @@ class DrugDurationInitTest {
         .whenInit(model)
         .then(assertThatFirst(
             hasModel(model),
-            hasEffects(SetDrugDuration(duration) as DrugDurationEffect)
+            hasNoEffects()
```
vinaysshenoy, line 21: With this, the name of the test is out of sync with the assertion. Please rename the test (or even delete it)?

vinaysshenoy (changes requested):

app/src/test/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationInitTest.kt
msasikanth: Oh yes, completely forgot. Removing the test

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationUiRenderer.kt
```diff
@@ -3,13 +3,22 @@ package org.simple.clinic.teleconsultlog.drugduration
 import org.simple.clinic.mobius.ViewRenderer
 import org.simple.clinic.teleconsultlog.drugduration.DrugDurationValidationResult.BLANK
 import org.simple.clinic.teleconsultlog.drugduration.DrugDurationValidationResult.NOT_VALIDATED
+import org.simple.clinic.util.ValueChangedCallback
 
 class DrugDurationUiRenderer(private val ui: DrugDurationUi) : ViewRenderer<DrugDurationModel> {
 
+  private val durationCallback = ValueChangedCallback<String>()
```
vinaysshenoy, line 10: It's perfectly fine to render this again. We use this only when migrating screens in the old arch to Mobius if we don't want to update the tests.

I'm approving this for now, we can remove this later.

vinaysshenoy (approved):

app/src/main/java/org/simple/clinic/teleconsultlog/drugduration/DrugDurationUiRenderer.kt
msasikanth: I have tried without it but the screen was freezing, because I am updating the model every time text changes and setting the text every time model changes, so it will be stuck in an endless loop. Am I missing something? 🤔

vinaysshenoy: Is this a Text field and this is to prefill the field? If so, the previous method of using an effect might have been the correct approach since the state management is being done by the Android framework 😅

msasikanth: So it's actually doing 2 things. We are prefilling the text field when the sheet is created, then we are saving the duration to model so that we retain the newly entered duration instead of going back to the previous value.

msasikanth: We talked about it in #1640

vinaysshenoy: Ah, why did we make it part of the UI renderer this time? 🤔

msasikanth: Oh, this felt like the "correct" approach instead of using UiActions. So basically, I wanted to have UiActions for only one time actions like dialogs, toast and snackbars. Also, what's the best way we can differentiate between what methods to go in UiActions and UiRenderer? or shall I continue with my approach?

msasikanth: Should I have not gone with this approach? 🤔

vinaysshenoy: I think "Prefilling" is a "one-time" action since you only need it to happen once when the screen is launched.

This is okay for now, let's get back to this once we are done with the feature.

PR #1658, opened 2020-08-25
Save and restore the state of the Home screen correctly

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1074/save-and-restore-the-state-of-the-homescreen-correctly


app/src/main/java/org/simple/clinic/home/HomeScreen.kt
```diff
@@ -91,6 +91,7 @@ class HomeScreen(context: Context, attrs: AttributeSet) : RelativeLayout(context
     TabLayoutMediator(homeTabLayout, viewPager) { tab, position ->
       tab.text = resources.getString(tabs[position].title)
     }.attach()
+    viewPager.isUserInputEnabled = false
```
msasikanth, line 94: Have we confirmed this behaviour change with designers since this might affect users?

app/src/main/java/org/simple/clinic/home/HomeScreen.kt
```diff
@@ -34,6 +35,8 @@ class HomeScreen(context: Context, attrs: AttributeSet) : RelativeLayout(context
   @Inject
   lateinit var effectHandlerFactory: HomeScreenEffectHandler.Factory
 
+  private val tabs = listOf(PATIENTS, OVERDUE, REPORTS)
```
msasikanth, line 38: Is there a specific reason why we are no longer using `HomeTab.values()`?

msasikanth (commented):
PR looks good overall. Left couple of comments PTAL

msasikanth:
Now that we have a home fragment, should we also take a look at opening patient summary screen from overdue?

app/src/main/java/org/simple/clinic/home/HomeScreen.kt
vinaysshenoy: Good point. Will inform.

app/src/main/java/org/simple/clinic/home/HomeScreen.kt
vinaysshenoy: It creates a new array each time, plus changing the order of declarations will change the order of the tabs. I thought it would be better to make the order explicit in the screen.

vinaysshenoy:
> Now that we have a home fragment, should we also take a look at opening patient summary screen from overdue?

Yup, I was thinking we could do this once the telemed MVP was done.

app/src/main/java/org/simple/clinic/home/HomeScreen.kt
msasikanth: Okay 👍🏾, the reason I asked is now we have to make sure every time we added a new enum to the `HomeTab` we have to remember to add it over here as well.

vinaysshenoy: Yeah, I think that's fine because anyone will see that the tabs are not updating.

vinaysshenoy: This also gives us the flexibility to conditionally change the screens based on say, a remote config flag.

msasikanth: Got it ✅

app/src/main/java/org/simple/clinic/home/HomeScreen.kt
vinaysshenoy: It's fine for now, please see the conversation at https://simpledotorg.slack.com/archives/CFM321423/p1598347202005400

msasikanth (approved):
LGTM

PR #1659, opened 2020-08-25
Add support for the teleconsultation record deeplink

msasikanth:
https://app.clubhouse.io/simpledotorg/story/884/add-support-for-the-teleconsultation-record-deeplink

### DIFF
--- a/app/src/main/java/org/simple/clinic/deeplink/DeepLinkUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/deeplink/DeepLinkUpdate.kt
@@ -10,7 +10,7 @@ class DeepLinkUpdate : Update<DeepLinkModel, DeepLinkEvent, DeepLinkEffect> {
   override fun update(model: DeepLinkModel, event: DeepLinkEvent): Next<DeepLinkModel, DeepLinkEffect> {
     return when (event) {
       is UserFetched -> userFetched(model, event)
-      is PatientFetched -> patientFetched(event)
+      is PatientFetched -> patientFetched(event, model)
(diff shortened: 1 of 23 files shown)


app/src/main/java/org/simple/clinic/deeplink/DeepLinkUpdate.kt
```diff
@@ -34,14 +34,22 @@ class DeepLinkUpdate : Update<DeepLinkModel, DeepLinkEvent, DeepLinkEffect> {
...
+    val effect = if (event.patient != null) {
+      navigateToPatientSummary(model)
     } else {
       ShowPatientDoesNotExist
     }
     return dispatch(effect)
   }
+
+  private fun navigateToPatientSummary(model: DeepLinkModel): DeepLinkEffect {
+    return if (model.isLogTeleconsultDeepLink) {
```
Honey14, line 49: Why would we want to remove `isLogTeleconsultDeepLink` later? We still would want to check if navigating to `NavigateToPatientSummaryWithTeleconsultLog` or `NavigateToPatientSummary`, right ?

msasikanth: WDYM? We will continue using it 🤔

Honey14: Oops, my mistake, there would PatientSummaryScreen with Teleconsult log only.

Honey14 (approved):
Looks good

PR #1660, opened 2020-08-25
Improve sync integration tests

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1080/improve-integration-tests-for-sync


vinaysshenoy:
@msasikanth @Honey14 I have added both of you as reviewers for this PR since it's quite a large PR and involves sync. However, most of it is mostly duplicated code except with some changes for different sync resources, so hopefully it should not be too much of a problem to review.

The instrumented tests will currently be failing because of [THIS](https://simpledotorg.slack.com/archives/CFHMC60P5/p1598420681013400), but it need not stop review as we can trigger the tests again later.

app/src/main/java/org/simple/clinic/remoteconfig/RemoteConfigSync.kt
```diff
@@ -28,7 +27,8 @@ class RemoteConfigSync @Inject constructor(
   override fun syncConfig(): Single<SyncConfig> {
     return Single.just(SyncConfig(
         syncInterval = SyncInterval.FREQUENT,
-        batchSize = BatchSize.SMALL,
-        syncGroup = SyncGroup.FREQUENT))
+        batchSize = 0, // Unused for remote config sync
```
msasikanth, line 29: Correct me if I am wrong, we are doing this because we want to sync the entire remote config data at once?

vinaysshenoy: Not really, what the case is here is the design of the sync framework assumes that anything that is synced has batches, but that is not the case for things like remote config, help, and reports. So the batch size does not matter here because it is never used for remote config sync.

app/src/androidTest/java/org/simple/clinic/sync/BaseSyncCoordinatorAndroidTest.kt
```diff
@@ -1,75 +0,0 @@
...
-import io.reactivex.Single
-import org.junit.Test
-import org.simple.clinic.patient.SyncStatus
-import org.simple.clinic.patient.SyncStatus.DONE
-import org.simple.clinic.patient.SyncStatus.PENDING
-import org.simple.clinic.util.None
-import org.simple.clinic.util.Optional
-
-// TODO: Use this class for all data sync tests.
-abstract class BaseSyncCoordinatorAndroidTest<T, P> {
```
msasikanth, line 15: Why are we removing this? 🤔

msasikanth (approved):
LGTM

app/src/androidTest/java/org/simple/clinic/sync/BaseSyncCoordinatorAndroidTest.kt
vinaysshenoy: Because the tests were badly written and deeply coupled to the implementation, they added no value. What these tests were meant to cover are now being handled by the new integration tests we have added.

Leaving these in will get in the way when refactoring, so I removed them because we don't need them anymore.

app/src/test/java/org/simple/clinic/sync/DataSyncTest.kt
```diff
@@ -18,18 +17,20 @@ class DataSyncTest {
 
   @Test
   fun `when syncing everything, all the model syncs should be invoked`() {
-    val modelSync1 = mock<ModelSync>()
-    val modelSync2 = mock<ModelSync>()
-    val modelSync3 = mock<ModelSync>()
-
-    whenever(modelSync1.syncConfig()).thenReturn(createSyncConfig(SyncGroup.DAILY))
-    whenever(modelSync1.sync()).thenReturn(Completable.complete())
+    val modelSync1 = FakeModelSync(
```
Honey14, line 32: Mocking or creating a fake class would do the same thing, right?

Honey14 (commented):
Left a few questions. Can you take a look?

app/src/main/java/org/simple/clinic/bloodsugar/sync/BloodSugarSync.kt
```diff
@@ -19,7 +19,7 @@ class BloodSugarSync @Inject constructor(
     private val repository: BloodSugarRepository,
     private val userSession: UserSession,
     @Named("last_blood_sugar_pull_token") private val lastPullToken: Preference<Optional<String>>,
-    @Named("sync_config_frequent") private val configProvider: Single<SyncConfig>
```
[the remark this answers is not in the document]
vinaysshenoy: This is mainly to reduce the complexity of using reactive classes. The other override will get removed in a later PR.

app/src/test/java/org/simple/clinic/sync/DataSyncTest.kt
vinaysshenoy: The problem with mocks is that I need to keep my mocks updated when changing implementation. Creating a fake object allows me to change implementation without needing to update the fake as much.

Honey14: Should'nt the mock be updating itself when we change the implementation of the class?

vinaysshenoy: We might get compilation errors in some cases, and other times we may not. Plus, since mocks need you to provide the stub each time, there might be scenarios where a test is passing because the mock stub is not behaving the same as a change we introduced, which gives a false sense of security.

PR #1666, opened 2020-08-28
Update the `PrescribedDrug` to include drug duration and frequency

msasikanth:
https://app.clubhouse.io/simpledotorg/story/1086/update-the-prescribeddrug-to-include-drug-duration-and-frequency

### DIFF
--- a/app/src/main/java/org/simple/clinic/drugs/sync/PrescribedDrugPayload.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/sync/PrescribedDrugPayload.kt
@@ -4,6 +4,7 @@ import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
 import org.simple.clinic.drugs.PrescribedDrug
 import org.simple.clinic.patient.SyncStatus
+import org.simple.clinic.teleconsultlog.medicinefrequency.MedicineFrequency
 import java.time.Instant
 import java.util.UUID
 
@@ -41,7 +42,13 @@ data class PrescribedDrugPayload(
     val updatedAt: Instant,
 
     @Json(name = "deleted_at")
-    val deletedAt: Instant?
+    val deletedAt: Instant?,
+
+    @Json(name = "frequency")
+    val frequency: MedicineFrequency?,
+
+    @Json(name = "duration_in_days")
+    val durationInDays: Int?
 ) {
 
   fun toDatabaseModel(syncStatus: SyncStatus): PrescribedDrug {
@@ -57,7 +64,9 @@ data class PrescribedDrugPayload(
         syncStatus = syncStatus,
         createdAt = createdAt,
         updatedAt = updatedAt,
-        deletedAt = deletedAt
+        deletedAt = deletedAt,
+        frequency = frequency,
+        durationInDays = durationInDays
     )
   }
 }
--- a/app/src/main/java/org/simple/clinic/teleconsultlog/medicinefrequency/MedicineFrequency.kt
+++ b/app/src/main/java/org/simple/clinic/teleconsultlog/medicinefrequency/MedicineFrequency.kt
@@ -1,8 +1,53 @@
 package org.simple.clinic.teleconsultlog.medicinefrequency
 
-enum class MedicineFrequency {
-  OD,
-  BD,
-  TDS,
-  QDS
+import android.os.Parcelable
+import androidx.room.TypeConverter
+import com.squareup.moshi.FromJson
+import com.squareup.moshi.ToJson
+import kotlinx.android.parcel.Parcelize
+import org.simple.clinic.util.room.SafeEnumTypeAdapter
+
+sealed class MedicineFrequency : Parcelable {
+
+  @Parcelize
+  object OD : MedicineFrequency()
+
+  @Parcelize
+  object BD : MedicineFrequency()
+
+  @Parcelize
+  object TDS : MedicineFrequency()
+
+  @Parcelize
+  object QDS : MedicineFrequency()
+
+  @Parcelize
+  data class Unknown(val actualValue: String) : MedicineFrequency()
+
+  object TypeAdapter : SafeEnumTypeAdapter<MedicineFrequency>(
+      knownMappings = mapOf(
+          OD to "od",
+          BD to "bd",
+          TDS to "tds",
+          QDS to "qds"
+      ),
+      unknownStringToEnumConverter = { Unknown(it) },
+      unknownEnumToStringConverter = { (it as Unknown).actualValue }
+  )
+
+  class RoomTypeConverter {
+    @TypeConverter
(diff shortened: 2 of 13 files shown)


app/src/main/java/org/simple/clinic/teleconsultlog/medicinefrequency/MedicineFrequency.kt
```diff
@@ -1,8 +1,53 @@
...
-  TDS,
-  QDS
+import android.os.Parcelable
+import androidx.room.TypeConverter
+import com.squareup.moshi.FromJson
+import com.squareup.moshi.ToJson
+import kotlinx.android.parcel.Parcelize
+import org.simple.clinic.util.room.SafeEnumTypeAdapter
+
+sealed class MedicineFrequency : Parcelable {
```
vinaysshenoy, line 10: Since we change to a sealed class from an enum, the earlier `.toString()` implementations will be broken since this will be different for every instance of the JVM process. Can we manually override the `.toString` for each of the `object` types to make them constant?

app/src/main/java/org/simple/clinic/teleconsultlog/medicinefrequency/MedicineFrequencySheet.kt
```diff
@@ -117,6 +117,7 @@ class MedicineFrequencySheet : BottomSheetActivity(), MedicineFrequencySheetUiAc
       BD -> R.id.medicineFrequencyBdRadioButton
       TDS -> R.id.medicineFrequencyTdsRadioButton
       QDS -> R.id.medicineFrequencyQdsRadioButton
+      is MedicineFrequency.Unknown -> throw AssertionError("Unknown medicine frequency $medicineFrequency")
```
vinaysshenoy, line 120: The reason we add this as an `Unknown` is so that we can have a graceful fallback in cases where an older version of the app may not know how to handle this new value. Throwing an error here defeats the purpose since it will cause an app crash.

We should just ignore `Unknown` values here instead of throwing an error.

app/src/main/java/org/simple/clinic/drugs/sync/PrescribedDrugPayload.kt
```diff
@@ -41,7 +42,13 @@ data class PrescribedDrugPayload(
...
 
     @Json(name = "deleted_at")
-    val deletedAt: Instant?
+    val deletedAt: Instant?,
+
+    @Json(name = "frequency")
+    val frequency: MedicineFrequency?,
+
+    @Json(name = "duration_in_days")
+    val durationInDays: Int?
```
vinaysshenoy, line 51: I was under the impression that we were only going to store this in the teleconsult record. If we make this change here, how will it work with the regular edit medicines flow (from the patient summary) where there is no UI to change the frequency and duration?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/drugs/sync/PrescribedDrugPayload.kt
msasikanth: We have decided to move it to `PrescribedDrug` to avoid having any metadata about the prescribed drugs when sending record and just send the prescribed drug UUID. You can find the conversation [here](https://simpledotorg.slack.com/archives/CV9V1DWLV/p1597820621045100?thread_ts=1597730690.017500&cid=CV9V1DWLV). As for the medicines UI, for now, we are continuing with the existing UI, but we might change it in future if needed.

app/src/main/java/org/simple/clinic/teleconsultlog/medicinefrequency/MedicineFrequencySheet.kt
msasikanth: Yes, will do that.

app/src/main/java/org/simple/clinic/teleconsultlog/medicinefrequency/MedicineFrequency.kt
msasikanth: we weren't using `.toString()` for these classes, should I still override that method? 🤔

vinaysshenoy: They will be used implicitly if we print any data classes which contain these properties as part of a string template.

app/src/main/java/org/simple/clinic/teleconsultlog/medicinefrequency/MedicineFrequency.kt
```diff
@@ -1,8 +1,69 @@
...
+
+  @Parcelize
+  data class Unknown(val actualValue: String) : MedicineFrequency()
+
+  object TypeAdapter : SafeEnumTypeAdapter<MedicineFrequency>(
+      knownMappings = mapOf(
+          OD to OD.toString(),
+          BD to BD.toString(),
+          TDS to TDS.toString(),
+          QDS to QDS.toString()
```
vinaysshenoy, line 48: What I meant was to override `.toString()` to return the same values that they would have as enums ["OD", "BD", "TDS", "QDS"] and leave the rest as is. Sorry if I wasn't clear about this.

app/src/main/java/org/simple/clinic/teleconsultlog/medicinefrequency/MedicineFrequencySheet.kt
```diff
@@ -117,6 +117,7 @@ class MedicineFrequencySheet : BottomSheetActivity(), MedicineFrequencySheetUiAc
       BD -> R.id.medicineFrequencyBdRadioButton
       TDS -> R.id.medicineFrequencyTdsRadioButton
       QDS -> R.id.medicineFrequencyQdsRadioButton
+      is MedicineFrequency.Unknown -> -1 // Don't check any radio button
     }
     medicineFrequencyRadioGroup.check(id)
```
vinaysshenoy, line 122: If we pass -1 to this, it won't cause a crash or anything?

vinaysshenoy (commented):
Left a comment about the static constants used for medicine frequency. Please take a look?

app/src/main/java/org/simple/clinic/teleconsultlog/medicinefrequency/MedicineFrequencySheet.kt
msasikanth: Nope, it will return without doing anything if the id is -1 or same as already checked id

app/src/main/java/org/simple/clinic/teleconsultlog/medicinefrequency/MedicineFrequency.kt
msasikanth: Ah okay, will change it 👍🏾

vinaysshenoy (approved):
LGTM!

PR #1667, opened 2020-08-28
Add Capabilties to user

Honey14:
https://app.clubhouse.io/simpledotorg/story/880/add-a-check-if-the-current-mo-is-allowed-to-log-a-teleconsult-record-at-the-current-facility

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration75AndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration75AndroidTest.kt
@@ -1,0 +1,136 @@
+package org.simple.clinic.storage.migrations
+
+import org.junit.Test
+import org.simple.clinic.assertColumns
+import org.simple.clinic.assertValues
+import org.simple.clinic.insert
+import org.simple.clinic.user.User
+import org.simple.clinic.user.UserStatus
+import org.simple.clinic.util.randomOfEnum
+import java.time.Instant
+import java.util.UUID
+
+class Migration75AndroidTest : BaseDatabaseMigrationTest(74, 75) {
+
+  @Test
+  fun capabilities_should_be_added_to_the_user_table() {
+    before.assertColumns("LoggedInUser", setOf(
+        "uuid",
+        "fullName",
+        "phoneNumber",
+        "pinDigest",
+        "status",
+        "createdAt",
+        "updatedAt",
+        "loggedInStatus",
+        "registrationFacilityUuid",
+        "currentFacilityUuid",
+        "teleconsultPhoneNumber"
+    ))
+
+    after.assertColumns("LoggedInUser", setOf(
+        "uuid",
+        "fullName",
+        "phoneNumber",
+        "pinDigest",
+        "status",
+        "createdAt",
+        "updatedAt",
+        "loggedInStatus",
+        "registrationFacilityUuid",
+        "currentFacilityUuid",
+        "teleconsultPhoneNumber",
+        "capability_canTeleconsult"
+    ))
+  }
+
+  @Test
+  fun logged_in_user_data_is_retained_after_the_migration() {
+    val facilityUuid = UUID.fromString("fabf0ddf-9f32-40b7-bd3e-3d975783f337")
+    before.insert("Facility", mapOf(
+        "uuid" to facilityUuid,
+        "name" to "PHC Obvious",
+        "facilityType" to null,
+        "streetAddress" to null,
+        "villageOrColony" to null,
+        "district" to "Bathinda",
+        "state" to "Punjab",
+        "country" to "India",
+        "pinCode" to null,
+        "protocolUuid" to null,
+        "groupUuid" to null,
+        "location_latitude" to null,
+        "location_longitude" to null,
+        "createdAt" to Instant.parse("2018-01-01T00:00:00Z"),
+        "updatedAt" to Instant.parse("2018-01-01T00:00:01Z"),
+        "deletedAt" to null,
+        "config_diabetesManagementEnabled" to true,
+        "config_teleconsultationEnabled" to true,
+        "syncStatus" to "DONE"
+    ))
+
+    before.insert("LoggedInUser", mapOf(
+        "uuid" to UUID.fromString("bbcf13e8-1a13-40b0-8556-ce6478b2499f"),
+        "fullName" to "Joshua Dawson",
+        "phoneNumber" to "999555888",
+        "pinDigest" to "1111",
+        "status" to "WaitingForApproval",
+        "createdAt" to Instant.parse("2018-01-01T00:00:00Z"),
+        "updatedAt" to Instant.parse("2018-01-01T00:00:00Z"),
+        "loggedInStatus" to "LOGGED_IN",
+        "registrationFacilityUuid" to facilityUuid,
+        "currentFacilityUuid" to facilityUuid,
+        "teleconsultPhoneNumber" to "85459839458"
+    ))
+
+    after.query(""" SELECT * FROM LoggedInUser """).use { cursor ->
+      cursor.moveToNext()
+      cursor.assertValues(mapOf(
+          "uuid" to UUID.fromString("bbcf13e8-1a13-40b0-8556-ce6478b2499f"),
+          "fullName" to "Joshua Dawson",
+          "phoneNumber" to "999555888",
+          "pinDigest" to "1111",
+          "status" to "WaitingForApproval",
+          "createdAt" to Instant.parse("2018-01-01T00:00:00Z"),
+          "updatedAt" to Instant.parse("2018-01-01T00:00:00Z"),
+          "loggedInStatus" to "LOGGED_IN",
+          "registrationFacilityUuid" to facilityUuid,
+          "currentFacilityUuid" to facilityUuid,
+          "teleconsultPhoneNumber" to "85459839458",
+          "capability_canTeleconsult" to null
+      ))
+    }
+
+  }
+
+
+  @Test
+  fun capabilities_should_be_added_to_the_Ongoing_login_entry_table() {
+    before.assertColumns("OngoingLoginEntry", setOf(
+        "uuid",
+        "phoneNumber",
+        "pin",
+        "fullName",
+        "pinDigest",
+        "registrationFacilityUuid",
+        "status",
+        "createdAt",
+        "updatedAt",
+        "teleconsultPhoneNumber",
+    ))
+
+    after.assertColumns("OngoingLoginEntry", setOf(
+        "uuid",
+        "phoneNumber",
+        "pin",
+        "fullName",
+        "pinDigest",
+        "registrationFacilityUuid",
+        "status",
+        "createdAt",
+        "updatedAt",
+        "teleconsultPhoneNumber",
+        "capability_canTeleconsult"
+    ))
+  }
+}
--- a/app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration75AndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration75AndroidTest.kt
@@ -80,7 +80,7 @@ class Migration75AndroidTest : BaseDatabaseMigrationTest(74, 75) {
         "loggedInStatus" to "LOGGED_IN",
         "registrationFacilityUuid" to facilityUuid,
         "currentFacilityUuid" to facilityUuid,
-        "teleconsultPhoneNumber" to "85459839458"
+        "teleconsultPhoneNumber" to "1111111111"
     ))
 
     after.query(""" SELECT * FROM LoggedInUser """).use { cursor ->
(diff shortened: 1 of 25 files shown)


app/src/main/java/org/simple/clinic/user/User.kt
```diff
@@ -138,6 +143,53 @@ data class User(
     UNAUTHORIZED;
 
     class RoomTypeConverter : RoomEnumTypeConverter<LoggedInStatus>(LoggedInStatus::class.java)
+
+  }
+
+  @Parcelize
+  data class Capabilities(val canTeleconsult: CapabilityType) : Parcelable
+
+  sealed class CapabilityType : Parcelable {
```
msasikanth, line 162: Can we change the name to be more appropriate of the use case, please? 

`CapabilityType` seems like it will have different capabilities, but in fact, we are just setting the enable/disable status of the capabilities. WDYT?

app/src/main/java/org/simple/clinic/user/User.kt
```diff
@@ -138,6 +143,53 @@ data class User(
...
+      @TypeConverter
+      fun fromEnum(enum: CapabilityType?) = TypeAdapter.fromEnum(enum)
+    }
+
+    class MoshiTypeAdapter {
+      @FromJson
+      fun toEnum(value: String?) = TypeAdapter.toEnum(value)
+
+      @ToJson
+      fun fromEnum(enum: CapabilityType) = TypeAdapter.fromEnum(enum)
```
msasikanth, line 196: Let's handle null cases as well here.

app/src/main/java/org/simple/clinic/user/User.kt
```diff
@@ -138,6 +143,53 @@ data class User(
...
+    }
+
+    companion object {
+      @VisibleForTesting
+      fun random() = CapabilityType.TypeAdapter.knownMappings.keys.shuffled().first()
```
msasikanth, line 201: I don't think we need it since we are not using it anywhere.

app/src/main/java/org/simple/clinic/storage/migrations/Migration_73.kt
```diff
@@ -0,0 +1,17 @@
+package org.simple.clinic.storage.migrations
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+import org.simple.clinic.storage.inTransaction
+import javax.inject.Inject
+
+@Suppress("Classname")
+class Migration_73 @Inject constructor() : Migration(72, 73) {
```
msasikanth, line 9: Just an FYI, since we are squshing the commits when merging you can do the migration in a different commit so that it will be easier to review.

app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration73AndroidTest.kt
```diff
@@ -0,0 +1,39 @@
+package org.simple.clinic.storage.migrations
+
+import org.junit.Test
+import org.simple.clinic.assertColumns
+
+class Migration73AndroidTest : BaseDatabaseMigrationTest(72, 73) {
+
+  @Test
+  fun `capabilities_should_be_added_to_the_user_table`() {
```
msasikanth, line 9: Can we also add a test to make sure the `LoggedInUser ` data is retained after the migration?

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -89,7 +89,8 @@ import org.simple.clinic.util.room.UuidRoomTypeConverter
     PatientStatus.RoomTypeConverter::class,
     ReminderConsent.RoomTypeConverter::class,
     BloodSugarMeasurementType.RoomTypeConverter::class,
-    DeletedReason.RoomTypeConverter::class
+    DeletedReason.RoomTypeConverter::class,
+    User.CapabilityType.TypeAdapter::class
```
msasikanth: ~~We need to add `RoomTypeConverter` here.~~ NVM it has been added in next commit

app/src/main/java/org/simple/clinic/login/LoginUserWithOtp.kt
```diff
@@ -61,7 +61,8 @@ class LoginUserWithOtp @Inject constructor(
           loggedInStatus = User.LoggedInStatus.LOGGED_IN,
           registrationFacilityUuid = payload.registrationFacilityId,
           currentFacilityUuid = payload.registrationFacilityId,
-          teleconsultPhoneNumber = payload.teleconsultPhoneNumber
+          teleconsultPhoneNumber = payload.teleconsultPhoneNumber,
+          capabilities = null
```
msasikanth, line 65: We need to pass the capabilities from the payload.

app/src/main/java/org/simple/clinic/login/pin/LoginPinEffectHandler.kt
```diff
@@ -89,7 +89,8 @@ class LoginPinEffectHandler @AssistedInject constructor(
         loggedInStatus = User.LoggedInStatus.OTP_REQUESTED,
         registrationFacilityUuid = entry.registrationFacilityUuid!!,
         currentFacilityUuid = entry.registrationFacilityUuid,
-        teleconsultPhoneNumber = entry.teleconsultPhoneNumber
+        teleconsultPhoneNumber = entry.teleconsultPhoneNumber,
+        capabilities = null
```
msasikanth, line 93: Setting `null` here is wrong, we need to update `OngoingLoginEntry` to have capabilities as well. Please refer `UsersApi#activate` API endpoint which will return `LoggedInUserPayload` which will have the capabilities.

app/src/main/java/org/simple/clinic/storage/migrations/Migration_73.kt
```diff
@@ -0,0 +1,17 @@
...
+  override fun migrate(database: SupportSQLiteDatabase) {
+    database.inTransaction {
+      execSQL("""
+        ALTER TABLE "LoggedInUser" ADD COLUMN "capability_canTeleconsult" TEXT
```
msasikanth, line 13: Let's set the default value to `null`

app/src/main/java/org/simple/clinic/user/UserSession.kt
```diff
@@ -92,7 +93,8 @@ class UserSession @Inject constructor(
           loggedInStatus = status,
           registrationFacilityUuid = payload.registrationFacilityId,
           currentFacilityUuid = payload.registrationFacilityId,
-          teleconsultPhoneNumber = payload.teleconsultPhoneNumber
+          teleconsultPhoneNumber = payload.teleconsultPhoneNumber,
+          capabilities = null
```
msasikanth, line 97: we need to set capabilities from the payload here

app/src/main/java/org/simple/clinic/user/refreshuser/RefreshCurrentUser.kt
```diff
@@ -81,7 +81,8 @@ class RefreshCurrentUser @Inject constructor(
           loggedInStatus = loggedInStatus,
           registrationFacilityUuid = payload.registrationFacilityId,
           currentFacilityUuid = storedUser.currentFacilityUuid,
-          teleconsultPhoneNumber = payload.teleconsultPhoneNumber
+          teleconsultPhoneNumber = payload.teleconsultPhoneNumber,
+          capabilities = null
```
msasikanth, line 85: we need to set capabilities from the payload here

app/src/main/java/org/simple/clinic/user/registeruser/RegisterUser.kt
```diff
@@ -71,7 +71,8 @@ class RegisterUser @Inject constructor(
           status = status,
           createdAt = createdAt,
           updatedAt = updatedAt,
-          teleconsultPhoneNumber = user.teleconsultPhoneNumber
+          teleconsultPhoneNumber = user.teleconsultPhoneNumber,
+          capabilities = null
```
msasikanth, line 75: we need to set capabilities from the user here

app/src/main/java/org/simple/clinic/user/registeruser/RegisterUser.kt
```diff
@@ -100,7 +101,8 @@ class RegisterUser @Inject constructor(
           loggedInStatus = LOGGED_IN,
           registrationFacilityUuid = payload.registrationFacilityId,
           currentFacilityUuid = payload.registrationFacilityId,
-          teleconsultPhoneNumber = payload.teleconsultPhoneNumber
+          teleconsultPhoneNumber = payload.teleconsultPhoneNumber,
+          capabilities = null
```
msasikanth, line 105: we need to set capabilities from the payload here

app/src/main/java/org/simple/clinic/user/resetpin/ResetUserPin.kt
```diff
@@ -72,7 +72,8 @@ class ResetUserPin @Inject constructor(
           loggedInStatus = RESET_PIN_REQUESTED,
           registrationFacilityUuid = registrationFacilityId,
           currentFacilityUuid = registrationFacilityId,
-          teleconsultPhoneNumber = teleconsultPhoneNumber
+          teleconsultPhoneNumber = teleconsultPhoneNumber,
+          capabilities = null
```
msasikanth, line 76: we need to set capabilities from the payload here

msasikanth (changes requested):
Overall PR looks good, left a few comments PTAL.

app/src/main/java/org/simple/clinic/user/User.kt
Honey14: Should we name it as `Capability`?

msasikanth: Probably something like `CapabilityStatus`?

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -100,7 +100,8 @@ import org.simple.clinic.util.room.UuidRoomTypeConverter
     DeletedReason.RoomTypeConverter::class,
     MedicineFrequency.RoomTypeConverter::class,
     TeleconsultationType.RoomTypeConverter::class,
-    RoomTypeConverter::class
+    RoomTypeConverter::class,
+    User.CapabilityStatus.TypeAdapter::class
```
msasikanth: ~~You have to pass the `User. CapabilityStatus.RoomTypeConverter::class` here~~ NVM it has been done in next commit

app/src/main/java/org/simple/clinic/user/User.kt
```diff
@@ -138,6 +142,48 @@ data class User(
     UNAUTHORIZED;
 
     class RoomTypeConverter : RoomEnumTypeConverter<LoggedInStatus>(LoggedInStatus::class.java)
+
+  }
+
+  @Parcelize
+  data class Capabilities(val canTeleconsult: CapabilityStatus) : Parcelable
+
+  sealed class CapabilityStatus : Parcelable {
```
msasikanth, line 161: Let's override `toString` for the sealed classes types since we will be using it as an enum.

app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration75AndroidTest.kt
```diff
@@ -0,0 +1,136 @@
...
+          "fullName" to "Joshua Dawson",
+          "phoneNumber" to "999555888",
+          "pinDigest" to "1111",
+          "status" to "WaitingForApproval",
+          "createdAt" to Instant.parse("2018-01-01T00:00:00Z"),
+          "updatedAt" to Instant.parse("2018-01-01T00:00:00Z"),
+          "loggedInStatus" to "LOGGED_IN",
+          "registrationFacilityUuid" to facilityUuid,
+          "currentFacilityUuid" to facilityUuid,
+          "teleconsultPhoneNumber" to "85459839458",
```
msasikanth, line 99: Let's not use actual phone numbers in our code. Let's go with `1111111111` or something like that.

app/src/sharedTest/java/org/simple/clinic/TestData.kt
```diff
@@ -1071,7 +1071,9 @@ object TestData {
       registrationFacilityUuid = registrationFacilityUuid,
       status = status,
       createdAt = createdAt,
-      updatedAt = updatedAt
+      updatedAt = updatedAt,
+      capabilities = null,
+      teleconsultPhoneNumber = "+919995558883"
```
msasikanth, line 1076: This can be an actual phone number, let's go with `+911111111111`

msasikanth, line 1076: Let's provide the option to override these values from `ongoingLoginEntry` method params or else it would be a hidden behaviour.

msasikanth (changes requested):
PR looks good to me. Left couple of comments PTAL

app/src/test/java/org/simple/clinic/login/pin/LoginPinScreenLogicTest.kt
```diff
@@ -128,7 +128,8 @@ class LoginPinScreenLogicTest {
         loggedInStatus = User.LoggedInStatus.OTP_REQUESTED,
         registrationFacilityUuid = registrationFacilityUuid,
         currentFacilityUuid = registrationFacilityUuid,
-        teleconsultPhoneNumber = teleconsultPhoneNumber
+        teleconsultPhoneNumber = teleconsultPhoneNumber,
+        capabilities = null
```
msasikanth, line 132: You can do it in next PR, can we also pass capabilities here? Look at how we passed `teleconsultPhoneNumber`

msasikanth (approved):
LGTM

PR #1668, opened 2020-08-31
Add lint rule for warning `ImageView#src` usage

msasikanth:


lint/src/main/java/org/simple/clinic/lint/ImageSrcDetector.kt
```diff
@@ -0,0 +1,53 @@
...
+    const val IMAGE_VIEW = "ImageView"
+    const val ATTR_SRC = "src"
+
+    private const val ImageSrcIssueId = "ImageSrcIssueId"
+    private const val ImageSrcIssueDescription = "Please use `app:srcCompat` when setting an image resource"
+
+    val ImageSrcIssue = Issue.create(
+        id = ImageSrcIssueId,
+        briefDescription = ImageSrcIssueDescription,
+        explanation = ImageSrcIssueDescription,
```
Honey14, line 26: We've named it `ImageSrcIssueDescription` but we are describing how to solve the issue.

`briefDescription` should summarises the issue while `explanation` should also include on how the new lint rule will be solved.

Honey14 (changes requested):
Looks great. Left one comment. PTAL.

lint/src/main/java/org/simple/clinic/lint/ImageSrcDetector.kt
msasikanth: Ah, good point. Will change it

Honey14 (approved):

PR #1670, opened 2020-09-01
Fix non-2XX response codes not being handled correctly in sync

vinaysshenoy:


app/src/main/java/org/simple/clinic/util/RetrofitExtensions.kt
```diff
@@ -2,8 +2,14 @@ package org.simple.clinic.util
...
 import retrofit2.HttpException
+import retrofit2.Response
 
 inline fun <reified T : Any> readErrorResponseJson(error: HttpException, moshi: Moshi): T {
   val jsonAdapter = moshi.adapter(T::class.java)
   return jsonAdapter.fromJson(error.response()!!.errorBody()!!.source())!!
 }
+
+
+fun <T> Response<T>.read(): T? {
```
msasikanth, line 13: It seems like we are force casting it to non-null in sync classes, can't we just do it in this extension?

msasikanth (approved):
Looks good. Left one comment PTAL

app/src/main/java/org/simple/clinic/util/RetrofitExtensions.kt
vinaysshenoy: This is an extension method which could also be used in non-sync contexts, so I didn't want to make the assumption for those usages as well.

msasikanth: Got it 👍🏾

PR #1671, opened 2020-09-01
Remove `process_token` usage from `TeleconsultationSync`

msasikanth:
You can find the reference to the conversation where we decided to remove this [here](https://simpledotorg.slack.com/archives/CV9V1DWLV/p1598945680015200)


Honey14 (approved):
Looks good

PR #1672, opened 2020-09-02
Add lint fix option to replace `ImageView#src` usage

msasikanth:
Added an option to provide lint fix option that can be used in Android Studio options to replace the existing usages with `app:srcCompat` usage.

In order to try this, please run `app:lintQaDebug` once and then replace any `app:srcCompat` usage with `android:src` and Android Studio should show the error and option to replace the usage.

### DIFF
--- a/lint/src/main/java/org/simple/clinic/lint/ImageSrcDetector.kt
+++ b/lint/src/main/java/org/simple/clinic/lint/ImageSrcDetector.kt
@@ -3,6 +3,7 @@ package org.simple.clinic.lint
 import com.android.tools.lint.detector.api.Category
 import com.android.tools.lint.detector.api.Implementation
 import com.android.tools.lint.detector.api.Issue
+import com.android.tools.lint.detector.api.LintFix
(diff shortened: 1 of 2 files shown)


lint/src/main/java/org/simple/clinic/lint/ImageSrcDetector.kt
```diff
@@ -45,8 +47,15 @@ class ImageSrcDetector : ResourceXmlDetector() {
...
-        ImageSrcExplanation
+        context.getLocation(element.getAttributeNodeNS(ANDROID_URI, "src")),
+        ImageSrcExplanation,
+        LintFix.create().composite(
+            LintFix.create().set(
+                AUTO_URI, "srcCompat",
+                element.getAttributeNS(ANDROID_URI, "src")
+            ).build(),
+            LintFix.create().unset(ANDROID_URI, "src").build()
+        )
```
vinaysshenoy, line 58: 👏

vinaysshenoy (approved):
Looks good! Will try it out once and see.

PR #1674, opened 2020-09-02
Add `TeleconsultRecord` and `TeleconsultRecordPayload`

Honey14:
https://app.clubhouse.io/simpledotorg/story/907/add-support-for-teleconsult-record-sync-resource

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordTest.kt
@@ -1,0 +1,130 @@
+package org.simple.clinic.teleconsultlog.teleconsultrecord
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+import org.simple.clinic.TestClinicApp
+import org.simple.clinic.TestData
+import org.simple.clinic.storage.Timestamps
+import org.simple.clinic.util.TestUserClock
+import org.simple.clinic.util.toUtcInstant
+import java.time.Instant
+import java.time.LocalDate
+import java.util.UUID
+import javax.inject.Inject
+
+class TeleconsultRecordTest {
+
+  @Inject
+  lateinit var teleconsultRecordDao: TeleconsultRecord.RoomDao
+
+  @Inject
+  lateinit var teleconsultRecordPrescribedDrugDao: TeleconsultRecordPrescribedDrug.RoomDao
+
+  @Inject
+  lateinit var teleconsultRecordWithPrescribedDrugs: TeleconsultRecordWithPrescribedDrugs.RoomDao
+
+  private val userClock = TestUserClock()
+
+  @Before
+  fun setup() {
+    TestClinicApp.appComponent().inject(this)
+  }
+
+  @After
+  fun tearDown() {
+    teleconsultRecordDao.clear()
+    teleconsultRecordPrescribedDrugDao.clear()
+  }
+
+  @Test
+  fun teleconsultation_records_with_prescribed_drugs_should_be_fetched_correctly() {
+
+    // given
+    val teleconsultRecordId1 = UUID.fromString("9ee0bc81-26e6-4ced-9b33-fa76a0a995e3")
+    val teleconsultRecordId2 = UUID.fromString("f1ad859f-d076-4bae-b0f2-3fb23c60d880")
+    val patientId = UUID.fromString("a706dfbf-315a-4201-be42-10b784c6d0b4")
+    val medicalOffcerId = UUID.fromString("97c558bb-df4e-4c7a-a5ac-df2520ccd4b1")
+    val facilityId = UUID.fromString("f12c2316-85e4-466e-9c92-efa52605c6b9")
+    val requesterId = UUID.fromString("7a131902-c214-4aa9-800f-201513d77fd7")
+    val date = LocalDate.parse("2020-09-03")
+    val prescribedDrugUuid = UUID.fromString("a6e44fdd-d029-47b0-9964-2ad4747166ec")
+    val prescribedDrugUuid2 = UUID.fromString("4904369e-c425-4098-bde1-560e09eb4383")
+
+    val teleconsultRecordInfo = TestData.teleconsultRecordInfo(
+        recordedAt = Instant.parse("2020-09-03T00:00:00Z"),
+        teleconsultationType = TeleconsultationType.Audio,
+        patientTookMedicines = Answer.Yes,
+        patientConsented = Answer.No,
+        medicalOfficerNumber = "22222222"
+    )
+
+    val teleconsultRequestInfo = TestData.teleconsultRequestInfo(
+        requesterId = requesterId,
+        facilityId = facilityId,
+        requestedAt = Instant.parse("2020-09-02T00:00:00Z"),
+    )
+
+    val teleconsultRecord1 = TestData.teleconsultationRecordInfo(
+        id = teleconsultRecordId1,
+        patientId = patientId,
+        medicalOfficerId = medicalOffcerId,
+        teleconsultRecordInfo = teleconsultRecordInfo,
+        teleconsultRequestInfo = teleconsultRequestInfo,
+        timestamps = Timestamps(
+            createdAt = date.toUtcInstant(userClock),
+            updatedAt = date.toUtcInstant(userClock),
+            deletedAt = null
+        )
+    )
+
+    val teleconsultRecord2 = TestData.teleconsultationRecordInfo(
+        id = teleconsultRecordId2,
+        patientId = patientId,
+        medicalOfficerId = medicalOffcerId,
+        teleconsultRecordInfo = teleconsultRecordInfo,
+        teleconsultRequestInfo = teleconsultRequestInfo,
+        timestamps = Timestamps(
+            createdAt = date.toUtcInstant(userClock),
+            updatedAt = date.toUtcInstant(userClock),
+            deletedAt = null
+        )
+    )
+
+    teleconsultRecordDao.save(
+        listOf(
+            teleconsultRecord1,
+            teleconsultRecord2
+        )
+    )
+
+    val teleconsultRecordPrescribedDrug1 = TestData.teleconsultationRecordPrescribedDrug(
+        teleconsultRecordId = teleconsultRecordId1,
+        prescribedDrugUuid = prescribedDrugUuid,
+    )
+
+    val teleconsultRecordPrescribedDrug2 = TestData.teleconsultationRecordPrescribedDrug(
+        teleconsultRecordId = teleconsultRecordId1,
+        prescribedDrugUuid = prescribedDrugUuid2,
+    )
+
+    val expectedTeleconsultRecordPrescribedDrugList = listOf(
+        teleconsultRecordPrescribedDrug2,
+        teleconsultRecordPrescribedDrug1
+    )
+
+    teleconsultRecordPrescribedDrugDao.save(
+        expectedTeleconsultRecordPrescribedDrugList
+    )
+
+    // when
+    val teleconsultRecordWithPrescribedDrugs = teleconsultRecordWithPrescribedDrugs
+        .getPrescribedUuidForTeleconsultRecordUuid(teleconsultRecordId1)
+
+    // then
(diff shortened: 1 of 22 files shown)


app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecord.kt
```diff
@@ -0,0 +1,46 @@
...
+    val teleconsultRequestInfo: TeleconsultRequestInfo?,
+
+    @Embedded(prefix = "record_")
+    val teleconsultRecordInfo: TeleconsultRecordInfo?,
+
+    val createdAt: String,
+
+    val updatedAt: String,
+
+    val deletedAt: String
```
vinaysshenoy: We have a `Timestamps` class which we should be using for new entities to represent these timestamps. Can we use that here instead?

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRequestInfo.kt
```diff
@@ -0,0 +1,11 @@
+package org.simple.clinic.teleconsultlog.teleconsultrecord
+
+import java.util.UUID
+
+data class TeleconsultRequestInfo(
+    val requesterId: UUID,
+
+    val facilityId: UUID,
+
+    val requestedAt: String
```
vinaysshenoy, line 10: This is a timestamp and it should be an `Instant`, no?

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordInfo.kt
```diff
@@ -0,0 +1,16 @@
+package org.simple.clinic.teleconsultlog.teleconsultrecord
+
+import org.simple.clinic.medicalhistory.Answer
+
+data class TeleconsultRecordInfo(
+
+    val recordedAt: String,
```
vinaysshenoy, line 7: This is a timestamp and should be an `Instant`, no?

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordInfo.kt
```diff
@@ -0,0 +1,16 @@
+package org.simple.clinic.teleconsultlog.teleconsultrecord
+
+import org.simple.clinic.medicalhistory.Answer
```
vinaysshenoy, line 3: We should not use the Medical history answer here and should create a dedicated type for this. They are two different concepts and should be represented differently, even if they look the same.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordPrescribedDrug.kt
```diff
@@ -0,0 +1,30 @@
...
+import androidx.room.OnConflictStrategy.REPLACE
+import androidx.room.PrimaryKey
+import androidx.room.Query
+import java.util.UUID
+
+@Entity
+data class TeleconsultRecordPrescribedDrug(
+    @PrimaryKey
+    val teleconsultRecordId: UUID,
+    val prescribedDrugUuid: UUID
```
vinaysshenoy, line 15: Shouldn't this be a composite primary key based of both the IDs? Since a single teleconsult record might have multiple prescriptions, saving a list of teleconsult prescribed drug records will end up only saving one since each one will replace the previous one.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecord.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    val patientId: UUID,
+
+    val medicalOfficerId: UUID,
+
+    @Embedded(prefix = "request_")
+    val teleconsultRequestInfo: TeleconsultRequestInfo?,
+
+    @Embedded(prefix = "record_")
+    val teleconsultRecordInfo: TeleconsultRecordInfo?,
```
vinaysshenoy, line 27: Shouldn't the type here be `TeleconsultRecordWithPrescribedDrugs`?

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordWithPrescribedDrugs.kt
```diff
@@ -0,0 +1,26 @@
...
+        entityColumn = "teleconsultRecordId"
+    )
+    val prescribedDrugs: List<TeleconsultRecordPrescribedDrug>
+
+) {
+  @Dao
+  interface RoomDao {
+
+    @Query("SELECT * FROM TeleconsultRecord")
+    fun getPrescribedUuidForTeleconsultRecordUuid(): List<TeleconsultRecordWithPrescribedDrugs>
```
vinaysshenoy, line 24: The name of this method implies that we should be passing in a teleconsult record ID and get the list of prescribed drug UUIDs which have been created as part of that teleconsult record but the method signature and the SQL query do not match this expectation.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultationType.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.squareup.moshi.ToJson
+import org.simple.clinic.util.room.SafeEnumTypeAdapter
+
+sealed class TeleconsultationType {
+
+  object Audio : TeleconsultationType()
+
+  object Video : TeleconsultationType()
+
+  object Message : TeleconsultationType()
```
vinaysshenoy, line 14: Can we override the `.toString()` method on each of the `object` types so that they behave as if they were enums instead of sealed class values?

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordInfoPayload.kt
```diff
@@ -0,0 +1,28 @@
...
+    @Json(name = "teleconsultation_type")
+    val teleconsultationType: TeleconsultationType,
+
+    @Json(name = "patient_took_medicines")
+    val patientTookMedicines: Answer,
+
+    @Json(name = "patient_consented")
+    val patientConsented: Answer,
+
+    @Json(name = "medical_officier_number")
```
vinaysshenoy, line 22: There seems to be a typo in the JSON property name: `officier` instead of `officer`.

Since this has not gone into production and there are no consumers of this API, can we get it corrected instead of bubbling down the typo into the client code?

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordInfoPayload.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+import org.simple.clinic.medicalhistory.Answer
+import java.util.UUID
+
+@JsonClass(generateAdapter = true)
+data class TeleconsultRecordInfoPayload(
+    @Json(name = "recorded_at")
+    val recordedAt: String,
```
vinaysshenoy, line 11: This should be an `Instant`.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordPayload.kt
```diff
@@ -0,0 +1,33 @@
...
+    val teleconsultRecordInfo: TeleconsultRecordInfoPayload?,
+
+    @Json(name = "created_at")
+    val createdAt: String,
+
+    @Json(name = "updated_at")
+    val updatedAt: String,
+
+    @Json(name = "deleted_at")
+    val deletedAt: String
```
vinaysshenoy, line 32: These are timestamps and should be `Instant` types.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRequestInfoPayload.kt
```diff
@@ -0,0 +1,18 @@
...
+@JsonClass(generateAdapter = true)
+data class TeleconsultRequestInfoPayload(
+    @Json(name = "requester_id")
+    val requesterId: UUID,
+
+    @Json(name = "facility_id")
+    val facilityId: UUID,
+
+    @Json(name = "requested_at")
+    val requestedAt: String
```
vinaysshenoy, line 16: This should be `Instant`.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordPrescribedDrug.kt
```diff
@@ -0,0 +1,30 @@
...
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy.REPLACE
+import androidx.room.PrimaryKey
+import androidx.room.Query
+import java.util.UUID
+
+@Entity
+data class TeleconsultRecordPrescribedDrug(
+    @PrimaryKey
+    val teleconsultRecordId: UUID,
```
vinaysshenoy, line 14: This should also be a foreign key with a delete cascade action into the `TeleconsultRecord` table since we cannot have this without creating a record first.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecord.kt
```diff
@@ -0,0 +1,46 @@
...
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy.REPLACE
+import androidx.room.PrimaryKey
+import androidx.room.Query
+import androidx.room.Relation
+import org.simple.clinic.drugs.PrescribedDrug
+import java.util.UUID
+
+@Entity
+data class TeleconsultRecord(
```
vinaysshenoy, line 15: This is quite a complex model. Can we write a small integration test that constructs a complete object of this type and saves it into the database and then reads it back to verify that it was done correctly?

vinaysshenoy (changes requested):
Overall PR looks good, left some comments about the types being chosen and a modelling issue.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecord.kt
Honey14: No, `TeleconsultRecordInfo` contains record information while the `TeleconsultRecordWithPrescribedDrugs` is a table that will have teleconsultRecordId and prescribedDrugUuid

vinaysshenoy:
@Honey14 It seems like the database migration test has failed, please take a look?

app/src/androidTest/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordTest.kt
```diff
@@ -0,0 +1,126 @@
...
+    teleconsultRecordPrescribedDrugDao.save(
+        expectedTeleconsultRecordPrescribedDrugList
+    )
+
+    // when
+    val teleconsultRecordPrescribedDrugList = teleconsultRecordPrescribedDrugDao
+        .getAll()
+
+    // then
+    assertThat(teleconsultRecordPrescribedDrugList).isEqualTo(expectedTeleconsultRecordPrescribedDrugList)
```
vinaysshenoy, line 123: This does not test the complete picture since we are not verifying whether multiple prescribed drugs for a single teleconsult record are being saved correctly or not. We should just save a single complete teleconsult record with multiple prescribed drug UUIDs and then verify that loading that `TeleconsultRecordWithPrescribedDrugs` record is as expected.

vinaysshenoy (changes requested):
Overall PR looks good, there's a failing test and one improvement we can make to the tests and then it will be good to merge. 👍

vinaysshenoy (approved):

PR #1676, opened 2020-09-03
Add global error handler to catch RxJava errors that can be ignored

vinaysshenoy:
Since we were using Retrofit observable calls earlier, certain exceptions were being swallowed by them internally when the stream was disposed and a request was ongoing.

However, since we have moved to synchronous calls, it is up to us to manage disposal of the streams ourselves. I've elected to use the method prescribed by the RxJava error handling docs (See [HERE](https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling)) and swallow these exceptions globally.

### DIFF
--- a/app/src/main/java/org/simple/clinic/ClinicApp.kt
+++ b/app/src/main/java/org/simple/clinic/ClinicApp.kt
@@ -7,6 +7,8 @@ import androidx.camera.camera2.Camera2Config
 import androidx.camera.core.CameraXConfig
 import androidx.work.Configuration
 import androidx.work.WorkManager
+import io.reactivex.exceptions.UndeliverableException
+import io.reactivex.plugins.RxJavaPlugins
 import org.simple.clinic.activity.CloseActivitiesWhenUserIsUnauthorized
 import org.simple.clinic.analytics.UpdateAnalyticsUserId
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/ClinicApp.kt
```diff
@@ -51,6 +55,12 @@ abstract class ClinicApp : Application(), CameraXConfig.Provider {
 
     crashReporter.init(this)
     Timber.plant(CrashBreadcrumbsTimberTree(crashReporter))
+    RxJavaPlugins.setErrorHandler { error ->
+      if (!error.canBeIgnoredSafely()) {
+        val cause = if(error is UndeliverableException) error.cause else error
+        Thread.currentThread().uncaughtExceptionHandler?.uncaughtException(Thread.currentThread(), cause!!)
```
msasikanth, line 61: Can you tell me why we are doing this here, please?

msasikanth (approved):
LGTM

app/src/main/java/org/simple/clinic/ClinicApp.kt
vinaysshenoy: Some errors we can safely ignore, others we have to manually throw and ensure that it gets reported. This is what is happening here.

PR #1677, opened 2020-09-03
Bump SQLite version to 3.32.2

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1152/migrate-app-sqlite-library-to-the-newest-version

# Changes
Due to a change in the behaviour of the `ALTER TABLE` statement, I had to make some changes to the current set of migrations to stop them from failing tests. The changes that I made are fallback behaviour and should not be used going forward for new `ALTER TABLE` migrations and we should be setting the index and table names explicitly instead of using implicit behaviour.

The documentation on the changes and the process is described on the SQLite `ALTER TABLE` [page](https://sqlite.org/lang_altertable.html).


app/src/main/java/org/simple/clinic/storage/migrations/Migration_6_7.kt
```diff
@@ -51,6 +55,10 @@ class Migration_6_7 @Inject constructor() : Migration(6, 7) {
       FROM `LoggedInUser_v6`
       """)
       database.execSQL("DROP TABLE `LoggedInUser_v6`")
+
+      database.execSQL("PRAGMA legacy_alter_table = OFF")
+      database.execSQL("PRAGMA foreign_key_check")
```
Honey14, line 60: we would only to check the `PRAGMA foreign_key_check` when the foreign key constraints were originally enabled, right? Since we have disabled above, do we need this here?

app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration36AndroidTest.kt
```diff
@@ -0,0 +1,284 @@
...
+      'null',
+      'recorded-at')
+    """)
+
+    after.query("SELECT * FROM $tableName").use {
+      assertThat(it.count).isEqualTo(1)
+      assertThat(it.getColumnIndex("recordedAt")).isEqualTo(-1)
+      it.moveToNext()
+      assertThat(it.columnCount).isEqualTo(12)
+      assertThat(it.string("uuid")).isEqualTo(mhUuid)
```
Honey14, line 101: Can you explain what is happening here?

Honey14 (commented):
Looks good. Left one doubt and a comment. Please take a look

app/src/main/java/org/simple/clinic/storage/migrations/Migration_6_7.kt
vinaysshenoy: We do need it here, it verifies that the foreign key setup is correct and then we can enable them again. This is as per the "Making Other Kinds Of Table Schema Changes" section in the sqlite `ALTER TABLE` document linked earlier.

app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration36AndroidTest.kt
vinaysshenoy: This is mostly a copy from an older migration. In a previous migration, we had added the `recordedAt` column to the medical history table, but we decided not to use that field so we dropped it in this migration.

Honey14 (approved):

PR #1679, opened 2020-09-04
Change storage of app lock timestamps to be in-memory instead of in shared preferences

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1148/stop-storing-app-lock-timestamp-in-shared-preferences

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Add `ImageSrcDetector` lint for warning when using `ImageView#src`
 - Add `TeleconsultRecord` and `TeleconsultRecordPayload`
 - Bumped internal SQLite version to 3.32.2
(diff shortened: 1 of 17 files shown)


msasikanth (approved):
Looks good. 

Not an important change, but while we are at app lock screen can we make a small change to `pin_entry_card`, can we set the `pinEditText` `imeOptions` to `actionDone`? Currently, it shows return/newline key.

vinaysshenoy:
> Looks good.
> 
> Not an important change, but while we are at app lock screen can we make a small change to `pin_entry_card`, can we set the `pinEditText` `imeOptions` to `actionDone`? Currently, it shows return/newline key.

Let's make it a new PR since commits get squashed now.

PR #1682, opened 2020-09-07
Add support for recording a teleconsultation log for the patient #2

msasikanth:


app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordWithPrescribedDrugs.kt
```diff
@@ -21,6 +22,7 @@ data class TeleconsultRecordWithPrescribedDrugs(
   @Dao
   interface RoomDao {
 
+    @Transaction
```
vinaysshenoy, line 25: Is there a reason we are executing a select statement in a transaction? It normally is useful only when we are inserting data into multiple tables and we want it to be atomic.

One place where it might make sense is if we are simultaneously querying and writing to the same record, but that doesn't seem to be the case here.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
```diff
@@ -23,9 +32,36 @@ class TeleconsultRecordEffectHandler @AssistedInject constructor(
...
         .build()
   }
 
+  private fun createTeleconsultRecord(): ObservableTransformer<CreateTeleconsultRecord, TeleconsultRecordEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .map { effect ->
+            teleconsultRecordRepository.createTeleconsultRecordForMedicalOfficer(
+                teleconsultRecordId = uuidGenerator.v4(),
```
vinaysshenoy, line 45: For the medical officer, aren't we supposed to get the teleconsult record ID from the deeplink?

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
```diff
@@ -23,9 +32,36 @@ class TeleconsultRecordEffectHandler @AssistedInject constructor(
...
+                teleconsultRecordInfo = createTeleconsultRecordInfo(effect)
+            )
+          }
+          .map { TeleconsultRecordCreated }
+    }
+  }
+
+  private fun createTeleconsultRecordInfo(it: CreateTeleconsultRecord): TeleconsultRecordInfo {
+    return TeleconsultRecordInfo(
+        recordedAt = Instant.now(utcClock),
```
vinaysshenoy, line 57: There might be a couple of millisecond difference between this timestamp and the one used to create the timestamps in `TeleconsultRecordRepository#createTeleconsultRecordForMedicalOfficer`.

Since we are going to be merging multiple teleconsult records on the server side for reports, can we confirm with the server team whether they will depend on these timestamps being the same in order to merge them or select which records to use?

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
```diff
@@ -23,9 +32,36 @@ class TeleconsultRecordEffectHandler @AssistedInject constructor(
...
         .addTransformer(LoadTeleconsultRecordWithPrescribedDrugs::class.java, loadTeleconsultRecordWithPrescribedDrugs())
+        .addTransformer(CreateTeleconsultRecord::class.java, createTeleconsultRecord())
         .build()
   }
 
+  private fun createTeleconsultRecord(): ObservableTransformer<CreateTeleconsultRecord, TeleconsultRecordEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .map { effect ->
```
vinaysshenoy, line 43: Since we aren't using the return value of `createTeleconsultRecordForMedicalOfficer` in the next operator, should this be a `.doOnNext` instead?

app/src/test/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandlerTest.kt
```diff
@@ -78,4 +91,45 @@ class TeleconsultRecordEffectHandlerTest {
...
+            patientTookMedicines = Yes,
+            patientConsented = Yes,
+            medicalOfficerNumber = null
+        ),
+        teleconsultRequestInfo = null,
+        createdAt = Instant.now(utcClock),
+        updatedAt = Instant.now(utcClock)
+    )
+
+    doNothing().whenever(teleconsultRecordRepository).createTeleconsultRecordForMedicalOfficer(
```
vinaysshenoy, line 115: Is the `doNothing()` necessary here?

app/src/test/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandlerTest.kt
```diff
@@ -78,4 +91,45 @@ class TeleconsultRecordEffectHandlerTest {
...
+    // when
+    effectHandlerTestCase.dispatch(CreateTeleconsultRecord(
+        patientUuid = patientUuid,
+        teleconsultationType = Audio,
+        patientTookMedicine = Yes,
+        patientConsented = Yes
+    ))
+
+    // then
+    effectHandlerTestCase.assertOutgoingEvents(TeleconsultRecordCreated)
```
vinaysshenoy, line 131: We should also verify the side effect of creating the record on the repository mock, correct?

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordInit.kt
```diff
@@ -0,0 +1,12 @@
+package org.simple.clinic.teleconsultlog.teleconsultrecord.screen
+
+import com.spotify.mobius.First
+import com.spotify.mobius.Init
+import org.simple.clinic.mobius.first
+
+class TeleconsultRecordInit : Init<TeleconsultRecordModel, TeleconsultRecordEffect> {
+
+  override fun init(model: TeleconsultRecordModel): First<TeleconsultRecordModel, TeleconsultRecordEffect> {
+    return first(model, LoadTeleconsultRecordWithPrescribedDrugs(model.teleconsultRecordId))
```
vinaysshenoy, line 10: Don't we need to check for the screen restoration use case?

app/src/test/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordInitTest.kt
```diff
@@ -0,0 +1,24 @@
...
+import com.spotify.mobius.test.FirstMatchers.hasModel
+import com.spotify.mobius.test.InitSpec
+import com.spotify.mobius.test.InitSpec.assertThatFirst
+import org.junit.Test
+import java.util.UUID
+
+class TeleconsultRecordInitTest {
+
+  @Test
+  fun `when screen is created and record id is not null, then load teleconsult record with prescribed drugs`() {
```
vinaysshenoy, line 13: Can we not use implementation specific terms like `record id is not null` in the name of the test please?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordWithPrescribedDrugs.kt
msasikanth: It's defined in the Room relationships docs, it also complains as warning every time the annotation processor runs.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
msasikanth: Ah yes, good catch. Fixing it

app/src/test/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordInitTest.kt
msasikanth: Oh yes, I forgot to rename the test. We won't ever have a scenario with teleconsult record id being null.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordInit.kt
msasikanth: Yes, but I couldn't figure out which things to check in this particular scenario. Because there will be default values in the model from start. So we cannot just verify if it's null. Any ideas on how we can handle these scenarios?

app/src/test/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandlerTest.kt
msasikanth: Without this mock is failing for void return methods.

vinaysshenoy: Can we recheck? If I remember correctly, Mockito automatically handles stubs for void returns.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordInit.kt
vinaysshenoy: Ah, I see what you mean. Let's leave it as is for now and we can revisit this later.

app/src/test/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandlerTest.kt
msasikanth: I have tested with this and it fails with `UnfinishedStubbingException`
```kt
    whenever(teleconsultRecordRepository.createTeleconsultRecordForMedicalOfficer(
        teleconsultRecordId = teleconsultRecord.id,
        patientUuid = teleconsultRecord.patientId,
        medicalOfficerId = teleconsultRecord.medicalOfficerId,
        teleconsultRecordInfo = teleconsultRecord.teleconsultRecordInfo!!
    ))
```

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
msasikanth: The conversation is happening [here](https://simpledotorg.slack.com/archives/CV9V1DWLV/p1599472445007200).

app/src/test/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandlerTest.kt
vinaysshenoy: Huh, this is new. Checking.

vinaysshenoy: Oh wait, I mean you should just skip setting up a stub at all in the first place and use a `verify(teleconsultRepository).createTeleconsultRecordForMedicalOfficer()` verification call.

vinaysshenoy: If you don't start a stubbing with `whenever`, Mockito will handle it by returning `null` for methods which return something and `Unit` for methods that don't.

msasikanth: Ah okay, will do that 👍🏽

vinaysshenoy: Yup, I tested it out right now by just removing the entire stubbing and it works.

vinaysshenoy (approved):
LGTM!

PR #1684, opened 2020-09-08
Add support for recording a teleconsultation log for the patient #3

msasikanth:


msasikanth:
Screenshots of the feature, you can try the feature by creating a deep link for an existing patient with the new deep link format.

screenshot-1599545456248

screenshot-1599545462879

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEvent.kt
```diff
@@ -3,17 +3,26 @@ package org.simple.clinic.teleconsultlog.teleconsultrecord.screen
...
+}
 
-data class TeleconsultRecordWithPrescribedDrugsLoaded(val teleconsultRecordWithPrescribedDrugs: TeleconsultRecordWithPrescribedDrugs?) : TeleconsultRecordEvent()
+data class TeleconsultRecordWithPrescribedDrugsLoaded(val teleconsultRecordWithPrescribedDrugs: TeleconsultRecordWithPrescribedDrugs?) : TeleconsultRecordEvent() {
+  override val analyticsName: String = "Teleconsult Record:Record with Prescribed Drugs Loaded"
+}
 
-object TeleconsultRecordCreated : TeleconsultRecordEvent()
+object TeleconsultRecordCreated : TeleconsultRecordEvent() {
+  override val analyticsName: String = "Teleconsult Record:Record Created"
```
vinaysshenoy, line 20: We should not be overriding analytics names for events which are not UI events unless we specifically want to capture them.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordScreen.kt
```diff
@@ -0,0 +1,146 @@
...
+  override fun onRestoreInstanceState(state: Parcelable?) {
+    super.onRestoreInstanceState(delegate.onRestoreInstanceState(state))
+  }
+
+  override fun onFinishInflate() {
+    super.onFinishInflate()
+    if (isInEditMode) return
+
+    context.injector<Injector>().inject(this)
+    backClicks()
```
vinaysshenoy, line 98: Can we rename this to something like `setupBackClickListener()`?

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/Answer.kt
```diff
@@ -1,16 +1,36 @@
...
+      return if (boolean) Yes else No
+    }
+
+    fun toBoolean(answer: Answer): Boolean {
+      return when (answer) {
+        Yes -> true
+        No,
+        is Unknown -> false
+      }
+    }
```
vinaysshenoy, line 24: I would like to avoid making boolean functions like this a part of the domain since they hide a lot of information. Can we move these to the screen where they are being used so that the potential for misuse does not spread?

app/src/main/res/values/strings.xml
```diff
@@ -690,5 +690,13 @@
   <string name="sheet_medicine_frequency_qds">QDS</string>
   <string name="sheet_medicine_frequency_save">Save</string>
 
-
-</resources>
+  <!-- Teleconsult Record Screen -->
+  <string name="screen_teleconsult_record">TELECONSULT RECORD</string>
```
vinaysshenoy, line 694: If we want this to be displayed in all caps, can we add this as an attribute to the `TextView` with `textAllCaps=true` instead of in the string resource?

app/src/main/java/org/simple/clinic/deeplink/DeepLinkActivity.kt
```diff
@@ -131,7 +131,7 @@ class DeepLinkActivity : AppCompatActivity(), DeepLinkUiActions {
     finish()
   }
 
-  override fun navigateToPatientSummaryWithTeleconsultLog(patientUuid: UUID, teleconsultRecordId: UUID?) {
+  override fun navigateToPatientSummaryWithTeleconsultLog(patientUuid: UUID, teleconsultRecordId: UUID) {
```
vinaysshenoy, line 134: Will this change still work if a medical officer receives a teleconsult request from the current implementation (without the teleconsult record ID)?

We would still want the older flow to work if the medical officer has updated to the new version but the nurse has not.

app/src/test/java/org/simple/clinic/summary/PatientSummaryViewRendererTest.kt
```diff
@@ -279,7 +279,7 @@ class PatientSummaryViewRendererTest {
     // given
     val patientUuid = UUID.fromString("66bc73a4-29a1-4b07-8be1-9223dc5d7cbb")
     val model = PatientSummaryModel.from(
-        openIntention = OpenIntention.ViewExistingPatientWithTeleconsultLog(
+        openIntention = OpenIntention.  ViewExistingPatientWithTeleconsultLog(
```
vinaysshenoy, line 282: Was this extra space added accidentally?

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordWithPrescribedDrugs.kt
```diff
@@ -24,6 +24,6 @@ data class TeleconsultRecordWithPrescribedDrugs(
 
     @Transaction
     @Query("SELECT * FROM TeleconsultRecord WHERE id = :teleconsultRecordUuid")
-    fun getPrescribedUuidForTeleconsultRecordUuid(teleconsultRecordUuid: UUID): TeleconsultRecordWithPrescribedDrugs
+    fun getPrescribedUuidForTeleconsultRecordUuid(teleconsultRecordUuid: UUID): TeleconsultRecordWithPrescribedDrugs?
```
vinaysshenoy, line 27: Can we rename this method to something like `getCompleteTeleconsultLog()`?

vinaysshenoy (changes requested):
Overall PR looks good, left a few minor comments.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordScreen.kt
msasikanth: This is a temporary method, it is replaced in the final PR with a Observable stream so that we can pass it to Mobius loop.

app/src/main/java/org/simple/clinic/deeplink/DeepLinkActivity.kt
msasikanth: Yes, there is a different function for that called `navigateToPatientSummary`. So the older flow will continue to work as expected.

app/src/test/java/org/simple/clinic/summary/PatientSummaryViewRendererTest.kt
msasikanth: Yes. Will fix it 👍🏽

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordUiRenderer.kt
```diff
@@ -0,0 +1,19 @@
...
+
+  override fun render(model: TeleconsultRecordModel) {
+    renderTeleconsultationType(model)
+
+    ui.setPatientTookMedicines(model.patientTookMedicines)
+    ui.setPatientConsented(model.patientConsented)
+  }
+
+  private fun renderTeleconsultationType(model: TeleconsultRecordModel) {
+    ui.setTeleconsultationType(model.teleconsultationType)
```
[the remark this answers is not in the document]
msasikanth: Nope. Initially, I was planning on having individual methods for checking the radio buttons. But later decided to let `RadioGroup` handle it.

app/src/main/res/values/strings.xml
vinaysshenoy: That's something I found out today. But I still feel we should use the attribute to control this because it's a property of the view and not of the string resource itself.

vinaysshenoy: This is okay for now, we can change it in the final PR.

vinaysshenoy (approved):
LGTM!

PR #1687, opened 2020-09-09
Disable state saving and restoration for the search results screen

vinaysshenoy:
This is temporarily done to stop crashes from persisting lists of
search results in the saved state bundle until paging library v3 is ready.

### DIFF
--- a/app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchView.kt
+++ b/app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchView.kt
@@ -68,6 +68,7 @@ class PatientSearchView(context: Context, attrs: AttributeSet) : RelativeLayout(
 
     val screenDestroys = detaches().map { ScreenDestroyed() }
     setupScreen(screenDestroys)
+    isSaveEnabled = false
   }
 
   override fun onAttachedToWindow() {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchView.kt
```diff
@@ -68,6 +68,7 @@ class PatientSearchView(context: Context, attrs: AttributeSet) : RelativeLayout(
 
     val screenDestroys = detaches().map { ScreenDestroyed() }
     setupScreen(screenDestroys)
+    isSaveEnabled = false
```
Honey14, line 71: Do we still need to override `onSaveInstanceState` and `onRestoreInstanceState` if we're disabling state restoration?

Honey14 (commented):
Looks good. Left one question. PTAL

app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchView.kt
vinaysshenoy: This is only temporary and prevents them from being invoked. Once we add Paging v3, we will remove this and let this be saved and restored as usual.

Honey14 (approved):

PR #1690, opened 2020-09-09
Add support for recording a teleconsultation log for the patient - Final

msasikanth:


app/src/test/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandlerTest.kt
```diff
@@ -164,4 +164,26 @@ class TeleconsultRecordEffectHandlerTest {
...
+
+  @Test
+  fun `when validate teleconsult record effect is received, then check if the teleconsult record exists`() {
+    // given
+    val teleconsultRecordId = UUID.fromString("9b0cf077-ef4a-449a-9588-db115dc7eb7c")
+    val teleconsultRecordWithPrescribedDrugs = TestData.teleconsultRecordWithPrescribedDrugs(
+        teleconsultRecord = TestData.teleconsultRecord(
+            id = teleconsultRecordId
+        ),
+        prescribedDrugs = emptyList()
```
Honey14, line 176: There isn't a case where `teleconsultRecordId` will get stored in the `TeleconsultRecordPrescribedDrug` table without `prescribedDrugs` list in it. We shouldn't have `prescribedDrugs` as emptyList

Honey14 (changes requested):

app/src/test/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandlerTest.kt
msasikanth: Actually there is, when the record is first created on medical officer device we won't be adding the prescriptions. Those are done in the next screen. Even if we consider the sync resource, the prescriptions will be added back in the patient prescription screen on save.

Honey14: So the record in `TeleconsultRecordPrescribedDrug` will not be created yet, right?

msasikanth: Also, we are not saving directly to `TeleconsultRecordPrescribedDrug` table. It will just get teleconsult record and the prescribed drugs separately.

Honey14: So when the record first is created from `TeleconsultRecordScreen`, `TeleconsultRecordPrescribedDrug` will also have one record with empty prescribedDrug list, correct?

msasikanth: Not really. If you take a look at the [`createTeleconsultRecordForMedicalOfficer`](https://github.com/simpledotorg/simple-android/blob/3efa631202d560b632a10f7814a11865dc78409f/app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordRepository.kt#L18). We are only creating the teleconsult record and saving it. We are not creating any prescribed drug here, it will be created and saved in the patient prescriptions screen.

Honey14: Ah, got it.

Honey14 (approved):
Looks great

PR #1691, opened 2020-09-09
Block user from logging a teleconsult record if they are not a medical officer

Honey14:
https://app.clubhouse.io/simpledotorg/story/880/add-a-check-if-the-current-mo-is-allowed-to-log-a-teleconsult-record-at-the-current-facility


app/src/main/java/org/simple/clinic/deeplink/DeepLinkUpdate.kt
```diff
@@ -18,18 +18,19 @@ class DeepLinkUpdate : Update<DeepLinkModel, DeepLinkEvent, DeepLinkEffect> {
       model: DeepLinkModel,
       event: UserFetched
   ): Next<DeepLinkModel, DeepLinkEffect> {
-    return when {
-      event.user?.loggedInStatus == User.LoggedInStatus.LOGGED_IN -> {
-        val effect = if (model.patientUuid != null) {
+    return if (event.user?.loggedInStatus == User.LoggedInStatus.LOGGED_IN) {
+      val effect = if (event.user.capabilities == User.Capabilities(User.CapabilityStatus.Yes)) {
```
vinaysshenoy, line 22: Can we extract these boolean checks into computed properties on the `User` model?

This will make this easier to understand, like `event.user?.isUserLoggedIn` and `event.user?.capabilities?.canRecordTeleconsultLog`.

vinaysshenoy, line 22: This check is non-obvious and will get broken when we need to introduce a new capability. We should be explicitly checking the `canTeleconsult` property like

```kotlin
event.user.capabilities.canTeleconsult == CapabilityStatus.Yes
```

app/src/main/java/org/simple/clinic/deeplink/DeepLinkUpdate.kt
```diff
@@ -18,18 +18,19 @@ class DeepLinkUpdate : Update<DeepLinkModel, DeepLinkEvent, DeepLinkEffect> {
...
-        dispatch(effect)
-      }
-      else -> {
-        dispatch(NavigateToSetupActivity)
+      } else {
+        ShowTeleconsultLogNotAllowed
       }
+      dispatch(effect)
+    } else {
+      dispatch(NavigateToSetupActivity)
```
vinaysshenoy, line 33: CodeClimate is showing issues here with complex methods and nested control flows. We will need to extract some of these methods out into smaller ones.

app/src/main/java/org/simple/clinic/main/TheActivity.kt
```diff
@@ -338,4 +347,13 @@ class TheActivity : AppCompatActivity(), TheActivityUi {
         .setPositiveButton(R.string.deeplink_no_patient_positive_action, null)
         .show()
   }
+
+
+  private fun showTeleconsultNotAllowedErrorDialog() {
+    AlertDialog.Builder(this, R.style.Clinic_V2_DialogStyle)
+        .setTitle("You are not authorised to log teleconsult")
+        .setMessage("Please check with your supervisor")
+        .setPositiveButton("Okay", null)
```
vinaysshenoy, line 356: Can we not use hardcoded strings in the source code? These should be added as string resources.

app/src/test/java/org/simple/clinic/deeplink/DeepLinkUpdateTest.kt
```diff
@@ -30,7 +30,8 @@ class DeepLinkUpdateTest {
   fun `if there is a logged in user and patient uuid is null, then show no patient error`() {
     val user = TestData.loggedInUser(
         uuid = UUID.fromString("dc0a9d11-aee4-4792-820f-c5cb66ae5e47"),
-        loggedInStatus = User.LoggedInStatus.LOGGED_IN
+        loggedInStatus = User.LoggedInStatus.LOGGED_IN,
+        capabilities = User.Capabilities(User.CapabilityStatus.Yes)
     )
```
vinaysshenoy, line 35: What do you think of creating two different users as class properties:

```kotlin
val medicalOfficer = // Create user who can record teleconsult
val nurse = // Create user who cannot record teleconsult
```

We can then use these personas in the individual tests.

app/src/test/java/org/simple/clinic/deeplink/DeepLinkUpdateTest.kt
```diff
@@ -75,6 +77,24 @@ class DeepLinkUpdateTest {
         ))
   }
 
+  @Test
+  fun `if user is logged in and user does not have capabilities, then show user is not authorised to teleconsult log`(){
```
vinaysshenoy, line 81: The name of this test does not reflect the actual test setup, right? It says that the `user does not have capabilities` but in the test setup method we actually are adding capabilities.

I feel we need to add two different tests here even though the outcome is the same (showing the not allowed to log teleconsult dialog).

- One where we do not know if the user is allowed to log a teleconsult or not
- One where we know the user is not allowed to log a teleconsult

In addition, since this is a test for business logic, we should not be using implementation specific phrases like `user does not have capabilities` and should instead use the use-case specific phrases like `is not aware of recording teleconsult logs`.

vinaysshenoy (changes requested):
Overall PR looks good, but it needs some cleanup.

Could you also please reword the PR title since that will be the final commit message when it gets squashed, so we should be sticking to the same format of our commit messages (i.e, `if applied, this commit will <insert PR title here>`.

app/src/test/java/org/simple/clinic/deeplink/DeepLinkUpdateTest.kt
Honey14: > * One where we do not know if the user is allowed to log a teleconsult or not

We'll have capabilities set as `no` by default, so the user is always aware whether he/she is allowed to log teleconsult or not, right? Not sure if this test is required.

vinaysshenoy: This is not that scenario, this is the scenario where we upgrade from a version of the app where there is no capabilities saved for the user in the local database and the refresh user call fails and the capabilities are never fetched and saved.

vinaysshenoy:
The PR message, while better still does not read correctly. You can notice it if you append the PR title.

`If applied, this commit will When the current user is not a Medical Officer, they should not be allowed to log a teleconsult record.`

A better wording would be something like "Block user from logging a teleconsult record if they are not a medical officer".

app/src/main/java/org/simple/clinic/deeplink/DeepLinkUpdate.kt
vinaysshenoy: I don't see this change in branch. Was it not pushed? 🤔

app/src/main/java/org/simple/clinic/deeplink/DeepLinkUpdate.kt
```diff
@@ -18,18 +18,27 @@ class DeepLinkUpdate : Update<DeepLinkModel, DeepLinkEvent, DeepLinkEffect> {
...
+    return if (event.user?.loggedInStatus == User.LoggedInStatus.LOGGED_IN) {
+      checkIfMedicalOfficerHasCapabilitiesToTeleconsult(model, event)
+    } else {
+      dispatch(NavigateToSetupActivity)
+    }
+  }
+
+  private fun checkIfMedicalOfficerHasCapabilitiesToTeleconsult(model: DeepLinkModel, event: UserFetched): Next<DeepLinkModel, DeepLinkEffect> {
+    val effect = if (event.user?.capabilities?.canTeleconsult == User.CapabilityStatus.Yes) {
+      checkIfPatientIsNotNull(model)
```
vinaysshenoy, line 30: The name of the method is based on a check that it is making and not what the primary goal of the method (to fetch the patient details) is. Can we rename the method to reflect that?

app/src/main/java/org/simple/clinic/deeplink/DeepLinkUpdate.kt
```diff
@@ -18,18 +18,27 @@ class DeepLinkUpdate : Update<DeepLinkModel, DeepLinkEvent, DeepLinkEffect> {
...
-        dispatch(NavigateToSetupActivity)
-      }
+    return if (event.user?.loggedInStatus == User.LoggedInStatus.LOGGED_IN) {
+      checkIfMedicalOfficerHasCapabilitiesToTeleconsult(model, event)
+    } else {
+      dispatch(NavigateToSetupActivity)
+    }
+  }
+
+  private fun checkIfMedicalOfficerHasCapabilitiesToTeleconsult(model: DeepLinkModel, event: UserFetched): Next<DeepLinkModel, DeepLinkEffect> {
```
vinaysshenoy, line 28: Same as other comment that the name of this method is exposing a check that it is making and not what the primary goal of the method (to show the log teleconsult screen) is.

vinaysshenoy (changes requested):
Overall PR looks good, left a few comments around naming.

app/src/main/java/org/simple/clinic/deeplink/DeepLinkUpdate.kt
Honey14: Sorry, I think I did the capabilities part and missed this. I will re-check from now as well.

app/src/main/java/org/simple/clinic/deeplink/DeepLinkUpdate.kt
```diff
@@ -18,23 +18,23 @@ class DeepLinkUpdate : Update<DeepLinkModel, DeepLinkEvent, DeepLinkEffect> {
       model: DeepLinkModel,
       event: UserFetched
   ): Next<DeepLinkModel, DeepLinkEffect> {
-    return if (event.user?.loggedInStatus == User.LoggedInStatus.LOGGED_IN) {
-      checkIfMedicalOfficerHasCapabilitiesToTeleconsult(model, event)
+    return if (event.user?.isUserLoggedIn == true) {
```
vinaysshenoy, line 21: Do we need the `== true` since this returns a boolean?

Honey14: It had an error and was asking for not-null assertion, which wasn't a wise choice considering user can also have a different status. WDYT?

vinaysshenoy (approved):

app/src/main/java/org/simple/clinic/deeplink/DeepLinkUpdate.kt
vinaysshenoy: Ah, I see. Let's leave it in for now.

PR #1710, opened 2020-09-10
Stop querying redundant facility for HTTP request headers

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1197/stop-querying-additional-facility-in-headers-for-making-network-calls


Honey14 (approved):
looks good.

PR #1719, opened 2020-09-14
Pass `TeleconsultationType` to `DoneClicked.analyticsName`

msasikanth:
We would like to know which `TeleconsultationType` is selected when creating the teleconsultation record.


Honey14 (approved):
Looks good. :)

PR #1722, opened 2020-09-14
Change `TeleconsultFacilityInfoApi` endpoint

msasikanth:
the [conversation](https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1599211058076600) regarding this change.

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/teleconsultation/sync/TeleconsultFacilityInfoApi.kt
+++ b/app/src/main/java/org/simple/clinic/summary/teleconsultation/sync/TeleconsultFacilityInfoApi.kt
@@ -6,6 +6,6 @@ import retrofit2.http.Headers
 
 interface TeleconsultFacilityInfoApi {
   @Headers(value = ["X-RESYNC-TOKEN: 1"])
-  @GET("v4/teleconsultation_medical_officers/sync")
+  @GET("v4/facility_medical_officers/sync")
   fun pull(): Call<TeleconsultFacilityInfoPullResponse>
 }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/summary/teleconsultation/sync/TeleconsultFacilityInfoApi.kt
```diff
@@ -6,6 +6,6 @@ import retrofit2.http.Headers
 
 interface TeleconsultFacilityInfoApi {
   @Headers(value = ["X-RESYNC-TOKEN: 1"])
-  @GET("v4/teleconsultation_medical_officers/sync")
+  @GET("v4/facility_medical_officers/sync")
```
vinaysshenoy, line 9: Is this available in QA? If yes, we probably should also enable the sync for it since it's currently disabled.

vinaysshenoy (approved):
PR looks good.

app/src/main/java/org/simple/clinic/summary/teleconsultation/sync/TeleconsultFacilityInfoApi.kt
msasikanth: Not sure, let me check with server devs and enable it if it's available.

vinaysshenoy: When you enable it, there's also a sync integration test that we will need to enable again.

PR #1724, opened 2020-09-15
Add screen for updating the patient's prescription #2

msasikanth:
https://app.clubhouse.io/simpledotorg/story/889/add-screen-for-updating-the-patient-s-prescription


app/src/main/java/org/simple/clinic/teleconsultlog/prescription/patientinfo/TeleconsultPatientInfoEffectHandler.kt
```diff
@@ -16,9 +19,18 @@ class TeleconsultPatientInfoEffectHandler @Inject constructor(
...
         .subtypeEffectHandler<TeleconsultPatientInfoEffect, TeleconsultPatientInfoEvent>()
         .addTransformer(LoadPatientProfile::class.java, loadPatientProfile())
+        .addTransformer(LoadPrescriptionDate::class.java, loadPrescriptionDate())
         .build()
   }
 
+  private fun loadPrescriptionDate(): ObservableTransformer<LoadPrescriptionDate, TeleconsultPatientInfoEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .map { LocalDate.now(userClock) }
```
vinaysshenoy, line 29: While this works, I prefer to add the current timestamp in an event when it is generated by the user (if it's a UI event) and then derive the date from it in the effect handler when it needs to be used instead of adding a separate effect for loading the current date.

But this is a good approach if the source trigger for this effect is not a UI event.

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/patientinfo/TeleconsultPatientInfoModel.kt
```diff
@@ -1,22 +1,29 @@
...
 
 import org.simple.clinic.patient.PatientProfile
+import java.time.LocalDate
 import java.util.UUID
 
 data class TeleconsultPatientInfoModel(
     val patientUuid: UUID,
-    val patientProfile: PatientProfile?
+    val patientProfile: PatientProfile?,
+    val prescriptionDate: LocalDate?
```
vinaysshenoy, line 13: I looked into the later commits and saw that this date is only being used to render the current date on the UI. If this is the case, I think we can make the current date part of the factory method of the model and provide it when creating it since it is very unlikely the date will change while the medical officer is on this screen.

This will let us skip needing the effect to load the current date.

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/patientinfo/TeleconsultPatientInfoView.kt
```diff
@@ -0,0 +1,122 @@
...
+  lateinit var dateFormatter: DateTimeFormatter
+
+  @Inject
+  lateinit var effectHandler: TeleconsultPatientInfoEffectHandler
+
+  @Inject
+  lateinit var screenRouter: ScreenRouter
+
+  private val delegate by unsafeLazy {
+    val screenKey = screenRouter.key<TeleconsultPrescriptionScreenKey>(this)
```
vinaysshenoy, line 45: Not necessary to change this, but something to keep in mind here that we are coupling this view to the `TeleconsultPrescriptionScreen`.

vinaysshenoy: I am assuming that this is because this view is going to be embedded within that screen.

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/patientinfo/TeleconsultPatientInfoView.kt
```diff
@@ -0,0 +1,122 @@
...
+  private fun displayPatientAddress(address: PatientAddress) {
+    val addressFields = listOf(
+        address.streetAddress,
+        address.colonyOrVillage,
+        address.district,
+        address.state,
+        address.zone
+    ).filterNot { it.isNullOrBlank() }
+
+    patientAddressTextView.text = addressFields.joinToString()
```
vinaysshenoy, line 112: Is this bit of code that joins the address fields into a comma-separated string repeated in a few places? I feel like I've seen similar pieces of code earlier.

vinaysshenoy: If yes, we might just want to extract it to a method on the `PatientAddress` class.

app/src/main/res/values/strings.xml
```diff
@@ -712,4 +712,7 @@
 
   <!-- Teleconsult Prescription Screen -->
   <string name="screen_teleconsult_prescription_patient_details" translatable="false">%1$s, %2$s, %3$s</string>
-</resources>
+
+  <!-- Teleconsult Patient Info View -->
+  <string name="view_teleconsult_patient_info" translatable="false">%1$s, %2$s, %3$s</string>
```
vinaysshenoy, line 717: Something to think about for the future, but we have a lot of untranslatable strings now, and I'm wondering if we should extract those out into a separate string resources file for better separation at some point.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/patientinfo/TeleconsultPatientInfoView.kt
msasikanth: Yes. Is there a better way of handling these cases?

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/patientinfo/TeleconsultPatientInfoModel.kt
msasikanth: Yes, I was considering both these usages and thought having effect is better for testing. But yeah, it doesn't make much sense to have effect that loads just the date, will switch to the factory method.

app/src/main/res/values/strings.xml
msasikanth: Definitely, it's also worth considering using a single string for commonly used text like this one.

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/patientinfo/TeleconsultPatientInfoView.kt
msasikanth: Yes, will do that.

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/patientinfo/TeleconsultPatientInfoEffectHandler.kt
msasikanth: Removed this after placing the `prescriptionDate` as factory parameter for the model.

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/patientinfo/TeleconsultPatientInfoView.kt
vinaysshenoy: Not at the moment, I think it's find to keep this as is. A more scalable approach might be to have more find grained Dependency scopes which we can use to inject these properties into sub views.

app/src/main/res/values/strings.xml
vinaysshenoy: Can we raise a ticket for this so we can get to it at some point?

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/patientinfo/TeleconsultPatientInfoView.kt
msasikanth: Agreed ✅

app/src/main/res/values/strings.xml
msasikanth: Tracking stories here:
- https://app.clubhouse.io/simpledotorg/story/1236/extract-non-translatable-strings-into-separate-string-resource-file-for-better-separation
- https://app.clubhouse.io/simpledotorg/story/1237/extract-common-strings-into-a-single-string

app/src/main/java/org/simple/clinic/patient/PatientAddress.kt
```diff
@@ -36,6 +36,17 @@ data class PatientAddress(
 
     val deletedAt: Instant?
 ) : Parcelable {
+
+  val addressFieldsString: String
```
vinaysshenoy, line 40: Can we rename this to something like `completeAddress` and drop the `String` suffix since that is already in the type?

app/src/main/java/org/simple/clinic/patient/PatientAddress.kt
```diff
@@ -36,6 +36,17 @@ data class PatientAddress(
...
+    get() = listOf(
+        streetAddress,
+        colonyOrVillage,
+        district,
+        state,
+        zone
+    ).filterNot { it.isNullOrBlank() }
+        .joinToString()
```
vinaysshenoy, line 48: I don't like how this is formatted, but it seems like our current settings format it in this way. Maybe we can look at changing our settings later to format it better.

vinaysshenoy (changes requested):
PR looks good, left a comment about naming.

vinaysshenoy (approved):
LGTM

vinaysshenoy (approved):

PR #1725, opened 2020-09-15
Add methods to purge unnecessary patient records

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1201/prune-the-local-database-at-regular-intervals

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
@@ -124,4 +124,48 @@ class PurgeDatabaseAndroidTest {
     assertThat(phoneNumberDao.phoneNumber(notSyncedPatientProfile.patientUuid).blockingFirst()).containsExactly(deletedButUnsyncedPhoneNumber)
   }
 
+  @Test
+  fun purging_the_database_should_delete_soft_deleted_business_ids() {
+    // given
+    val syncedPatientProfile = TestData.patientProfile(
+        patientUuid = UUID.fromString("57d2ef99-59e7-4dc5-9cc5-4fe6917386b7"),
+        generateBusinessId = false,
+        syncStatus = SyncStatus.DONE
+    )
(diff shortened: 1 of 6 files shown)


app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
```diff
@@ -0,0 +1,83 @@
...
+        generateBusinessId = true,
+        patientDeletedAt = Instant.parse("2018-01-01T00:00:00Z"),
+        syncStatus = SyncStatus.PENDING
+    )
+    patientAddressDao.save(listOf(deletedPatientProfile.address, notDeletedPatientProfile.address, deletedButUnsyncedPatientProfile.address))
+    patientDao.save(listOf(deletedPatientProfile.patient, notDeletedPatientProfile.patient, deletedButUnsyncedPatientProfile.patient))
+    phoneNumberDao.save(deletedPatientProfile.phoneNumbers + notDeletedPatientProfile.phoneNumbers + deletedButUnsyncedPatientProfile.phoneNumbers)
+    businessIdDao.save(deletedPatientProfile.businessIds + notDeletedPatientProfile.businessIds + deletedButUnsyncedPatientProfile.businessIds)
+
+    assertThat(patientDao.patientProfileImmediate(deletedPatientProfile.patientUuid)).isEqualTo(deletedPatientProfile)
```
msasikanth, line 71: Just to know, is there a reason we need to assert if the returned values are the same for a specific ID here? Those are already tested methods, right?

msasikanth (approved):
LGTM

app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
vinaysshenoy: I added these specifically to handle cases where if we manually add a check for deleted records in the SQL query, the tests will always pass even if the hard delete stops working since the query is excluding soft deleted records.

Adding this check here will ensure that the query is reading deleted records and will fail if the query is modified to exclude them.

PR #1744, opened 2020-09-23
Refactor `SignatureActivity`

msasikanth:


msasikanth:
Waiting for #1743 to be merged in, to address some comments.

Honey14 (changes requested):
PR looks good. The App crashes when trying to load the empty bitmap in the `init` function, not sure whether it is accepting empty bitmap. PTAL

msasikanth:
> PR looks good. The App crashes when trying to load the empty bitmap in the `init` function, not sure whether it is accepting empty bitmap. PTAL

Good catch, fixed the issue. The library doesn't handle when setting null bitmaps. So in order to fix it if we load a `null` bitmap we do nothing now.

Honey14 (approved):
Looks good.

PR #1856, opened 2020-09-24
Add screen for updating the patient's prescription - Final

msasikanth:
https://app.clubhouse.io/simpledotorg/story/889/add-screen-for-updating-the-patient-s-prescription

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/drugs/PrescriptionRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/drugs/PrescriptionRepositoryAndroidTest.kt
@@ -12,13 +12,13 @@ import org.simple.clinic.facility.Facility
 import org.simple.clinic.patient.SyncStatus
 import org.simple.clinic.rules.LocalAuthenticationRule
 import org.simple.clinic.storage.Timestamps
+import org.simple.clinic.teleconsultlog.medicinefrequency.MedicineFrequency
 import org.simple.clinic.teleconsultlog.medicinefrequency.MedicineFrequency.BD
 import org.simple.clinic.teleconsultlog.medicinefrequency.MedicineFrequency.OD
 import org.simple.clinic.user.UserSession
 import org.simple.clinic.util.Rules
 import org.simple.clinic.util.TestUtcClock
 import java.time.Duration
-import java.time.Instant
 import java.time.LocalDate
 import java.time.Month
 import java.util.UUID
@@ -137,11 +137,7 @@ class PrescriptionRepositoryAndroidTest {
   fun updating_prescribed_drug_duration_should_work_correctly() {
     // given
     val durationToAdvanceBy = Duration.ofMinutes(10)
-    val timestamps = Timestamps(
-        createdAt = Instant.now(clock),
-        updatedAt = Instant.now(clock),
-        deletedAt = null,
-    )
+    val timestamps = Timestamps.create(clock)
     val prescribedDrug = TestData.prescription(
         uuid = UUID.fromString("c4b74e2b-2ea3-4c4d-a6ab-8e93faa66159"),
         name = "Taco",
@@ -172,11 +168,7 @@ class PrescriptionRepositoryAndroidTest {
   fun updating_prescribed_drug_frequency_should_work_correctly() {
     // given
     val durationToAdvanceBy = Duration.ofMinutes(10)
-    val timestamps = Timestamps(
-        createdAt = Instant.now(clock),
-        updatedAt = Instant.now(clock),
-        deletedAt = null,
-    )
+    val timestamps = Timestamps.create(clock)
     val prescribedDrug = TestData.prescription(
         uuid = UUID.fromString("2f9daf84-50fa-4955-ab18-ea48d8fc9fe1"),
         name = "Taco",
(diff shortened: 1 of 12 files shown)


app/src/androidTest/java/org/simple/clinic/drugs/PrescriptionRepositoryAndroidTest.kt
```diff
@@ -202,4 +203,115 @@ class PrescriptionRepositoryAndroidTest {
...
+  @Test
+  fun updating_prescribed_drug_teleconsultation_id_should_work_correctly() {
+    // given
+    val durationToAdvanceBy = Duration.ofMinutes(10)
+    val teleconsultationId = UUID.fromString("35756e86-4b31-44cb-bd5f-cd3bc8ad2448")
+    val timestamps = Timestamps(
+        createdAt = Instant.now(clock),
+        updatedAt = Instant.now(clock),
+        deletedAt = null,
+    )
```
vinaysshenoy, line 216: Just FYI, we already have a `Timestamps.create(UtcClock)` factory function which does the same thing.

app/src/androidTest/java/org/simple/clinic/drugs/PrescriptionRepositoryAndroidTest.kt
```diff
@@ -202,4 +203,115 @@ class PrescriptionRepositoryAndroidTest {
...
+        timestamps = timestamps,
+        syncStatus = SyncStatus.DONE,
+        teleconsultationId = null
+    )
+
+    database.prescriptionDao().save(listOf(prescribedDrug))
+
+    // when
+    clock.advanceBy(durationToAdvanceBy)
+    repository.updateTeleconsultationId(
```
vinaysshenoy, line 230: Why do we need this method which accepts a single drug UUID if we already one that accepts a list?

app/src/androidTest/java/org/simple/clinic/drugs/PrescriptionRepositoryAndroidTest.kt
```diff
@@ -202,4 +203,115 @@ class PrescriptionRepositoryAndroidTest {
...
+            timestamps = prescribedDrug1.timestamps.copy(updatedAt = prescribedDrug1.updatedAt.plus(durationToAdvanceBy)),
+            syncStatus = SyncStatus.PENDING
+        ),
+        prescribedDrug2.copy(
+            teleconsultationId = teleconsultationId,
+            timestamps = prescribedDrug2.timestamps.copy(updatedAt = prescribedDrug2.updatedAt.plus(durationToAdvanceBy)),
+            syncStatus = SyncStatus.PENDING
+        )
+    )
+    assertThat(repository.newestPrescriptionsForPatientImmediate(patient1Uuid)).isEqualTo(expectedPrescribedDrugs)
```
vinaysshenoy, line 315: Can we also assert that `prescrubedDrug3` is unaffected?

app/src/main/java/org/simple/clinic/drugs/PrescriptionRepository.kt
```diff
@@ -157,4 +157,22 @@ class PrescriptionRepository @Inject constructor(
...
+  }
+
+  fun updateTeleconsultationIds(drugUuids: List<UUID>, teleconsultationId: UUID) {
+    dao.updateTeleconsultationIds(
+        drugUuids = drugUuids,
+        teleconsultationId = teleconsultationId,
+        updatedAt = Instant.now(utcClock),
+        syncStatus = SyncStatus.PENDING
+    )
+  }
```
vinaysshenoy, line 177: Can we rename this to better reflect what it is doing? Ex: `addTeleconsultationIdToDrugs`.

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionEffectHandler.kt
```diff
@@ -27,9 +37,30 @@ class TeleconsultPrescriptionEffectHandler @AssistedInject constructor(
...
         .addTransformer(LoadDataForNextClick::class.java, loadDataForNextClick())
+        .addTransformer(CreatePrescription::class.java, createPrescription())
         .build()
   }
 
+  private fun createPrescription(): ObservableTransformer<CreatePrescription, TeleconsultPrescriptionEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .doOnNext { medicalRegistrationIdPreference.set(Optional.of(it.medicalRegistrationId)) }
```
vinaysshenoy, line 51: We're mixing creating a prescription and updating the current medical registration ID in the same effect, which should not be done. We can send out multiple effects in a single `dispatch` or `next`, which we should be using here to save the medical registration ID locally.

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionEffectHandler.kt
```diff
@@ -27,9 +37,30 @@ class TeleconsultPrescriptionEffectHandler @AssistedInject constructor(
...
+          .doOnNext { updatePrescribedDrugTeleconsultId(it.patientUuid, it.teleconsultRecordId) }
+          .map { PrescriptionCreated(it.medicalInstructions) }
+    }
+  }
+
+  private fun updatePrescribedDrugTeleconsultId(patientUuid: UUID, teleconsultRecordId: UUID) {
+    val prescribedDrugs = prescriptionRepository.newestPrescriptionsForPatientImmediate(patientUuid)
+
+    prescriptionRepository.updateTeleconsultationIds(
+        drugUuids = prescribedDrugs.map { it.uuid },
```
vinaysshenoy, line 62: We could move this `map {}` to the place where we are reading the prescriptions from the repository since we don't use the prescribed drugs anywhere else.

app/src/test/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionEffectHandlerTest.kt
```diff
@@ -100,4 +113,45 @@ class TeleconsultPrescriptionEffectHandlerTest {
...
+    val medicalRegistrationId = "ABC123456"
+    val instructions = "This is a medical instruction"
+
+    val prescribedDrug1 = TestData.prescription(
+        uuid = UUID.fromString("10eb023a-cd5d-4140-9a7e-0e08ac26f3d7"),
+        name = "Taco",
+        patientUuid = patientUuid,
+        timestamps = Timestamps(
+            createdAt = Instant.now(),
+            updatedAt = Instant.now(),
```
vinaysshenoy, line 131: Can we use a clock here to fix the time in the tests?

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionEvent.kt
```diff
@@ -21,3 +21,7 @@ data class DataForNextClickLoaded(
 data class NextButtonClicked(val medicalInstructions: String, val medicalRegistrationId: String) : TeleconsultPrescriptionEvent() {
   override val analyticsName: String = "Teleconsult Prescription:Next Clicked"
 }
+
+data class PrescriptionCreated(
+    val medicalInstructions: String
```
vinaysshenoy, line 26: What do you think of tracking this in `TeleconsultPrescriptionModel` so we don't need to forward this via an event?

app/src/test/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionUpdateTest.kt
```diff
@@ -93,4 +95,28 @@ class TeleconsultPrescriptionUpdateTest {
             ))
         ))
   }
+
+  @Test
+  fun `when data for next click is loaded, then create prescription`() {
+    val medicalInstructions = "This is a medical instructions"
+    val medicalRegistrationId = "ABC12345"
+    val bitmap = mock<Bitmap>()
```
vinaysshenoy, line 103: Something to think about later, but we can remove the need for mocking a platform type by changing the event to have a boolean property (`hasSavedSignature`) which can be set in the effect handler. This will let us have the added benefit of making `.update()` pure again.

vinaysshenoy (changes requested):

app/src/test/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionUpdateTest.kt
msasikanth: Yeah, I initially wanted to use that way. But then it seemed like it was a logical decision that is being made in effect handler. But yeah, I can make this change in this PR.

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionEvent.kt
msasikanth: That would be the only use case for it, like we won't be using it for anything else. But yeah, I can move it to model.

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionEffectHandler.kt
msasikanth: Yep, definitely. I was just afraid if it was going to be some hidden behavior

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionEffectHandler.kt
msasikanth: Okay yeah, since these are side effects and doesn't depend on each other they can be different effects right?

app/src/androidTest/java/org/simple/clinic/drugs/PrescriptionRepositoryAndroidTest.kt
msasikanth: Initially I was trying out a different approach, but left it as a way to just update single teleconsultation id. I can remove it since there is no use case for it currently.

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionEffectHandler.kt
vinaysshenoy: Yes

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionEvent.kt
vinaysshenoy: No need to do it in this PR, we can do it later if needed.

app/src/test/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionUpdateTest.kt
vinaysshenoy: The logical decision here is _what_ to do if there is no saved signature, which is still being decided in the `update()` function. What we are changing is that the effect handler will tell the `update()` function whether there is a saved signature or not.

vinaysshenoy (approved):

PR #1858, opened 2020-09-24
Add teleconsult record sync resource

Honey14:
https://app.clubhouse.io/simpledotorg/story/907/add-support-for-teleconsult-record-sync-resource

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordRepositoryAndroidTest.kt
@@ -8,10 +8,10 @@ import org.simple.clinic.TestClinicApp
 import org.simple.clinic.TestData
 import org.simple.clinic.storage.Timestamps
 import org.simple.clinic.teleconsultlog.teleconsultrecord.Answer.Yes
-import org.simple.clinic.util.TestUserClock
-import org.simple.clinic.util.toUtcInstant
+import org.simple.clinic.util.TestUtcClock
 import java.time.Instant
 import java.time.LocalDate
+import java.time.Month
 import java.util.UUID
 import javax.inject.Inject
 
@@ -20,11 +20,13 @@ class TeleconsultRecordRepositoryAndroidTest {
   @Inject
   lateinit var repository: TeleconsultRecordRepository
 
-  private val userClock = TestUserClock()
+  @Inject
+  lateinit var testUtcClock: TestUtcClock
 
   @Before
   fun setup() {
     TestClinicApp.appComponent().inject(this)
+    testUtcClock.setDate(LocalDate.of(2020, Month.SEPTEMBER, 15))
   }
 
   @After
@@ -40,11 +42,7 @@ class TeleconsultRecordRepositoryAndroidTest {
     val teleconsultRecordId3 = UUID.fromString("f1ad859f-d076-4bae-b0f2-3fb23c60d880")
     val patientUuid = UUID.fromString("91f230a6-e67f-428e-95f8-6090415a5c4e")
     val medicalOfficerId = UUID.fromString("5e2a46b8-6d77-47fa-9538-281e4992cb46")
-    val date = LocalDate.parse("2020-09-15")
-    val createdAt = date.toUtcInstant(userClock)
-    val updatedAt = date.toUtcInstant(userClock)
-    val deletedAt = null
-    val timestamps = Timestamps(createdAt, updatedAt, deletedAt)
+    val timestamps = Timestamps.create(testUtcClock)
 
     val teleconsultRecord1 = TestData.teleconsultRecord(
         id = teleconsultRecordId1,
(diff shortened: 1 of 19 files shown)


app/src/main/java/org/simple/clinic/storage/migrations/Migration_78.kt
```diff
@@ -0,0 +1,18 @@
...
+import androidx.sqlite.db.SupportSQLiteDatabase
+import org.simple.clinic.storage.inTransaction
+import javax.inject.Inject
+
+@Suppress("Classname")
+class Migration_78 @Inject constructor() : Migration(77, 78) {
+  override fun migrate(database: SupportSQLiteDatabase) {
+    database.inTransaction {
+      execSQL("""
+        ALTER TABLE "TeleconsultRecord" ADD COLUMN "syncStatus" TEXT NOT NULL
```
vinaysshenoy, line 13: Do we need to set a default value for existing records? This will set the value to existing records to `null` which will cause a crash if we try to read it.

vinaysshenoy (changes requested):

app/src/androidTest/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecordRepositoryAndroidTest.kt
```diff
@@ -0,0 +1,101 @@
...
+
+  @Test
+  fun saving_teleconsult_records_with_prescribed_drugs_should_work_properly() {
+    // given
+    val teleconsultRecordId1 = UUID.fromString("7631538a-7510-4147-b239-1c56c7d2ef70")
+    val teleconsultRecordId2 = UUID.fromString("9ee0bc81-26e6-4ced-9b33-fa76a0a995e3")
+    val teleconsultRecordId3 = UUID.fromString("f1ad859f-d076-4bae-b0f2-3fb23c60d880")
+    val patientUuid = UUID.fromString("91f230a6-e67f-428e-95f8-6090415a5c4e")
+    val medicalOfficerId = UUID.fromString("5e2a46b8-6d77-47fa-9538-281e4992cb46")
+    val date = LocalDate.parse("2020-09-15")
```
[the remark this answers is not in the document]
Honey14: `Timestamps.create()` asked for `TestUtcClock` clock, using that now

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecord.kt
```diff
@@ -34,6 +34,19 @@ data class TeleconsultRecord(
     val syncStatus: SyncStatus
 ) {
 
+  fun toPayload(): TeleconsultRecordPayload {
```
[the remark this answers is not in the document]
Honey14: How can we get the `teleconsultRecord` object in the `TeleconsultRecordSync` class?
Ps. We've consistently converted the object into network representation class in the domain class across the project but you have a fair point as well.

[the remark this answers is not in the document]
vinaysshenoy: We used to do that initially but we decided to move this conversion to the sync classes later on. I don't think we ended up moving all the existing resources, sadly.

> How can we get the teleconsultRecord object in the TeleconsultRecordSync class?

You already have the instance since `TeleconsultRecordSync` calls `.toPayload()` on it, right? What I mean is that we should instead have a `recordToPayload(TeleconsultRecordSync): TeleconsultRecordPayload` method in `TeleconsultRecordSync` itself.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/TeleconsultRecord.kt
```diff
@@ -27,9 +29,53 @@ data class TeleconsultRecord(
...
+  }
+
+  private fun getTeleconsultRecordInfoPayload(): TeleconsultRecordInfoPayload? {
+    return if (teleconsultRecordInfo != null) {
+      TeleconsultRecordInfoPayload(
+          recordedAt = teleconsultRecordInfo.recordedAt,
+          teleconsultationType = teleconsultRecordInfo.teleconsultationType,
+          patientTookMedicines = teleconsultRecordInfo.patientTookMedicines,
+          patientConsented = teleconsultRecordInfo.patientConsented,
+          medicalOfficerNumber = teleconsultRecordInfo.medicalOfficerNumber.toString(),
```
vinaysshenoy, line 60: I just realised that this `.toString()` call on a nullable string will return the string `"null"` if it is `null`, which is incorrect behaviour and we should be sending a blank string instead. The `.orEmpty()` extension method is the right one to use here.

app/src/main/java/org/simple/clinic/storage/migrations/Migration_78.kt
```diff
@@ -0,0 +1,18 @@
...
+import androidx.sqlite.db.SupportSQLiteDatabase
+import org.simple.clinic.storage.inTransaction
+import javax.inject.Inject
+
+@Suppress("Classname")
+class Migration_78 @Inject constructor() : Migration(77, 78) {
+  override fun migrate(database: SupportSQLiteDatabase) {
+    database.inTransaction {
+      execSQL("""
+        ALTER TABLE "TeleconsultRecord" ADD COLUMN "syncStatus" TEXT
```
vinaysshenoy, line 13: If you try to run the migration test locally, it will fail since the database schema is now different. The version in the `78.json` schema file has the column as `NOT NULL` but the migration does not add that qualifier.

We should be setting the default value as `PENDING` to existing records as part of the migration.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/storage/migrations/Migration_78.kt
Honey14: Updated to include `PENDING` as a default not null value in the migration. PTAL

vinaysshenoy (approved):
LGTM!

PR #1865, opened 2020-09-29
Remove foreign keys on facility table from User table

vinaysshenoy:
`FacilitySync` currently uses 'Insert or Replace' when we save facilities into the local database during sync. However, the implementation of this in SQLite ends up triggering a DELETE on conflict and then an insertion.

A consequence of this is that any other table that has a foreign key reference to the facility table will end up getting deleted if a row in the facility table is replaced that has a reference to this row.

This commit removes the foreign key reference on the facility table from the user table and we will rely on the flow in the application logic to ensure that facilities are present.


msasikanth (approved):
LGTM

PR #1866, opened 2020-09-29
Update QUIRKS

Honey14:
Starting android 8.0 and above we don't need to use `threeTenBp` for working with date and time, we are instead using the built-in `java-time`. Hence removing the explicit data mentioning `threeTenBp`


vinaysshenoy (approved):
Looks good! Thanks for doing this!

PR #1871, opened 2020-09-29
Check the current screen before updating the app lock time

vinaysshenoy:
We introduced a bug in a previous commit where the app lock screen could be bypassed by hitting back on the app lock screen and then opening it again before the app lock timer runs out.

This was happening because `TheActivity` would update the app lock timer in `onStop` regardless of whether the app lock was successfully completed or not and the lock at time would be advanced to some point in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,8 @@
 - Add teleconsult record sync
 
 ### Fixes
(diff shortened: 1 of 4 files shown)


msasikanth (approved):
LGTM

PR #1873, opened 2020-09-30
Save teleconsult record id to prescribed drugs when creating TeleconsultRecord

msasikanth:
https://app.clubhouse.io/simpledotorg/story/1257/save-teleconsult-record-id-to-prescribed-drugs-when-creating-teleconsultrecord

### DIFF
--- a/app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
@@ -9,8 +9,6 @@ import io.reactivex.ObservableTransformer
 import org.simple.clinic.drugs.PrescriptionRepository
 import org.simple.clinic.facility.Facility
 import org.simple.clinic.patient.PatientRepository
-import org.simple.clinic.patient.SyncStatus
-import org.simple.clinic.storage.Timestamps
 import org.simple.clinic.teleconsultlog.teleconsultrecord.TeleconsultRecordInfo
 import org.simple.clinic.teleconsultlog.teleconsultrecord.TeleconsultRecordRepository
 import org.simple.clinic.user.User
@@ -57,16 +55,12 @@ class TeleconsultRecordEffectHandler @AssistedInject constructor(
           .filter { (prescriptions, _) -> prescriptions.isNotEmpty() }
           .doOnNext { (prescriptions, _) -> prescriptionRepository.softDeletePrescriptions(prescriptions) }
           .flatMap { (prescriptions, effect) ->
-            val clonedPrescriptions = prescriptions
-                .map {
-                  it.copy(
+            val clonedPrescriptions = prescriptions.map { prescribedDrug ->
+                  prescribedDrug.refillForTeleconsultation(
                       uuid = uuidGenerator.v4(),
                       facilityUuid = currentFacility.get().uuid,
-                      syncStatus = SyncStatus.PENDING,
-                      timestamps = Timestamps.create(utcClock),
-                      frequency = null,
-                      durationInDays = null,
-                      teleconsultationId = effect.teleconsultRecordId
+                      teleconsultationId = effect.teleconsultRecordId,
+                      utcClock = utcClock
                   )
                 }
 
--- a/app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
@@ -6,6 +6,7 @@ import com.squareup.inject.assisted.AssistedInject
 import dagger.Lazy
 import io.reactivex.Observable
 import io.reactivex.ObservableTransformer
+import org.simple.clinic.drugs.PrescribedDrug
 import org.simple.clinic.drugs.PrescriptionRepository
 import org.simple.clinic.facility.Facility
 import org.simple.clinic.patient.PatientRepository
@@ -52,21 +53,28 @@ class TeleconsultRecordEffectHandler @AssistedInject constructor(
       effects
           .observeOn(schedulersProvider.io())
           .map { effect -> prescriptionRepository.newestPrescriptionsForPatientImmediate(effect.patientUuid) to effect }
-          .filter { (prescriptions, _) -> prescriptions.isNotEmpty() }
-          .doOnNext { (prescriptions, _) -> prescriptionRepository.softDeletePrescriptions(prescriptions) }
-          .flatMap { (prescriptions, effect) ->
-            val clonedPrescriptions = prescriptions.map { prescribedDrug ->
-                  prescribedDrug.refillForTeleconsultation(
-                      uuid = uuidGenerator.v4(),
-                      facilityUuid = currentFacility.get().uuid,
-                      teleconsultationId = effect.teleconsultRecordId,
-                      utcClock = utcClock
-                  )
-                }
+          .doOnNext { (prescriptions, effect) -> clonePrescriptions(prescriptions, effect) }
+          .map { PatientPrescriptionsCloned }
+    }
+  }
 
-            prescriptionRepository.save(clonedPrescriptions)
-                .andThen(Observable.just(PatientPrescriptionsCloned))
-          }
+  private fun clonePrescriptions(
+      prescriptions: List<PrescribedDrug>,
+      effect: ClonePatientPrescriptions
+  ) {
+    if (prescriptions.isNotEmpty()) {
+      prescriptionRepository.softDeletePrescriptions(prescriptions)
+
+      val clonedPrescriptions = prescriptions.map { prescribedDrug ->
+        prescribedDrug.refillForTeleconsultation(
+            uuid = uuidGenerator.v4(),
+            facilityUuid = currentFacility.get().uuid,
+            teleconsultationId = effect.teleconsultRecordId,
+            utcClock = utcClock
+        )
+      }
+
+      prescriptionRepository.saveImmediate(clonedPrescriptions)
     }
   }
 
--- a/app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
@@ -4,21 +4,28 @@ import com.spotify.mobius.rx2.RxMobius
 import com.squareup.inject.assisted.Assisted
 import com.squareup.inject.assisted.AssistedInject
 import dagger.Lazy
+import io.reactivex.Observable
 import io.reactivex.ObservableTransformer
+import org.simple.clinic.drugs.PrescriptionRepository
 import org.simple.clinic.patient.PatientRepository
+import org.simple.clinic.patient.SyncStatus
+import org.simple.clinic.storage.Timestamps
 import org.simple.clinic.teleconsultlog.teleconsultrecord.TeleconsultRecordInfo
 import org.simple.clinic.teleconsultlog.teleconsultrecord.TeleconsultRecordRepository
 import org.simple.clinic.user.User
 import org.simple.clinic.util.UtcClock
 import org.simple.clinic.util.scheduler.SchedulersProvider
+import org.simple.clinic.uuid.UuidGenerator
 import java.time.Instant
 
 class TeleconsultRecordEffectHandler @AssistedInject constructor(
     private val user: Lazy<User>,
     private val teleconsultRecordRepository: TeleconsultRecordRepository,
     private val patientRepository: PatientRepository,
+    private val prescriptionRepository: PrescriptionRepository,
     private val schedulersProvider: SchedulersProvider,
     private val utcClock: UtcClock,
+    private val uuidGenerator: UuidGenerator,
     @Assisted private val uiActions: UiActions
 ) {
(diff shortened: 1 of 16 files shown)


app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordModel.kt
```diff
@@ -48,4 +49,8 @@ data class TeleconsultRecordModel(
   fun patientLoaded(patient: Patient): TeleconsultRecordModel {
     return copy(patient = patient)
   }
+
+  fun patientPrescriptionsCloned(): TeleconsultRecordModel {
```
vinaysshenoy, line 54: How about naming this `teleconsultRecordCreated()` instead?

vinaysshenoy: Reason: It's the `update` function that knows that cloning the prescriptions means that the teleconsult record has been created. The model here does not know anything about the prescriptions at all.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordUiRenderer.kt
```diff
@@ -11,6 +12,15 @@ class TeleconsultRecordUiRenderer(
     ui.setTeleconsultationType(model.teleconsultationType)
     ui.setPatientTookMedicines(model.patientTookMedicines)
     ui.setPatientConsented(model.patientConsented)
+    renderDoneProgress(model.teleconsultRecordCreateState)
+  }
+
+  private fun renderDoneProgress(teleconsultRecordCreateState: TeleconsultRecordCreateState?) {
+    if (teleconsultRecordCreateState == CREATING_TELECONSULT_RECORD) {
```
vinaysshenoy, line 19: Can we extract this logic to a computed property on the model, say `isCreatingTeleconsultRecord`?

app/src/test/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordUiRendererTest.kt
```diff
@@ -63,6 +64,51 @@ class TeleconsultRecordUiRendererTest {
     verify(ui).setPatientTookMedicines(Yes)
     verify(ui).setPatientConsented(Yes)
     verify(ui).renderPatientDetails(patient)
+    verify(ui).hideDoneProgress()
+    verifyNoMoreInteractions(ui)
+  }
+
+  @Test
+  fun `when teleconsult record create state is creating, then show done progress`() {
```
vinaysshenoy, line 72: We can rename the situation for this test to `when the teleconsult record is being created`.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordUi.kt
```diff
@@ -9,4 +9,6 @@ interface TeleconsultRecordUi {
   fun setPatientTookMedicines(patientTookMedicines: Answer)
   fun setPatientConsented(patientConsented: Answer)
   fun renderPatientDetails(patient: Patient)
+  fun showDoneProgress()
+  fun hideDoneProgress()
```
vinaysshenoy, line 13: Can we drop the `Done` part of the method names? I don't think they add any additional meaning and are just noise at this point.

app/src/test/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordUiRendererTest.kt
```diff
@@ -63,6 +64,51 @@ class TeleconsultRecordUiRendererTest {
...
+    // then
+    verify(ui).setTeleconsultationType(Audio)
+    verify(ui).setPatientTookMedicines(Yes)
+    verify(ui).setPatientConsented(Yes)
+    verify(ui).showDoneProgress()
+    verifyNoMoreInteractions(ui)
+  }
+
+  @Test
+  fun `when cloning prescriptions is completed, then hide progress`() {
```
vinaysshenoy, line 94: It is an implementation detail that cloning the prescriptions completes the recording of a teleconsult. We can rename this to `when the teleconsult record is created, then hide the progress`.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordUi.kt
msasikanth: That was intentional, `showProgress` & `hideProgress` sounded generic for the screen. That's why I added where we will be showing the progress. Should I continue renaming it? (or perhaps I can rename it to `showDoneButtonProgress` & `hideDoneButtonProgress`

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordModel.kt
msasikanth: Good point, renaming it 👍🏽

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
```diff
@@ -36,9 +43,35 @@ class TeleconsultRecordEffectHandler @AssistedInject constructor(
...
+          .flatMap { (prescriptions, effect) ->
+            val clonedPrescriptions = prescriptions
+                .map {
+                  it.copy(
+                      uuid = uuidGenerator.v4(),
+                      syncStatus = SyncStatus.PENDING,
+                      timestamps = Timestamps.create(utcClock),
+                      frequency = null,
+                      durationInDays = null,
+                      teleconsultationId = effect.teleconsultRecordId
```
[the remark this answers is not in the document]
msasikanth: we should? 😅 . I was not aware of that, making the change.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
```diff
@@ -36,9 +43,35 @@ class TeleconsultRecordEffectHandler @AssistedInject constructor(
...
+        .addTransformer(ClonePatientPrescriptions::class.java, clonePatientPrescriptions())
         .build()
   }
 
+  private fun clonePatientPrescriptions(): ObservableTransformer<ClonePatientPrescriptions, TeleconsultRecordEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .map { effect -> prescriptionRepository.newestPrescriptionsForPatientImmediate(effect.patientUuid) to effect }
+          .doOnNext { (prescriptions, _) -> prescriptionRepository.softDeletePrescriptions(prescriptions) }
```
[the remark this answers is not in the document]
msasikanth: Yep, filtering prescriptions that are empty before soft deleting or cloning them. Is that fine?

```kt
filter { (prescriptions, _) -> prescriptions.isNotEmpty() }
```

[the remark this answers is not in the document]
vinaysshenoy: This should work, one concern is whether we are dependent on the event emitted from this effect handler to proceed with business logic.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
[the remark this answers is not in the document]
vinaysshenoy: When we say clone a prescription, what we _mean_ is that it should as if the current user has manually prescribed the same drugs again. This means updating these fields which would be considered "meta" information, but cloning the others which are the actual prescription itself.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordUi.kt
vinaysshenoy: The way I am looking at this is that it is up to the screen whether to show this progress state on the done button or not and the UI renderer itself doesn't really care where it is shown, just that it is shown and hidden in some way.

vinaysshenoy: We shouldn't have to change these names if we decide to say, show a progress modal instead of doing it as part of the button.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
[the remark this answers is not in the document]
msasikanth: I don't understand the concern 😅

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordUi.kt
msasikanth: Ah okay, makes sense. Renaming the methods

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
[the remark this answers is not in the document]
vinaysshenoy: Are we dependent on the `PrescriptionCloned` event that is emitted from this effect to do some work in the `update` function?

If yes, then this filter will not work because the event will never be emitted which will stop processing of important business logic.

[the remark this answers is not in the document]
msasikanth: Ah okay, yeah we need to navigate to teleconsult success screen after cloning is completed. Is there a way to send a event after the filter or is there other way I can achieve this?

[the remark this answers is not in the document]
vinaysshenoy: We can still send the `PrescriptionCloned` event, just not call any of the methods if there are no prescriptions available.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
```diff
@@ -52,20 +52,23 @@ class TeleconsultRecordEffectHandler @AssistedInject constructor(
...
+                prescribedDrug.refillForTeleconsultation(
+                    uuid = uuidGenerator.v4(),
+                    facilityUuid = currentFacility.get().uuid,
+                    teleconsultationId = effect.teleconsultRecordId,
+                    utcClock = utcClock
+                )
+              }
+
+              prescriptionRepository.saveImmediate(clonedPrescriptions)
+            }
```
vinaysshenoy: This can be moved to a `.doOnNext {}` block, right?

vinaysshenoy: The Rx chain then becomes a little simpler since we can avoid the `.flatMap {}`.

```kotlin
.doOnNext { /* Delete old prescriptions, clone and save new ones */ }
.map { PatientPrescriptionsCloned } 
```

vinaysshenoy (changes requested):
PR looks good, left a comment. In addition, there is one CodeClimate issue that we'll need to fix and then it'll be ready to go.

app/src/main/java/org/simple/clinic/teleconsultlog/teleconsultrecord/screen/TeleconsultRecordEffectHandler.kt
msasikanth: Don't we have to wait until cloning is completed before moving to next screen? 🤔

vinaysshenoy: `.doOnNext` will wait.

vinaysshenoy: Operators wait unless you explicitly change the scheduler using a `.flatMap {}`.

msasikanth: Ah okay, I think I got confused with the terminology. I thought being side effect meant it will not wait for the operations to be completed, but actually it will wait but doesn't send the results to downstream right?

vinaysshenoy: Yup.

vinaysshenoy (approved):

PR #1875, opened 2020-09-30
Show an error message if no medicines are added on the Rx screen.

msasikanth:
https://app.clubhouse.io/simpledotorg/story/1429/show-an-error-message-if-no-medicines-are-added-on-the-rx-screen


app/src/test/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionUpdateTest.kt
```diff
@@ -52,14 +52,34 @@ class TeleconsultPrescriptionUpdateTest {
...
   }
 
+  @Test
+  fun `when there are no medicines added, then show medicines required error`() {
+    val medicalInstructions = "This is a medical instructions"
+    val medicalRegistrationId = "ABC12345"
+
+    updateSpec
+        .given(model)
+        .whenEvent(DataForNextClickLoaded(
```
Honey14, line 71: Can we add `DataForNextClickLoaded` event as a local property for better readability?

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionUpdate.kt
```diff
@@ -31,21 +32,27 @@ class TeleconsultPrescriptionUpdate : Update<TeleconsultPrescriptionModel, Telec
...
-              medicalInstructions = event.medicalInstructions
-          )
-      )
-    } else {
-      dispatch(ShowSignatureRequiredError)
+    return when {
+      event.hasSignatureBitmap.not() -> {
+        dispatch(ShowSignatureRequiredError)
+      }
+      event.hasMedicines.not() -> {
```
Honey14, line 39: Error scenario : 
Step 1: There's a possibility where the medical officer has not added both Signature and Medicines. In this case, when I click on `Next` button, only signature error is enabled.
Step 2: I add the signature then click on `Next`, we get the medicine error. 

Expected behaviour: On click of `Next`, we should be getting both the errors together if empty.

Honey14 (changes requested):
Looks great. Left few comments. PTAL

app/src/test/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionUpdateTest.kt
msasikanth: we can, but we are not reusing it anywhere and IMO the implementation is not large enough to extract to an variable 🤔

app/src/main/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionUpdate.kt
msasikanth: Ah good catch, will add this.

app/src/test/java/org/simple/clinic/teleconsultlog/prescription/TeleconsultPrescriptionUpdateTest.kt
Honey14: Sure, no problem. Had readability in mind., but this works as well.

Honey14 (approved):
Looks good

PR #1876, opened 2020-09-30
Fix ViewPager2 nested elements scrolling

msasikanth:


app/src/main/java/org/simple/clinic/util/NestedScrollableHost.kt
```diff
@@ -0,0 +1,74 @@
...
+import android.widget.FrameLayout
+import kotlin.math.absoluteValue
+
+/**
+ * Layout to wrap a scrollable component inside a ViewPager2. Provided as a solution to the problem
+ * where pages of ViewPager2 have nested scrollable elements that scroll in the opposite direction as
+ * ViewPager2 like RecyclerView, WebView, etc. The scrollable element needs to be the immediate and
+ * only child of this host layout.
+ *
+ * Based on: https://github.com/android/views-widgets-samples/blob/master/ViewPager2/app/src/main/java/androidx/viewpager2/integration/testapp/NestedScrollableHost.kt
```
vinaysshenoy, line 32: Instead of using `master` as the branch in the link, can we instead link to a specific commit hash in the tree. Otherwise the main repo could end up removing this link in `master` and this link would lead to nowhere.

vinaysshenoy (changes requested):

Can we also update the changelog to say that this is something that we fixed?

app/src/main/java/org/simple/clinic/util/NestedScrollableHost.kt
msasikanth: Good point, fixed it.

vinaysshenoy (approved):

PR #1877, opened 2020-10-01
Switch the positions of the teleconsultation and save buttons in `PatientSummaryScreen`

Honey14:
https://app.clubhouse.io/simpledotorg/story/883/switch-the-positions-of-the-teleconsultation-and-save-buttons-on-the-patient-summary-screen


app/src/main/res/layout/screen_patient_summary.xml
```diff
@@ -203,32 +203,32 @@
...
       android:layout_weight="1"
-      android:text="@string/patientsummary_finish_saving_patient"
+      android:text="@string/patientsummary_contact_doctor"
+      android:textAllCaps="true"
       android:textAppearance="@style/Clinic.V2.TextAppearance.Button1"
-      app:icon="@drawable/ic_done_24dp"
-      app:iconGravity="textStart" />
+      android:visibility="gone"
+      app:backgroundTint="@color/contact_doctor_button"
+      app:iconTint="@android:color/white" />
```
msasikanth, line 218: Since we are no longer using icon for this button, we can remove the icon tint attr.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -222,7 +222,7 @@ class PatientSummaryScreen(
```
msasikanth, line 222: Can we rename the commit message to something like this

"Change tele-consult `ProgressMaterialButton` to `MaterialButton`"

in commit message you can mention about renaming `doctorButton` to `teleconsultButton`

msasikanth (changes requested):
PR looks good overall. Left couple of comments PTAL

app/src/main/res/layout/screen_patient_summary.xml
```diff
@@ -203,32 +203,32 @@
     android:paddingTop="@dimen/spacing_8"
     android:paddingBottom="@dimen/spacing_8">
 
-    <com.google.android.material.button.MaterialButton
-      android:id="@+id/doneButton"
-      style="?attr/materialButtonStyle"
+    <org.simple.clinic.widgets.ProgressMaterialButton
```
msasikanth: I don't think we need to switch to `ProgressMaterialButton` here since, we need it to be `MaterialButton`

app/src/main/res/layout/screen_patient_summary.xml
```diff
@@ -203,32 +203,32 @@
...
+      android:id="@+id/doctorButton"
       android:layout_width="0dp"
       android:layout_height="wrap_content"
       android:layout_marginStart="@dimen/spacing_8"
       android:layout_marginEnd="@dimen/spacing_8"
       android:layout_weight="1"
-      android:text="@string/patientsummary_finish_saving_patient"
+      android:text="@string/patientsummary_contact_doctor"
+      android:textAllCaps="true"
```
msasikanth, line 214: I think I missed this during initial review, we don't have to use `textAllCaps` attr. Using the default material button style will automatically do this for us `style="?attr/materialButtonStyle" `

app/src/main/res/layout/screen_patient_summary.xml
```diff
@@ -214,8 +214,7 @@
       android:textAllCaps="true"
       android:textAppearance="@style/Clinic.V2.TextAppearance.Button1"
       android:visibility="gone"
-      app:backgroundTint="@color/contact_doctor_button"
-      app:iconTint="@android:color/white" />
+      app:backgroundTint="@color/contact_doctor_button" />
```
msasikanth, line 217: We can stop using this background now, instead we can directly use the `green1` color since we won't have any disable button state now.

msasikanth (changes requested):
I think I missed couple of things during initial review, PTAL

app/src/main/res/layout/screen_patient_summary.xml
Honey14: This has been changed in the later commit, PTAL

msasikanth: Yeah but is that even necessary? 🤔 . We are adding `ProgressMaterialButton` in this commit and removing it in the next one.

Honey14: No It's not. I'll edit that commit

app/src/main/res/values/strings.xml
```diff
@@ -214,7 +214,7 @@
   <string name="patientsummary_toolbar_title">%1$s, %2$s, %3$s</string>
   <string name="patientsummary_bangladesh_national_id">ID</string>
 
-  <string name="patientsummary_contact_doctor">Doctor</string>
+  <string name="patientsummary_contact_doctor" translatable="false">Tele-consult</string>
```
msasikanth: There is no point settings this string as `translatable="false"`, we can create a new string called `patientsummary_teleconsult` and remove this string in `Remove unused resources` commit.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -222,7 +222,7 @@ class PatientSummaryScreen(
 
   private fun doneClicks() = doneButton.clicks().map { PatientSummaryDoneClicked(screenKey.patientUuid) }
```
msasikanth, line 223: Since we are no longer changing the `ProgressMaterialButton` to `MaterialButton` let's change the commit message to `Rename doctorButton to teleconsultButton`.

app/src/main/res/values/strings.xml
```diff
@@ -214,7 +214,7 @@
   <string name="patientsummary_toolbar_title">%1$s, %2$s, %3$s</string>
   <string name="patientsummary_bangladesh_national_id">ID</string>
 
-  <string name="patientsummary_contact_doctor" translatable="false">Tele-consult</string>
+  <string name="patientsummary_teleconsult">Tele-consult</string>
```
msasikanth, line 217: Since we are now adding the `patientsummary_teleconsult` string in the first commit, we can combine the last 3 commits into one single commit called `Remove unused resources`.

msasikanth (changes requested):
Left couple of comments PTAL

msasikanth (approved):
LGTM

PR #1878, opened 2020-10-01
Fix issues with sync where sync events were being triggered at the sync level instead of the group level

vinaysshenoy:
This will be important for block level sync since we will need to trigger certain events when the sync of an entire group completes.


msasikanth (approved):
LGTM

PR #1882, opened 2020-10-05
Add `syncGroup` property to the `Facility` resource

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1457/add-syncgroup-property-to-the-facilityresource

### DIFF
--- a/app/src/main/java/org/simple/clinic/storage/migrations/Migration_80.kt
+++ b/app/src/main/java/org/simple/clinic/storage/migrations/Migration_80.kt
@@ -1,0 +1,16 @@
+package org.simple.clinic.storage.migrations
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+import org.simple.clinic.storage.inTransaction
+import javax.inject.Inject
+
+@Suppress("ClassName")
+class Migration_80 @Inject constructor(): Migration(79, 80) {
+
+  override fun migrate(database: SupportSQLiteDatabase) {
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/simple/clinic/storage/migrations/Migration_80.kt
```diff
@@ -0,0 +1,16 @@
...
+import androidx.sqlite.db.SupportSQLiteDatabase
+import org.simple.clinic.storage.inTransaction
+import javax.inject.Inject
+
+@Suppress("ClassName")
+class Migration_80 @Inject constructor(): Migration(79, 80) {
+
+  override fun migrate(database: SupportSQLiteDatabase) {
+    database.inTransaction {
+      execSQL(""" ALTER TABLE "Facility" ADD COLUMN "syncGroup" TEXT """)
```
msasikanth, line 13: Should we add a default value or will it be done automatically during migration?

msasikanth (commented):
PR looks good overall. Left one comment PTAL

app/src/main/java/org/simple/clinic/storage/migrations/Migration_80.kt
vinaysshenoy: The default value is `NULL`, which will be set automatically since the field is nullable. A later PR will trigger a resync so that the sync groups are fetched from the server.

msasikanth (approved):

msasikanth (approved):

PR #1884, opened 2020-10-05
Use manifest API endpoint in tests instead of the fallback API endpoint

vinaysshenoy:
We have less than five users who are current on a build before the country selection screen was added. This means that the fallback country mechanism is mostly unnecessary and we can think of removing it.

The first step in this would be to use the manifest endpoint in tests instead of the fallback. We will remove the fallback mechanism after reaching out to the users who are on older app versions in person and asking them to update.

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/di/TestRetrofitModule.kt
+++ b/app/src/androidTest/java/org/simple/clinic/di/TestRetrofitModule.kt
@@ -37,7 +37,7 @@ class TestRetrofitModule {
       moshi: Moshi,
       okHttpClient: OkHttpClient
   ): Retrofit {
-    val baseUrl = BuildConfig.FALLBACK_ENDPOINT
+    val baseUrl = BuildConfig.MANIFEST_ENDPOINT
 
     return Retrofit.Builder()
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/org/simple/clinic/di/TestRetrofitModule.kt
```diff
@@ -37,7 +37,7 @@ class TestRetrofitModule {
       moshi: Moshi,
       okHttpClient: OkHttpClient
   ): Retrofit {
-    val baseUrl = BuildConfig.FALLBACK_ENDPOINT
+    val baseUrl = BuildConfig.MANIFEST_ENDPOINT
```
Honey14, line 40: They both have the same link to it? Can you tell me what is the difference?

Honey14 (approved):
Looks good. Left one question

app/src/androidTest/java/org/simple/clinic/di/TestRetrofitModule.kt
vinaysshenoy: The explanation is in `gradle.properties` if you look for the `qaFallbackApiEndpoint` property.

PR #1890, opened 2020-10-07
Delete blood pressure measurements without a patient when deleting data from other sync groups

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1413/regularly-purge-data-outside-the-current-sync-group

### DIFF
--- a/app/src/main/java/org/simple/clinic/AppDatabase.kt
+++ b/app/src/main/java/org/simple/clinic/AppDatabase.kt
@@ -217,6 +217,7 @@ abstract class AppDatabase : RoomDatabase() {
       val facilityIdsInCurrentSyncGroup = facilityDao().facilityIdsInSyncGroup(currentFacility.syncGroup)
 
       patientDao().deletePatientsNotInFacilities(facilityIdsInCurrentSyncGroup)
+      bloodPressureDao().deleteWithoutLinkedPatient()
     }
   }
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -217,6 +217,7 @@ abstract class AppDatabase : RoomDatabase() {
       val facilityIdsInCurrentSyncGroup = facilityDao().facilityIdsInSyncGroup(currentFacility.syncGroup)
 
       patientDao().deletePatientsNotInFacilities(facilityIdsInCurrentSyncGroup)
+      bloodPressureDao().deleteWithoutLinkedPatient()
```
Honey14, line 220: In what scenario would BP measurements be linked without patient? Shouldn't this be deleted in the use-case where the patient is removed?

Honey14 (changes requested):
Looks good. Can we update the changelog for this?

app/src/main/java/org/simple/clinic/AppDatabase.kt
vinaysshenoy: Yes.

vinaysshenoy:
> Looks good. Can we update the changelog for this?

This has been done in a later PR which will add deletions for the other resources.

Honey14 (approved):

PR #1893, opened 2020-10-09
Fix prescribed drug frequency type adapter

msasikanth:


harimohanraj89 (approved):
That looks correct to me. Are the unit tests that need to be updated?

msasikanth:
> 
> 
> That looks correct to me. Are the unit tests that need to be updated?

I don't think we need to update any unit tests for this. But I want to wait until Monday before we merge this in to make sure if need to handle the data that has been entered in the SBX builds (I doubt we need to).

It won't effect the release, we will merge this is in first and then make a DEMO release on Monday.

PR #1901, opened 2020-10-13
Fix download path when getting image using file provider

msasikanth:


vinaysshenoy (approved):
Whoops, looks like I missed this when reviewing!

msasikanth:
> Whoops, looks like I missed this when reviewing!

Oh no, I introduced this in yesterdays PR. I forgot absolute path doesn't have end file seprators. I wanted to avoid using direct strings and instead switch to `Environment`

PR #1902, opened 2020-10-13
Change `ModelSync` instances to be tagged instead of grouped

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1570/change-modelsync-to-have-a-set-of-tags-instead-of-a-single-sync-group


vinaysshenoy:
Closing this because I realised that this method would not work for what I had in mind for block level sync.

PR #1908, opened 2020-10-14
App freezes when we press enter with empty pin in lock screen

msasikanth:
https://app.clubhouse.io/simpledotorg/story/589/app-freezes-when-we-press-enter-with-empty-pin-in-lock-screen

### DIFF
--- a/app/src/main/java/org/simple/clinic/security/pin/PinEntryEvent.kt
+++ b/app/src/main/java/org/simple/clinic/security/pin/PinEntryEvent.kt
@@ -17,3 +17,5 @@ data class PinAuthenticated(val data: Any?) : PinEntryEvent() {
 }
 
 data class PinVerified(val result: PinVerificationMethod.VerificationResult) : PinEntryEvent()
+
+object PinEntryDoneClicked : PinEntryEvent()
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/simple/clinic/security/pin/PinEntryEvent.kt
```diff
@@ -17,3 +17,5 @@ data class PinAuthenticated(val data: Any?) : PinEntryEvent() {
 }
 
 data class PinVerified(val result: PinVerificationMethod.VerificationResult) : PinEntryEvent()
+
+object PinEntryDoneClicked : PinEntryEvent()
```
vinaysshenoy, line 21: Can we add an analytics event name for this?

vinaysshenoy (changes requested):
PR looks good, left a comment. Please take a look?

app/src/main/java/org/simple/clinic/security/pin/PinEntryEvent.kt
msasikanth: Currently, we are not verifying or do anything on this click. So I was planning on adding this once we start doing something? 🤔

vinaysshenoy: Ah, good point. It looks good then.

vinaysshenoy (approved):

PR #1909, opened 2020-10-15
Turn off `Logger` for `com.spotify.mobius.ControllerStateBase`

msasikanth:


app/src/main/java/org/simple/clinic/ReleaseClinicApp.kt
```diff
@@ -15,6 +18,12 @@ class ReleaseClinicApp : ClinicApp() {
     listOf(MixpanelAnalyticsReporter(this).swallowErrors())
   }
 
+  override fun onCreate() {
+    super.onCreate()
+    val logger = LoggerFactory.getLogger("com.spotify.mobius.ControllerStateBase") as Logger
+    logger.level = Level.OFF
```
vinaysshenoy, line 24: Can we leave a comment here which explains why we're disabling this logging explicitly?

msasikanth: Done 👍🏽

Honey14 (dismissed):
Looks good.

Honey14 (approved):

PR #1910, opened 2020-10-15
Change the `syncGroup` property on `Facility` to be non-nullable

vinaysshenoy:


app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration81AndroidTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        "groupUuid" to UUID.fromString("24f5ecca-da1f-4c1d-951f-055db4cb1c2b"),
+        "createdAt" to Instant.parse("2018-01-01T00:00:00Z"),
+        "updatedAt" to Instant.parse("2018-01-01T00:00:01Z"),
+        "syncStatus" to "DONE",
+        "deletedAt" to Instant.parse("2018-01-01T00:00:02Z"),
+        "location_latitude" to 12.9653,
+        "location_longitude" to 77.5955413,
+        "config_diabetesManagementEnabled" to true,
+        "config_teleconsultationEnabled" to false,
+        "syncGroup" to null
```
msasikanth, line 37: If we are migrate from a DB version where `syncGroup` is not present, it will first add a `null` sync group and then migrate to an empty string right?

msasikanth (approved):
LGTM

app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration81AndroidTest.kt
vinaysshenoy: Yup, pending migrations are run in sequence.

PR #1911, opened 2020-10-15
Purge data from a different sync group whenever all syncs complete

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1413/regularly-purge-data-outside-the-current-sync-group

### DIFF
--- a/app/src/main/java/org/simple/clinic/sync/DataSync.kt
+++ b/app/src/main/java/org/simple/clinic/sync/DataSync.kt
@@ -202,10 +202,20 @@ class DataSync(
     }.exhaustive()
   }
 
+  private fun purgeOnCompletedSyncs(): SingleTransformer<List<SyncResult>, List<SyncResult>> {
+    return SingleTransformer { syncResultsStream ->
+      syncResultsStream
+          .doOnSuccess { syncResults -> if (syncResults.haveAllCompleted()) purgeOnSync.purgeUnusedData() }
+    }
+  }
+
   @WorkerThread
   @Throws(IOException::class) // This is only needed so Mockito can generate mocks for this method correctly
   fun syncTheWorld() {
-    allSyncs().ignoreElement().blockingAwait()
+    allSyncs()
+        .compose(purgeOnCompletedSyncs())
+        .ignoreElement()
+        .blockingAwait()
   }
 
   @WorkerThread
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/simple/clinic/sync/DataSync.kt
```diff
@@ -48,7 +48,7 @@ class DataSync(
 
   @Inject
   constructor(
-      modelSyncs: ArrayList<ModelSync>,
+      modelSyncs: List<@JvmSuppressWildcards ModelSync>,
```
Honey14, line 52: When do we need `JvmSuppressWildcards` ?

app/src/test/java/org/simple/clinic/sync/DataSyncTest.kt
```diff
@@ -687,6 +691,151 @@ class DataSyncTest {
...
+    )
+
+    val modelSync3 = FakeModelSync(
+        _name = "sync3",
+        config = dailySyncConfig,
+        pushError = RuntimeException()
+    )
+
+    val purgeOnSync = mock<PurgeOnSync>()
+    val dataSync = DataSync(
```
Honey14, line 864: Can we make `modelSync1` , `modelSync2` , `modelSync3`, `purgeOnSync` , `dataSync` as global variables? We're using them in most of the tests?

Honey14 (changes requested):
Looks great. Left one question and a comment. PTAL

app/src/main/java/org/simple/clinic/sync/DataSync.kt
vinaysshenoy: Dagger has an issue with injecting wildcard generics (If you look at `SyncModule`, it provides `frequentlySyncingRepositories` as `List<SynceableRepositories<*, *>>` which are wildcards. This is a problem only when you use Kotlin, which has this annotation that you can use to fix this problem.

app/src/test/java/org/simple/clinic/sync/DataSyncTest.kt
vinaysshenoy: The `modelSync` instances are not the same in the tests. They have the same names, but they behave differently.

Honey14 (approved):
Looks good

PR #1914, opened 2020-10-16
Show a dialog when registering a patient or viewing patient summary if diagnosis is not selected

Honey14:
https://app.clubhouse.io/simpledotorg/story/593/show-a-dialog-when-registering-a-patient-or-scheduling-an-appointment-if-diagnosis-is-not-selected


app/src/main/java/org/simple/clinic/medicalhistory/SelectDiagnosisErrorDialog.kt
```diff
@@ -0,0 +1,31 @@
...
+import androidx.appcompat.app.AppCompatDialogFragment
+import androidx.fragment.app.FragmentManager
+import org.simple.clinic.R
+
+class SelectDiagnosisErrorDialog : AppCompatDialogFragment() {
+
+  companion object {
+
+    private const val FRAGMENT_TAG = "select_diagnosis_error_alert"
+    fun show(fragmentManager: FragmentManager) {
```
vinaysshenoy, line 15: Can you add a line break here before the method declaration, please?

app/src/main/java/org/simple/clinic/medicalhistory/SelectDiagnosisErrorDialog.kt
```diff
@@ -0,0 +1,31 @@
...
+      val fragment = SelectDiagnosisErrorDialog().apply {
+        isCancelable = false
```
vinaysshenoy, line 17: Is there a need to make this uncancelable? I think it's fine since tapping on the "Save" button with bring up the dialog again.

app/src/main/java/org/simple/clinic/summary/medicalhistory/MedicalHistorySummaryView.kt
```diff
@@ -46,7 +46,7 @@ class MedicalHistorySummaryView(
   lateinit var activity: AppCompatActivity
 
   @Inject
-  lateinit var effectHandlerFactory: MedicalHistorySummaryEffectHandler.Factory
+  lateinit var effectHandlerFactory: MedicalHistorySummaryEffectHandler
```
vinaysshenoy, line 49: Can we rename this property since it's no longer a factory but the effect handler itself?

vinaysshenoy (changes requested):
Overall PR looks good, left a few small comments.

app/src/main/java/org/simple/clinic/medicalhistory/SelectDiagnosisErrorDialog.kt
Honey14: Fair point

vinaysshenoy (approved):

PR #1915, opened 2020-10-19
Bump AGP to 4.1.0

msasikanth:


app/src/main/AndroidManifest.xml
```diff
@@ -24,7 +35,7 @@
     android:label="@string/app_name"
     android:supportsRtl="true"
     android:theme="@style/Clinic.V2.Theme"
-    tools:ignore="AllowBackup,GoogleAppIndexingWarning"
+    tools:ignore="AllowBackup,GoogleAppIndexingWarning, LockedOrientationActivity"
```
Honey14, line 38: What would `LockedOrientationActivity` do?

lint/build.gradle
```diff
@@ -1,5 +1,4 @@
 apply plugin: 'java-library'
-apply plugin: 'kotlin'
```
Honey14, line 2: Why did we remove `kotlin` plugin from the lint `build.gradle` file?

app/build.gradle
```diff
@@ -273,7 +273,7 @@ dependencies {
   implementation "androidx.camera:camera-core:$versions.camerax"
```
Honey14, line 273: The commit message is correct but it has multiple changes and it's hard to understand which change is for which plugin or dependency

Honey14 (commented):
Overall looks good. Left few comments. PTAL

app/src/main/AndroidManifest.xml
msasikanth: We are basically ignoring a lint error for `android:screenOrientation` being locked to portrait.

lint/build.gradle
msasikanth: The build system was complaining about not finding this plugin in the `lint` module and removing it doesn't seem to break anything.  I may add it back in future if needed.

app/build.gradle
msasikanth: Oh, those are all for the plugin. Before the plugin update, the dependency issue was not visible in logs, only after updating to AGP 4.1.0 it started showing up. That's why I combined it in this commit instead of a separate one.

lint/build.gradle
Honey14: I'm not sure why we want and removed both but unless it doesn't break anything, it works i guess.

msasikanth:
Moving this to WIP, custom lint rules are not running.

msasikanth:
> Moving this to WIP, custom lint rules are not running.

This has been resolved. In order to test build the project and verify if the lint rules are being generated and we can user them.

Honey14 (approved):
Looks good.

PR #2023, opened 2020-10-21
Delay purging of sync group data for a specified time after switching

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1491/delay-purging-of-data-from-an-older-sync-group-if-the-blocks-have-been-switched-within-a-certain-time-window

### DIFF
--- a/app/src/main/java/org/simple/clinic/facility/change/confirm/ConfirmFacilityChangeEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/facility/change/confirm/ConfirmFacilityChangeEffectHandler.kt
@@ -12,6 +12,7 @@ import org.simple.clinic.main.TypedPreference.Type.FacilitySyncGroupSwitchedAt
 import org.simple.clinic.reports.ReportsRepository
 import org.simple.clinic.reports.ReportsSync
 import org.simple.clinic.util.Optional
+import org.simple.clinic.util.UtcClock
 import org.simple.clinic.util.scheduler.SchedulersProvider
 import java.time.Instant
 import javax.inject.Named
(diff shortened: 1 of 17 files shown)


app/src/main/java/org/simple/clinic/facility/change/confirm/ConfirmFacilityChangeEffectHandler.kt
```diff
@@ -31,6 +31,7 @@ class ConfirmFacilityChangeEffectHandler @AssistedInject constructor(
         .subtypeEffectHandler<ConfirmFacilityChangeEffect, ConfirmFacilityChangeEvent>()
         .addTransformer(ChangeFacilityEffect::class.java, changeFacility(schedulersProvider.io()))
         .addAction(CloseSheet::class.java, { uiActions.closeSheet() }, schedulersProvider.ui())
+        .addTransformer(LoadCurrentFacility::class.java, loadCurrentFacility())
```
msasikanth, line 41: Why aren't we using the current facility dagger dependency for this, Is it because it won't be updated when the facility is changed or is there a different reason? 🤔

app/src/test/java/org/simple/clinic/facility/change/confirm/ConfirmFacilityChangeUpdateTest.kt
```diff
@@ -10,13 +10,13 @@ import java.util.UUID
...
 
-  private val updateSpec = UpdateSpec<ConfirmFacilityChangeModel, ConfirmFacilityChangeEvent, ConfirmFacilityChangeEffect>(ConfirmFacilityChangeUpdate())
+  private val updateSpec = UpdateSpec(ConfirmFacilityChangeUpdate())
 
   @Test
   fun `when facility change is confirmed then user's facility should be changed`() {
     val selectedFacility = TestData.facility(UUID.fromString("ef47531f-9b8c-4045-8578-eda31f0952c4"))
     updateSpec
-        .given(ConfirmFacilityChangeModel())
+        .given(ConfirmFacilityChangeModel.create())
```
msasikanth: ~~This we are using the default model in a couple of tests, shall we extract these out into a global variable?~~ NVM this has been done in later commits.

msasikanth (commented):
LGTM. Left one comment PTAL

app/src/main/java/org/simple/clinic/facility/change/confirm/ConfirmFacilityChangeEffectHandler.kt
vinaysshenoy: We already have the facility repository which is being used to update the current facility, so I just decided to reuse that dependency instead of adding another one.

msasikanth (approved):

PR #2033, opened 2020-10-21
Add `TeleconsultStatusSheet` #1

msasikanth:
https://app.clubhouse.io/simpledotorg/story/1623/create-teleconsultstatusdialog


app/src/test/java/org/simple/clinic/summary/teleconsultation/status/TeleconsultStatusInitTest.kt
```diff
@@ -6,13 +6,17 @@ import com.spotify.mobius.test.InitSpec
...
 
 class TeleconsultStatusInitTest {
 
   @Test
   fun `when screen is created, then prefill teleconsult status if it exists`() {
     val initSpec = InitSpec(TeleconsultStatusInit())
-    val model = TeleconsultStatusModel.create(TeleconsultStatus.Yes)
+    val model = TeleconsultStatusModel.create(
+        teleconsultRecordId = UUID.fromString("60af7f4a-b61f-40ae-a2b9-9f5273a70d4e"),
+        teleconsultStatus = TeleconsultStatus.Yes
```
vinaysshenoy, line 18: Creating a model in the tests should follow the process they do in production. In production, we will never have a teleconsult status ready when creating the model and we'll end up constructing it via the events. We should follow the same flow here.

```kotlin
val model = TeleconsultStatusModel
  .create(UUID.fromString())
  .teleconsultStatusLoaded()
```

app/src/main/java/org/simple/clinic/summary/teleconsultation/status/TeleconsultStatusModel.kt
```diff
@@ -1,14 +1,17 @@
...
 
 data class TeleconsultStatusModel(
+    val teleconsultRecordId: UUID,
     val teleconsultStatus: TeleconsultStatus?
 ) {
 
   companion object {
 
-    fun create(teleconsultStatus: TeleconsultStatus?) = TeleconsultStatusModel(
+    fun create(teleconsultRecordId: UUID, teleconsultStatus: TeleconsultStatus?) = TeleconsultStatusModel(
```
vinaysshenoy, line 13: We've mixed two changes in this commit:

- Adding the teleconsult record ID as a property of the model
- Handling the done click event

These should be two separate commits for easier review.

vinaysshenoy (changes requested):

app/src/test/java/org/simple/clinic/summary/teleconsultation/status/TeleconsultStatusInitTest.kt
msasikanth: If user selects the still waiting status, won't we have to reselect it next time the sheet is shown? 🤔 . That was the only reason why I included default teleconsult status option, or else we can remove prefill effect as well.

app/src/test/java/org/simple/clinic/summary/teleconsultation/status/TeleconsultStatusUpdateTest.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.spotify.mobius.test.UpdateSpec.assertThatNext
+import org.junit.Test
+import org.simple.clinic.teleconsultlog.teleconsultrecord.TeleconsultStatus
+
+class TeleconsultStatusUpdateTest {
+
+  @Test
+  fun `when teleconsult status is changed, the update the model`() {
+    val updateSpec = UpdateSpec(TeleconsultStatusUpdate())
+    val model = TeleconsultStatusModel.create(null)
```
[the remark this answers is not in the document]
msasikanth: The initial idea behind this was to allow setting default `TeleconsultStatus`, but if we don't have to set default option I can remove this and prefill teleconsult status.

app/src/test/java/org/simple/clinic/summary/teleconsultation/status/TeleconsultUiRendererTest.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.nhaarman.mockitokotlin2.mock
+import com.nhaarman.mockitokotlin2.verify
+import com.nhaarman.mockitokotlin2.verifyNoMoreInteractions
+import org.junit.Test
+import org.simple.clinic.teleconsultlog.teleconsultrecord.TeleconsultStatus
+
+class TeleconsultUiRendererTest {
+
+  @Test
+  fun `when teleconsult status is not null, then enable done button`() {
```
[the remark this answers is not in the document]
msasikanth: By default, we are not doing anything with model data. We have the button disabled in the layout initially, user cannot deselect the radio buttons anyway, so it didn't make sense to have a disable button in ui renderer WDYT?

app/src/test/java/org/simple/clinic/summary/teleconsultation/status/TeleconsultStatusInitTest.kt
vinaysshenoy: Isn't that taken care of by the prefill status effect? When the effect is triggered, it prefills the UI which triggers a checked change event which then updates the model again.

app/src/test/java/org/simple/clinic/summary/teleconsultation/status/TeleconsultUiRendererTest.kt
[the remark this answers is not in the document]
vinaysshenoy: Ah, makes sense.

app/src/test/java/org/simple/clinic/summary/teleconsultation/status/TeleconsultStatusInitTest.kt
msasikanth: NVM, turns out we don't have to prefill the status in sheet. Confirmed with Pragati, sorry 😅 

Will remove the prefill effect as well default status in the factory method.

msasikanth: Correct me if I am wrong, State restoration should automatically work with the RadioGroup so we don't manually have to prefill it in those scenarios as well right?

vinaysshenoy: Yup, that should be handled by the system

vinaysshenoy (approved):

PR #2034, opened 2020-10-21
Add `TeleconsultStatusSheet`

msasikanth:
https://app.clubhouse.io/simpledotorg/story/1623/add-teleconsultstatussheet


msasikanth:
Merge this in once #2033 is merged in

app/src/main/java/org/simple/clinic/summary/teleconsultation/status/TeleconsultStatusEvent.kt
```diff
@@ -1,11 +1,14 @@
 package org.simple.clinic.summary.teleconsultation.status
 
 import org.simple.clinic.teleconsultlog.teleconsultrecord.TeleconsultStatus
+import org.simple.clinic.widgets.UiEvent
 
-sealed class TeleconsultStatusEvent
+sealed class TeleconsultStatusEvent : UiEvent
 
 data class TeleconsultStatusChanged(val teleconsultStatus: TeleconsultStatus) : TeleconsultStatusEvent()
```
vinaysshenoy, line 8: Should we add an analytics name for this as well since it's triggered via a UI event?

app/src/main/java/org/simple/clinic/summary/teleconsultation/status/TeleconsultStatusSheet.kt
```diff
@@ -0,0 +1,147 @@
...
+  @Inject
+  lateinit var effectHandlerFactory: TeleconsultStatusEffectHandler.Factory
+
+  private lateinit var component: TeleconsultStatusComponent
+
+  private val teleconsultRecordId by unsafeLazy {
+    intent.getSerializableExtra(EXTRA_TELECONSULT_RECORD_ID) as UUID
+  }
+
+  private val idToTeleconsultStatus = mapOf(
```
vinaysshenoy, line 57: Can we rename the property to reflect that it is the ID of the teleconsult radio buttons instead of just `id`?

app/src/main/res/values/styles_v2.xml
```diff
@@ -649,7 +649,7 @@
   </style>
 
   <style name="Clinic.V2.MaterialButton" parent="Widget.MaterialComponents.Button">
-    <item name="backgroundTint">?attr/colorSecondary</item>
+    <item name="backgroundTint">@color/material_button_background</item>
```
vinaysshenoy, line 652: This change will apply to all buttons where this style is being used. That is expected, right?

vinaysshenoy (changes requested):
Overall PR looks good. Left a couple of minor comments.

app/src/main/java/org/simple/clinic/summary/teleconsultation/status/TeleconsultStatusEvent.kt
msasikanth: My only concern was what if the user keep changes that radio button in the same session we will have few analytics events for that. If that's fine I can implement it.

app/src/main/res/values/styles_v2.xml
msasikanth: Yes, that's expected behaviour. When a button is disabled we want it to be grey and not just keep the actual button color which will confuse the users right?

app/src/main/java/org/simple/clinic/summary/teleconsultation/status/TeleconsultStatusEvent.kt
vinaysshenoy: I'd say that was the desired behaviour since it's user analytics.

app/src/main/res/values/styles_v2.xml
vinaysshenoy: Yup, just wanted to confirm.

vinaysshenoy (changes requested):
PR looks good, let's just add the analytics event in and we'll be good to go.

vinaysshenoy (approved):

PR #2041, opened 2020-10-26
Simplify generation of Business ID metadata strings

vinaysshenoy:


app/src/main/java/org/simple/clinic/patient/businessid/BusinessId.kt
```diff
@@ -85,6 +89,17 @@ data class BusinessId(
 
       @VisibleForTesting(otherwise = VisibleForTesting.NONE)
       fun values() = TypeAdapter.knownMappings.keys
+
+      fun forIdentifierType(
+          type: IdentifierType
+      ): Optional<MetaDataVersion> {
+        return Optional.ofNullable(when (type) {
```
msasikanth, line 96: Should we extract out the `when` block into a separate variable for readability purposes?

msasikanth (changes requested):
Looks good overall. Left one comment PTAL

app/src/main/java/org/simple/clinic/patient/businessid/BusinessId.kt
vinaysshenoy: Yes, I can do that.

msasikanth (approved):

PR #2042, opened 2020-10-26
Clean up patient repository

vinaysshenoy:
This is being done in preparation for some future refactoring work before we get to the "Faster patient registration" feature.

- Extract patient registration to a use case outside the `PatientRepository` scoped to the patient entry screens
- Unify the patient entry and edit forms in some fashion so we don't need to duplicate most of the code

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -3,7 +3,6 @@ package org.simple.clinic.patient
 import androidx.arch.core.executor.testing.InstantTaskExecutorRule
 import com.google.common.truth.Truth.assertThat
 import com.squareup.moshi.JsonAdapter
-import io.reactivex.Single
 import org.junit.After
 import org.junit.Before
 import org.junit.Rule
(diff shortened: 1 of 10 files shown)


app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -3275,19 +3256,18 @@ class PatientRepositoryAndroidTest {
...
             loggedInUser = loggedInUser,
             facility = facilityToSavePatientAt,
             patientUuid = UUID.fromString("0d8b5e31-2695-49b3-ba23-d4f198012870"),
             addressUuid = UUID.fromString("05dfcc13-b855-4e06-be91-7c09c79d98ca"),
             supplyUuidForBpPassport = { UUID.fromString("b6915d20-d396-4c5d-b25c-4b2d9b81fd2b") },
-            supplyUuidForAlternativeId = { UUID.fromString("91539c5f-70a0-4e69-9740-fa8b37fa2f16") },
-            supplyUuidForPhoneNumber = { UUID.fromString("b842da76-26f8-4d7d-814a-415209335ecb") }
-        )
+            supplyUuidForAlternativeId = { UUID.fromString("91539c5f-70a0-4e69-9740-fa8b37fa2f16") }
+        ) { UUID.fromString("b842da76-26f8-4d7d-814a-415209335ecb") }
```
Honey14, line 3267: Can you tell me what is this UUID string for? phone number UUID?

Honey14 (approved):
Looks good overall. Left few questions. Please take a look

app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryEffectHandler.kt
```diff
@@ -37,7 +37,7 @@ class NewMedicalHistoryEffectHandler @AssistedInject constructor(
         .subtypeEffectHandler<NewMedicalHistoryEffect, NewMedicalHistoryEvent>()
         .addConsumer(OpenPatientSummaryScreen::class.java, { effect -> uiActions.openPatientSummaryScreen(effect.patientUuid) }, schedulersProvider.ui())
         .addTransformer(RegisterPatient::class.java, registerPatient(schedulersProvider.io()))
-        .addTransformer(LoadOngoingPatientEntry::class.java, loadOngoingNewPatientEntry(schedulersProvider.io()))
```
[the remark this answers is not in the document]
vinaysshenoy: It was an asynchronous call earlier. We're slowly converting most of our calls which return Rx based types to synchronously returning them since the API is a lot simpler and effect handlers work on a background thread anyway.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
vinaysshenoy: Yes, I don't know why but it seems like the IDE auto format is moving the last lambda parameter out of the method invocation automatically. I'll fix this in the next PR in the refactoring work.

PR #2044, opened 2020-10-28
When teleconsult status is changed, then go to home screen

msasikanth:


app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -275,6 +276,11 @@ class PatientSummaryScreen(
         .subscribe(::openContinuation)
   }
 
+  private fun teleconsultStatusChanged() = screenRouter.streamScreenResults()
+      .ofType<ActivityResult>()
+      .extractSuccessful(TELECONSULT_STATUS_SHEET) {}
```
vinaysshenoy, line 281: This extension is meant to extract a result of the result intent. If you only care whether it was successful and not what the result is, maybe a new extension method like `.filterIfSuccess()` is a better option here than adding a blank lambda?

vinaysshenoy (changes requested):
Overall PR looks good. One change that we can discuss and see whether it needs to be done or not.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: That's a good point, will add an extension.

vinaysshenoy (approved):

PR #2054, opened 2020-11-03
Enable ViewBinding

msasikanth:


app/src/main/java/org/simple/clinic/splash/SplashScreen.kt
```diff
@@ -5,25 +5,34 @@ import android.util.AttributeSet
...
 import androidx.navigation.findNavController
 import com.airbnb.lottie.LottieDrawable
-import kotlinx.android.synthetic.main.screen_splash.view.*
 import org.simple.clinic.R
+import org.simple.clinic.databinding.ScreenSplashBinding
 
 class SplashScreen(context: Context, attributeSet: AttributeSet) : ConstraintLayout(context, attributeSet) {
 
+  private var _binding: ScreenSplashBinding? = null
+  private val binding get() = _binding!!
```
vinaysshenoy, line 14: Maybe we can add accessor properties for the child views instead of the `binding` property?

```kotlin
private val splashLottieView
  get() = binding!!.splashLottieView
```

vinaysshenoy: In this way, we will not need to change the event setup, just add whatever views we need to reference in one place.

msasikanth: Oh yeah, good idea. I was just going with the default view binding setup, will change it.

msasikanth: Done

Honey14 (approved):
Looks good

PR #2055, opened 2020-11-03
Disable running the app on rooted devices

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1814/add-root-detection-at-launch


build.gradle
```diff
@@ -78,7 +78,8 @@ buildscript {
       desugarJdk               : '1.0.9',
       signaturePad             : '1.3.1',
       viewpager2               : '1.0.0',
-      lint                     : '27.0.1'
+      lint                     : '27.0.1',
+      rootbeer                 : '5fef9f62f9'
```
msasikanth, line 82: Are we using Jitpack version rather than maven?

msasikanth (commented):
PR LGTM. Left one comment PTAL

build.gradle
vinaysshenoy: The library is not pushed to maven central, and the latest official release is months old. A bunch of fixes for false positives have gotten into `master` without an official release, so I opted to fetch the current `master` commit instead.

msasikanth (approved):

PR #2068, opened 2020-11-04
Migration app screens to use ViewBinding #1

Honey14:


vinaysshenoy (approved):
LGTM

PR #2069, opened 2020-11-05
Add recipes doc

msasikanth:


doc/recipes.md
```diff
@@ -0,0 +1,73 @@
...
+@Test
+fun `the feature is overriden in this test`() {
+  Features.overrides[Feature.SecureCall] = true
+}
+```
+
+You can read more about this usage [here](https://github.com/simpledotorg/simple-android/issues/1364)
+
+
+## Dagger Qualifiers
```
Honey14, line 39: We already have a well documented `010-dagger-qualifier-migration.md`. Do we need to have this here as well?

Honey14 (commented):
Looks great. Left one comment. PTAL

doc/recipes.md
msasikanth: Good point, we don't need to capture it again here. Removing it 👍🏽

Honey14 (approved):
Looks great

PR #2073, opened 2020-11-09
Update patient address label field for Ethiopia

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1881/change-the-label-of-district-to-zone-for-ethiopia-in-patient-entry-fields


msasikanth (approved):
LGTM

PR #2076, opened 2020-11-09
Convert the blood sugar values while rendering when the blood sugar unit is changed

msasikanth:
https://app.clubhouse.io/simpledotorg/story/1792/convert-the-blood-sugar-values-while-rendering-when-the-blood-sugar-unit-is-changed


app/src/main/java/org/simple/clinic/bloodsugar/BloodSugarReading.kt
```diff
@@ -44,6 +44,17 @@ data class BloodSugarReading(val value: String, val type: BloodSugarMeasurementT
...
+    get() {
+      return when (type) {
+        Random -> value.toFloat().toString()
+        PostPrandial -> value.toFloat().toString()
+        Fasting -> value.toFloat().toString()
+        HbA1c -> value.toFloat().toString()
+        is Unknown -> value.toInt().toString()
+      }
+    }
+
```
vinaysshenoy, line 57: We could change `displayValue` and `displayUnit` to functions which accept the `BloodSugarUnitPreference` as a parameter and then do the conversions within that. This will allow us to keep all the conditional logic in a single place instead of spreading it across multiple places.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/bloodsugar/BloodSugarReading.kt
msasikanth: But didn't we talk about `BloodSugarReading` not knowing about `BloodSugarUnitPreference`? That's why I went with this approach. Changing `displayValue` and `displayUnit` to functions.

vinaysshenoy: No, what I meant was that the properties of `BloodSugarReading` should always be in `mg/DL`.

msasikanth: Ah okay. Doing this change 👍🏽

vinaysshenoy: In this scenario, when we receive a `BloodSugarReading`, we have absolutely no idea if it's in mmol/L or `mg/DL`. Having functions that accept it as parameters is perfectly fine since those functions can be moved to other places later if needed.

app/src/main/java/org/simple/clinic/bloodsugar/BloodSugarReading.kt
```diff
@@ -33,16 +35,32 @@ data class BloodSugarReading(val value: String, val type: BloodSugarMeasurementT
...
+    return when (type) {
+      Random -> value.toFloat().toInt().toString()
+      PostPrandial -> value.toFloat().toInt().toString()
+      Fasting -> value.toFloat().toInt().toString()
+      HbA1c -> value.toFloat().toString()
+      is Unknown -> value.toInt().toString()
     }
+  }
+
+  private fun displayValueMmol(): String {
```
vinaysshenoy, line 55: Did we miss doing the conversion from mg/dl to mmol/l here?

vinaysshenoy (changes requested):
Overall PR looks good, left a comment.

app/src/main/java/org/simple/clinic/bloodsugar/BloodSugarReading.kt
msasikanth: OMG yes, Since we are no longer using `toMmol` we should be doing that here. Fixing it 😅

vinaysshenoy (approved):

PR #2080, opened 2020-11-10
Convert the blood sugar values to mg/dL while saving/updating

msasikanth:
https://app.clubhouse.io/simpledotorg/story/1791/convert-the-blood-sugar-values-to-mg-dl-while-saving-updating


app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffect.kt
```diff
@@ -33,7 +33,7 @@ object Dismiss : BloodSugarEntryEffect()
 
 object ShowDateEntryScreen : BloodSugarEntryEffect()
 
-data class ShowBloodSugarValidationError(val result: ValidationResult) : BloodSugarEntryEffect()
+data class ShowBloodSugarValidationError(val result: ValidationResult, val unitPreference: BloodSugarUnitPreference) : BloodSugarEntryEffect()
```
Honey14, line 36: Can we rename `BloodUnitPreference` to `BloodSugarUnitPreference` in the commit message?

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
```diff
@@ -259,13 +262,26 @@ class BloodSugarEntrySheet : BottomSheetActivity(), BloodSugarEntryUi, RemoveBlo
     showBloodSugarErrorMessage(getString(R.string.bloodsugarentry_error_empty))
   }
 
-  override fun showBloodSugarHighError(measurementType: BloodSugarMeasurementType) {
+  override fun showBloodSugarHighError(measurementType: BloodSugarMeasurementType, unitPreference: BloodSugarUnitPreference) {
+    when (unitPreference) {
+      Mg -> showBloodSugarHighErrorForMg(measurementType)
```
Honey14, line 267: We're adding the error for both unit preferences here and the commit message suggests for mmol/L only? Can we change it to add for both the unit preferences?

Honey14 (changes requested):
Looks good.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffect.kt
msasikanth: Good catch. Fixed it

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
msasikanth: Here we are reusing existing mg error and added a new error for mmol, so that why I didn't mention mg in the commit message.

Honey14 (approved):
Looks great

PR #2081, opened 2020-11-10
Add blood sugar unit switching button

Honey14:
https://app.clubhouse.io/simpledotorg/story/1789/add-blood-sugar-unit-switching-button-in-bloodsugarentrysheet


app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandler.kt
```diff
@@ -68,9 +71,19 @@ class BloodSugarEntryEffectHandler @AssistedInject constructor(
...
         .addConsumer(ShowConfirmRemoveBloodSugarDialog::class.java, { ui.showConfirmRemoveBloodSugarDialog(it.bloodSugarMeasurementUuid) }, schedulersProvider.ui())
+        .addTransformer(LoadBloodSugarUnitPreference::class.java, loadBloodSugarUnitPreference())
         .build()
   }
 
+  private fun loadBloodSugarUnitPreference(): ObservableTransformer<LoadBloodSugarUnitPreference, BloodSugarEntryEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .map { bloodSugarUnitPreference.get() }
```
msasikanth, line 82: This is not an observable stream, we want this to be an observable stream so that whenever the unit changes we can update the layout accordingly.

msasikanth: So currently it's an immediate call and it will only send out `BloodSugarUnitPreferenceLoaded` once whenever the effect is called.

app/src/main/res/layout/sheet_blood_sugar_entry.xml
```diff
@@ -104,9 +104,28 @@
...
+          android:visibility="gone"
+          app:layout_constraintEnd_toEndOf="@+id/bloodSugarReadingEditText"
+          app:layout_constraintStart_toStartOf="@+id/bloodSugarReadingEditText"
+          app:layout_constraintTop_toBottomOf="@+id/bloodSugarReadingEditText" />
+
+        <com.google.android.material.button.MaterialButton
+          android:id="@+id/bloodSugarReadingUnitButton"
+          android:layout_width="wrap_content"
+          android:layout_height="wrap_content"
+          android:layout_marginTop="@dimen/spacing_10"
```
msasikanth, line 122: The margin for the button is 16dp, please change this.

app/src/main/res/layout/sheet_blood_sugar_entry.xml
```diff
@@ -104,9 +104,28 @@
...
           android:labelFor="@+id/bloodSugarReadingEditText"
           android:text="@string/bloodsugarentry_mg_dl"
           android:textAppearance="@style/Clinic.V2.TextAppearance.Body2Left.Grey1"
+          app:cornerRadius="30dp"
```
msasikanth, line 126: It's not a good idea to hard code corner radius for the button. We want the corner radius to be half the height of the button. We should be using shape appearance here to set the corner radius to 50% of the height.

app/src/main/res/layout/sheet_blood_sugar_entry.xml
```diff
@@ -104,9 +104,28 @@
...
+          android:textAppearance="@style/Clinic.V2.TextAppearance.Body2Left.Grey1"
+          android:visibility="gone"
+          app:layout_constraintEnd_toEndOf="@+id/bloodSugarReadingEditText"
+          app:layout_constraintStart_toStartOf="@+id/bloodSugarReadingEditText"
+          app:layout_constraintTop_toBottomOf="@+id/bloodSugarReadingEditText" />
+
+        <com.google.android.material.button.MaterialButton
+          android:id="@+id/bloodSugarReadingUnitButton"
+          android:layout_width="wrap_content"
+          android:layout_height="wrap_content"
```
msasikanth, line 121: The button height is a bit large compared to the mockups, can we create a new style for this button and adjust minimun height to 32dp.

In that style, we can override and provide a shape appearance overlay for setting the corner radius.

https://material.io/develop/android/theming/shape

app/src/main/res/values/strings.xml
```diff
@@ -511,6 +511,7 @@
 
   <!-- Blood sugar entry -->
   <string name="bloodsugarentry_mg_dl" translatable="false">mg/dL</string>
+  <string name="bloodsugarentry_mmol_l" translatable="false">Mmol/L</string>
```
msasikanth, line 514: I think the unit is supposed to start with a small case `m`. `mmol/L`

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
```diff
@@ -383,6 +383,16 @@ class BloodSugarEntrySheet : BottomSheetActivity(), BloodSugarEntryUi, RemoveBlo
     bloodSugarReadingUnitButton.text = getString(R.string.bloodsugarentry_mg_dl)
   }
 
+  override fun showBloodSugarUnitPreferenceButton() {
+    bloodSugarReadingUnitButton.visibility = View.VISIBLE
+    bloodSugarReadingUnitLabel.visibility = View.GONE
```
msasikanth, line 373: Can we separate hiding the `bloodSugarReadingUnitLabel` visibility to it's own method please?

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
```diff
@@ -383,6 +383,16 @@ class BloodSugarEntrySheet : BottomSheetActivity(), BloodSugarEntryUi, RemoveBlo
...
+  }
+
+  override fun hideBloodSugarUnitPreferenceButton() {
+    bloodSugarReadingUnitButton.visibility = View.GONE
+    bloodSugarReadingUnitLabel.visibility = View.VISIBLE
```
msasikanth, line 378: Same comment as above, can we move the showing on `bloodSugarReadingUnitLabel` into its own method.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRenderer.kt
```diff
@@ -15,6 +19,15 @@ class BloodSugarEntryUiRenderer(
...
     manageProgress(model)
     manageBloodSugarUnitPreferenceButtonText(model)
+    manageBloodSugarUnitPreferenceButtonVisibility(model)
+  }
+
+  private fun manageBloodSugarUnitPreferenceButtonVisibility(model: BloodSugarEntryModel) {
+    when (model.openAs.measurementType) {
+      Random, PostPrandial, Fasting -> ui.showBloodSugarUnitPreferenceButton()
+      HbA1c -> ui.hideBloodSugarUnitPreferenceButton()
+      else -> ui.showBloodSugarUnitPreferenceButton()
```
msasikanth, line 50: I think we should hide the blood sugar unit preference button for unknown measurement types to avoid any unnecessary conversions. Please check with Vinay once.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
```diff
@@ -393,6 +393,18 @@ class BloodSugarEntrySheet : BottomSheetActivity(), BloodSugarEntryUi, RemoveBlo
     bloodSugarReadingUnitLabel.visibility = View.VISIBLE
   }
 
+  override fun setInputTypeForBloodSugarMmol() {
```
msasikanth, line 381: Probably instead of creating 3 methods for switching the input types, we can have `decimalBloodSugarInputType` and `numberBloodSugarInputType` (you can think of better naming). That way we will have 2 methods that are not tied to specific blood sugar type or unit.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
```diff
@@ -145,22 +145,7 @@ class BloodSugarEntrySheet : BottomSheetActivity(), BloodSugarEntryUi, RemoveBlo
```
msasikanth, line 145: Just an FYI, we usually name the commit `Remove unused resources` if we are removing something from the `res` folder like layouts, strings, color, etc., We usually name this as `Remove unused code in -file-name-`

msasikanth (changes requested):
PR looks good overall. Left a few comments PTAL

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
Honey14: Great observation. Changing this to have two methods.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRenderer.kt
Honey14: Pushing the updated changes now. Will change this when Vinay replied in the next PR if required.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandler.kt
```diff
@@ -68,9 +71,18 @@ class BloodSugarEntryEffectHandler @AssistedInject constructor(
...
         .addTransformer(UpdateBloodSugarEntry::class.java, updateBloodSugarEntryTransformer(schedulersProvider.io()))
         .addConsumer(ShowConfirmRemoveBloodSugarDialog::class.java, { ui.showConfirmRemoveBloodSugarDialog(it.bloodSugarMeasurementUuid) }, schedulersProvider.ui())
+        .addTransformer(LoadBloodSugarUnitPreference::class.java, loadBloodSugarUnitPreference())
         .build()
   }
 
+  private fun loadBloodSugarUnitPreference(): ObservableTransformer<LoadBloodSugarUnitPreference, BloodSugarEntryEvent> {
+    return ObservableTransformer { effect ->
+      effect
+          .switchMap { bloodSugarUnitPreference.asObservable().subscribeOn(schedulersProvider.io()) }
```
msasikanth, line 81: I think we can observe on io scheduler here, instead of subscribing on io for the preference only.

```kt
      effect
          .switchMap { bloodSugarUnitPreference.asObservable() }
          .map(::BloodSugarUnitPreferenceLoaded)
```

app/src/main/res/values/shape.xml
```diff
@@ -8,4 +8,14 @@
     <item name="cornerSizeBottomLeft">@dimen/spacing_0</item>
     <item name="cornerSizeBottomRight">@dimen/spacing_0</item>
   </style>
+
+  <style name="ShapeAppearanceOverlay.Clinic.V2.UnitPreferenceButton" parent="">
+    <item name="cornerFamily">rounded</item>
+    <item name="cornerSize">50%</item>
+  </style>
+
+  <style name="Clinic.V2.MaterialButton.UnitPreference">
```
msasikanth, line 17: This should be in `styles_v2.xml`

app/src/main/res/values/shape.xml
```diff
@@ -8,4 +8,14 @@
...
+    <item name="cornerRadius">@null</item>
```
msasikanth, line 18: Also, it seems like you forgot to set the min-height for this button, can we also set the text appearance in this style instead of at the view, please? Please check the mockups or we can pair on this if you need more info.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
```diff
@@ -376,6 +375,14 @@ class BloodSugarEntrySheet : BottomSheetActivity(), BloodSugarEntryUi, RemoveBlo
     bloodSugarDateButton.visibleOrGone(isVisible = true)
   }
 
+  override fun showBloodSugarUnitPreferenceMmol() {
```
msasikanth, line 363: Do you think we should name these methods as `setBloodSugarUnitPreferenceMmol` or `setBloodSugarUnitPreferenceMg`, since we are not actually showing the button, but rather just setting the unit preference. WDYT?

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRenderer.kt
msasikanth: Let's just hide this, for now, we will continue saving them as `mg` anyway.

msasikanth (changes requested):
Left a few comments PTAL

app/src/main/res/values/shape.xml
msasikanth: The current button height in the mockups is 32dp. The current default button height in Simple is 48dp, that's why the button looks a bit big compared to mockups.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryEffectHandler.kt
Honey14: So we observe on the effect instead of observing just the preference, right?

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntrySheet.kt
Honey14: Yes this makes sense and we're also not setting but instead setting the label for it. So changed it to `setBloodSugarUnitPreferenceLabelToMmol` and `setBloodSugarUnitPreferenceLabelToMg`

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRenderer.kt
msasikanth: It doesn't seem like we are hiding the blood sugar unit button for unknow measurement type, please check? Also let's use `Unknown` instead of an else block.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRenderer.kt
```diff
@@ -20,6 +20,24 @@ class BloodSugarEntryUiRenderer(
...
+    setInputTypeBloodSugarUnitLabel(model)
+  }
+
+  private fun setInputTypeBloodSugarUnitLabel(model: BloodSugarEntryModel) {
+    when (model.openAs.measurementType) {
+      Random, PostPrandial, Fasting -> {
+        setInputTypeUnitPreferenceSelectionBased(model)
+      }
+      HbA1c -> ui.decimalOrNumericBloodSugarInputType()
+      else -> ui.numericBloodSugarInputType()
```
msasikanth, line 35: Instead of else block let's define `Unknown`.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRenderer.kt
```diff
@@ -28,7 +28,10 @@ class BloodSugarEntryUiRenderer(
       Random, PostPrandial, Fasting -> {
         setInputTypeUnitPreferenceSelectionBased(model)
       }
-      HbA1c -> ui.decimalOrNumericBloodSugarInputType()
+      HbA1c -> {
+        ui.decimalOrNumericBloodSugarInputType()
+        ui.setLabelForHbA1c()
+      }
       else -> ui.numericBloodSugarInputType()
```
msasikanth, line 35: If we are going to hide the unit preference button in unknow measurement type, we should be showing the label as `mg/dL`. WDYT?

msasikanth (changes requested):
Left couple of comments PTAL

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRenderer.kt
Honey14: Hey, Can you check this again? I can see this has been updated. I'll update it to use `Unknown` instead of `else`

msasikanth: Oh okay, I missed it. you have changed it a different commit where we are handling decimal input.

app/src/main/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUiRenderer.kt
Honey14: Yes I think mg/dL is better than showing nothing there.

msasikanth (approved):

msasikanth (dismissed):

msasikanth (approved):

PR #2116, opened 2020-11-13
Migrate app to use View Binding instead of Kotlin synthetic view accessors - 3

msasikanth:


Honey14 (approved):
Looks good

PR #2118, opened 2020-11-13
Add `BloodSugarUnitSelectionDialog`

Honey14:
https://app.clubhouse.io/simpledotorg/story/1790/create-bloodsugarunitselectiondialog


app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionDialog.kt
```diff
@@ -10,7 +10,7 @@ import androidx.appcompat.app.AppCompatDialogFragment
 import androidx.fragment.app.FragmentManager
 import org.simple.clinic.R
 
-class BloodSugarUnitSelectionDialog : AppCompatDialogFragment() {
+class BloodSugarUnitSelectionDialog : AppCompatDialogFragment(), BloodSugarUnitSelectionUiActions {
```
vinaysshenoy, line 29: This could actually be a separate commit which just implements the UI actions on the dialog since this change is not necessary for handling the event in the `update()` function.

app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionDialog.kt
```diff
@@ -23,17 +40,87 @@ class BloodSugarUnitSelectionDialog : AppCompatDialogFragment(), BloodSugarUnitS
...
+    )
+  }
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    delegate.onRestoreInstanceState(savedInstanceState)
+  }
+
+  override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+    return layout
```
vinaysshenoy, line 121: Is this necessary? By default, it should return the same layout which the dialog already has, right?

app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionDialog.kt
```diff
@@ -23,17 +40,87 @@ class BloodSugarUnitSelectionDialog : AppCompatDialogFragment(), BloodSugarUnitS
...
+  }
+
+  @SuppressLint("CheckResult", "InflateParams")
   override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
     layout = LayoutInflater.from(context).inflate(R.layout.dialog_bloodsugar_selectionunit, null)
+
     return AlertDialog.Builder(requireContext())
         .setTitle(getString(R.string.blood_sugar_unit_selection_choose))
```
vinaysshenoy, line 129: Do we need to get the string resource manually? As far as I know, there's a method overload for `setTitle` which accepts a string resource ID.

app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionUiActions.kt
```diff
@@ -1,5 +1,8 @@
 package org.simple.clinic.bloodsugar.unitselection
 
+import org.simple.clinic.bloodsugar.BloodSugarUnitPreference
+
 interface BloodSugarUnitSelectionUiActions {
   fun closeDialog()
+  fun prefilledBloodSugarUnitSelection(blodSugarUnitPreference: BloodSugarUnitPreference)
```
vinaysshenoy, line 7: The name of this method is in the past tense, that is it implies that _something has happened_ instead of telling the UI action of an operation that must be performed. Can we rename this method to better reflect this, please?

app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionDialog.kt
```diff
@@ -36,10 +37,29 @@ class BloodSugarUnitSelectionDialog : AppCompatDialogFragment(), BloodSugarUnitS
...
+      }
+
+      val arguments = Bundle()
+      arguments.putSerializable(KEY_UNIT_PREF, bloodSugarUnitPreference)
 
-    fun show(fragmentManager: FragmentManager) {
       val fragment = BloodSugarUnitSelectionDialog()
-      fragment.show(fragmentManager, FRAGMENT_TAG)
+      fragment.arguments = arguments
+
```
vinaysshenoy, line 67: Extra blank line?

app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionDialog.kt
```diff
@@ -132,6 +153,15 @@ class BloodSugarUnitSelectionDialog : AppCompatDialogFragment(), BloodSugarUnitS
     dismiss()
   }
 
+  override fun prefilledBloodSugarUnitSelection(blodSugarUnitPreference: BloodSugarUnitPreference) {
+    val bloodSugarUnitPreferenceButton: RadioButton? = when (blodSugarUnitPreference) {
```
vinaysshenoy, line 168: This does not need to be nullable, right? If the value is anything other than `Mg` or `Mmol` (which is impossible since it's an enum), this will not be a valid state and we can raise an error instead.

app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionDialog.kt
```diff
@@ -132,6 +153,15 @@ class BloodSugarUnitSelectionDialog : AppCompatDialogFragment(), BloodSugarUnitS
...
+      BloodSugarUnitPreference.Mg -> bloodSugarUnitMg
+      BloodSugarUnitPreference.Mmol -> bloodSugarUnitMmol
+    }
+
+    bloodSugarUnitPreferenceButton?.isChecked = true
```
vinaysshenoy, line 173: Instead of this, there's a method on the `RadioGroup` called [`check()`](https://developer.android.com/reference/android/widget/RadioGroup#check(int)), which is the right method to use in this scenario since the radio buttons are wrapped in a radio group.

app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionModel.kt
```diff
@@ -2,6 +2,18 @@ package org.simple.clinic.bloodsugar.unitselection
 
 import android.os.Parcelable
 import kotlinx.android.parcel.Parcelize
+import org.simple.clinic.bloodsugar.BloodSugarUnitPreference
 
 @Parcelize
-class BloodSugarUnitSelectionModel : Parcelable
+data class BloodSugarUnitSelectionModel(
+    val bloodSugarUnitPreference: BloodSugarUnitPreference
```
vinaysshenoy, line 9: If this is added to the model, it should be updated when the selection is changed via the radio group. I don't see this being done, can we update this please?

app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionInit.kt
```diff
@@ -0,0 +1,11 @@
+package org.simple.clinic.bloodsugar.unitselection
+
+import com.spotify.mobius.First
+import com.spotify.mobius.Init
+
+class BloodSugarUnitSelectionInit : Init<BloodSugarUnitSelectionModel, BloodSugarUnitSelectionEffect> {
+  override fun init(model: BloodSugarUnitSelectionModel): First<BloodSugarUnitSelectionModel, BloodSugarUnitSelectionEffect> {
+    val effect = mutableSetOf(PreFillBloodSugarUnitSelected(model.bloodSugarUnitPreference))
+    return First.first(model, effect as Set<BloodSugarUnitSelectionEffect>)
```
vinaysshenoy, line 9: If we have only a single effect, we do not need to create a set of effects since we have our function `MobiusFoo#first` which accepts `vararg` parameters.

app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionInit.kt
```diff
@@ -0,0 +1,11 @@
+package org.simple.clinic.bloodsugar.unitselection
+
+import com.spotify.mobius.First
+import com.spotify.mobius.Init
+
+class BloodSugarUnitSelectionInit : Init<BloodSugarUnitSelectionModel, BloodSugarUnitSelectionEffect> {
+  override fun init(model: BloodSugarUnitSelectionModel): First<BloodSugarUnitSelectionModel, BloodSugarUnitSelectionEffect> {
+    val effect = mutableSetOf(PreFillBloodSugarUnitSelected(model.bloodSugarUnitPreference))
```
vinaysshenoy, line 8: This is also why we need to update the model when the preference changes because if the screen gets restored _after_ the user has changed the selection, but _before_ the "Done" button is clicked, the preference will be reset back to the initial one instead of the one that the user has selected.

app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionDialog.kt
```diff
@@ -8,29 +8,39 @@ import android.os.Bundle
...
 import org.simple.clinic.util.unsafeLazy
 import org.simple.clinic.widgets.ScreenDestroyed
-import org.simple.clinic.widgets.UiEvent
 import javax.inject.Inject
 
 class BloodSugarUnitSelectionDialog : AppCompatDialogFragment(), BloodSugarUnitSelectionUiActions {
 
   private var layout: View? = null
 
+  private var binding: DialogBloodsugarSelectionunitBinding? = null
```
vinaysshenoy, line 33: This will crash since we are not initializing the binding anywhere.

vinaysshenoy, line 33: We also need to clear this in the appropriate lifecycle method because the fragment will leak memory otherwise.

vinaysshenoy (changes requested):
Overall PR looks good, we need to make a few changes before it's ready to merge.

app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionDialog.kt
```diff
@@ -0,0 +1,36 @@
...
+      fragment.show(fragmentManager, FRAGMENT_TAG)
+    }
+  }
+  
+  @SuppressLint("InflateParams")
+  override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+    layout = LayoutInflater.from(context).inflate(R.layout.dialog_bloodsugar_selectionunit, null)
+    return AlertDialog.Builder(requireContext())
+        .setTitle(getString(R.string.blood_sugar_unit_selection_choose))
+        .setView(layout)
```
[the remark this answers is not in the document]
Honey14: We have a customised Ui, yes the single-list and done button.

app/src/test/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionEffectHandlerTest.kt
```diff
@@ -27,5 +32,17 @@ class BloodSugarUnitSelectionEffectHandlerTest {
...
+    verifyZeroInteractions(uiActions)
+  }
+
+  @Test
+  fun `when close dialog effect is received, then close the dialog`() {
+    // when
+    effectHandlerTestCase.dispatch(CloseDialog)
+
+    // then
+    verify(uiActions).closeDialog()
```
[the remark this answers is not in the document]
Honey14: It does but we have an event on click of done button, which saves the updated unit preference and then leads to closing the dialog explicitly.

app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionDialog.kt
vinaysshenoy: I don't see the `binding` variable being cleared anywhere. Can we clear this to avoid a memory leak?

vinaysshenoy (changes requested):
Overall PR looks good to merge. One thing we need to correct is that the binding variable needs to be cleared when the dialog is dismissed.

app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionDialog.kt
vinaysshenoy: Is `onStop()` the right place to do this? If we put the app in the background and then resume it, the binding should be reinitialized, but it won't. Would `onDismiss()` be a better place to do this?

Honey14: Can we do this in `onDetach` ? Dialog would be detached after this and binding would be reinitialized when the dialog is opened again. What do you think?

vinaysshenoy: If we do this in `onDetach()`, then we should be initializing the binding in the corresponding `onAttach()` as well.

Honey14: Understood

vinaysshenoy (approved):

PR #2122, opened 2020-11-19
Open `BloodSugarUnitSelectionDialog`

Honey14:
https://app.clubhouse.io/simpledotorg/story/1789/add-blood-sugar-unit-switching-button-in-bloodsugarentrysheet

### DIFF
--- a/app/src/test/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUpdateTest.kt
+++ b/app/src/test/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUpdateTest.kt
@@ -391,4 +391,17 @@ class BloodSugarEntryUpdateTest {
             )
         )
   }
+
+  @Test
+  fun `When the blood sugar unit button is clicked then show blood sugar unit selection dialog`() {
+    updateSpec
+        .given(defaultModel)
+        .whenEvent(BloodSugarReadingUnitButtonClicked)
+        .then(
+            assertThatNext(
+                hasNoModel(),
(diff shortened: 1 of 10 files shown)


app/src/test/java/org/simple/clinic/bloodsugar/entry/BloodSugarEntryUpdateTest.kt
```diff
@@ -391,4 +391,21 @@ class BloodSugarEntryUpdateTest {
             )
         )
   }
+
+  @Test
+  fun `When the blood sugar unit button is clicked then show blood sugar unit selection dialog`() {
+    val bloodSugarUnitPreference = BloodSugarUnitPreference.Mmol
+    defaultModel
```
vinaysshenoy, line 398: The result of this method call is not being used anywhere. Was this a mistake?

vinaysshenoy (changes requested):
Overall PR looks good, left one comment about an unused statement. Please take a look?

In addition, I think we need to make some changes to the `CHANGELOG`.

If we look at the "Features" section for this release, the relevant items are 

- Add blood sugar unit switching button in `BloodSugarEntrySheet`
- Add `BloodSugarUnitSelectionDialog`

The problem with this wording is that they are very developer-centric. The features and changes section are supposed to be user-centric, i.e., they should make sense to a nurse who is reading them.

A suggestion would be to replace these two with a single one:
`Add support for recording RBS, FBS, and PPBS blood sugars in two units: mmol/L or mg/dL`

vinaysshenoy:
Can you try running the app and triggering the dialog? The app crashes when the dialog is triggered in `onAttach()`.

vinaysshenoy (approved):

PR #2164, opened 2020-12-02
Update CHANGELOG

vinaysshenoy:


CHANGELOG.md
```diff
@@ -4,11 +4,17 @@
...
 - [In Progress: 11 Nov 2020] Migrate app screens to use ViewBinding
 - Bump appcompat -> 1.2.0
-- Move BP passport scan results handling to `HomeScreen`
 
 ### Changes
 - Disable the change language feature on devices running Lollipop (API level 21, 22)
 
+## 2020-12-01-7532
+### Internal
+- Move BP passport scan results handling to `HomeScreen`
```
msasikanth, line 13: The 2 changes are for the same thing, do we need to track both of them?

vinaysshenoy: Yeah, they're essentially the same. But I thought that I would include it as a different one since the internal change is the one that lead to the bug being fixed. But I can remove it as well.

msasikanth: Okay, let's leave it then 👍🏽

msasikanth (approved):

PR #2175, opened 2020-12-07
Add instrumentation for search queries

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/2020/add-instrumentation-for-patient-search


app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
```diff
@@ -3,7 +3,6 @@ package org.simple.clinic.patient.filter
 import info.debatty.java.stringsimilarity.CharacterInsDelInterface
```
msasikanth, line 3: I had no idea, this class was a thing 😮 . Is this ranking for results?

msasikanth (dismissed):
LGTM

app/src/main/java/org/simple/clinic/patient/filter/WeightedLevenshteinSearch.kt
vinaysshenoy: Yup

msasikanth:
> /instrumentedTests

FYI this triggers an completely different Action, so it won't update the required check status because it never ran.

vinaysshenoy:
> > /instrumentedTests
> 
> FYI this triggers an completely different Action, so it won't update the required check status because it never ran.

Can you explain this, please? I see the `qa_android_tests` job running. 🤔

vinaysshenoy:
https://github.com/simpledotorg/simple-android/actions/runs/409737643

msasikanth:
> > > /instrumentedTests
> > 
> > 
> > FYI this triggers an completely different Action, so it won't update the required check status because it never ran.
> 
> Can you explain this, please? I see the `qa_android_tests` job running. 🤔

If you look at the actions page there are 2 sets of CI workflows that are running. One is for this [PR](https://github.com/simpledotorg/simple-android/actions/runs/409728116) and [another](https://github.com/simpledotorg/simple-android/actions/runs/409737643) one is for the comment command. When we marked a action as required for a PR merge we expect it to be started when the PR is created. But because that test now triggeres when we leave a comment command it longer updates the status of the original action flow that is started when we openend a PR.

vinaysshenoy:
> > > > /instrumentedTests
> > > 
> > > 
> > > FYI this triggers an completely different Action, so it won't update the required check status because it never ran.
> > 
> > 
> > Can you explain this, please? I see the `qa_android_tests` job running. thinking
> 
> If you look at the actions page there are 2 sets of CI workflows that are running. One is for this [PR](https://github.com/simpledotorg/simple-android/actions/runs/409728116) and [another](https://github.com/simpledotorg/simple-android/actions/runs/409737643) one is for the comment command. When we marked a action as required for a PR merge we expect it to be started when the PR is created. But because that test now triggeres when we leave a comment command it longer updates the status of the original action flow that is started when we openend a PR.

Hmm, then the current implementation is not yet ready to use. We make to make running instrumented tests mandatory to merge a PR, but not necessarily want to run it every time a PR is updated. Until we figure out how to make that happen, I think we should revert to the older flow for now.

msasikanth (approved):

PR #2208, opened 2020-12-08
Add PR comment commands for running instrumented tests and rebasing

msasikanth:


.github/scripts/rebase_and_push.sh
```diff
@@ -0,0 +1,17 @@
+readonly GIT_HAS_NO_CHANGES=0 # 0 = clean, 1 = dirty
+
+git config user.name "$USER_NAME"
+git config user.email "$USER_EMAIL"
```
vinaysshenoy, line 4: Where are these shall variables getting set from?

msasikanth: From the secrets, we already have these setup for `commit_and_push` & `amend_and_push`

msasikanth: So if you look at actual GH Action job, we are passing these variables from there

.github/workflows/ci_check_pr_comment.yml
```diff
@@ -0,0 +1,40 @@
...
+on:
+  issue_comment:
+    types: [ created ]
+
+jobs:
+  # reactivecircus/android-emulator-runner@v2 requires MacOS to run on
+  # https://github.com/ReactiveCircus/android-emulator-runner
+  qa_android_tests:
+    # This job only runs for pull request comments and when the comment starts with `/instrumentedTests`
+    if: contains(github.event.comment.html_url, '/pull/') && startsWith(github.event.comment.body, '/instrumentedTests')
```
Honey14, line 13: So this will run when our url contains `/pull/` as well?

.github/workflows/ci_checks.yml
```diff
@@ -163,7 +142,7 @@ jobs:
 
   qa_purge_env:
     runs-on: [ubuntu-latest]
-    needs: [qa_lint, qa_unit_tests, mobius_migration_tests, qa_android_tests]
```
Honey14, line 166: We still want `qa-android-tests` as required, right? Why are we removing it from here?

Honey14 (commented):
Looks good overall. Cannot see the instrumented tests running as github actions?

.github/workflows/ci_check_pr_comment.yml
msasikanth: We are basically checking if the comment is for a pull request or not.

.github/workflows/ci_checks.yml
msasikanth: We are, if you look at `ci_checks_pr_comment.yml`, there is another `qa_purge_env` job that needs `qa_android_tests`. We removed it from here, because it's not longer part of this GH Action.

msasikanth:
I think this PR is in a weird state now, where the `qa_android_tests` have been removed from the `ci_checks` file in this PR and so it cannot figure out a way to run that test yet. 🤔 

@vinaysshenoy I am marking the android tests as not required until we merge this in and re-enable it once the PR comment commands are available on `master`.

Honey14 (approved):
Looks good.

PR #2219, opened 2020-12-09
Implement the sub-string based search

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/1967/implement-instant-search

I will be raising a different PR to return a list of patients that are assigned to the current facility and not the ones who have visited it.

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
@@ -1,12 +1,12 @@
 package org.simple.clinic.patient
 
 import android.os.Parcelable
+import androidx.paging.DataSource
 import androidx.room.Dao
 import androidx.room.DatabaseView
 import androidx.room.Embedded
 import androidx.room.Query
 import io.reactivex.Flowable
-import io.reactivex.Single
 import kotlinx.android.parcel.Parcelize
 import java.time.Instant
 import java.time.LocalDate
@@ -97,7 +97,7 @@ data class PatientSearchResult(
 
     @Embedded(prefix = "lastSeen_")
     val lastSeen: LastSeen?
-): Parcelable {
+) : Parcelable {
 
   override fun toString(): String {
     return "Name: $fullName, UUID: $uuid, Facility UUID: ${lastSeen?.lastSeenAtFacilityUuid}"
@@ -144,6 +144,27 @@ data class PatientSearchResult(
         ORDER BY P.fullName COLLATE NOCASE ASC LIMIT :limit
     """)
     fun searchByPhoneNumber(phoneNumber: String, limit: Int): List<PatientSearchResult>
+
+    @Suppress("AndroidUnresolvedRoomSqlReference")
+    @Query("""
+        SELECT * FROM 
+        PatientSearchResult searchResult
+        LEFT JOIN Patient P ON P.uuid = searchResult.uuid
+        WHERE phoneNumber LIKE '%' || :phoneNumber || '%' AND P.deletedAt IS NULL
+        ORDER BY P.fullName COLLATE NOCASE ASC 
+    """)
+    fun searchByPhoneNumberPaginated(phoneNumber: String): DataSource.Factory<Int, PatientSearchResult>
+
+    @Suppress("AndroidUnresolvedRoomSqlReference")
+    @Query("""
+    SELECT * FROM 
+    PatientSearchResult searchResult
+    LEFT JOIN Patient P ON P.uuid = searchResult.uuid
+    WHERE P.fullName LIKE '%' || :name || '%' AND P.deletedAt IS NULL
+    ORDER BY P.fullName COLLATE NOCASE ASC 
+    """
+    )
+    fun searchByNamePaginated(name: String): DataSource.Factory<Int, PatientSearchResult>
   }
 
   data class PatientNameAndId(val uuid: UUID, val fullName: String)
--- a/app/src/main/java/org/simple/clinic/patient/PatientSearchResultDataSource.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientSearchResultDataSource.kt
@@ -55,9 +55,25 @@ class PatientSearchResultDataSource(
     val loadParamsForDatabaseSource = LoadRangeParams(params.startPosition, params.loadSize)
     source.loadRange(loadParamsForDatabaseSource, object : LoadRangeCallback<PatientSearchResult>() {
       override fun onResult(patientSearchResult: MutableList<PatientSearchResult>) {
+        val patientSearchResults = partitionSearchResultsByFacility(patientSearchResult, currentFacility.get())
+        val searchResultsItems = SearchResultsItemType.from(patientSearchResults)
+        callback.onResult(searchResultsItems)
       }
     })
 
+  }
+
+  private fun partitionSearchResultsByFacility(
+      searchResults: List<PatientSearchResult>,
+      facility: Facility
+  ): PatientSearchResults {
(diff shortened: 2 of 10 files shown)


app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
```diff
@@ -144,6 +144,27 @@ data class PatientSearchResult(
...
+
+    @Suppress("AndroidUnresolvedRoomSqlReference")
+    @Query("""
+        SELECT * FROM 
+        PatientSearchResult searchResult
+        LEFT JOIN Patient P ON P.uuid = searchResult.uuid
+        WHERE phoneNumber LIKE '%' || :phoneNumber || '%' AND P.deletedAt IS NULL
+        ORDER BY P.fullName COLLATE NOCASE ASC 
+    """)
+    fun searchByPhoneNumberNew(phoneNumber: String): DataSource.Factory<Int, PatientSearchResult>
```
vinaysshenoy, line 156: Can we rename this to something like `searchByPhoneNumberPaginated`? Same for `searchByName` since we are going to have both the flows present for some time in the app.

app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
```diff
@@ -144,6 +144,27 @@ data class PatientSearchResult(
...
+
+    @Suppress("AndroidUnresolvedRoomSqlReference")
+    @Query("""
+    SELECT * FROM 
+    PatientSearchResult searchResult
+    LEFT JOIN Patient P ON P.uuid = searchResult.uuid
+    WHERE P.fullName LIKE '%' || :name || '%' AND P.deletedAt IS NULL
+    ORDER BY P.fullName COLLATE NOCASE ASC 
+  """
+    )
```
vinaysshenoy, line 166: Can you reformat this to retain the consistency with the rest of the file, please?

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
```diff
@@ -73,6 +73,12 @@ class ShortCodeSearchResultScreen(
 
   private var viewBinding: ScreenShortcodeSearchResultBinding? = null
 
+  private val toolBar
+    get() = viewBinding!!.toolBar
+
+  private val patientSearchView
+    get() = viewBinding!!.patientSearchView
+
```
vinaysshenoy: Can you squash this commit with the older commit where this was deleted?

app/src/main/java/org/simple/clinic/patient/PatientSearchResultDataSource.kt
```diff
@@ -55,11 +55,27 @@ class PatientSearchResultDataSource(
     val loadParamsForDatabaseSource = LoadRangeParams(params.startPosition, params.loadSize)
     source.loadRange(loadParamsForDatabaseSource, object : LoadRangeCallback<PatientSearchResult>() {
       override fun onResult(patientSearchResult: MutableList<PatientSearchResult>) {
+        val patientSearchResults = partitionSearchResultsByFacility(patientSearchResult, currentFacility.get())
+        val searchResultsItems = SearchResultsItemType.from(patientSearchResults)
```
vinaysshenoy, line 61: This will not work since `SearchResultsItemType#from` will only work correctly if we are loading the entire set of patient results at once. Here when it's paginated, since we are loading data in chunks, using this factory method will return the wrong set of results.

Consider the following scenario, where we have patients 1..10, the first five of which are in the current facility and the latter five are not.

Regardless of whether we are loading them in a paginated fashion or not, the complete set of results should look like this:

```
currentFacilityHeader
patient1
patient2
patient3
patient4
patient5
NotInCurrentFacilityHeader
patient6
patient7
patient8
patient9
patient10
```

Consider a scenario where we are loading search results in chunks of 3 patients at a time. What will happen with the paginated data source in the current implementation is:

```
currentFacilityHeader
patient1
patient2
patient3    ---> Page 1
currentFacilityHeader
patient4
patient5
NotInCurrentFacilityHeader
patient6    ---> Page 2
currentFacilityHeader
NoPatientsInCurrentFacility
NotInCurrentFacilityHeader
patient7
patient8
patient9    --> Page 3
currentFacilityHeader
NoPatientsInCurrentFacility
NotInCurrentFacilityHeader
patient10    --> Page 4
```

This will happen because the factory method in `SearchResultsItemType` has no concept of data being loaded in chunks and assumes that the set it receives is the complete set.

app/src/main/java/org/simple/clinic/patient/PatientSearchResultDataSource.kt
```diff
@@ -55,11 +55,27 @@ class PatientSearchResultDataSource(
     val loadParamsForDatabaseSource = LoadRangeParams(params.startPosition, params.loadSize)
     source.loadRange(loadParamsForDatabaseSource, object : LoadRangeCallback<PatientSearchResult>() {
```
vinaysshenoy, line 58: We cannot call this with the params directly. This is because we have added additional items to the final set of items (the facility header, etc) which are unknown to the data source created for Room. We will need to account for these additional items when passing a set of params to load the next page.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/patient/PatientSearchResultDataSource.kt
msasikanth: Wait, I thought it would retain data as it loads 🤔

msasikanth: Yep, as far as I can tell it would retain the loaded data and make it part of the paged list. If it's not present in the paged list it would try to load. So this might work?

vinaysshenoy: The problem is that the chunk of data that is loaded in the `loadRange` callback is being passed to the factory method, which assumes the list it receives the entire list of data so the additional headers that we compute and add to the subset again.

vinaysshenoy: > Yep, as far as I can tell it would retain the loaded data and make it part of the paged list. If it's not present in the paged list it would try to load. So this might work?

It does do this, but this is done after the `loadRange` callback will be invoked, which will then append to the current list.

msasikanth: Ah okay. Without keeping a reference of the list inside the data source, this might be difficult to implement here then? Since we convert the data source to a paged list at the screen, is possible to do this there? or is there a better approach? 🤔

vinaysshenoy: I'm considering writing a `DataSource` base class to abstract away this kind of logic. Maybe we can think about this problem separately and come back tomorrow to figure out the best approach to take here.

janhavisinghh: @vinaysshenoy @msasikanth let's get on a call to discuss the approach we are planning to follow here?

janhavisinghh:
After some discussions on adding pagination to instant search, we realised that it is a lot more complicated than initially anticipated. We have decided to leave out pagination for the initial release of instant search. I am closing this PR for now and will be implementing the rest of the things on a new branch. Check out [this](https://app.clubhouse.io/simpledotorg/story/1967/implement-instant-search) clubhouse story for more information.

PR #2220, opened 2020-12-10
Fix app going back to home screen after scanning BP passport

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2063/app-shows-home-screen-after-scanning-a-bp-passport

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/HomeScreenEvent.kt
+++ b/app/src/main/java/org/simple/clinic/home/HomeScreenEvent.kt
@@ -1,14 +1,11 @@
 package org.simple.clinic.home
 
 import org.simple.clinic.facility.Facility
-import org.simple.clinic.patient.Patient
 import org.simple.clinic.patient.businessid.Identifier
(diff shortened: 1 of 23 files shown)


app/src/main/java/org/simple/clinic/home/HomeScreenEvent.kt
```diff
@@ -5,6 +5,7 @@ import org.simple.clinic.patient.Patient
 import org.simple.clinic.patient.businessid.Identifier
```
Honey14, line 4: When we don't find the patient we open the search screen. Can we update the commit message to reflect that?

Honey14 (changes requested):

app/src/main/java/org/simple/clinic/home/HomeScreenEvent.kt
msasikanth: Oh wait, I thought I pushed that change. My bad will fix it 😅

Honey14 (approved):
Looks good

PR #2222, opened 2020-12-10
Implement instant search screen - 1

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2029/implement-instant-search-screen

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -2413,7 +2413,7 @@ class PatientRepositoryAndroidTest {
     // when
     fun patientsInFacility(facility: Facility): List<PatientSearchResult> {
       return patientRepository
-          .allPatientsInFacility(facility)
+          .allPatientsInFacility_Old(facility)
           .blockingFirst()
     }
 
@@ -2482,7 +2482,7 @@ class PatientRepositoryAndroidTest {
     // when
     fun patientsInFacility(facility: Facility): List<PatientSearchResult> {
       return patientRepository
-          .allPatientsInFacility(facility)
+          .allPatientsInFacility_Old(facility)
           .blockingFirst()
     }
 
@@ -2594,7 +2594,7 @@ class PatientRepositoryAndroidTest {
     // when
     fun patientsInFacility(facility: Facility): List<PatientSearchResult> {
       return patientRepository
-          .allPatientsInFacility(facility)
+          .allPatientsInFacility_Old(facility)
           .blockingFirst()
     }
 
@@ -3163,7 +3163,7 @@ class PatientRepositoryAndroidTest {
     )
 
     //when
-    val allPatientsInFacility = patientRepository.allPatientsInFacility(currentFacility).blockingFirst()
+    val allPatientsInFacility = patientRepository.allPatientsInFacility_Old(currentFacility).blockingFirst()
 
     //then
     assertThat(allPatientsInFacility).hasSize(2)
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -3878,24 +3878,20 @@ class PatientRepositoryAndroidTest {
         phoneNumber: String?,
         assignedFacilityId: UUID?
     ) {
-      val patientProfile = testData
-          .patientProfile(patientUuid = patientUuid, generatePhoneNumber = !phoneNumber.isNullOrBlank(), patientAssignedFacilityId = assignedFacilityId)
-          .let { patientProfile ->
-            patientProfile.copy(
-                patient = patientProfile.patient.copy(fullName = "Patient", assignedFacilityId = assignedFacilityId),
-                phoneNumbers = patientProfile
-                    .phoneNumbers
-                    .take(1)
-                    .map { it.copy(number = phoneNumber!!) }
-            )
-          }
+      val patientProfile = TestData
+          .patientProfile(
+              patientUuid = patientUuid,
+              generatePhoneNumber = false,
+              patientPhoneNumber = phoneNumber,
+              patientAssignedFacilityId = assignedFacilityId
+          )
 
       patientRepository.save(listOf(patientProfile)).blockingAwait()
     }
 
     // given
-    val currentFacility = testData.facility(uuid = UUID.fromString("c0056c11-105d-4079-80fa-c745128f7fc5"), name = "CHC Bucho")
-    val facility2 = testData.facility(uuid = UUID.fromString("532264e6-b358-445e-8b48-c6253677db24"), name = "Test")
+    val currentFacility = TestData.facility(uuid = UUID.fromString("c0056c11-105d-4079-80fa-c745128f7fc5"), name = "CHC Bucho")
+    val facility2 = TestData.facility(uuid = UUID.fromString("532264e6-b358-445e-8b48-c6253677db24"), name = "Test")
     database.facilityDao().save(listOf(currentFacility, facility2))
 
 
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
```diff
@@ -95,7 +95,9 @@ data class PatientSearchResult(
     val phoneUpdatedAt: Instant?,
 
     @Embedded(prefix = "lastSeen_")
-    val lastSeen: LastSeen?
+    val lastSeen: LastSeen?,
+
+    val isAtAssignedFacility: Boolean?
```
vinaysshenoy, line 100: Is it necessary to add this as a property in the search result view? I thought we would only need it in the SQL query to sort the results, right?

vinaysshenoy, line 100: It seems like this is nullable only because we cannot add this in the view, right? If this is the case, we should not be adding this here since it should never be null from a modelling perspective (a patient can either be assigned to the current facility or not).

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -3980,4 +3980,78 @@ class PatientRepositoryAndroidTest {
...
+        patientUuid: UUID,
+        patientName: String,
+        assignedFacilityId: UUID?
+    ) {
+      val patientProfile = testData
+          .patientProfile(patientUuid = patientUuid, patientName = patientName, patientAssignedFacilityId = assignedFacilityId)
+          .let { patientProfile ->
+            patientProfile.copy(
+                patient = patientProfile.patient.copy(fullName = patientName, assignedFacilityId = assignedFacilityId),
+            )
```
vinaysshenoy, line 3996: Isn't this copy method redundant? We are already setting these properties when constructing the patient profile.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -3980,4 +3980,78 @@ class PatientRepositoryAndroidTest {
...
+          }
+
+      patientRepository.save(listOf(patientProfile)).blockingAwait()
+    }
+
+    fun searchResults(facility: Facility): List<String> {
+      return patientRepository
+          .allPatientsInFacility2(facility)
```
vinaysshenoy, line 4004: Can we think of a different name than `allPatientsInFacility2`?

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -3980,4 +3980,78 @@ class PatientRepositoryAndroidTest {
...
+          .map { it.fullName }
+    }
+
+    // given
+    val currentFacility = TestData.facility(
+        uuid = UUID.fromString("c0056c11-105d-4079-80fa-c745128f7fc5"),
+        name = "CHC Bucho"
+    )
+    val facility2 = TestData.facility(
```
vinaysshenoy, line 4013: Since we have one facility named `currentFacility`, can we name the other to `otherFacility`? This also removes confusion when reading `patientWithCurrentFacilityAsAssignedFacility2` because I initially thought this patient was assigned to `faciltiy2` instead of `patient 2 assigned to current facility`.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -646,6 +646,12 @@ class PatientRepository @Inject constructor(
         .toObservable()
   }
 
+  fun allPatientsInFacility2(facility: Facility): List<PatientSearchResult> {
+    return database
+        .patientSearchDao()
+        .searchInFacilityAndSortByName2(facility.uuid, PatientStatus.Active)
```
vinaysshenoy, line 652: Can we think of a different name than `searchInFacilityAndSortByName2`?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
msasikanth: It would return the sorted results but we still need to decide where to place the header right? currently, there is no indicator to decide headers in between the assigned and non-assigned facility patients. Do you think there is some other way we can achieve this?

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
msasikanth: My plan is to eventually remove the original `allPatientsInFacility` and rename `allPatientsInFacility2` to `allPatientsInFacility`. But I can rename these to `allPatientsInFacility_Old` & `allPatientsInFacility` if this is a problem. WDYT?

app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
msasikanth: Hmm, but we cannot do the assigned facility check when creating the view since it's dynamic. So that's why I am setting these values in the query. Is there a way we can do this?

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
msasikanth: Ah right, I copied the above implementation since it's mostly the same. Will fix it 👍🏽

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
msasikanth: Good point will do that 👍🏽

app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
msasikanth: This is how I am planning to add the headers.

Screenshot 2020-12-28 150559

vinaysshenoy: Ah, I see. I was thinking we could add the patient assigned facility ID separately to the `PatientSearchResult` type.

vinaysshenoy: Does this help?

https://github.com/simpledotorg/simple-android/pull/2222#discussion_r549331706

We can then leave the boolean flag only at the SQL level and then add the assigned facility ID in the search result as a way to test this condition.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
vinaysshenoy: Ah, if we are planning to remove it later, then it doesn't really matter.

app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
msasikanth: I am unsure how that's going to work out. If we pass the `assignedFacilityId` via `PatientSearchResult`, we will still need to have a conditional to verify whether or not the current facility id matches the assigned facility id and then place the headers. I thought having a boolean value would better in this scenario since the data is already sorted at SQL level.

vinaysshenoy: The issue that I am talking about here is irrelevant to whether we are sorting in SQL or not. Sorting and grouping here are runtime operations since we will not know whether a patient is assigned to the given facility or not until we run the query runtime. My problem is adding the `isAssignedToCurrentFacility` boolean property to the `PatientSearchResult` type.

When it comes to a database view, it cannot have runtime properties since the SQL statement is a part of the view declaration. So something like "is the patient assigned to the current facility" should not be tracked in the view since the view itself cannot tell what the current facility is. However, something like "what is the patient's assigned facility" is something that can be tracked since it is a static property tied to the patient.

What we _should_ be doing here is tracking each patient's assigned facility in the search results and then doing a runtime check to split them into groups.

msasikanth: My only concern is it doesn't matter if we are sorting them at SQL level anymore since we will be doing the sorting and partition when generating the item view types list. Is that right?

vinaysshenoy: Yeah, that's a redundant sorting for now. But it will still make the runtime sort operation more efficient since the source results are already sorted.

In any case, once we eventually move this class to a paginated fashion, the sorting at the SQL layer will be needed again, so 🤷🏼

vinaysshenoy (approved):
LGTM

PR #2223, opened 2020-12-10
Add BP passport sheet

Honey14:
https://app.clubhouse.io/simpledotorg/story/1977/show-bp-passport-sheet-in-instant-search-screen


app/src/main/res/values/styles_v2.xml
```diff
@@ -84,6 +84,12 @@
     <item name="android:fontFamily">sans-serif-medium</item>
   </style>
 
+  <style name="Clinic.V2.TextAppearance.Black">
```
msasikanth, line 87: We already have a text style for this. If you look at the Figma file, they are using `h6` with `grey0` text color. We have `Clinic.V2.TextAppearance.H6.Grey0`

app/src/main/res/values/strings.xml
```diff
@@ -772,4 +772,8 @@
   <string name="blood_sugar_unit_selection_choose">Choose</string>
   <string name="blood_sugar_unit_selection_done">DONE</string>
 
-</resources>
+  <!-- BP passport sheet -->
+  <string name="sheet_bp_passport_register_new_patient">Register New Patient</string>
+  <string name="sheet_bp_passport_add_to_existing_patient">Add to existing Patient</string>
```
msasikanth, line 777: Can we check the word casing here, please?  Only the first word should be captalised.

app/src/main/res/layout/sheet_bp_passport.xml
```diff
@@ -0,0 +1,53 @@
...
+
+  <com.google.android.material.button.MaterialButton
+    android:id="@+id/registerNewPatientButton"
+    style="?attr/materialButtonStyle"
+    android:layout_width="@dimen/spacing_0"
+    android:layout_height="wrap_content"
+    android:layout_marginTop="@dimen/spacing_16"
+    android:layout_marginBottom="@dimen/spacing_16"
+    android:text="@string/sheet_bp_passport_register_new_patient"
+    android:textAllCaps="true"
```
msasikanth, line 31: Can we use the text appearance here, please? It will be `Clinic.V2.TextAppearance.Button1`

app/src/main/res/layout/sheet_bp_passport.xml
```diff
@@ -0,0 +1,53 @@
...
+
+  <com.google.android.material.button.MaterialButton
+    android:id="@+id/addToExistingPatientButton"
+    style="@style/Clinic.V2.MaterialButton.Blue2"
+    android:layout_width="@dimen/spacing_0"
+    android:layout_height="wrap_content"
+    android:layout_marginBottom="@dimen/spacing_16"
+    android:text="@string/sheet_bp_passport_add_to_existing_patient"
+    android:textAllCaps="true"
+    app:drawableTint="@color/white100"
```
msasikanth, line 47: We don't have an icon, so we don't need to add `drawableTint` here.

app/src/main/res/layout/sheet_bp_passport.xml
```diff
@@ -0,0 +1,53 @@
...
+    app:layout_constraintTop_toBottomOf="@+id/bpPassportNumberTextview" />
+
+  <com.google.android.material.button.MaterialButton
+    android:id="@+id/addToExistingPatientButton"
+    style="@style/Clinic.V2.MaterialButton.Blue2"
+    android:layout_width="@dimen/spacing_0"
+    android:layout_height="wrap_content"
+    android:layout_marginBottom="@dimen/spacing_16"
+    android:text="@string/sheet_bp_passport_add_to_existing_patient"
+    android:textAllCaps="true"
```
msasikanth, line 46: Same comment as above, let's use text appearance for this. It will be `Clinic.V2.TextAppearance.Button1`

app/src/main/res/layout/sheet_bp_passport.xml
```diff
@@ -0,0 +1,53 @@
...
+  <com.google.android.material.button.MaterialButton
+    android:id="@+id/registerNewPatientButton"
+    style="?attr/materialButtonStyle"
+    android:layout_width="@dimen/spacing_0"
+    android:layout_height="wrap_content"
+    android:layout_marginTop="@dimen/spacing_16"
+    android:layout_marginBottom="@dimen/spacing_16"
+    android:text="@string/sheet_bp_passport_register_new_patient"
+    android:textAllCaps="true"
+    app:drawableTint="@color/white100"
```
msasikanth, line 32: The icon is already white, is there a reason we are adding the drawable tint here?

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import android.os.Bundle
+import org.simple.clinic.R
+import org.simple.clinic.widgets.BottomSheetActivity
+
+class BpPassportSheet : BottomSheetActivity() {
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.sheet_bp_passport)
```
msasikanth: We should be using Viewbinding.

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportEffect.kt
```diff
@@ -0,0 +1,5 @@
+package org.simple.clinic.bp.assignbppassport
+
+sealed class BpPassportEffect
+
+data class OpenPatientEntryScreen(val bpPassportNumber: Int) : BpPassportEffect()
```
msasikanth, line 5: The BP passport number we use is not an int. It's an `Identifier`, can you take a look at the existing register new patient functionality from the search. Looks like we don't even have to use activity results to open the screen, but instead use the `OngoingLoginEntry`. So, please take a look at the `PatientSearchResultsScreen` Mobius implementation to see how the new patient registration works.

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportEvent.kt
```diff
@@ -1,3 +1,5 @@
 package org.simple.clinic.bp.assignbppassport
 
 sealed class BpPassportEvent
+
+data class RegisterNewPatientClicked(val bpPassportNumber: Int) : BpPassportEvent()
```
msasikanth, line 5: Instead of passing the BP passport identifier with the event. We can save it in the model during screen init, since we will have to BP passport identifier when opening the screen. WDYT?

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
```diff
@@ -1,13 +1,122 @@
...
+  private fun addToExistingPatientClicks(): Observable<BpPassportEvent> {
+    return addToExistingPatientButton
+        .clicks()
+        .map { AddToExistingPatientClicked }
+  }
+
+  private fun registerNewPatientClicks(): Observable<BpPassportEvent> {
+    return registerNewPatientButton
+        .clicks()
+        .map { RegisterNewPatientClicked(0) }
```
msasikanth: We should not be passing a "dummy" data like this. If we need to pass some information from the screen, we can use intent extras for activities and screen args for the screens.

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
```diff
@@ -1,13 +1,122 @@
...
-    setContentView(R.layout.sheet_bp_passport)
+    binding = SheetBpPassportBinding.inflate(layoutInflater)
+    setContentView(binding.root)
+
+    delegate.onRestoreInstanceState(savedInstanceState)
+  }
+
+  override fun openPatientEntryScreen(bpPassportNumber: Int) {
+    val intent = Intent()
+    intent.putExtra(KEY_BP_PASSPORT_NUMBER, bpPassportNumber)
```
msasikanth, line 97: Let's go with the existing search screen implementation, where we need to save the `OngoingLoginEntry` and then we can open the patient entry screen without needing to pass the BP passport identifier.

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
```diff
@@ -21,6 +22,15 @@ class BpPassportSheet : BottomSheetActivity(), BpPassportUiActions {
 
   companion object {
     private const val KEY_BP_PASSPORT_NUMBER = "bpPassportNumber"
+
+    fun intent(
+        context: Context,
+        bpPassportNumber: Int
```
msasikanth, line 28: We should be using `Identifier` here for the BP passport.

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
```diff
@@ -1,13 +1,122 @@
...
+
+  private val events: Observable<BpPassportEvent> by unsafeLazy {
+    Observable
+        .merge(
+            registerNewPatientClicks(),
+            addToExistingPatientClicks()
+        )
+  }
+
+  private fun addToExistingPatientClicks(): Observable<BpPassportEvent> {
```
msasikanth, line 60: Can we re-arrange these functions so that these are not between variables?

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
```diff
@@ -1,13 +1,122 @@
...
+    return addToExistingPatientButton
+        .clicks()
+        .map { AddToExistingPatientClicked }
+  }
+
+  private fun registerNewPatientClicks(): Observable<BpPassportEvent> {
```
msasikanth, line 66: Same comment as above. Can we re-arrange these functions so that these are not between variables?

msasikanth (changes requested):
Looks good overall. We need to make a few changes to the approach and we are good to merge this in.

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
Honey14: ViewBinding is used in the later commits

app/src/main/res/layout/sheet_bp_passport.xml
```diff
@@ -0,0 +1,53 @@
...
+
+  <com.google.android.material.button.MaterialButton
+    android:id="@+id/registerNewPatientButton"
+    style="?attr/materialButtonStyle"
+    android:layout_width="@dimen/spacing_0"
+    android:layout_height="wrap_content"
+    android:layout_marginTop="@dimen/spacing_16"
+    android:layout_marginBottom="@dimen/spacing_16"
+    android:text="@string/sheet_bp_passport_register_new_patient"
+    android:textAllCaps="true"
```
msasikanth, line 31: Looks like you missed removing `textAllCaps`. can you remove it, please?

app/src/main/res/layout/sheet_bp_passport.xml
```diff
@@ -0,0 +1,53 @@
...
+    app:layout_constraintTop_toBottomOf="@+id/bpPassportNumberTextview" />
+
+  <com.google.android.material.button.MaterialButton
+    android:id="@+id/addToExistingPatientButton"
+    style="@style/Clinic.V2.MaterialButton.Blue2"
+    android:layout_width="@dimen/spacing_0"
+    android:layout_height="wrap_content"
+    android:layout_marginBottom="@dimen/spacing_16"
+    android:text="@string/sheet_bp_passport_add_to_existing_patient"
+    android:textAllCaps="true"
```
msasikanth, line 46: Same comment as above

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportEffectHandler.kt
```diff
@@ -1,23 +1,46 @@
...
         .addTransformer(SaveNewOngoingPatientEntry::class.java, saveNewPatientEntry())
+        .addTransformer(OpenPatientEntryScreen::class.java, openPatientEntryScreen())
         .build()
   }
 
+  private fun openPatientEntryScreen(): ObservableTransformer<OpenPatientEntryScreen, BpPassportEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .map { currentFacility.get() }
```
msasikanth, line 38: Here we are doing 2 tasks, getting current facility and opening the patient entry screen. Let's separate them out into 2 separate effects, WDYT?

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
```diff
@@ -1,13 +1,128 @@
...
+            registerNewPatientClicks(),
+            addToExistingPatientClicks()
+        )
+  }
+
+  private val delegate by unsafeLazy {
+
+    MobiusDelegate.forActivity(
+        events = events,
+        defaultModel = BpPassportModel(),
```
msasikanth: We cannot create a `BpPassportModel` like this, it has a constructor param here as well. So let's try to use the factory method here.

msasikanth (changes requested):
LGTM left a few comments PTAL

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
Honey14: This is updated in the prefilling commit. Please check

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportEffectHandler.kt
Honey14: Makes sense. Will update it to have two separate effects.

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
msasikanth: Yes, but this commit will have a compilation error won't it? Because we are not providing a constructor param for the `BpPassportModel`

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
```diff
@@ -1,13 +1,128 @@
...
-    setContentView(R.layout.sheet_bp_passport)
+    binding = SheetBpPassportBinding.inflate(layoutInflater)
+    setContentView(binding.root)
+
+    delegate.onRestoreInstanceState(savedInstanceState)
+  }
+
+  override fun openPatientEntryScreen(facility: Facility) {
+    val intent = Intent()
+    intent.putExtra(FACILITY_CHANGE, facility.name)
```
msasikanth, line 92: Instead of passing the facility directly, should we wrap it in a result class? WDYT? You can take a look at `ScanResult`, which would allow us to handle multiple activity results from this sheet.

msasikanth (changes requested):

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
msasikanth: Not necessary in this PR, you can make this change in a different one.

msasikanth (approved):

PR #2224, opened 2020-12-11
Prepare the short code search results screen for mobius migration

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/1955/migrate-short-code-search-results-screen-to-mobius

### DIFF
--- a/app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
+++ b/app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
@@ -1,6 +1,7 @@
 package org.simple.clinic.shortcodesearchresult
 
 import android.content.Context
+import android.os.Parcelable
 import android.util.AttributeSet
 import android.view.View
 import android.widget.RelativeLayout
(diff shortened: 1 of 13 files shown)


app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
```diff
@@ -64,7 +67,15 @@ class ShortCodeSearchResultScreen(context: Context, attributes: AttributeSet) :
...
   private val adapter = ItemAdapter(SearchResultsItemType.DiffCallback())
 
-  private lateinit var disposable: Disposable
+  private val events by unsafeLazy {
+    Observable
+        .merge(
+            searchPatientClicks(),
+            patientItemClicks()
+        )
+        .compose(ReportAnalyticsEvents())
```
msasikanth, line 81: Doesn't the controller already have reporting at this point?

msasikanth (commented):
LGTM. Left one comment PTAL

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
vinaysshenoy: There is no controller in this intermediate architecture. We never ended up reporting these events to analytics at all, looks like.

msasikanth: Ah okay 👍🏽

msasikanth (approved):

PR #2229, opened 2020-12-15
Migrate app to use View Binding instead of Kotlin synthetic view accessors - 5

msasikanth:

### DIFF
--- a/app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchView.kt
+++ b/app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchView.kt
@@ -4,13 +4,14 @@ import android.annotation.SuppressLint
 import android.content.Context
 import android.os.Parcelable
 import android.util.AttributeSet
+import android.view.LayoutInflater
 import android.widget.RelativeLayout
 import androidx.recyclerview.widget.LinearLayoutManager
 import com.jakewharton.rxbinding3.view.detaches
 import io.reactivex.Observable
 import io.reactivex.rxkotlin.ofType
-import kotlinx.android.synthetic.main.patient_search_view.view.*
 import org.simple.clinic.R
+import org.simple.clinic.databinding.PatientSearchViewBinding
 import org.simple.clinic.di.injector
 import org.simple.clinic.mobius.DeferredEventSource
 import org.simple.clinic.mobius.MobiusDelegate
@@ -33,6 +34,26 @@ class PatientSearchView(context: Context, attrs: AttributeSet) : RelativeLayout(
 
   @Inject
   lateinit var effectHandler: SearchResultsEffectHandler
+
+  private var binding: PatientSearchViewBinding? = null
+
+  private val resultsRecyclerView
+    get() = binding!!.resultsRecyclerView
+
+  private val newPatientButton
+    get() = binding!!.newPatientButton
+
+  private val loader
+    get() = binding!!.loader
+
+  private val newPatientContainer
+    get() = binding!!.newPatientContainer
+
+  private val emptyStateView
+    get() = binding!!.emptyStateView
+
+  private val newPatientRationaleTextView
+    get() = binding!!.newPatientRationaleTextView
 
   var registerNewPatientClicked: RegisterNewPatientClicked? = null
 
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchView.kt
```diff
@@ -59,11 +80,13 @@ class PatientSearchView(context: Context, attrs: AttributeSet) : RelativeLayout(
   @SuppressLint("CheckResult")
   override fun onFinishInflate() {
     super.onFinishInflate()
-    inflate(context, R.layout.patient_search_view, this)
```
Honey14, line 62: How would the layout file be attached if we're removing it from here?

app/src/main/res/layout/activity_setup.xml
```diff
@@ -1,10 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
-<fragment xmlns:android="http://schemas.android.com/apk/res/android"
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
Honey14, line 2: Why do we need a framelayout?

app/src/main/java/org/simple/clinic/signature/SignatureActivity.kt
```diff
@@ -49,9 +48,25 @@ class SignatureActivity : AppCompatActivity(), SignatureUiActions {
...
+    get() = binding.drawSignatureFrame
+
+  private val clearSignature
+    get() = binding.clearSignature
+
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
-    setContentView(R.layout.activity_signature)
+
+    binding = ActivitySignatureBinding.inflate(layoutInflater)
```
Honey14, line 68: Can we nullify binding when the screen is destroyed?

Honey14 (changes requested):
Looks good overall. App is crashing on `PatientsTabScreen` and left a few comments. Could you please take a look?

app/src/main/java/org/simple/clinic/signature/SignatureActivity.kt
msasikanth: We don't need to do that for Activities. We only need to do that for Fragments and Views.

app/src/main/res/layout/activity_setup.xml
msasikanth: View binding will not generate a reference for `fragment` when it's root layout. So we are just wrapping it in `FrameLayout` to generate that.

app/src/main/java/org/simple/clinic/searchresultsview/PatientSearchView.kt
msasikanth: We are still inflating the layout. If you look at a couple of lines below. We are using this `PatientSearchViewBinding.inflate(layoutInflater, this)` to inflate the layout.

app/src/main/java/org/simple/clinic/signature/SignatureActivity.kt
Honey14: I see. Why is that activities don't need?

msasikanth: You can refer to the official [docs](https://developer.android.com/topic/libraries/view-binding). But essentially Fragments can outlive their views, so to avoid any memory leaks we clear the binding.

msasikanth:
> Looks good overall. App is crashing on `PatientsTabScreen` and left a few comments. Could you please take a look?

Dropped the patient tab screen migration for now. There seem to be some lifecycle issues with it. Please try out the app again to see if there are any other issues with these screens.

Honey14 (changes requested):
Looks good. App is crashing on `OverDueScreen`. PTAL

msasikanth:
> `OverDueScreen`

Oh yes, we just need to drop that commit. It's related to the FragmentStateAdapter we use for ViewPager2, we will migrate these once the new navigation framework is implemented.

Honey14 (approved):
looks good

PR #2231, opened 2020-12-16
Report error when UUID generation fails when scanning a BP passport

msasikanth:


app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -43,6 +50,7 @@ class ScanSimpleIdUpdate : Update<ScanSimpleIdModel, ScanSimpleIdEvent, ScanSimp
       val identifier = Identifier(bpPassportCode.toString(), BpPassport)
       next(model = model.searching(), SearchPatientByIdentifier(identifier))
     } catch (e: IllegalArgumentException) {
+      crashReporter.report(e)
```
vinaysshenoy, line 53: This should ideally be reported as an effect because it makes the `update()` function impure. I'll approve it for now since it's urgent, but we should fix it in a later PR.

vinaysshenoy (approved):

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
msasikanth: Yes, this seemed like the easiest approach to save some time. Will do this correctly after the hotfix.

PR #2232, opened 2020-12-17
Migrate app to use View Binding instead of Kotlin synthetic view accessors - 6

msasikanth:


msasikanth:
The current `CHANGELOG` in `master` already captures the latest view binding migration change. So we don't need to update the `CHANGELOG` again here.

janhavisinghh (approved):
lgtm

PR #2233, opened 2020-12-17
Migrate app to use View Binding instead of Kotlin synthetic view accessors - 7

msasikanth:


app/src/main/java/org/simple/clinic/widgets/BindingPagingItemAdapter.kt
```diff
@@ -0,0 +1,50 @@
...
+import androidx.viewbinding.ViewBinding
+import io.reactivex.Observable
+import io.reactivex.subjects.PublishSubject
+import io.reactivex.subjects.Subject
+import org.simple.clinic.widgets.recyclerview.BindingViewHolder
+
+// TODO (SM) : Rename to PagingItemAdapter once migration to ViewBinding is finished
+open class BindingPagingItemAdapter<I : BindingPagingItemAdapter.Item<E>, E>(
+    diffCallback: DiffUtil.ItemCallback<I>,
+    private val bindings: Map<Int, (layoutInflater: LayoutInflater, parent: ViewGroup) -> ViewBinding>
```
vinaysshenoy: What do you think of extracting the type `(layoutInflater: LayoutInflater, parent: ViewGroup) -> ViewBinding` to a type alias for better readability?

vinaysshenoy: Can we add a `@LayoutRes` annotation to the key parameter type declaration, please?

app/src/main/java/org/simple/clinic/widgets/BindingPagingItemAdapter.kt
```diff
@@ -0,0 +1,50 @@
...
+  private val eventSubject = PublishSubject.create<E>()
+  val itemEvents: Observable<E> = eventSubject.hide()
+
+  override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): BindingViewHolder {
+    val layoutInflater = LayoutInflater.from(parent.context)
+
+    val binding = bindings
+        .filter { it.key == viewType }
+        .values
+        .first()
```
vinaysshenoy: Since `bindings` is a map, can't we just use `bindings.getValue(viewType)` here?

app/src/main/java/org/simple/clinic/widgets/BindingPagingItemAdapter.kt
```diff
@@ -0,0 +1,50 @@
...
+    return BindingViewHolder(binding = binding.invoke(layoutInflater, parent))
+  }
+
+  override fun onBindViewHolder(holder: BindingViewHolder, position: Int) {
+    getItem(position)?.render(holder, eventSubject)
+  }
+
+  override fun getItemViewType(position: Int): Int {
+    val item = getItem(position)
+    return item?.layoutResId() ?: throw NullPointerException("Failed to get item")
```
vinaysshenoy: Maybe we could also add the position to the exception message.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/widgets/BindingPagingItemAdapter.kt
msasikanth: We cannot place that inside the `Map` definition.

msasikanth: Yeah, good point. Will do that

app/src/main/java/org/simple/clinic/widgets/BindingPagingItemAdapter.kt
msasikanth: Yep, I forgot about that extension 😅

app/src/main/java/org/simple/clinic/widgets/BindingPagingItemAdapter.kt
msasikanth: This is the error. `This annotation is not applicable to target 'type usage'`

vinaysshenoy: Ah cool.

vinaysshenoy (approved):

PR #2236, opened 2020-12-18
Migrate app to use View Binding instead of Kotlin synthetic view accessors - 8

Honey14:


Honey14:
Assigning it to @vinaysshenoy as @msasikanth is on leave.

vinaysshenoy (approved):

PR #2240, opened 2020-12-21
Migrate app to use View Binding instead of Kotlin synthetic view accessors - 9

Honey14:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 ### Internal
 - Add shell env comment to `pre-push` hook
 - Add the Router for the new navigation framework 
-- [In Progress: 18 Dec 2020] Migrate app to use ViewBinding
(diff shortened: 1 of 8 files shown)


vinaysshenoy (changes requested):
Overall PR looks good, but there's a crash that's happening with `SyncIndicatorView`. All you need to do is open the patient summary screen, press "Save" to go back to the home screen and the app will crash. Please take a look?

msasikanth:
> 
> 
> Overall PR looks good, but there's a crash that's happening with `SyncIndicatorView`. All you need to do is open the patient summary screen, press "Save" to go back to the home screen and the app will crash. Please take a look?

Since `SyncIndicatorView` is part of `HomeScreen` view pager, we cannot use ViewBinding for those views yet. For now, you can drop the migration for that view @Honey14 .

vinaysshenoy (approved):
LGTM

PR #2242, opened 2020-12-23
Add a compatibility layer that lets us use the older screens in the new navigation framework

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/2131/add-support-for-the-new-navigation-framework

This is a follow up PR to #2241.


msasikanth:
@janhavisinghh looks like you used merge instead of rebasing when rolling this PR on top of `master`, so there seems to be a merge commit added. Can we remove that before merging it to `master`, please?

janhavisinghh (approved):
LGTM.

PR #2243, opened 2020-12-23
Add new home illustration for India

msasikanth:


vinaysshenoy:
@msasikanth If the new homescreen graphic contains the "IHCI" logo, we cannot directly replace the india graphic in the app since that graphic is used as the default graphic for all countries unless otherwise specified. IHCI is an India specific initiative so it makes no sense to have it as the default.

We should instead be renaming the current graphic to something like the "homescreen_default" and then adding this new graphic specifically for India.

msasikanth:
> 
> 
> @msasikanth If the new homescreen graphic contains the "IHCI" logo, we cannot directly replace the india graphic in the app since that graphic is used as the default graphic for all countries unless otherwise specified. IHCI is an India specific initiative so it makes no sense to have it as the default.
> 
> We should instead be renaming the current graphic to something like the "homescreen_default" and then adding this new graphic specifically for India.

But we are already getting the illustration based on country in `PatientsTabScreen#illustrationResourceId`, so I just replaced the `ic_homescreen_india`. For any countries that are not defined in that method we are setting `ic_homescreen_default`. 🤔

vinaysshenoy:
> > @msasikanth If the new homescreen graphic contains the "IHCI" logo, we cannot directly replace the india graphic in the app since that graphic is used as the default graphic for all countries unless otherwise specified. IHCI is an India specific initiative so it makes no sense to have it as the default.
> > We should instead be renaming the current graphic to something like the "homescreen_default" and then adding this new graphic specifically for India.
> 
> But we are already getting the illustration based on country in `PatientsTabScreen#illustrationResourceId`, so I just replaced the `ic_homescreen_india`. For any countries that are not defined in that method we are setting `ic_homescreen_default`. thinking

Huh, did we already have an India specific graphic different from the default one :O

vinaysshenoy:
I swear, I don't know half of the things that have changed in this project, sorry for the earlier comment about needing to change it. 😅

Honey14 (approved):
Looks good

PR #2244, opened 2020-12-23
Implement Instant Search Query

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/1967/implement-instant-search-query

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [In Progress: 21 Dec 2020] Instant search in Faster registration
 - [In Progress: 23 Dec 2020] Migrate app to use ViewBinding
 - Track code style and lint rules for project in VCS
+- Add Instant search query
 
 ### Changes
 - Change home illustration for India
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -3922,4 +3922,62 @@ class PatientRepositoryAndroidTest {
         patientWithCurrentFacilityAsAssignedFacility, patientWithAnotherFacilityAsAssignedFacility
     )
   }
+
+  @Test
+  fun `searching_for_a_patient_by_name_must_return_list_of_patient_search_results_paritioned_by_assigned_and_not_assigned_facility`() {
+    fun createPatientWithNameAndAssignedFacilityID(
+        patientUuid: UUID,
+        patientName: String,
+        assignedFacilityId: UUID?
+    ) {
+      val patientProfile = testData
+          .patientProfile(patientUuid = patientUuid, patientName = patientName, patientAssignedFacilityId = assignedFacilityId)
+          .let { patientProfile ->
+            patientProfile.copy(
+                patient = patientProfile.patient.copy(fullName = patientName, assignedFacilityId = assignedFacilityId),
+            )
+          }
+
+      patientRepository.save(listOf(patientProfile)).blockingAwait()
+    }
+
+    fun searchResults(patientName: String, facilityUuid: UUID): List<UUID> {
+      return patientRepository
+          .search2(Name(patientName), facilityUuid)
+          .map { it.uuid }
+    }
+
+    // given
+
+    val currentFacility = testData.facility(uuid = UUID.fromString("c0056c11-105d-4079-80fa-c745128f7fc5"), name = "CHC Bucho")
+    val facility2 = testData.facility(uuid = UUID.fromString("532264e6-b358-445e-8b48-c6253677db24"), name = "Test")
+    database.facilityDao().save(listOf(currentFacility, facility2))
+
+
+    val patientWithCurrentFacilityAsAssignedFacility = UUID.fromString("d47ae2e7-7453-4a6f-806e-88eb130823d8")
+    createPatientWithNameAndAssignedFacilityID(
+        patientUuid = patientWithCurrentFacilityAsAssignedFacility,
+        patientName = "Test Name",
+        assignedFacilityId = currentFacility.uuid)
+
+    val patientWithAnotherFacilityAsAssignedFacility = UUID.fromString("f6cfd657-bf32-40c1-9dd2-4c956fd910a8")
+    createPatientWithNameAndAssignedFacilityID(
+        patientUuid = patientWithAnotherFacilityAsAssignedFacility,
+        patientName = "Tahello Name",
+        assignedFacilityId = facility2.uuid
+    )
+
+    //when
+    val searchResults = searchResults("T", currentFacility.uuid)
+    val searchResults2 = searchResults("T", facility2.uuid)
+
+    //then
+    assertThat(searchResults).containsExactly(
+        patientWithCurrentFacilityAsAssignedFacility, patientWithAnotherFacilityAsAssignedFacility
+    )
+
+    assertThat(searchResults2).containsExactly(
+        patientWithAnotherFacilityAsAssignedFacility, patientWithCurrentFacilityAsAssignedFacility
+    )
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -63,6 +63,18 @@ class PatientRepository @Inject constructor(
     }
   }
 
+  @WorkerThread
+  fun search2(criteria: PatientSearchCriteria, facilityId: UUID): List<PatientSearchResult> {
+    return when (criteria) {
+      is Name -> TODO()
```
msasikanth: FYI, We don't use Kotlin `TODO` class to define todo's since it will crash the app. Here we could've just returned the results from `searchByPhoneNumber2` directly instead of using the conditionals in this commit and define it in next commit where we will be using `searchByName2`.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -3870,4 +3870,62 @@ class PatientRepositoryAndroidTest {
...
+          }
+
+      patientRepository.save(listOf(patientProfile)).blockingAwait()
+    }
+
+    fun searchResults(phoneNumber: String, facilityUuid: UUID): Set<UUID> {
+      return patientRepository
+          .search2(PhoneNumber(phoneNumber), facilityUuid)
+          .map { it.uuid }
+          .toSet()
```
msasikanth, line 3900: Is there a reason we are converting them to `Set` here?

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -3928,4 +3928,63 @@ class PatientRepositoryAndroidTest {
...
+          }
+
+      patientRepository.save(listOf(patientProfile)).blockingAwait()
+    }
+
+    fun searchResults(patientName: String, facilityUuid: UUID): Set<UUID> {
+      return patientRepository
+          .search2(Name(patientName), facilityUuid)
+          .map { it.uuid }
+          .toSet()
```
msasikanth, line 3954: Is there a reason we are converting them to `Set` here?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Add the Router for the new navigation framework 
 - [In Progress: 21 Dec 2020] Instant search in Faster registration
 - [In Progress: 22 Dec 2020] Migrate app to use ViewBinding
+- Instant search query in Faster Registration
```
msasikanth, line 9: Can we rename this to `Add Instant search query` please?

msasikanth (changes requested):

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
janhavisinghh: Not here as I am only creating one search result object

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
janhavisinghh: Yes, because I am setting multiple values of search results.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
msasikanth: yeah, but wouldn't a list suffice for this?

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
msasikanth: yeah, but wouldn't a list suffice for this?

janhavisinghh: I mean, sure. Is there a reason we should not be using a set?

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
janhavisinghh: I have removed it from here as I'm only using this method once so I inlined it.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
msasikanth: When we are using `toSet` we are constructing a new `Set<T>` or ` Collection<T>`, while it's not that big of a problem here we don't have to do that if we don't need it right?

janhavisinghh: But, while using a list we will be creating a new list as well right?

msasikanth: Yes, but the `toSet` operation is done on that particular list. Here we have used the `map` operator to generate a list of specific fields, and we are applying `toSet` on top of that. Is there is no specific reason for using the `toSet` extension, we can use the list from `map` function. So here we can remove the `toSet` extension and it should still behave the same way.

janhavisinghh: Got it

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -63,6 +63,18 @@ class PatientRepository @Inject constructor(
     }
   }
 
+  @WorkerThread
+  fun search2(criteria: PatientSearchCriteria, facilityId: UUID): List<PatientSearchResult> {
+    return when (criteria) {
+      is Name -> searchByPhoneNumber2(criteria.patientName, facilityId) //Change to searchByName2
+      is PhoneNumber -> searchByPhoneNumber2(criteria.phoneNumber, facilityId)
+    }
+  }
```
msasikanth: This is not what I suggested 😅 . What I meant was something like this.

```kt
  @WorkerThread
  fun search2(criteria: PatientSearchCriteria, facilityId: UUID): List<PatientSearchResult> {
    return  searchByPhoneNumber2(criteria.phoneNumber, facilityId)
  }
```

Here at this point, we are only returning phone number results. Once we are adding the search by name functionality then we introduce the conditional to switch between those two. So we are following the [YAGNI principle](https://martinfowler.com/bliki/Yagni.html).

msasikanth (changes requested):
Left one comment PTAL

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
janhavisinghh: That's not gonna work as criteria can either be PhoneNumber of Name. If it's PhoneNumber then we get the number string. I can call the old searchByName() method here.

msasikanth: Ah, you can cast the criteria. So it will be like
```kt
val phoneNumber = (criteria as PhoneNumber).phoneNumber
searchByPhoneNumber2(phoneNumber, facilityId)
```

msasikanth: Since in this commit we are only searching for phone number this should be fine. Once we add the search by name in next commit it should be fine, we can add that conditional there.

msasikanth (approved):

PR #2245, opened 2020-12-23
Migrate app to use View Binding instead of Kotlin synthetic view accessors - 10

Honey14:

### DIFF
--- a/app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
+++ b/app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
@@ -3,6 +3,7 @@ package org.simple.clinic.widgets.qrcodescanner
 import android.content.Context
 import android.util.AttributeSet
 import android.util.DisplayMetrics
+import android.view.LayoutInflater
 import android.view.View
 import android.widget.FrameLayout
 import androidx.appcompat.app.AppCompatActivity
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -53,7 +61,9 @@ constructor(
   private val cameraProviderFuture = ProcessCameraProvider.getInstance(context.applicationContext)
 
   init {
-    View.inflate(context, R.layout.view_qrcode_scanner, this)
+    val layoutInflater = LayoutInflater.from(context)
+    binding = ViewQrcodeScannerBinding.inflate(layoutInflater, this, true)
+    //    View.inflate(context, R.layout.view_qrcode_scanner, this)
```
vinaysshenoy, line 66: Can we delete this line?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
vinaysshenoy: Making this change since Honey is on leave and it's quite a trivial one.

vinaysshenoy (approved):

PR #2248, opened 2020-12-29
Add convenience screens for working with the new navigation framework

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/2131/add-support-for-the-new-navigation-framework

### DIFF
--- a/app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseBottomSheet.kt
+++ b/app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseBottomSheet.kt
@@ -29,6 +29,7 @@ import org.simple.clinic.mobius.eventSources
 import org.simple.clinic.mobius.first
 import org.simple.clinic.navigation.v2.ScreenKey
 import org.simple.clinic.util.unsafeLazy
+import org.simple.clinic.util.overrideCancellation
 
 abstract class BaseBottomSheet<K : ScreenKey, B: ViewBinding, M : Parcelable, E, F> : BottomSheetDialogFragment() {
 
@@ -87,21 +88,10 @@ abstract class BaseBottomSheet<K : ScreenKey, B: ViewBinding, M : Parcelable, E,
   override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
     val dialog = super.onCreateDialog(savedInstanceState)
 
-    with(dialog) {
-      // This is needed because the router is not aware of the changes
-      // in the history when the bottom sheet dialog is dismissed in the
-      // normal fashion.
-      setCancelable(false)
-      setCanceledOnTouchOutside(false)
-      setOnKeyListener { _, keyCode, event ->
-        if (event.action == ACTION_UP && keyCode == KEYCODE_BACK) {
-          backPressed()
-          true
-        } else {
-          false
-        }
-      }
-    }
+    // This is needed because the router is not aware of the changes
+    // in the history when the bottom sheet dialog is dismissed in the
+    // normal fashion.
+    dialog.overrideCancellation(::backPressed)
 
     return dialog
   }
--- a/app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseBottomSheet.kt
+++ b/app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseBottomSheet.kt
@@ -1,0 +1,161 @@
+package org.simple.clinic.navigation.v2.fragments
+
+import android.app.Dialog
+import android.os.Bundle
+import android.os.Parcelable
+import android.view.KeyEvent.ACTION_UP
+import android.view.KeyEvent.KEYCODE_BACK
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.annotation.LayoutRes
+import androidx.fragment.app.DialogFragment
+import androidx.viewbinding.ViewBinding
+import com.google.android.material.bottomsheet.BottomSheetDialogFragment
+import com.spotify.mobius.Connection
+import com.spotify.mobius.EventSource
+import com.spotify.mobius.Init
+import com.spotify.mobius.MobiusLoop
+import com.spotify.mobius.Next.noChange
+import com.spotify.mobius.Update
+import com.spotify.mobius.android.MobiusAndroid
+import com.spotify.mobius.extras.Connectables
+import com.spotify.mobius.functions.Consumer
+import com.spotify.mobius.rx2.RxMobius
+import io.reactivex.Observable
+import io.reactivex.ObservableTransformer
+import org.simple.clinic.R
+import org.simple.clinic.mobius.eventSources
+import org.simple.clinic.mobius.first
+import org.simple.clinic.navigation.v2.ScreenKey
+import org.simple.clinic.util.unsafeLazy
+
+abstract class BaseBottomSheet<K : ScreenKey, B: ViewBinding, M : Parcelable, E, F> : BottomSheetDialogFragment() {
+
+  companion object {
+    private const val KEY_MODEL = "org.simple.clinic.navigation.v2.fragments.BaseScreen.KEY_MODEL"
+  }
+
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
```diff
@@ -19,12 +19,12 @@ import com.spotify.mobius.functions.Consumer
...
 import io.reactivex.Observable
 import io.reactivex.ObservableTransformer
-import org.simple.clinic.util.unsafeLazy
 import org.simple.clinic.mobius.eventSources
 import org.simple.clinic.mobius.first
 import org.simple.clinic.navigation.v2.ScreenKey
+import org.simple.clinic.util.unsafeLazy
 
-abstract class BaseScreen<K: ScreenKey, M: Parcelable, E, F>: Fragment() {
+abstract class BaseScreen<K : ScreenKey, B, M : Parcelable, E, F> : Fragment() {
```
msasikanth, line 27: We can make the binding generic extending `ViewBinding`.

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  private val controller: MobiusLoop.Controller<M, E> by unsafeLazy {
+    MobiusAndroid.controller(loop, defaultModel())
+  }
+
+  protected val screenKey by unsafeLazy { ScreenKey.key<K>(this) }
+
+  @LayoutRes
+  abstract fun layoutResId(): Int
```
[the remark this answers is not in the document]
msasikanth: Since we are providing the binding generic in next commit, we can make the binding as constructor parameter and pass `binding.root` to the `FragmentConstructor`.

app/src/main/res/values/styles_v2.xml
```diff
@@ -737,4 +737,12 @@
     <item name="shapeAppearanceOverlay">@style/ShapeAppearanceOverlay.Clinic.V2.UnitPreferenceButton</item>
   </style>
 
+  <style name="Clinic.V2.Theme.BottomSheetFragment" parent="Theme.Design.Light.BottomSheetDialog">
```
msasikanth, line 740: Perhaps we can use the Material styles here?

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseDialog.kt
```diff
@@ -0,0 +1,129 @@
+package org.simple.clinic.navigation.v2.fragments
```
msasikanth, line 1: Can we rename the commit message to `Add BaseDialog as a base class for all dialogs supporting `

msasikanth (changes requested):
Looks good overall, left a couple of comments PTAL.

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
vinaysshenoy: Can we do that? I didn't know that the binding classes extend a common type. If they do, that might make some of the stuff simpler. I'll take a look.

app/src/main/res/values/styles_v2.xml
vinaysshenoy: Hmm, not sure which style I should be using. Which one is it?

msasikanth: It will be `Theme.MaterialComponents.Light.BottomSheetDialog`. Most of the time we just have to replace `Design` or `AppCompat` to `MaterialComponents`

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
msasikanth: Yep. unfortunately, it doesn't have `bind` or `inflate` methods but it provides us with `getRoot` method and a way to make sure they only provide binding classes.

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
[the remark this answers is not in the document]
vinaysshenoy: We cannot pass the binding as a constructor parameter because that is not something that can be saved and restored, so we won't be able to add a default constructor. This is also why we cannot pass the layout res ID as a constructor parameter because this will stop us from being able to construct a view binding instance, right?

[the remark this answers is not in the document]
msasikanth: Ah right, looks like we can only pass `LayoutResId` via the `Fragment` constructor. We can stick with existing approach then 👍🏽

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
```diff
@@ -61,7 +72,11 @@ abstract class BaseScreen<K: ScreenKey, M: Parcelable, E, F>: Fragment() {
   open fun additionalEventSources(): List<EventSource<E>> = emptyList()
 
   override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
-    return inflater.inflate(layoutResId(), container, false)
+    val view = inflater.inflate(layoutResId(), container, false)
+
+    _binding = bindView(view)
+
+    return view
```
msasikanth, line 79: Instead of this could we have used the return value from `bindView` method? 🤔 . Something like this perhaps.

```kt
  abstract fun bindView(layoutInflater: LayoutInflater, container: ViewGroup?): B

  override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
    return bindView(inflater, container).root
  }
```

That way we don't have to provide the `LayoutResId` and do the binding inflate method in the classes that extend `BaseScreen`. WDYT?

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseBottomSheet.kt
```diff
@@ -0,0 +1,166 @@
...
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setStyle(DialogFragment.STYLE_NORMAL, R.style.Clinic_V2_Theme_BottomSheetFragment)
+  }
+
+  override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+    val view = inflater.inflate(layoutResId(), container, false)
+
+    _binding = bindView(view)
```
msasikanth, line 88: Same comment as above, we can avoid using the `LayoutResId`'s

app/src/main/res/values/styles_v2.xml
msasikanth: Looks like you forgot to change it for `BottomSheetFragment`

msasikanth: NVM, you changed it in next commit.

msasikanth (commented):
Looks good overall. Left one comment, if that's something we don't want to do or will be doing that later, I can merge this in now.

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
vinaysshenoy: I am not entirely sure how that would work? I decided to follow the pattern that we use in our current screens where we call `Binding.bind(this)` which means we have to inflate the view and then instantiate the binding by providing it a reference to the view.

msasikanth: It's slightly different for fragments, we can directly inflate from the binding and use `binding.root` for returning the view.

https://developer.android.com/topic/libraries/view-binding

msasikanth (approved):

PR #2257, opened 2021-01-05
Implement instant search screen - 4

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2029/implement-instant-search-screen

### DIFF
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
@@ -25,3 +25,5 @@ data class OpenLinkIdWithPatientScreen(val patientId: UUID, val identifier: Iden
 data class OpenBpPassportSheet(val identifier: Identifier) : InstantSearchEffect()
 
 data class ShowNoPatientsInFacility(val facility: Facility) : InstantSearchEffect()
+
+object ShowNoSearchResults : InstantSearchEffect()
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
@@ -3,6 +3,7 @@ package org.simple.clinic.instantsearch
 import org.simple.clinic.facility.Facility
 import org.simple.clinic.patient.PatientSearchCriteria
 import org.simple.clinic.patient.PatientSearchResult
+import java.util.UUID
 
 sealed class InstantSearchEffect
 
@@ -15,3 +16,5 @@ data class SearchWithCriteria(val criteria: PatientSearchCriteria, val facility:
 data class ShowPatientSearchResults(val patients: List<PatientSearchResult>, val facility: Facility) : InstantSearchEffect()
 
 data class ValidateSearchQuery(val searchQuery: String) : InstantSearchEffect()
+
+data class OpenPatientSummary(val patientId: UUID) : InstantSearchEffect()
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
@@ -29,3 +29,5 @@ data class ShowNoPatientsInFacility(val facility: Facility) : InstantSearchEffec
 object ShowNoSearchResults : InstantSearchEffect()
 
 object HideNoPatientsInFacility : InstantSearchEffect()
+
+object HideNoSearchResults : InstantSearchEffect()
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
@@ -21,3 +21,5 @@ data class ValidateSearchQuery(val searchQuery: String) : InstantSearchEffect()
 data class OpenPatientSummary(val patientId: UUID) : InstantSearchEffect()
 
 data class OpenLinkIdWithPatientScreen(val patientId: UUID, val identifier: Identifier) : InstantSearchEffect()
+
+data class OpenBpPassportSheet(val identifier: Identifier) : InstantSearchEffect()
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
@@ -3,6 +3,7 @@ package org.simple.clinic.instantsearch
 import org.simple.clinic.facility.Facility
 import org.simple.clinic.patient.PatientSearchCriteria
 import org.simple.clinic.patient.PatientSearchResult
+import org.simple.clinic.patient.businessid.Identifier
 import java.util.UUID
 
 sealed class InstantSearchEffect
@@ -18,3 +19,5 @@ data class ShowPatientSearchResults(val patients: List<PatientSearchResult>, val
 data class ValidateSearchQuery(val searchQuery: String) : InstantSearchEffect()
 
 data class OpenPatientSummary(val patientId: UUID) : InstantSearchEffect()
+
+data class OpenLinkIdWithPatientScreen(val patientId: UUID, val identifier: Identifier) : InstantSearchEffect()
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
@@ -23,3 +23,5 @@ data class OpenPatientSummary(val patientId: UUID) : InstantSearchEffect()
 data class OpenLinkIdWithPatientScreen(val patientId: UUID, val identifier: Identifier) : InstantSearchEffect()
 
 data class OpenBpPassportSheet(val identifier: Identifier) : InstantSearchEffect()
+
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/simple/clinic/instantsearch/InstantSearchUpdate.kt
```diff
@@ -58,7 +58,12 @@ class InstantSearchUpdate : Update<InstantSearchModel, InstantSearchEvent, Insta
   private fun searchResultsLoaded(model: InstantSearchModel, event: SearchResultsLoaded): Next<InstantSearchModel, InstantSearchEffect> {
     if (!model.hasSearchQuery) return noChange()
 
-    return dispatch(ShowPatientSearchResults(event.patientsSearchResults, model.facility!!))
+    val effect = if (event.patientsSearchResults.isNotEmpty())
+      ShowPatientSearchResults(event.patientsSearchResults, model.facility!!)
+    else
+      ShowNoSearchResults
```
vinaysshenoy, line 64: Should this be an effect? Seems to be more appropriate as an implementation detail of the UI renderer which looks at the search results, right?

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchUpdate.kt
```diff
@@ -58,7 +58,12 @@ class InstantSearchUpdate : Update<InstantSearchModel, InstantSearchEvent, Insta
   private fun searchResultsLoaded(model: InstantSearchModel, event: SearchResultsLoaded): Next<InstantSearchModel, InstantSearchEffect> {
     if (!model.hasSearchQuery) return noChange()
 
-    return dispatch(ShowPatientSearchResults(event.patientsSearchResults, model.facility!!))
+    val effect = if (event.patientsSearchResults.isNotEmpty())
+      ShowPatientSearchResults(event.patientsSearchResults, model.facility!!)
```
vinaysshenoy, line 62: Search results should be a part of the model and rendered via the UI renderer. Why are we loading them via effects? 🤔

app/src/test/java/org/simple/clinic/instantsearch/InstantSearchUpdateTest.kt
```diff
@@ -82,6 +82,22 @@ class InstantSearchUpdateTest {
         ))
   }
 
+  @Test
+  fun `when search results are empty, then show no search results`() {
+    val facility = TestData.facility()
```
vinaysshenoy, line 87: Can we fix the IDs used in tests?

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffectHandler.kt
```diff
@@ -36,6 +36,7 @@ class InstantSearchEffectHandler @AssistedInject constructor(
       .addConsumer(ShowNoPatientsInFacility::class.java, { uiActions.showNoPatientsInFacility(it.facility) }, schedulers.ui())
       .addAction(ShowNoSearchResults::class.java, { uiActions.showNoSearchResults() }, schedulers.ui())
       .addAction(HideNoPatientsInFacility::class.java, uiActions::hideNoPatientsInFacility, schedulers.ui())
+      .addAction(HideNoSearchResults::class.java, uiActions::hideNoSearchResults, schedulers.ui())
```
vinaysshenoy, line 39: These should be handled in the UI renderer and not via effects. UI actions are meant for things that are not part of the screen as such, like showing a modal or navigating to a different screen.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchUpdate.kt
```diff
@@ -41,7 +41,7 @@ class InstantSearchUpdate : Update<InstantSearchModel, InstantSearchEvent, Insta
         dispatch(HideNoPatientsInFacility, HideNoSearchResults, SearchWithCriteria(criteria, model.facility!!))
       }
       InstantSearchValidator.Result.LengthTooShort -> noChange()
-      InstantSearchValidator.Result.Empty -> dispatch(LoadAllPatients(model.facility!!))
+      InstantSearchValidator.Result.Empty -> dispatch(HideNoPatientsInFacility, HideNoSearchResults, LoadAllPatients(model.facility!!))
```
vinaysshenoy, line 44: We can statically import `InstantSearchValidator.Result` here for better readability.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffectHandler.kt
```diff
@@ -30,16 +30,19 @@ class InstantSearchEffectHandler @AssistedInject constructor(
...
-          .map { instantSearchValidator.validate(
-              searchQuery = it.searchQuery,
-              minLengthForSearchQuery = instantSearchConfig.minLengthOfSearchQuery
-          ) }
+          .map {
+            instantSearchValidator.validate(
+                searchQuery = it.searchQuery,
+                minLengthForSearchQuery = instantSearchConfig.minLengthOfSearchQuery
+            )
+          }
```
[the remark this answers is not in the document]
msasikanth: Ah, I must have accidentally added this here it was supposed to go in with the original commit. I will keep in mind for future PRs 😅

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
```diff
@@ -21,3 +21,5 @@ data class ValidateSearchQuery(val searchQuery: String) : InstantSearchEffect()
 data class OpenPatientSummary(val patientId: UUID) : InstantSearchEffect()
 
 data class OpenLinkIdWithPatientScreen(val patientId: UUID, val identifier: Identifier) : InstantSearchEffect()
+
+data class OpenBpPassportSheet(val identifier: Identifier) : InstantSearchEffect()
```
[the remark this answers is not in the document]
msasikanth: Hmm, I was following the naming conventions of `OpenPatientSummary` to indicate which screen/sheet we will be opening. But `CheckNewRegistrationSheet` name sounds good as well, I can change it if you want.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchUpdate.kt
msasikanth: We discussed this before since we will be replacing these lists with paging data, we cannot place them in models. So for that reason, I had to switch to using effects for handling lists and error cases. Should we change it?

https://simpledotorg.slack.com/archives/C01FN9GCNAJ/p1609295944003000

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffectHandler.kt
msasikanth: Hopefully, this [comment ](https://github.com/simpledotorg/simple-android/pull/2257#discussion_r551903893) clarifies why I went with this usage.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchUpdate.kt
msasikanth: Hopefully, this [comment ](https://github.com/simpledotorg/simple-android/pull/2257#discussion_r551903893) clarifies why I went with this usage.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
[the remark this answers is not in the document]
vinaysshenoy: It's not about whether I prefer this name, it's about what makes more sense from a business use case perspective.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchUpdate.kt
vinaysshenoy: Ah, ok. I guess that makes sense.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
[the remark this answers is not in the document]
vinaysshenoy: In any case, we can discuss this later. It's not critical to this PR.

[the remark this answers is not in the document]
msasikanth: Ah okay, my only issue with `CheckNewRegistrationSheet` is its kind of misleading since we are not actually checking or validating anything. We are just opening a sheet, where the user has to select one of the 2 buttons to either create a patient or assign it to a patient. So `OpenBpPassportSheet` makes more sense in my opinion. WDYT?

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchUpdate.kt
vinaysshenoy: Makes sense, let's leave it as is.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffect.kt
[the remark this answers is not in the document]
vinaysshenoy: Ok, I don't think `OpenBpPassport` is the right name for this feature if that's the scenario but `CheckNewRegistration` is also not the right one. Let's table this for now and figure out a better name later.

vinaysshenoy (approved):

PR #2259, opened 2021-01-06
Rename `newbranch` script file

msasikanth:
Windows PowerShell cannot run the shell script directly without the `.sh` suffix. You can test by `bash newbranch.sh branch-name` in Windows PowerShell to see the output or else you can run `.\newbranch.sh branch-name` if you don't want any output printed in the shell.

Adding @vinaysshenoy as a reviewer since he is more familiar with the script.


vinaysshenoy:
Looks good, one issue is that renaming the file reset the executable flag. Will fix this when rerolling.

vinaysshenoy (approved):

msasikanth:
> Looks good, one issue is that renaming the file reset the executable flag. Will fix this when rerolling.

Ah right, I forgot to verify it in linux/unix environment. Will keep this in mind for any future changes.

PR #2274, opened 2021-01-07
Purge the QA environment before running the instrumented tests in CI

vinaysshenoy:
Since we use static patient IDs in tests, but the facility is randomly selected, it was possible to get into a situation where instrumented tests would continuously fail until the server was manually purged. This resolves the issue by running the purge task before triggering the instrumented tests.

More details here: https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1609913473022000

### DIFF
--- a/.github/workflows/ci_checks.yml
+++ b/.github/workflows/ci_checks.yml
@@ -35,10 +35,20 @@ jobs:
       - name: QA Unit Tests
         run: ./gradlew testQaDebugUnitTest
 
+  qa_purge_env:
+    runs-on: [ubuntu-latest]
+    steps:
+      - name: Purge QA server
+        run: |
+          curl -v --request DELETE \
+          --url https://api-qa.simple.org/qa/purge \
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_checks.yml
```diff
@@ -160,13 +170,3 @@ jobs:
           USER_EMAIL: ${{ secrets.USER_EMAIL }}
           BRANCH_NAME: ${{ github.event.pull_request.head.ref }}
         run: .github/scripts/amend_and_push.sh
-
-  qa_purge_env:
-    runs-on: [ubuntu-latest]
-    needs: [qa_lint, qa_unit_tests, qa_android_tests, mobius_migration_tests]
```
msasikanth, line 166: So I am guessing since purge is only needed for Android tests we longer have to depend on all of the other jobs?

msasikanth (approved):
LGTM

.github/workflows/ci_checks.yml
vinaysshenoy: Yup.

PR #2276, opened 2021-01-08
Fix garbled gradle logs when running pre-push hook in Git Bash

msasikanth:
Git Bash is printing garbled logs because it cannot figure out how to print the rich content from Gradle logs (ex: progress) when running the pre-push hook. In order to resolve that and make the logs readable, we set the console type for the pre-push hook.

**Reproduction steps**:
- Open Git Bash
- Run pre-push hook without these changes
> Prints garbled logs
- Run pre-push hook with these changes
> Prints plain logs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Add Instant search query
 - Add `ScreenFragmentCompat` to support using the older view-based screens in the v2 navigation framework
 - Add convenience classes for creating screens for the new navigation framework
+- Fix garbled gradle logs when running pre-push hook in Git Bash
 
 ### Fixes
 - Fix BPpassport prefill value to have display value
--- a/quality/pre-push
+++ b/quality/pre-push
@@ -1,2 +1,13 @@
 #!/bin/sh
-GIT_WORK_TREE=".." ./gradlew lintQaDebug assembleQaDebugUnitTest assembleQaDebugAndroidTest
+readonly MSYS="msys"
+readonly OSTYPE="$OSTYPE"
+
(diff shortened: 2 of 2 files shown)


msasikanth:
*garbled logs screenshot*

Screenshot 2021-01-08 085658

msasikanth:
The issue is specific to the Git Bash, when ran on Powershell the `rich` console output is displaying properly.

vinaysshenoy:
> The issue is specific to the Git Bash, when ran on Powershell the `rich` console output is displaying properly. This change would disable rich console output on Windows (msys) OS type. So we need to decide if we can ignore Git Bash and continue with other terminals like CMD or PowerShell, or do this change and print plain logs on all the terminals.

If the issue is specific to Git Bash, then let's print rich logs everywhere and default to plain ones on Git Bash. No need to limit other platforms.

msasikanth:
It will work fine on other platforms, the console type change is only for Windows platform terminals and only for the pre-push hook. Normal gradle commands works fine. But I will investigate why it's happening on Git Bash further, so marking this PR as WIP until then. Timeboxing it until afternoon.

vinaysshenoy:
> It will work fine on other platforms, the console type change is only for Windows platform terminals and only for the pre-push hook. Normal gradle commands works fine. But I will investigate why it's happening on Git Bash further, so marking this PR as WIP until then. Timeboxing it until afternoon.

Cool, am approving this for now. Feel free to merge it in if your experiment does not work out.

vinaysshenoy (approved):

vinaysshenoy:
Huh, looks like you enabled auto-merge for it. I did not notice 😅

PR #2285, opened 2021-01-12
Show keyboard in Instant Search only when there is no identifier

msasikanth:


app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreen.kt
```diff
@@ -155,7 +155,11 @@ class InstantSearchScreen(context: Context, attrs: AttributeSet) : ConstraintLay
 
     binding = ScreenInstantSearchBinding.bind(this)
 
-    searchQueryEditText.showKeyboard()
+    // We only need to open keyboard when there is not addition identifier (BP Passport identifier),
```
vinaysshenoy, line 158: This comment explains the "what", not the "why". Looking at the code, we already can see that the check is for the absence of the additional identifier so we don't need to write this down again in the comment.

What we should do instead is explain something like:
"When a scanned BP Passport does not result in a match, we bring up a bottom sheet which asks whether this is a new registration or an existing patient. If we show the keyboard in these cases, the UI is janky since the keyboard pops up and immediately another bottom sheet pops up. This improves the experience by showing the keyboard only if we have arrived here by searching for a patient by the name".

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreen.kt
msasikanth: Ah, good point. Will make this change.

vinaysshenoy (approved):

PR #2289, opened 2021-01-13
Highlight instant search results

Honey14:
https://app.clubhouse.io/simpledotorg/story/2035/highlight-patient-name-and-phone-number-based-on-search-query

### DIFF
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchResultsItemType.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchResultsItemType.kt
@@ -19,29 +19,35 @@ sealed class InstantSearchResultsItemType : ItemAdapter.Item<InstantSearchResult
 
     fun from(
         patientSearchResults: List<PatientSearchResult>,
-        currentFacility: Facility
+        currentFacility: Facility,
+        searchQuery: String?
     ): List<InstantSearchResultsItemType> {
       val (assignedFacilityPatients, nearbyFacilitiesPatients) = patientSearchResults
           .partition { it.assignedFacilityId == currentFacility.uuid }
 
-      val assignedFacilityPatientsGroup = generateAssignedFacilityPatientsGroup(assignedFacilityPatients, currentFacility)
-      val nearbyFacilitiesPatientsGroup = generateNearbyFacilitiesPatientsGroup(nearbyFacilitiesPatients, currentFacility)
+      val assignedFacilityPatientsGroup = generateAssignedFacilityPatientsGroup(assignedFacilityPatients, currentFacility, searchQuery)
+      val nearbyFacilitiesPatientsGroup = generateNearbyFacilitiesPatientsGroup(nearbyFacilitiesPatients, currentFacility, searchQuery)
 
       return assignedFacilityPatientsGroup + nearbyFacilitiesPatientsGroup
     }
 
     private fun generateAssignedFacilityPatientsGroup(
         assignedFacilityPatients: List<PatientSearchResult>,
-        currentFacility: Facility
+        currentFacility: Facility,
+        searchQuery: String?
     ) = if (assignedFacilityPatients.isNotEmpty()) {
-      listOf(AssignedFacilityHeader(currentFacility.name)) + SearchResult.forSearchResults(assignedFacilityPatients, currentFacility.uuid)
+      listOf(AssignedFacilityHeader(currentFacility.name)) + SearchResult.forSearchResults(assignedFacilityPatients, currentFacility.uuid, searchQuery)
     } else {
       emptyList()
     }
 
-    private fun generateNearbyFacilitiesPatientsGroup(nearbyFacilitiesPatients: List<PatientSearchResult>, currentFacility: Facility) =
+    private fun generateNearbyFacilitiesPatientsGroup(
+        nearbyFacilitiesPatients: List<PatientSearchResult>,
+        currentFacility: Facility,
+        searchQuery: String?
+    ) =
         if (nearbyFacilitiesPatients.isNotEmpty()) {
-          listOf(NearbyFacilitiesHeader) + SearchResult.forSearchResults(nearbyFacilitiesPatients, currentFacility.uuid)
+          listOf(NearbyFacilitiesHeader) + SearchResult.forSearchResults(nearbyFacilitiesPatients, currentFacility.uuid, searchQuery)
         } else {
           emptyList()
         }
@@ -81,17 +87,19 @@ sealed class InstantSearchResultsItemType : ItemAdapter.Item<InstantSearchResult
 
   data class SearchResult(
       val searchResultViewModel: PatientSearchResultViewModel,
-      val currentFacilityId: UUID
+      val currentFacilityId: UUID,
+      val searchQuery: String?
   ) : InstantSearchResultsItemType() {
 
     companion object {
       fun forSearchResults(
           searchResults: List<PatientSearchResult>,
-          currentFacilityId: UUID
+          currentFacilityId: UUID,
+          searchQuery: String?
       ): List<SearchResult> {
         return searchResults
             .map(::mapPatientSearchResultToViewModel)
-            .map { searchResultViewModel -> SearchResult(searchResultViewModel, currentFacilityId) }
+            .map { searchResultViewModel -> SearchResult(searchResultViewModel, currentFacilityId, searchQuery) }
       }
 
       private fun mapPatientSearchResultToViewModel(searchResult: PatientSearchResult): PatientSearchResultViewModel {
@@ -117,7 +125,7 @@ sealed class InstantSearchResultsItemType : ItemAdapter.Item<InstantSearchResult
       binding.patientSearchResultView.setOnClickListener {
         subject.onNext(Event.ResultClicked(searchResultViewModel.uuid))
       }
-      binding.patientSearchResultView.render(searchResultViewModel, currentFacilityId)
+      binding.patientSearchResultView.render(searchResultViewModel, currentFacilityId, searchQuery)
     }
   }
 
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreen.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreen.kt
@@ -190,15 +190,15 @@ class InstantSearchScreen :
 
   override fun showAllPatients(patients: List<PatientSearchResult>, facility: Facility) {
     searchResultsView.visibility = View.VISIBLE
-    allPatientsAdapter.submitList(InstantSearchResultsItemType.from(patients, facility))
+    allPatientsAdapter.submitList(InstantSearchResultsItemType.from(patients, facility, searchQuery = null))
 
     searchResultsView.swapAdapter(allPatientsAdapter, false)
     searchResultsView.scrollToPosition(0)
   }
 
-  override fun showPatientsSearchResults(patients: List<PatientSearchResult>, facility: Facility) {
+  override fun showPatientsSearchResults(patients: List<PatientSearchResult>, facility: Facility, searchQuery: String) {
     searchResultsView.visibility = View.VISIBLE
-    searchResultsAdapter.submitList(InstantSearchResultsItemType.from(patients, facility))
+    searchResultsAdapter.submitList(InstantSearchResultsItemType.from(patients, facility, searchQuery))
 
     searchResultsView.swapAdapter(searchResultsAdapter, false)
     searchResultsView.scrollToPosition(0)
--- a/app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
+++ b/app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
@@ -5,6 +5,7 @@ import android.text.Spannable
 import android.text.SpannableStringBuilder
 import android.text.style.BackgroundColorSpan
 import android.util.AttributeSet
+import android.view.View
 import com.google.android.material.card.MaterialCardView
 import kotlinx.android.synthetic.main.view_patient_search_result.view.*
 import org.simple.clinic.R
@@ -51,9 +52,26 @@ class PatientSearchResultItemView(
     renderPatientNameAgeAndGender(searchQuery, model)
     renderPatientAddress(model.address)
     renderPatientDateOfBirth(model.dateOfBirth)
-    renderPatientPhoneNumber(model.phoneNumber)
+    renderPatientPhoneNumber(searchQuery, model)
     renderVisited(model.lastSeen)
     renderLastSeen(model.lastSeen, currentFacilityId)
+  }
+
+  private fun getPatientPhoneNumber(
+      phoneNumber: String,
+      searchQuery: String?
+  ): PhoneNumber {
+    val canHighlightNumber = !searchQuery.isNullOrBlank() && phoneNumber.contains(searchQuery)
+
+    return if (canHighlightNumber) {
+      val indexOfSearchedPhoneNumber = phoneNumber.indexOf(searchQuery!!)
+      PhoneNumber.Highlighted(
+          patientNumber = phoneNumber,
+          highlightStart = indexOfSearchedPhoneNumber,
+          highlightEnd = indexOfSearchedPhoneNumber + searchQuery.length)
+    } else {
+      PhoneNumber.Plain(phoneNumber)
+    }
   }
 
   private fun renderLastSeen(lastSeen: PatientSearchResult.LastSeen?, currentFacilityId: UUID) {
@@ -76,11 +94,26 @@ class PatientSearchResultItemView(
     }
   }
 
-  private fun renderPatientPhoneNumber(phoneNumber: String?) {
-    phoneNumberContainer.visibleOrGone(phoneNumber.isNullOrBlank().not())
-    if (phoneNumber != null) {
-      phoneNumberTextView.text = phoneNumber
+  private fun renderPatientPhoneNumber(searchQuery: String?, patientSearchResult: PatientSearchResultViewModel) {
+    if (patientSearchResult.phoneNumber == null) {
+      phoneNumberContainer.visibility = View.GONE
+      return
     }
+
+    phoneNumberContainer.visibility = View.VISIBLE
+
+    val patientPhoneNumber = when (val number = getPatientPhoneNumber(patientSearchResult.phoneNumber, searchQuery)) {
+      is PhoneNumber.Highlighted -> {
+        val highlightNumber = SpannableStringBuilder(number.patientNumber)
+        val highlightColor = context.resolveColor(colorRes = R.color.simple_light_blue_100)
+        highlightNumber.setSpan(BackgroundColorSpan(highlightColor), number.highlightStart, number.highlightEnd, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
+        highlightNumber
+      }
+      is PhoneNumber.Plain -> {
+        number.patientNumber
+      }
+    }
+    phoneNumberTextView.text = patientPhoneNumber
   }
 
   private fun renderPatientDateOfBirth(dateOfBirth: LocalDate?) {
@@ -142,6 +175,12 @@ class PatientSearchResultItemView(
     data class Highlighted(override val patientName: String, val highlightStart: Int, val highlightEnd: Int) : Name(patientName)
     data class Plain(override val patientName: String) : Name(patientName)
   }
+
+  sealed class PhoneNumber(open val patientNumber: String?) {
+    data class Highlighted(override val patientNumber: String, val highlightStart: Int, val highlightEnd: Int) : PhoneNumber(patientNumber)
+    data class Plain(override val patientNumber: String) : PhoneNumber(patientNumber)
+  }
+
 
   interface Injector {
     fun inject(target: PatientSearchResultItemView)
--- a/app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
+++ b/app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
@@ -43,7 +43,7 @@ class PatientSearchResultItemView(
     context.injector<Injector>().inject(this)
   }
 
-  fun render(model: PatientSearchResultViewModel, currentFacilityId: UUID) {
+  fun render(model: PatientSearchResultViewModel, currentFacilityId: UUID, searchQuery: String?) {
     renderPatientNameAgeAndGender(model.fullName, model.gender, DateOfBirth.fromPatientSearchResultViewModel(model, userClock))
     renderPatientAddress(model.address)
     renderPatientDateOfBirth(model.dateOfBirth)
--- a/app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
+++ b/app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
@@ -1,6 +1,9 @@
 package org.simple.clinic.widgets
 
 import android.content.Context
+import android.text.Spannable
+import android.text.SpannableStringBuilder
+import android.text.style.BackgroundColorSpan
 import android.util.AttributeSet
 import com.google.android.material.card.MaterialCardView
 import kotlinx.android.synthetic.main.view_patient_search_result.view.*
@@ -13,6 +16,7 @@ import org.simple.clinic.patient.PatientAddress
 import org.simple.clinic.patient.PatientSearchResult
 import org.simple.clinic.patient.displayIconRes
 import org.simple.clinic.patient.displayLetterRes
+import org.simple.clinic.router.util.resolveColor
 import org.simple.clinic.util.UserClock
 import org.simple.clinic.util.toLocalDateAtZone
 import java.time.LocalDate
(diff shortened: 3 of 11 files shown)


app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreen.kt
```diff
@@ -196,15 +196,15 @@ class InstantSearchScreen(context: Context, attrs: AttributeSet) : ConstraintLay
 
   override fun showAllPatients(patients: List<PatientSearchResult>, facility: Facility) {
     searchResultsView.visibility = View.VISIBLE
-    allPatientsAdapter.submitList(InstantSearchResultsItemType.from(patients, facility))
+    allPatientsAdapter.submitList(InstantSearchResultsItemType.from(patients, facility, ""))
```
msasikanth, line 199: Instead of passing an empty string that doesn't say what it is. Let's pass a `null` value and use parameter names.

app/src/test/java/org/simple/clinic/instantsearch/InstantSearchUpdateTest.kt
```diff
@@ -92,6 +92,7 @@ class InstantSearchUpdateTest {
         uuid = UUID.fromString("34eb57a9-d80a-4f43-9f89-1e2dade3de3f"),
         name = "PHC Obvious"
     )
+    val searchQuery = "P"
```
msasikanth, line 95: Here the search query that is entered into the model is "Pat". Let's change it so that they are consistent.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchResultsItemType.kt
```diff
@@ -19,29 +20,35 @@ sealed class InstantSearchResultsItemType : ItemAdapter.Item<InstantSearchResult
...
     }
 
     private fun generateAssignedFacilityPatientsGroup(
         assignedFacilityPatients: List<PatientSearchResult>,
-        currentFacility: Facility
+        currentFacility: Facility,
+        searchQuery: String
     ) = if (assignedFacilityPatients.isNotEmpty()) {
-      listOf(AssignedFacilityHeader(currentFacility.name)) + SearchResult.forSearchResults(assignedFacilityPatients, currentFacility.uuid)
+      listOf(AssignedFacilityHeader(currentFacility.name)) + forSearchResults(assignedFacilityPatients, currentFacility.uuid, searchQuery)
```
msasikanth, line 40: I think you imported the `SearchResults.Companion.forSearchResults` directly. In this scenario it makes sense to leave it, or else we are missing the context onto why were are doing this.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -91,13 +95,41 @@ class PatientSearchResultItemView(
...
+    } else {
+      Name.Plain(renderPatient(patientSearchResult.fullName, patientSearchResult.gender, dateOfBirth))
+    }
+  }
+
+  private fun renderPatientNameAgeAndGender(name: Name) {
+    when (name) {
+      is Name.Highlighted -> {
+        val highlightName = SpannableStringBuilder(name.patientName)
+        val highlightColor = ContextCompat.getColor(context, R.color.facility_search_query_highlight)
```
msasikanth, line 146: Let's create a separate color for this? or use `simple_light_blue_100`. Also, you can now start using the `context.resolveColor` method to get the color, theme attribute color.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -44,8 +47,9 @@ class PatientSearchResultItemView(
     context.injector<Injector>().inject(this)
   }
 
-  fun render(model: PatientSearchResultViewModel, currentFacilityId: UUID) {
-    renderPatientNameAgeAndGender(model.fullName, model.gender, DateOfBirth.fromPatientSearchResultViewModel(model, userClock))
+  fun render(model: PatientSearchResultViewModel, currentFacilityId: UUID, searchQuery: String) {
+    val name = highlightPatientName(searchQuery, model, DateOfBirth.fromPatientSearchResultViewModel(model, userClock))
```
msasikanth, line 52: Since it's just being used by `renderPatientNameAgeAndGender` let's move this into that method.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -91,13 +95,41 @@ class PatientSearchResultItemView(
...
+        highlightName.setSpan(BackgroundColorSpan(highlightColor), name.highlightStart, name.highlightEnd, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
+        patientNameAgeGenderLabel.text = highlightName
+      }
+      is Name.Plain -> {
+        patientNameAgeGenderLabel.text = name.patientName
+      }
+    }
+  }
+
+  private fun renderPatient(fullName: String, gender: Gender, dateOfBirth: DateOfBirth): String {
```
msasikanth, line 156: Here we are not actually rendering the patient information, we are just getting the patient name, age, and dob string. Can we rename this to represent that?

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -77,10 +92,17 @@ class PatientSearchResultItemView(
...
 
-  private fun renderPatientPhoneNumber(phoneNumber: String?) {
-    phoneNumberContainer.visibleOrGone(phoneNumber.isNullOrBlank().not())
-    if (phoneNumber != null) {
-      phoneNumberTextView.text = phoneNumber
+  private fun renderPatientPhoneNumber(number: PhoneNumber) {
+    when (number) {
+      is PhoneNumber.Highlighted -> {
+        val highlightNumber = SpannableStringBuilder(number.patientNumber)
+        val highlightColor = ContextCompat.getColor(context, R.color.blue2)
```
msasikanth, line 109: This color will no longer exist, can we create a new color for representing the instant search highlight, or you can use `simple_light_blue_100`

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -52,11 +52,26 @@ class PatientSearchResultItemView(
     renderPatientNameAgeAndGender(name)
     renderPatientAddress(model.address)
     renderPatientDateOfBirth(model.dateOfBirth)
-    renderPatientPhoneNumber(model.phoneNumber)
+    val number = highlightPatientPhoneNumber(searchQuery, model)
```
msasikanth: We can move this inside the `renderPatientPhoneNumber` method

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -52,15 +53,24 @@ class PatientSearchResultItemView(
...
-    val canHighlightNumber = searchQuery.isNotBlank()
-        && patientSearchResult.phoneNumber.toString().contains(searchQuery)
+  private fun highlightPatientPhoneNumber(searchQuery: String, patientSearchResult: PatientSearchResultViewModel) {
+    if (patientSearchResult.phoneNumber != null) {
+      val highlightNumber = searchQuery.isNotBlank()
+          && patientSearchResult.phoneNumber.toString().contains(searchQuery)
+      val number = canHighlightNumber(highlightNumber, patientSearchResult, searchQuery)
+      renderPatientPhoneNumber(number)
+    } else {
+      phoneNumberContainer.visibility = View.GONE
```
msasikanth, line 69: We are already setting the phone number container visibility in the renderPhoneNumber method. Why are we doing this again here? 🤔 

Also, why did we remove the `renderPatientPhoneNumber` method?

msasikanth (changes requested):
LGTM. Left few comments PTAL

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
Honey14: We are checking the visibility before we highlight or show the container. We did not remove `renderPatientPhoneNumber` method. We're just called it when the phoneNumber is not null.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
Honey14: Improved in the later commit, please check.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -90,13 +94,42 @@ class PatientSearchResultItemView(
     addressLabel.text = address.completeAddress
   }
 
-  private fun renderPatientNameAgeAndGender(fullName: String, gender: Gender, dateOfBirth: DateOfBirth) {
+  private fun highlightPatientName(searchQuery: String?, patientSearchResult: PatientSearchResultViewModel, dateOfBirth: DateOfBirth): Name {
+    val canHighlight = searchQuery?.let { searchQuery.isNotBlank() && patientSearchResult.fullName.contains(it, ignoreCase = true) } == true
```
msasikanth, line 127: We can make this simple
```kt
val canHighlightNumber = !searchQuery.isNullOrBlank() && patientSearchResult.phoneNumber.contains(searchQuery)
```

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -90,13 +94,42 @@ class PatientSearchResultItemView(
...
+    return if (canHighlight) {
+      Name.Highlighted(
+          patientName = showPatientNameAgeAndGender(patientSearchResult.fullName, patientSearchResult.gender, dateOfBirth),
+          highlightStart = patientSearchResult.fullName.indexOf(searchQuery!!, ignoreCase = true),
```
msasikanth, line 131: We can perhaps extract out this index as a variable since this is being used as part of `highlightEnd` as well?

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -90,13 +94,42 @@ class PatientSearchResultItemView(
...
+  private fun showPatientNameAgeAndGender(fullName: String, gender: Gender, dateOfBirth: DateOfBirth): String {
     genderLabel.setImageResource(gender.displayIconRes)
 
     val ageValue = dateOfBirth.estimateAge(userClock)
 
     val genderLetter = resources.getString(gender.displayLetterRes)
-    patientNameAgeGenderLabel.text = resources.getString(R.string.patientsummary_toolbar_title, fullName, genderLetter, ageValue.toString())
+
+    return resources.getString(R.string.patientsummary_toolbar_title, fullName, genderLetter, ageValue.toString())
+
```
msasikanth, line 161: Unnecessary empty space

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -90,13 +94,42 @@ class PatientSearchResultItemView(
     addressLabel.text = address.completeAddress
   }
 
-  private fun renderPatientNameAgeAndGender(fullName: String, gender: Gender, dateOfBirth: DateOfBirth) {
+  private fun highlightPatientName(searchQuery: String?, patientSearchResult: PatientSearchResultViewModel, dateOfBirth: DateOfBirth): Name {
+    val canHighlight = searchQuery?.let { searchQuery.isNotBlank() && patientSearchResult.fullName.contains(it, ignoreCase = true) } == true
+    return if (canHighlight) {
+      Name.Highlighted(
+          patientName = showPatientNameAgeAndGender(patientSearchResult.fullName, patientSearchResult.gender, dateOfBirth),
```
msasikanth, line 130: We are calling `showPatientNameAgeAndGender` irrespective of whether it's highlighted or not. Let's extract it out as a variable inside this function. Also, do we need to have it as a separate function considering it's just getting string resources?

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -90,13 +94,42 @@ class PatientSearchResultItemView(
     addressLabel.text = address.completeAddress
   }
 
-  private fun renderPatientNameAgeAndGender(fullName: String, gender: Gender, dateOfBirth: DateOfBirth) {
+  private fun highlightPatientName(searchQuery: String?, patientSearchResult: PatientSearchResultViewModel, dateOfBirth: DateOfBirth): Name {
```
msasikanth, line 126: Let's call this function `patientName` considering it can also return a plain patient name and not just highlighted ones.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -51,11 +51,25 @@ class PatientSearchResultItemView(
...
     renderPatientDateOfBirth(model.dateOfBirth)
-    renderPatientPhoneNumber(model.phoneNumber)
+    val number = highlightPatientPhoneNumber(searchQuery, model)
+    renderPatientPhoneNumber(number)
     renderVisited(model.lastSeen)
     renderLastSeen(model.lastSeen, currentFacilityId)
   }
 
+  private fun highlightPatientPhoneNumber(searchQuery: String?, patientSearchResult: PatientSearchResultViewModel): PhoneNumber {
+    val canHighlightNumber = searchQuery?.let { searchQuery.isNotBlank() && patientSearchResult.phoneNumber.toString().contains(it) } == true
```
msasikanth: We can simplify this
```kt
val canHighlightNumber = !searchQuery.isNullOrBlank() && patientSearchResult.phoneNumber.contains(searchQuery)
```

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -51,11 +51,25 @@ class PatientSearchResultItemView(
...
     renderPatientAddress(model.address)
     renderPatientDateOfBirth(model.dateOfBirth)
-    renderPatientPhoneNumber(model.phoneNumber)
+    val number = highlightPatientPhoneNumber(searchQuery, model)
+    renderPatientPhoneNumber(number)
     renderVisited(model.lastSeen)
     renderLastSeen(model.lastSeen, currentFacilityId)
   }
 
+  private fun highlightPatientPhoneNumber(searchQuery: String?, patientSearchResult: PatientSearchResultViewModel): PhoneNumber {
```
msasikanth: Same comment as other highlight function, let's name this `patientPhoneNumber` since it can return a plain phone number as well.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -51,11 +51,25 @@ class PatientSearchResultItemView(
...
+    val canHighlightNumber = searchQuery?.let { searchQuery.isNotBlank() && patientSearchResult.phoneNumber.toString().contains(it) } == true
+    return if (canHighlightNumber) {
+      PhoneNumber.Highlighted(
+          patientNumber = patientSearchResult.phoneNumber.toString(),
+          highlightStart = patientSearchResult.phoneNumber.toString().indexOf(searchQuery!!, ignoreCase = true),
```
msasikanth, line 67: We can extract out the index to a variable since it's being used for both `highlightStart` & `highlightEnd`

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -51,11 +51,25 @@ class PatientSearchResultItemView(
...
+    renderPatientPhoneNumber(number)
     renderVisited(model.lastSeen)
     renderLastSeen(model.lastSeen, currentFacilityId)
   }
 
+  private fun highlightPatientPhoneNumber(searchQuery: String?, patientSearchResult: PatientSearchResultViewModel): PhoneNumber {
+    val canHighlightNumber = searchQuery?.let { searchQuery.isNotBlank() && patientSearchResult.phoneNumber.toString().contains(it) } == true
+    return if (canHighlightNumber) {
+      PhoneNumber.Highlighted(
+          patientNumber = patientSearchResult.phoneNumber.toString(),
```
msasikanth, line 66: Since phone number can be nullable, calling `toString` might be problematic considering it will print literal `null` as text. You can perhaps use `orEmpty`.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -90,16 +92,22 @@ class PatientSearchResultItemView(
...
-      is PhoneNumber.Highlighted -> {
-        val highlightNumber = SpannableStringBuilder(number.patientNumber)
-        val highlightColor = context.resolveColor(colorRes = R.color.simple_light_blue_100)
-        highlightNumber.setSpan(BackgroundColorSpan(highlightColor), number.highlightStart, number.highlightEnd, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
-        phoneNumberTextView.text = highlightNumber
-      }
-      is PhoneNumber.Plain -> {
-        phoneNumberTextView.text = number.patientNumber
+  private fun renderPatientPhoneNumber(searchQuery: String?, patientSearchResult: PatientSearchResultViewModel) {
+    phoneNumberContainer.visibleOrGone(patientSearchResult.phoneNumber.isNullOrBlank().not())
```
msasikanth, line 96: Since this function is causing the code climate nested conditional issue, instead of separting them out to a separate function. we can do something like this (this solution is just for this scenario).

```kt
if (patientSearchResult.phoneNumber == null) {
  phoneNumberContainer.visibility = View.GONE
  return
}

phoneNumberContainer.visibility = View.VISIBLE
```

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -51,14 +51,16 @@ class PatientSearchResultItemView(
     renderPatientNameAgeAndGender(searchQuery, model)
```
msasikanth, line 51: Could we have made the last commit and this commit into one? Considering the last commit is where we removed the visibility of the `phoneNumberContainer`. WDYT?

msasikanth (changes requested):
Left few comments PTAL

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
Honey14: This commit only has the patient name highlighting. I can squash the phone number highlight and the visibility commit into one.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
Honey14: How about `getPatientName` ?

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
Honey14: I think yes, we already are doing something in this function. Keeping a different function for this makes sense. WDYT?

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
msasikanth: Oh, I was not talking about this function. I was talking about the commit. I had to select somewhere to leave a comment so I selected the top of the file in this commit. and yeah let's combine these 2 commits.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
msasikanth: Sure.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
Honey14: If phoneNumber is null, it wouldn't go inside the `if` condition at all.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
msasikanth: What we are doing is creating the patient name. So it makes sense to include it in this function rather than having it as a separate function. As long as it doesn't trigger a code climate issue we can include getting strings in this function.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
msasikanth: I think this comment was left in this commit where the null check was not explicitly set. So it can print out the `Plain` number as `null`.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
Honey14: So we're not creating patient name, instead checking whether the patient name is supposed to be highlighted or not. Setting the patient name into a different function just looks for readable to me.

msasikanth: We are constructing `Name.Highlighted` and `Name.Plain` here, right? and we also need to pass in the string of patient name, age and gender to those classes. Doesn't that mean we constructing/creating the patient name here? Even if we want to leave it as a separate function then `showPatientNameAgeAndGender` is getting the strings from resources, the function name implies that we are displaying the text.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -76,10 +93,25 @@ class PatientSearchResultItemView(
...
-    if (phoneNumber != null) {
-      phoneNumberTextView.text = phoneNumber
+  private fun renderPatientPhoneNumber(searchQuery: String?, patientSearchResult: PatientSearchResultViewModel) {
+    if (patientSearchResult.phoneNumber == null) {
+      phoneNumberContainer.visibility = View.GONE
+      return
+    }
+
+    phoneNumberContainer.visibility = View.VISIBLE
+    val canHighlightNumber = !searchQuery.isNullOrBlank() && patientSearchResult.phoneNumber.toString().contains(searchQuery)
```
msasikanth, line 103: We don't have to call `toString` here. We can directly call `contains`. Since we are already checking for the `phonenumber` null above. The phone number will be smart casted to non-null.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -90,13 +94,36 @@ class PatientSearchResultItemView(
...
+    } else {
+      Name.Plain(getPatientNameAgeAndGender)
+    }
+  }
 
-    val genderLetter = resources.getString(gender.displayLetterRes)
-    patientNameAgeGenderLabel.text = resources.getString(R.string.patientsummary_toolbar_title, fullName, genderLetter, ageValue.toString())
+  private fun renderPatientNameAgeAndGender(searchQuery: String?, model: PatientSearchResultViewModel) {
+
+    when (val name = getPatientName(searchQuery, model, DateOfBirth.fromPatientSearchResultViewModel(model, userClock))) {
```
msasikanth, line 148: Let's try to do this instead.

```kt
val patientName = when (val name = getPatientName()) {
// return the highlighted and plain patient name accordingly
}
patientNameAgeGenderLabel.text = patientName
```

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -76,10 +93,25 @@ class PatientSearchResultItemView(
...
+  private fun renderPatientPhoneNumber(searchQuery: String?, patientSearchResult: PatientSearchResultViewModel) {
+    if (patientSearchResult.phoneNumber == null) {
+      phoneNumberContainer.visibility = View.GONE
+      return
+    }
+
+    phoneNumberContainer.visibility = View.VISIBLE
+    val canHighlightNumber = !searchQuery.isNullOrBlank() && patientSearchResult.phoneNumber.toString().contains(searchQuery)
+
+    when (val number = getPatientPhoneNumber(canHighlightNumber, patientSearchResult, searchQuery)) {
```
msasikanth, line 105: Let's try to do this instead

```kt
val phoneNumber= when (val number= getPatientPhoneNumber()) {
// return the highlighted and plain patient name accordingly
}
phoneNumberTextView.text = phoneNumber
```

msasikanth (changes requested):

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
msasikanth, line 103: We can move this inside `getPatientPhoneNumber` since that's the only place it's being used and we are already passing `PatientSearchResultViewModel` and search query there.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -51,11 +52,27 @@ class PatientSearchResultItemView(
...
     renderVisited(model.lastSeen)
     renderLastSeen(model.lastSeen, currentFacilityId)
   }
 
+  private fun getPatientPhoneNumber(
+      canHighlightNumber: Boolean,
+      patientSearchResult: PatientSearchResultViewModel,
+      searchQuery: String?
+  ): PhoneNumber {
+    val getIndexOfSearchedPhoneNumber = patientSearchResult.phoneNumber.toString().indexOf(searchQuery!!)
```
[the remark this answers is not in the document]
Honey14: We don't need to have ignoreCase for the number, right?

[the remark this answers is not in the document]
msasikanth: Ah, right. I completely forgot this is a number 😂

msasikanth (approved):

PR #2290, opened 2021-01-13
Move handling of the UI renderer to the base screens in the new navigation framework

vinaysshenoy:


app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
```diff
@@ -79,7 +79,11 @@ abstract class BaseScreen<K : ScreenKey, B: ViewBinding, M : Parcelable, E, F, R
   override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
     super.onViewCreated(view, savedInstanceState)
 
-    controller.connect(Connectables.contramap({ it }, ::connectViews))
+    val uiRenderer = uiRenderer()
+    controller.connect(
+        Connectables.contramap({ it })
```
msasikanth: Since we are not modifying the model before rendering, should be stop using `contramap` now?

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
```diff
@@ -77,11 +72,8 @@ abstract class BaseScreen<K : ScreenKey, B: ViewBinding, M : Parcelable, E, F, R
   override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
     super.onViewCreated(view, savedInstanceState)
 
-    val uiRenderer = uiRenderer()
-    controller.connect(
-        Connectables.contramap({ it })
-        { eventConsumer -> connectViews(eventConsumer, uiRenderer) }
-    )
+    val rxBridge = RxMobiusBridge(events(), uiRenderer())
+    controller.connect(Connectables.contramap({ it }, rxBridge))
```
msasikanth, line 76: Same comment as above, should we stop using `contramap`? we can re-introduce it if needed. At this point it seems like we are not modifying the model data before rendering anywhere.

msasikanth (changes requested):
LGTM, left a comment PTAL. Also let's update the changelog for this change?

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
vinaysshenoy: Yeah, this makes sense. I'll do this.

msasikanth (approved):

PR #2296, opened 2021-01-18
Change `TheActivity` to load the initial screen via the Mobius loop

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/2270/migrate-theactivity-to-the-new-navigation-framework


vinaysshenoy:
Closing this because it was more complex than I thought. I managed to figure out a different approach to fixing the problem I faced.

PR #2300, opened 2021-01-20
Material Theming Migration - 4

msasikanth:
https://app.clubhouse.io/simpledotorg/story/1780/material-theming-migration


app/src/main/java/org/simple/clinic/main/TheActivity.kt
```diff
@@ -271,7 +271,7 @@ class TheActivity : AppCompatActivity(), TheActivityUi {
     screenRouter.registerKeyChanger(FullScreenKeyChanger(
         activity = this,
         screenLayoutContainerRes = android.R.id.content,
-        screenBackgroundRes = R.color.window_background,
```
vinaysshenoy, line 274: Do we need to change this since we are going to be removing this class once we finish migrating to the new navigation framework?

app/src/main/res/layout/screen_manual_patient_entry.xml
```diff
@@ -133,8 +133,8 @@
           android:padding="@dimen/spacing_4"
           android:textAppearance="?attr/textAppearanceBody1Numeric"
           android:textColor="?attr/colorOnBackground"
-          tools:drawableStart="@drawable/patient_id_card"
-          tools:drawableTint="@color/color_on_surface_67"
+          app:drawableStartCompat="@drawable/patient_id_card"
+          app:drawableTint="@color/color_on_surface_67"
```
vinaysshenoy, line 137: We still use a `TextView` here, so how come this attribute is available now? 🤔

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
```diff
@@ -234,19 +229,8 @@ class ShortCodeSearchResultScreen(
...
-    return Truss()
-        .pushSpan(textSpan)
-        .append(formattedShortCode)
-        .popSpan()
-        .build()
-
-  }
-
-  private fun formatShortCodeForDisplay(context: Context, shortCode: String): CharSequence {
-    val textSpacingSpan = TextAppearanceWithLetterSpacingSpan(context, R.style.TextAppearance_Simple_Body0_Numeric)
```
vinaysshenoy, line 248: The reason we were doing this in this way was because the short code had a different letter spacing than the rest of the text.

Have we changed the design to no longer require a different letter spacing for the short code?

app/src/main/res/values/type.xml
```diff
@@ -5,6 +5,7 @@
     <item name="android:textSize">40sp</item>
```
vinaysshenoy, line 5: The commit message seems incomplete, please recheck?

app/src/main/res/layout/link_id_with_patient_view.xml
```diff
@@ -24,8 +24,7 @@
     android:background="?attr/colorSurface"
     android:elevation="20dp"
     app:cardCornerRadius="4dp"
-    app:contentPaddingBottom="4dp"
-    tools:ignore="UnusedAttribute">
+    app:shapeAppearanceOverlay="">
```
vinaysshenoy: Was this change meant to be part of the next commit?

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseBottomSheet.kt
```diff
@@ -67,7 +67,9 @@ abstract class BaseBottomSheet<K : ScreenKey, B : ViewBinding, M : Parcelable, E
 
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
-    setStyle(DialogFragment.STYLE_NORMAL, org.simple.clinic.R.style.Clinic_V2_Theme_BottomSheetFragment)
+    // TODO : It's not necessary since we applied the default style in theme.
+    //  Once we can verify the bottom sheet with new navigation framework. We can remove this.
```
vinaysshenoy, line 71: One thing I have started doing is no longer adding `TODO` statements in the code and putting them in Clubhouse with a `maintenance` label since nobody looks at `TODO` statements. Can we do that here?

quality/lint.xml
```diff
@@ -3,29 +3,7 @@
   <issue id="GoogleAppIndexingWarning" severity="informational" />
   <issue id="AllowBackup" severity="informational" />
   <issue id="UnusedResources">
-    <!-- TODO: Remove these once Material theming migration is finished -->
-    <ignore path="src/main/res/color/contact_doctor_button.xml" />
-    <ignore path="src/main/res/color/material_button_background.xml" />
-    <ignore path="src/main/res/color/selectcountry_radio_button.xml" />
     <ignore path="src/main/res/values/dimens.xml" />
```
vinaysshenoy, line 6: Hmm, why is this file in unused resources? It is something that is referenced, right?

app/src/main/res/layout-land/activity_signature.xml
```diff
@@ -98,6 +99,6 @@
 
     </androidx.constraintlayout.widget.ConstraintLayout>
 
-  </androidx.cardview.widget.CardView>
```
vinaysshenoy, line 101: We should consider adding lint checks to stop people from using the `androidx.cardview.widget.CardView` widget.

vinaysshenoy (changes requested):

app/src/main/res/layout/screen_manual_patient_entry.xml
msasikanth: Material inflater will automatically converts this to `MaterialTextView` which extends `AppCompatTextView`. So it's safe to use this. AppCompat should have been doing the same thing as well, but I am not sure what the problem was earlier 🤷🏾

app/src/main/res/layout/list_medical_history_question.xml
```diff
@@ -29,47 +30,29 @@
       android:textColor="?attr/colorOnSurface"
       tools:text="Kidney disease" />
 
-    <LinearLayout
+    <com.google.android.material.chip.ChipGroup
```
vinaysshenoy, line 33: One thing which I believe we used to do earlier was set the layout such that the width of the "Yes/No" widget would be consistent regardless of whether it was checked or not. What I see with the new chip widgets is that the width changes a lot when checking and unchecking the buttons. Would you be able to check this?

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
msasikanth: The design is the same, instead of applying the text appearance at runtime, I have applied it to the `Toolbar` directly. `app:titleTextAppearance="?attr/textAppearanceBody0Numeric"` Since we don't have 2 different styles for the title it seemed easier to apply it directly rather than using spans.

app/src/main/res/layout/link_id_with_patient_view.xml
msasikanth: Ah, yes. My bad, I can discard this if you want.

quality/lint.xml
msasikanth: I think it's probably some of the dimens that are being unused. Let me clean that up as well while at it.

app/src/main/res/layout-land/activity_signature.xml
msasikanth: Yep, we need to actually do that for few of the widgets that have MDC counter parts

app/src/main/java/org/simple/clinic/main/TheActivity.kt
msasikanth: This was just so I can remove all the older colors. It's not actually required but paves way for me to add the `resolveColor` method that can fetch color for a given theme attribute.

app/src/main/res/values/type.xml
msasikanth: Actually, that is it 😅 I have added the reasoning in the description.

app/src/main/res/layout/list_medical_history_question.xml
msasikanth: There will be some width change, according to the mockups and the normal component. But the reason why it seems too much is probably because of the 12dp start and end padding I have applied. I have to change it to 8dp and it seems fine.

app/src/main/res/layout/link_id_with_patient_view.xml
msasikanth: Fixed it

app/src/main/res/layout-land/activity_signature.xml
vinaysshenoy: Cool, can we add a task in Clubhouse for this?

app/src/main/res/values/type.xml
vinaysshenoy: I meant the description itself seems to be incomplete.

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
vinaysshenoy: Right, I think originally the title had two different text appearances which was why we went with this approach.

vinaysshenoy (changes requested):
Left a couple of comments, please take a look?

vinaysshenoy (approved):

PR #2301, opened 2021-01-20
Move to Dagger assisted inject

msasikanth:
- Bump dagger dependency
- Move to Dagger AssistedInject
- Drop assisted inject dependency


janhavisinghh (approved):
LGTM

PR #2307, opened 2021-01-25
Fix BP Passport sheet reopens after it is dismissed

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/2421/the-bp-passport-scan-sheet-is-continuously-being-shown


Honey14 (approved):
Looks good

PR #2310, opened 2021-01-27
Migrate `BpPassportSheet` to bottom sheet screen

msasikanth:


app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
```diff
@@ -77,9 +83,27 @@ class BpPassportSheet : BottomSheetActivity(), BpPassportUiActions {
...
+  override fun bindView(inflater: LayoutInflater, container: ViewGroup?): SheetBpPassportBinding {
+    return SheetBpPassportBinding.inflate(layoutInflater)
+  }
+
+  override fun events(): Observable<BpPassportEvent> {
+    return Observable
+        .merge(
+            registerNewPatientClicks(),
+            addToExistingPatientClicks()
+        )
```
vinaysshenoy: We would need to report these to analytics. I see that this was not done in the original implementation as well, so analytics events from this screen are absent even in the latest build.

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
```diff
@@ -1,6 +1,5 @@
 package org.simple.clinic.bp.assignbppassport
```
vinaysshenoy, line 1: Typo in commit message. It should be "sending".

app/src/main/java/org/simple/clinic/bp/assignbppassport/BpPassportSheet.kt
```diff
@@ -77,9 +83,27 @@ class BpPassportSheet : BottomSheetActivity(), BpPassportUiActions {
...
 
   private val bpPassportIdentifier: Identifier by lazy {
-    intent.getParcelableExtra(KEY_BP_PASSPORT_NUMBER)!!
+    screenKey.identifier
+  }
+
+  override fun defaultModel() = BpPassportModel.create(bpPassportIdentifier)
+
+  override fun bindView(inflater: LayoutInflater, container: ViewGroup?): SheetBpPassportBinding {
+    return SheetBpPassportBinding.inflate(layoutInflater)
```
vinaysshenoy, line 58: We should use the version of the `inflate()` method that takes in the `container` parameter as well so that the layout inflater can set view attributes correctly.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreen.kt
```diff
@@ -322,3 +339,6 @@ class InstantSearchScreen :
     fun inject(target: InstantSearchScreen)
   }
 }
+
+@Parcelize
+object BlankBpPassport : Parcelable
```
vinaysshenoy, line 321: We can make this an inner type in `InstantSearchScreen` instead of a global one.

vinaysshenoy (changes requested):
Overall PR looks good. Left a few comments.

Can we also update the CHANGELOG?

vinaysshenoy (approved):

PR #2311, opened 2021-01-28
Bump Mobius dependency

msasikanth:


msasikanth:
@vinaysshenoy while bumping the Mobius dependency I have also tried out using `MobiusLoopViewModel`. Take a look at this approach when possible.

One of the advantages of this approach is it would allow us to continue any in-flight transactions during config changes, so the loop gets disposed of when the ViewModel is cleared. We will have a new type called `ViewEffect` that would let us do the UI actions in a lifecycle safe manner now that the loop doesn't get disposed of during config changes.

All the changes for model and view effects will be returned in a lifecycle aware manner (either as LiveData or LiveQueue) and allow us to observe based on the `viewLifecycleOwner` of the `Fragment`. view effects observer will also give us the option to listen to all the UI actions that happened when the app is in a pause state and let us handle it when the app goes back to the resumed state.

This is mainly an experiment to see what kind of changes we have to make to support this. We have to move to use ViewEffects instead of UI actions, adjust the test case to assert outgoing view effects as well. Apart from that, there aren't any changes as far as I can tell in the implementation details for individual screens/views.

We don't have to merge this in if you think these changes are too much, I will raise a different PR with just the dependency update.

cc: @Honey14 @janhavisinghh

msasikanth:
Closing this will make changes to the Mobius approach post navigation framework migration.

PR #2316, opened 2021-02-01
DRAFT: Jetpack Compose Experiment

msasikanth:
This is just an experiment and will not go into production. Just trying out different things to evaluate the complexity involved when migrating things over.

This PR is just for collecting thoughts about some of the changes that are going to happen.

TODO:
- [x] Create a separate module for the Simple design system using Compose 
- [ ] Create a sample app to demonstrate the Simple Design System. 
- [x]  Create Simple Design System Theme (`SimpleTheme`). While `MaterialTheme` offers certain theme attributes that are useful it will not cover all the colors & type attributes we have. In order to do that we need to implement a custom theme.


msasikanth:
>~~- [ ] Move compose experiments into a separate module allowing us to not use the new Kotlin IR in the main module~~ looks like it doesn't matter since modules that depend on another module with new IR will automatically need new IR as well.

On the other hand, it makes sense to have it in a separate module. That way we can continue experimenting with it using the demo app and integrate it into the main `app` module when the compose hits beta/stable. It would also allow us to do small scale experiment with our custom widgets. Will make a separate module for now.

msasikanth:
So if we decided to go with custom `SimpleTheme` implementation, we would have to override components from material library to take the colours from our `SimpleColors` instead of material color palette. While it can provide a type safety for us to make sure everyone is using the components from our design system. It would also be work for us to create & constantly maintain them if there are any changes from designers. If our whole point is just to have a few custom attributes on top of material theme and only override components where needed and create custom components for common usages would be a better choice.

msasikanth:
Closing this. Will pick up later when we are implementing Compose in the app.

PR #2317, opened 2021-02-01
Fix app crash when returning to `PatientEntryScreen` from `MedicalHistoryScreen`

Honey14:
https://app.clubhouse.io/simpledotorg/story/1274/app-crashes-when-returning-to-the-register-patient-screen-before-completing-registration-when-patient-date-of-birth-is-entered

### DIFF
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
@@ -210,10 +210,14 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
 
   // FIXME This is temporally coupled to `scrollToFirstFieldWithError()`.
   private val allTextInputFields: List<EditText> by unsafeLazy {
+    val ageOrDateOfBirthEditText = if (ageEditTextInputLayout.visibility == View.VISIBLE) {
+      ageEditText
+    } else {
+      dateOfBirthEditText
+    }
     listOf(
         fullNameEditText,
-        ageEditText,
-        dateOfBirthEditText,
+        ageOrDateOfBirthEditText,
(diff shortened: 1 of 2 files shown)


vinaysshenoy (changes requested):
This change fixes the issue of the age field being visible when returning to the screen, but the crash still happens if you attempt to type. Can you look at this again, please?

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
```diff
@@ -210,10 +210,14 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
 
   // FIXME This is temporally coupled to `scrollToFirstFieldWithError()`.
   private val allTextInputFields: List<EditText> by unsafeLazy {
+    val ageOrDateOfBirthEditText = if (ageEditTextInputLayout.visibility == View.VISIBLE) {
+      ageEditText
+    } else {
+      dateOfBirthEditText
+    }
```
vinaysshenoy, line 212: This is not a good solution because it won't handle the case where we clear whatever field we were typing in and then start typing in the other, right?

vinaysshenoy (changes requested):
Left a comment about the implemented solution not handling some cases.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
Honey14: Just tried testing the scenario and it works for both of them. Can you try testing the same as well?

vinaysshenoy: Yup, it seems to be working. LGTM.

vinaysshenoy (approved):

PR #2337, opened 2021-02-03
Update `PatientStatus` to migrated

Honey14:
https://app.clubhouse.io/simpledotorg/story/2359/update-patient-migrated-status-from-call-result

### DIFF
--- a/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientEffect.kt
+++ b/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientEffect.kt
@@ -47,3 +47,7 @@ data class CancelAppointment(
 data class MarkPatientAsMovedToPrivate(
     val patientUuid : UUID
 ): ContactPatientEffect()
+
+data class MarkPatientAsTransferredToAnotherFacility(
+    val patientUuid: UUID
+) : ContactPatientEffect()
--- a/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientEffect.kt
+++ b/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientEffect.kt
@@ -6,7 +6,7 @@ import java.util.UUID
 
 sealed class ContactPatientEffect
 
-data class LoadPatientProfile(val patientUuid: UUID): ContactPatientEffect()
+data class LoadPatientProfile(val patientUuid: UUID) : ContactPatientEffect()
 
 data class LoadLatestOverdueAppointment(val patientUuid: UUID) : ContactPatientEffect()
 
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/simple/clinic/contactpatient/ContactPatientEffect.kt
```diff
@@ -47,3 +47,7 @@ data class CancelAppointment(
 data class MarkPatientAsMovedToPrivate(
     val patientUuid : UUID
 ): ContactPatientEffect()
+
```
vinaysshenoy, line 50: Same feedback as earlier about a non-descriptive commit message. Please reword?

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientEffect.kt
```diff
@@ -47,3 +47,7 @@ data class CancelAppointment(
...
+data class RemoveAppointment(
```
vinaysshenoy, line 51: We already have a `CancelAppointment` effect. Do we need a new effect here, when that one accomplishes the same task?

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientEffectHandler.kt
```diff
@@ -46,9 +46,19 @@ class ContactPatientEffectHandler @AssistedInject constructor(
...
+        .addTransformer(MarkPatientAsTransferredToAnotherFacility::class.java, markPatientAsMovedToPublic(schedulers.io()))
         .build()
   }
 
+  private fun markPatientAsMovedToPublic(scheduler: Scheduler): ObservableTransformer<MarkPatientAsTransferredToAnotherFacility, ContactPatientEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(scheduler)
+          .doOnNext { patientRepository.updatePatientStatusToMigrated(patientUuid = it.patientUuid) }
+          .map { PatientMarkAsMigrated }
```
vinaysshenoy, line 58: The `RemoveAppointment` effect which is triggered by this event will end up hard-coding the appointment cancel reason to `MovedToPrivatePractioner` which is the wrong reason here.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientEffectHandler.kt
Honey14: Got it. Removing the `RemoveAppointment` effect and continue using the `CancelAppointment` effect

vinaysshenoy (approved):

PR #2344, opened 2021-02-08
During Sign-up, add an explicit 'NEXT' button while user is on the phone number input step.

msasikanth:
https://app.clubhouse.io/simpledotorg/story/1636/during-sign-up-and-an-explicit-next-button-while-user-is-on-the-phone-number-input-step


app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreen.kt
```diff
@@ -110,6 +115,11 @@ class RegistrationPhoneScreen :
           .editorActions { it == EditorInfo.IME_ACTION_DONE }
           .map { RegistrationPhoneDoneClicked() }
 
+  private fun nextClicks() =
+      nextButton
+          .clicks()
+          .map { RegistrationPhoneDoneClicked() }
+
```
vinaysshenoy, line 122: Instead of doing this, should we instead have a single stream for `RegistrationPhoneDoneClicked` where the two separate streams are merged into one?

The reason why this might be a better option is that if we need to modify behavior of these streams in some way, it's easy to miss doing it in one of the two separate methods.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneScreen.kt
msasikanth: Yeah, good point. I forgot we can merge streams 😅

vinaysshenoy (approved):

PR #2350, opened 2021-02-08
Migrate `ShortCodeSearchResultScreen` to a fragment

vinaysshenoy:


app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
```diff
@@ -144,16 +146,20 @@ class ShortCodeSearchResultScreen :
...
+    super.onDestroyView()
+    patientSearchViewBinding = null
+  }
 
-    viewBinding = ScreenShortcodeSearchResultBinding.bind(this)
-    patientSearchViewBinding = PatientSearchViewBinding.bind(viewBinding!!.root)
-    
+  override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+    super.onViewCreated(view, savedInstanceState)
+    (view as ViewGroup).hideKeyboard()
```
msasikanth, line 136: We don't have to change it now, but we could have used `binding.root` instead of casting it to `ViewGroup`

msasikanth (approved):
LGTM

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
vinaysshenoy: Oh yes, I'll change that in another PR.

PR #2375, opened 2021-02-10
Fix prescription entry sheet UI

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2594/fix-prescription-entry-sheet-ui


Honey14 (approved):
Looks good

PR #2376, opened 2021-02-11
Use progress indicator for loading in facility picker

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2569/use-progress-indicator-for-loading-in-facility-picker


app/src/main/java/org/simple/clinic/facilitypicker/FacilityPickerUi.kt
```diff
@@ -5,7 +5,5 @@ import org.simple.clinic.facility.change.FacilityListItem
 interface FacilityPickerUi {
   fun showProgressIndicator()
   fun hideProgressIndicator()
-  fun showToolbarWithSearchField()
-  fun showToolbarWithoutSearchField()
   fun updateFacilities(facilityItems: List<FacilityListItem>)
 }
```
janhavisinghh, line 9: Can you tell me how is switching toolbar type related to progress indicator or shimmer effect in this PR?

janhavisinghh (approved):
Left one comment PTAL. LGTM otherwise.

janhavisinghh (approved):

app/src/main/java/org/simple/clinic/facilitypicker/FacilityPickerUi.kt
msasikanth: Previously we had 2 different toolbars. One is with search bar and another one is standard toolbar. Since in facility picker we always need search bar I have removed the standard toolbar and then used the progress indicator for search toolbar.

https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1612939938210800

PR #2378, opened 2021-02-12
Migrate `PatientSummaryScreen` to a fragment

msasikanth:


app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -220,6 +226,49 @@ class PatientSummaryScreen :
...
+    return PatientSummaryModel.from(screenKey.intention, screenKey.patientUuid)
+  }
+
+  override fun bindView(layoutInflater: LayoutInflater, container: ViewGroup?): ScreenPatientSummaryBinding {
+    return ScreenPatientSummaryBinding.inflate(layoutInflater, container, false)
+  }
+
+  override fun uiRenderer(): ViewRenderer<PatientSummaryModel> {
+    return PatientSummaryViewRenderer(this)
+  }
```
vinaysshenoy: Not necessary to change this, but adding that I prefer to convert these to single line expression functions which reduces a lot of the verbosity.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -259,6 +260,11 @@ class PatientSummaryScreen :
     return effectHandlerFactory.create(this).build()
   }
 
+  override fun onAttach(context: Context) {
+    context.injector<Injector>().inject(this)
```
vinaysshenoy: I generally prefer to do this _after_ the `super.onAttach` call. The reason being that is possible that dependencies get introduced which are dependent on the attach to be completed before they get injected.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -273,18 +276,13 @@ class PatientSummaryScreen :
...
-      return
-    }
+  override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+    super.onViewCreated(view, savedInstanceState)
 
     // Not sure why but the keyboard stays visible when coming from search.
     rootLayout.hideKeyboard()
 
-    val screenDestroys: Observable<ScreenDestroyed> = detaches().map { ScreenDestroyed() }
+    val screenDestroys: Observable<ScreenDestroyed> = view.detaches().map { ScreenDestroyed() }
```
vinaysshenoy: If we have situations like this when migrating the older screens, the right way to do it now is to manually track the subscriptions in a `CompositeDisposable` and manage them explicitly in the `onViewCreated()`/`onDestroyView()` lifecycle methods. Please see `InstantSearchScreen` for an example.

We can move this to the `BaseScreen` classes at some point, but for now this is what we need to do.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -194,6 +194,7 @@ class PatientSummaryScreen :
   }
 
   override fun uiRenderer(model: PatientSummaryModel): ViewRenderer<PatientSummaryModel> {
+    modelUpdateCallback?.invoke(model)
```
vinaysshenoy, line 197: Just FYI, this method is only invoked once when the Mobius loop is setup. This seems like something we want to be invoked for every model update, right?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: I have done that in a later commit, where I traced `ScreenDestroyed` using `PublishSubject`.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: Ah, okay. I have gotten used to this approach since it was like that in Dagger docs. But I can move it below the `super` call.

https://dagger.dev/dev-guide/android

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: Good point. Will start doing that for future migrations.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: I have changed the `RxMobiusBridge` to trigger every time model changes happen in the connection. It seems to be working correctly 🤔

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: What I mean is that we should not be using `detaches()` anymore in a fragment based screen and instead should be using the `onViewCreated()` and `onDestroyView()` callbacks to manage subscriptions like this.

vinaysshenoy: Oh wait, I see what you mean. In any case, can we follow one method of maintaining these kind of subscriptions? I am planning to make this kind of thing easier to manage in the base class so doing it in one way would make it faster to migrate in the future.

msasikanth: Yes, that would be great. We can expose observables from the base screen that gives the lifecycles events. That should make it easier when migrating since all we have to do is use the correct lifecycle variable. WDYT?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: This will end up creating a new instance of the UI renderer every time the model changes for all screens which seems a little wasteful. 

If what we want is for the `modelUpdateCallback` to be invoked every time the `PatientSummaryModel` is updated, we can instead change `PatientSummaryViewRenderer` to expose a `(PatientSummaryModel) -> Unit` constructor property which we can then invoke whenever the model changes and trigger the callback from there.

What do you think?

vinaysshenoy (changes requested):
Left some more comments clarifying what I was asking for earlier, please take a look?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: We could think of that, yes. I was thinking of creating a `DelegateableFragment` at some point that allows us to delegate lifecycle method callbacks, and then extract common functionality from all our `BaseScreen` types into composable delegate types. This could be a good fit for that.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: Yeah, let's follow a single approach in our codebase.

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: So if I understood that correctly, we will trigger that constructor property in the `ViewRenderer#render` method right?

vinaysshenoy: Yup

vinaysshenoy: The Mobius concept for this is a `Model Update Listener`, it just so happens for us that until now, all our `Model Update Listener` types were view renderers. A better solution would be to be able to accept a list of model update listeners, which will generally be a list with one item, a `ViewRenderer`. In specialised cases like this, we could send more than one listener, a `ViewRenderer` and maybe something that can invoke a callback like we need here.

^ Can you add a refactoring ticket to track this and the Delegatable fragment I mentioned in a previous comment?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: Tracking here: https://app.clubhouse.io/simpledotorg/story/2646/implement-delegateablefragment-to-delegate-lifecycle-method-callbacks

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: Tracking here: https://app.clubhouse.io/simpledotorg/story/2647/accept-list-of-model-update-listeners-when-creating-screens-with-mobius

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: Didn't we make the change to remove the `screenDestroys`? I still see it in the latest change set 🤔

app/src/main/java/org/simple/clinic/summary/PatientSummaryViewRenderer.kt
```diff
@@ -3,10 +3,13 @@ package org.simple.clinic.summary
 import org.simple.clinic.mobius.ViewRenderer
 
 class PatientSummaryViewRenderer(
-    private val ui: PatientSummaryScreenUi
+    private val ui: PatientSummaryScreenUi,
+    private val modelUpdateCallback: PatientSummaryModelUpdateCallback?
```
vinaysshenoy, line 7: Is this nullable only because we don't want to pass it in during the tests? In this case, it should not be nullable and we should have a `mock` or a no-op used in tests.

app/src/test/java/org/simple/clinic/summary/PatientSummaryViewRendererTest.kt
```diff
@@ -36,7 +36,7 @@ class PatientSummaryViewRendererTest {
   private val defaultModel = PatientSummaryModel.from(ViewExistingPatient, UUID.fromString("6fdf088e-f6aa-40e9-9cc2-22e197b83470"))
   private val ui = mock<PatientSummaryScreenUi>()
 
-  private val uiRenderer = PatientSummaryViewRenderer(ui)
+  private val uiRenderer = PatientSummaryViewRenderer(ui, modelUpdateCallback = null)
```
vinaysshenoy, line 39: Can we make a mock object instead of `null` and verify that that the listener was invoked whenever the model changes with the correct model?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
     - "Transferred to another public health facility"
     - "Moved to private practitioner"
 - Build only the `PRODUCTION` release APK as part of the CI checks
+- Migrate `PatientSummaryScreen` to a fragment
```
vinaysshenoy, line 14: Can we put this in a different section since it's not part of the new `DEMO` release?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: I did. Well, I am no longer using detaches instead I am triggering a `PublishSubject` in `onDestroyView`
https://github.com/simpledotorg/simple-android/pull/2378/commits/cef1f65bcfebf74ad51684d893ca3927b9289b75

app/src/main/java/org/simple/clinic/summary/PatientSummaryViewRenderer.kt
msasikanth: No, the `modelUpdateCallback` used in the `PatientSummaryScreen` is nullable. So that's why I had to make it nullable here as well. Should I instead make this non-null and have a use lambda and safe call in the screen?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
vinaysshenoy: I mean, can we follow a single approach to doing this sort of thing, like we did in `InstantSearchScreen`  by using a `CompositeDisposable` to track the subscriptions instead?

CHANGELOG.md
msasikanth: I am waiting for the CHANGELOG PR to be merged in so that I can fix this when rebasing.

app/src/main/java/org/simple/clinic/summary/PatientSummaryViewRenderer.kt
msasikanth: Made it non-null and went with no-op approach in tests.

vinaysshenoy (approved):
LGTM

PR #2379, opened 2021-02-12
Remove non-critical release variant builds from the CI checks

vinaysshenoy:


msasikanth (approved):
LGTM. Should we add the manual action for security build later?

vinaysshenoy:
> LGTM. Should we add the manual action for security build later?

We don't need it because we already have it on Bitrise.

PR #2380, opened 2021-02-15
Update sorting order of prescribed drugs

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2644/change-sorting-order-of-prescription-drugs


app/src/main/java/org/simple/clinic/widgets/DividerItemDecorator.kt
```diff
@@ -5,25 +5,30 @@ import android.graphics.Canvas
...
 
   private var divider: Drawable? = null
 
   init {
-    divider = ContextCompat.getDrawable(context, dividerDrawable)
+    context.obtainStyledAttributes(ATTRS).use {
+      divider = it.getDrawable(0)
+    }
+    if (divider == null) {
+      Timber.w("@android:attr/listDivider was not set in the theme used")
```
Honey14, line 30: Should we keep the timber statements?

app/src/main/res/layout/list_prescribeddrugs_custom_drug.xml
```diff
@@ -9,7 +9,8 @@
   android:focusable="true"
   android:foreground="?selectableItemBackground"
   android:orientation="vertical"
-  app:cardCornerRadius="0dp">
+  app:cardCornerRadius="0dp"
+  app:cardElevation="0dp">
```
Honey14: Do we need card elevation if we're setting it as 0dp?

Honey14 (changes requested):
Looks good. Left few comments. 
The PRD says, selected prescribed drugs should also be in alphabetical order. Please take a look.

app/src/main/java/org/simple/clinic/widgets/DividerItemDecorator.kt
msasikanth: Yes, this is a warning log to indicate the theme level attribute is not set. Since this is not an important operation, we should not crash it. So we are showing a warning so that we can resolve this during the development cycle.

app/src/main/res/layout/list_prescribeddrugs_custom_drug.xml
msasikanth: Yes, by default card elevation will be 4dp. So we are specifically setting it to 0dp. I think at this point we can just make this a `FrameLayout` since we have a wrapper card view around the recyclerview.

msasikanth:
> Looks good. Left few comments.
> The PRD says, selected prescribed drugs should also be in alphabetical order. Please take a look.

They only need to be alphabetical in order for protocol drugs that are not assigned to the patient. So for any drugs that are assigned to the patient, we order them based on the `updatedAt`. Assigned drugs will be at top of the list and in order of `updatedAt`. Below will be protocol drugs in alphabetic order.

app/src/main/res/layout/list_prescribeddrugs_custom_drug.xml
msasikanth: I have already done it in this commit. https://github.com/simpledotorg/simple-android/pull/2380/commits/f002030a472f4be3ad9d1615451cc3ee9ca34a17

Honey14 (approved):
Looks good

PR #2407, opened 2021-02-17
Ensure all Room query threads are available

vinaysshenoy:


app/src/main/java/org/simple/clinic/storage/StorageModule.kt
```diff
@@ -35,9 +35,9 @@ class StorageModule {
     // internally.
     val sqliteThreadPoolCount = SQLiteGlobal.getWALConnectionPoolSize() / 2
     val queryExecutor = ThreadPools.create(
-        corePoolSize = 1,
+        corePoolSize = sqliteThreadPoolCount,
         maxPoolSize = sqliteThreadPoolCount,
-        threadPrefix = "room-query-"
+        threadPrefix = "room-query"
```
msasikanth, line 40: Any particular reason for removing the hyphen from the thread prefix? 🤔

vinaysshenoy: The reason is explained in the commit message.

> Remove '-' from the Room thread name prefix
>
> This was done because the `ThreadPools` factory method already adds its own '-' where needed.****

msasikanth: Oh, I missed it. Sorry 😅

janhavisinghh:
@vinaysshenoy how do I test this?

janhavisinghh (approved):

vinaysshenoy:
> @vinaysshenoy how do I test this?

I used the Android Studio profiler to look at the Room query threads to verify that it's working fine. Since it's urgent, maybe we can merge this in now and I can show you the method I used to verify this tomorrow morning?

PR #2433, opened 2021-02-22
Implement auto complete text view in patient entry/edit screens

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/2574/implement-auto-complete-text-view-in-patient-entry-edit-screens

### DIFF
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientEffectHandler.kt
@@ -1,10 +1,10 @@
 package org.simple.clinic.editpatient
 
 import com.spotify.mobius.rx2.RxMobius
+import dagger.Lazy
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
-import dagger.Lazy
 import io.reactivex.Maybe
 import io.reactivex.Observable
 import io.reactivex.ObservableTransformer
@@ -73,9 +73,19 @@ class EditPatientEffectHandler @AssistedInject constructor(
         .addTransformer(FetchBpPassportsEffect::class.java, fetchBpPassports(schedulersProvider.io()))
         .addTransformer(SavePatientEffect::class.java, savePatientTransformer(schedulersProvider.io()))
         .addTransformer(LoadInputFields::class.java, loadInputFields())
-        .addConsumer(SetupUi::class.java, { ui.setupUi(it.inputFields)}, schedulersProvider.ui())
+        .addConsumer(SetupUi::class.java, { ui.setupUi(it.inputFields) }, schedulersProvider.ui())
+        .addTransformer(FetchColonyOrVillagesEffect::class.java, fetchColonyOrVillages())
         .build()
   }
+
+  private fun fetchColonyOrVillages(): ObservableTransformer<FetchColonyOrVillagesEffect, EditPatientEvent> {
+    return ObservableTransformer { fetchColonyOrVillagesEffect ->
+      fetchColonyOrVillagesEffect
+          .map { patientRepository.allColoniesOrVillagesInPatientAddress() }
+          .map(::ColonyOrVillagesFetched)
+    }
+  }
+
 
   private fun displayBpPassports(bpPassports: List<BusinessId>) {
     val identifiers = bpPassports.map { it.identifier.displayValue() }
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientEvent.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientEvent.kt
@@ -70,3 +70,5 @@ data class AlternativeIdChanged(val alternativeId: String) : EditPatientEvent() 
 }
 
 data class InputFieldsLoaded(val inputFields: InputFields) : EditPatientEvent()
+
+data class ColonyOrVillagesFetched(val colonyOrVillages: List<String>) : EditPatientEvent()
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientModel.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientModel.kt
@@ -20,7 +20,8 @@ data class EditPatientModel(
     val savedAddress: PatientAddress,
     val savedPhoneNumber: PatientPhoneNumber?,
     val savedBangladeshNationalId: BusinessId?,
-    val saveButtonState: EditPatientState?
+    val saveButtonState: EditPatientState?,
+    val colonyOrVillagesList: List<String>?
 ) : Parcelable {
   companion object {
     fun from(
@@ -40,9 +41,12 @@ data class EditPatientModel(
           saveButtonState
       )
       val ongoingEntry = savedEntry.copy()
-      return EditPatientModel(savedEntry, ongoingEntry, patient, address, phoneNumber, bangladeshNationalId, saveButtonState)
+      return EditPatientModel(savedEntry, ongoingEntry, patient, address, phoneNumber, bangladeshNationalId, saveButtonState, null)
     }
   }
+
+  val hasColonyOrVillagesList: Boolean
+    get() = !colonyOrVillagesList.isNullOrEmpty()
 
   fun updateName(name: String): EditPatientModel =
       copy(ongoingEntry = ongoingEntry.updateName(name))
@@ -79,4 +83,7 @@ data class EditPatientModel(
 
   fun buttonStateChanged(buttonState: EditPatientState?): EditPatientModel =
       copy(saveButtonState = buttonState)
+
+  fun updateColonyOrVillagesList(colonyOrVillages: List<String>): EditPatientModel =
+      copy(colonyOrVillagesList = colonyOrVillages)
 }
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
@@ -273,6 +273,10 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
     deletePatient.setOnClickListener { router.push(DeletePatientScreenKey(screenKey.patient.uuid).wrap()) }
   }
 
+  override fun showColonyOrVillagesList(colonyOrVillageList: List<String>) {
+
+  }
+
   private fun showOrHideInputFields(inputFields: InputFields) {
     val allTypesOfInputFields: Map<Class<*>, View> = mapOf(
         PatientNameField::class.java to fullNameInputLayout,
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
@@ -5,6 +5,7 @@ import android.os.Parcelable
 import android.util.AttributeSet
 import android.view.LayoutInflater
 import android.view.View
+import android.widget.ArrayAdapter
 import android.widget.RadioButton
 import android.widget.RelativeLayout
 import androidx.appcompat.app.AppCompatActivity
@@ -274,7 +275,13 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
   }
 
   override fun showColonyOrVillagesList(colonyOrVillageList: List<String>) {
+    ArrayAdapter<String>(context, R.layout.village_typeahead_list_item,  R.id.villageTypeAheadItemTextView, colonyOrVillageList).also { adapter ->
+      colonyOrVillageEditText.setAdapter(adapter)
+    }
 
+    colonyOrVillageEditText.setOnClickListener {
+      colonyOrVillageEditText.showDropDown()
+    }
   }
 
   private fun showOrHideInputFields(inputFields: InputFields) {
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientUi.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientUi.kt
@@ -29,4 +29,5 @@ interface EditPatientUi {
   fun showProgress()
   fun hideProgress()
   fun setupUi(inputFields: InputFields)
+  fun showColonyOrVillagesList(colonyOrVillageList: List<String>)
 }
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientUpdate.kt
@@ -56,6 +56,7 @@ class EditPatientUpdate(
       is AlternativeIdChanged -> next(model.updateAlternativeId(event.alternativeId))
       is BpPassportsFetched -> dispatch(DisplayBpPassportsEffect(event.bpPasssports))
       is InputFieldsLoaded -> dispatch(SetupUi(event.inputFields) as EditPatientEffect)
+      is ColonyOrVillagesFetched -> next(model.updateColonyOrVillagesList(event.colonyOrVillages))
     }
   }
 
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientViewRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientViewRenderer.kt
@@ -21,6 +21,10 @@ class EditPatientViewRenderer(private val ui: EditPatientUi) : ViewRenderer<Edit
     ui.setDateOfBirthAndAgeVisibility(fieldToShow)
 
     manageButtonState(model)
+
+    if (model.hasColonyOrVillagesList) {
+      ui.showColonyOrVillagesList(model.colonyOrVillagesList!!)
+    }
   }
 
   private fun manageButtonState(model: EditPatientModel) {
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
@@ -81,7 +81,16 @@ class PatientEntryEffectHandler @AssistedInject constructor(
         .addAction(OpenMedicalHistoryEntryScreen::class.java, ui::openMedicalHistoryEntryScreen, schedulersProvider.ui())
         .addTransformer(LoadInputFields::class.java, loadInputFields())
         .addConsumer(SetupUi::class.java, { ui.setupUi(it.inputFields) }, schedulersProvider.ui())
+        .addTransformer(FetchColonyOrVillagesEffect::class.java, fetchColonyOrVillages())
         .build()
+  }
+
+  private fun fetchColonyOrVillages(): ObservableTransformer<FetchColonyOrVillagesEffect, PatientEntryEvent> {
+    return ObservableTransformer { fetchColonyOrVillagesEffect ->
+      fetchColonyOrVillagesEffect
+          .map { patientRepository.allColoniesOrVillagesInPatientAddress() }
+          .map(::ColonyOrVillagesFetched)
+    }
   }
 
   private fun fetchOngoingEntryTransformer(scheduler: Scheduler): ObservableTransformer<FetchPatientEntry, PatientEntryEvent> {
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryEvent.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryEvent.kt
@@ -1,5 +1,6 @@
 package org.simple.clinic.newentry
 
+import org.simple.clinic.editpatient.EditPatientEvent
 import org.simple.clinic.newentry.country.InputFields
 import org.simple.clinic.patient.Gender
 import org.simple.clinic.patient.OngoingNewPatientEntry
@@ -71,3 +72,5 @@ data class ReminderConsentChanged(val reminderConsent: ReminderConsent) : Patien
 data class AlternativeIdChanged(val identifier: Identifier) : PatientEntryEvent()
 
 data class InputFieldsLoaded(val inputFields: InputFields) : PatientEntryEvent()
+
+data class ColonyOrVillagesFetched(val colonyOrVillages: List<String>) : PatientEntryEvent()
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryInit.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryInit.kt
@@ -5,6 +5,13 @@ import com.spotify.mobius.First.first
 import com.spotify.mobius.Init
 
 class PatientEntryInit : Init<PatientEntryModel, PatientEntryEffect> {
-  override fun init(model: PatientEntryModel): First<PatientEntryModel, PatientEntryEffect> =
-      first(model, setOf(FetchPatientEntry, LoadInputFields))
+  override fun init(model: PatientEntryModel): First<PatientEntryModel, PatientEntryEffect> {
+    val effects = mutableSetOf(FetchPatientEntry, LoadInputFields)
+
+    if (!model.hasColonyOrVillagesList) {
+      effects.add(FetchColonyOrVillagesEffect)
+    }
+
+    return first(model, effects)
+  }
 }
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
@@ -283,6 +283,10 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
     showOrHideGenderRadioButtons(inputFields)
   }
 
+  override fun showColonyOrVillagesList(colonyOrVillageList: List<String>) {
+
+  }
+
   private fun showOrHideInputFields(inputFields: InputFields) {
     val allTypesOfInputFields: Map<Class<*>, View> = mapOf(
         PatientNameField::class.java to fullNameInputLayout,
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
@@ -8,6 +8,7 @@ import android.text.style.StyleSpan
 import android.util.AttributeSet
 import android.view.View
 import android.view.inputmethod.EditorInfo
+import android.widget.ArrayAdapter
 import android.widget.EditText
 import android.widget.RadioButton
 import android.widget.RelativeLayout
@@ -284,7 +285,13 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
   }
 
   override fun showColonyOrVillagesList(colonyOrVillageList: List<String>) {
+    ArrayAdapter<String>(context, R.layout.village_typeahead_list_item, R.id.villageTypeAheadItemTextView, colonyOrVillageList).also { adapter ->
+      colonyOrVillageEditText.setAdapter(adapter)
+    }
 
+    colonyOrVillageEditText.setOnClickListener {
+      colonyOrVillageEditText.showDropDown()
+    }
   }
 
   private fun showOrHideInputFields(inputFields: InputFields) {
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryUi.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryUi.kt
@@ -16,4 +16,5 @@ interface PatientEntryUi {
   fun nextButtonShowInProgress()
   fun enableNextButton()
   fun setupUi(inputFields: InputFields)
+  fun showColonyOrVillagesList(colonyOrVillageList: List<String>)
 }
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryUiRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryUiRenderer.kt
@@ -19,6 +19,10 @@ class PatientEntryUiRenderer(val ui: PatientEntryUi) : ViewRenderer<PatientEntry
     val patientEntry = model.patientEntry
     identifierValueChangedCallback.pass(patientEntry.identifier.toOptional()) { renderIdentifier(patientEntry.identifier) }
 
+    if (model.hasColonyOrVillagesList) {
+      ui.showColonyOrVillagesList(model.colonyOrVillagesList!!)
+    }
+
     val personalDetails = patientEntry.personalDetails ?: return
     changeDateOfBirthAndAgeVisibility(personalDetails)
 
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryUpdate.kt
@@ -40,6 +40,7 @@ class PatientEntryUpdate(
       is SaveClicked -> onSaveClicked(model)
       is PatientEntrySaved -> next(model.buttonStateChanged(ButtonState.SAVED), OpenMedicalHistoryEntryScreen)
       is InputFieldsLoaded -> dispatch(SetupUi(event.inputFields))
+      is ColonyOrVillagesFetched -> next(model.colonyOrVillageListUpdated(event.colonyOrVillages))
     }
   }
 
--- a/app/src/main/res/layout/screen_edit_patient.xml
+++ b/app/src/main/res/layout/screen_edit_patient.xml
@@ -227,12 +227,12 @@
 
       <com.google.android.material.textfield.TextInputLayout
         android:id="@+id/colonyOrVillageInputLayout"
-        style="@style/Widget.Simple.TextField.Layout.PatientEntry"
+        style="@style/Widget.Simple.TextField.Layout.PatientEntry.AutoComplete"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_marginTop="@dimen/spacing_16">
 
-        <com.google.android.material.textfield.TextInputEditText
+        <com.google.android.material.textfield.MaterialAutoCompleteTextView
           android:id="@+id/colonyOrVillageEditText"
           android:layout_width="match_parent"
           android:layout_height="wrap_content"
--- a/app/src/main/res/layout/village_typeahead_list_item.xml
+++ b/app/src/main/res/layout/village_typeahead_list_item.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:tools="http://schemas.android.com/tools"
+  android:layout_width="match_parent"
+  android:layout_height="wrap_content"
+  android:background="@color/white"
+  android:orientation="vertical">
+
+  <TextView
+    android:id="@+id/villageTypeAheadItemTextView"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:gravity="center_vertical"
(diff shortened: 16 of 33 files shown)


app/src/main/res/layout/screen_edit_patient.xml
```diff
@@ -227,18 +227,19 @@
...
-        <com.google.android.material.textfield.TextInputEditText
+        <com.google.android.material.textfield.MaterialAutoCompleteTextView
           android:id="@+id/colonyOrVillageEditText"
           android:layout_width="match_parent"
           android:layout_height="wrap_content"
           android:hint="@string/patientedit_colony_or_village"
           android:imeOptions="flagNoFullscreen"
-          android:inputType="textPostalAddress|textCapSentences" />
+          android:inputType="textPostalAddress|textCapSentences"
+          />
```
msasikanth, line 242: Unnecessary line break

app/src/main/java/org/simple/clinic/editpatient/EditPatientEffectHandler.kt
```diff
@@ -73,10 +73,19 @@ class EditPatientEffectHandler @AssistedInject constructor(
...
+        .addTransformer(FetchColonyOrVillagesEffect::class.java, fetchColonyOrVillages())
         .build()
   }
 
+  private fun fetchColonyOrVillages(): ObservableTransformer<FetchColonyOrVillagesEffect, EditPatientEvent> {
+    var colonyOrVillages: List<String>? = null
+    return ObservableTransformer { fetchColonyOrVillagesEffect ->
+      fetchColonyOrVillagesEffect.doOnNext { colonyOrVillages = patientRepository.allColoniesOrVillagesInPatientAddress() }
+          .map { ColonyOrVillagesFetched(colonyOrVillages) }
+    }
```
msasikanth, line 87: We should not be using a `var` variable here to pass the list information downstream. Instead, use `map`.

```kt
fetchColonyOrVillagesEffect
    .map { patientRepository.allColoniesOrVillagesInPatientAddress() }
    .map(::ColonyOrVillagesFetched)
```

app/src/main/java/org/simple/clinic/editpatient/EditPatientEvent.kt
```diff
@@ -70,3 +70,5 @@ data class AlternativeIdChanged(val alternativeId: String) : EditPatientEvent()
 }
 
 data class InputFieldsLoaded(val inputFields: InputFields) : EditPatientEvent()
+
+data class ColonyOrVillagesFetched(val colonyOrVillages: List<String>?) : EditPatientEvent()
```
msasikanth, line 74: The `List` will never be `null`. We can make the `getColonyOrVillages` as non-null. If there is no data in the db, Room database will return an empty list instead of a nullable object for lists. So we don't have to handle nullable here.

app/src/test/java/org/simple/clinic/editpatient/EditPatientEffectHandlerTest.kt
```diff
@@ -304,4 +304,17 @@ class EditPatientEffectHandlerTest {
     verify(ui).setupUi(inputFields)
     verifyNoMoreInteractions(ui)
   }
+
+  @Test
+  fun `when fetch colony or village names effect is received, then fetch them from the database`() {
+    // given
+    val colonyOrVillages = listOf("Colony1, Colony2, Colony3, Colony 4")
```
msasikanth, line 311: I think you accidentally added the `,` in the String. Here it's a list of single string, it should be `listof("Colony1", "Colony2", "Colony3", "Colony4")`

app/src/main/java/org/simple/clinic/editpatient/EditPatientUpdate.kt
```diff
@@ -56,6 +56,7 @@ class EditPatientUpdate(
       is AlternativeIdChanged -> next(model.updateAlternativeId(event.alternativeId))
       is BpPassportsFetched -> dispatch(DisplayBpPassportsEffect(event.bpPasssports))
       is InputFieldsLoaded -> dispatch(SetupUi(event.inputFields) as EditPatientEffect)
+      is ColonyOrVillagesFetched -> next(model.updateColonyOrVillagesList(event.colonyOrVillages))
```
msasikanth: Currently, it seems like you're combining both the effect handler and update functionality into a single commit. Can we separate them into different commits, going forward?

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -480,7 +480,7 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
         ColonyOrVillageEmpty -> showEmptyColonyOrVillageError(true)
         DistrictEmpty -> showEmptyDistrictError(true)
         StateEmpty -> showEmptyStateError(true)
-        BothDateOfBirthAndAgeAdsent -> showAgeEmptyError(true)
+        BothDateOfBirthAndAgeAbsent -> showAgeEmptyError(true)
```
msasikanth, line 496: Since it's not related to this PR, can we not do this here? Please open a different PR for this.

app/src/main/java/org/simple/clinic/editpatient/EditPatientUpdate.kt
```diff
@@ -56,7 +56,7 @@ class EditPatientUpdate(
       is AlternativeIdChanged -> next(model.updateAlternativeId(event.alternativeId))
       is BpPassportsFetched -> dispatch(DisplayBpPassportsEffect(event.bpPasssports))
       is InputFieldsLoaded -> dispatch(SetupUi(event.inputFields) as EditPatientEffect)
-      is ColonyOrVillagesFetched -> next(model.updateColonyOrVillagesList(event.colonyOrVillages))
+      is ColonyOrVillagesFetched -> next(model.updateColonyOrVillagesList(event.colonyOrVillages), event.colonyOrVillages?.let { ShowColonyOrVillagesListEfect(it) } as EditPatientEffect)
```
msasikanth, line 59: If we are saving it to the model already, why not use `ViewRenderer`? The effect way makes sense if we are not saving it to the model. Since you're saving it to the model, let's use the `ViewRenderer` for updating the autocomplete adapter, that way it will also be consistent across config changes and we don't have to make the DB call when the model already contains the names.

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -274,7 +275,15 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
   }
 
   override fun showColonyOrVillagesList(colonyOrVillageList: List<String>) {
+    ArrayAdapter<String>(context, android.R.layout.simple_list_item_1, colonyOrVillageList).also { adapter ->
+      colonyOrVillageEditText.setAdapter(adapter)
+    }
 
+    colonyOrVillageEditText.threshold = 0
```
msasikanth, line 282: This doesn't make any difference, by default Android applies a threshold of 1 when the threshold is 0. So we can remove this.

https://developer.android.com/reference/android/widget/AutoCompleteTextView?authuser=1#setThreshold(int)

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -274,7 +275,15 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
...
+    colonyOrVillageEditText.setOnClickListener {
+      colonyOrVillageEditText.maxLines = 6
```
msasikanth, line 284: Any particular reason for setting `maxLines`? If not please remove this.

app/src/test/java/org/simple/clinic/newentry/PatientEntryInitTest.kt
```diff
@@ -0,0 +1,24 @@
...
+    val defaultModel = PatientEntryModel.DEFAULT
+
+    initSpec.whenInit(defaultModel).then(assertThatFirst(
+        hasModel(defaultModel),
+        hasEffects(
+            FetchPatientEntry, LoadInputFields, FetchColonyOrVillagesEffect
+        )
+    ))
+
+
```
msasikanth, line 22: Unnecessary line breaks

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
```diff
@@ -81,9 +81,18 @@ class PatientEntryEffectHandler @AssistedInject constructor(
...
+        .addTransformer(FetchColonyOrVillagesEffect::class.java, fetchColonyOrVillages())
         .build()
   }
 
+  private fun fetchColonyOrVillages(): ObservableTransformer<FetchColonyOrVillagesEffect, PatientEntryEvent> {
+    var colonyOrVillages: List<String>? = null
+    return ObservableTransformer { fetchColonyOrVillagesEffect ->
+      fetchColonyOrVillagesEffect.doOnNext { colonyOrVillages = patientRepository.allColoniesOrVillagesInPatientAddress() }
+          .map { ColonyOrVillagesFetched(colonyOrVillages) }
+    }
```
msasikanth, line 94: We should not be using a `var` variable here to pass the list information downstream. Instead, use `map`.

```kt
fetchColonyOrVillagesEffect
    .map { patientRepository.allColoniesOrVillagesInPatientAddress() }
    .map(::ColonyOrVillagesFetched)
```

app/src/main/java/org/simple/clinic/newentry/PatientEntryEvent.kt
```diff
@@ -71,3 +72,5 @@ data class ReminderConsentChanged(val reminderConsent: ReminderConsent) : Patien
 data class AlternativeIdChanged(val identifier: Identifier) : PatientEntryEvent()
 
 data class InputFieldsLoaded(val inputFields: InputFields) : PatientEntryEvent()
+
+data class ColonyOrVillagesFetched(val colonyOrVillages: List<String>?) : PatientEntryEvent()
```
msasikanth, line 76: The `List` will never be `null`. We can make the `getColonyOrVillages` as non-null. If there is no data in the db, Room database will return an empty list instead of a nullable object for lists. So we don't have to handle nullable here.

app/src/main/java/org/simple/clinic/newentry/PatientEntryUpdate.kt
```diff
@@ -40,6 +40,7 @@ class PatientEntryUpdate(
       is SaveClicked -> onSaveClicked(model)
       is PatientEntrySaved -> next(model.buttonStateChanged(ButtonState.SAVED), OpenMedicalHistoryEntryScreen)
       is InputFieldsLoaded -> dispatch(SetupUi(event.inputFields))
+      is ColonyOrVillagesFetched -> next(model.colonyOrVillageListUpdated(event.colonyOrVillages))
```
msasikanth: We should ideally be having 2 different commits for handling the effect and update function.

app/src/test/java/org/simple/clinic/newentry/PatientEntryEffectHandlerTest.kt
```diff
@@ -100,6 +100,20 @@ class PatientEntryEffectHandlerTest {
     verifyZeroInteractions(validationActions)
   }
 
+  @Test
+  fun `when fetch colony or village names effect is received, then fetch them from the database`() {
+    // given
+    val colonyOrVillages = listOf("Colony1, Colony2, Colony3, Colony 4")
```
msasikanth, line 106: I think you accidentally added the `,` in the String. Here it's a list of single string, it should be `listof("Colony1", "Colony2", "Colony3", "Colony4")`

app/src/test/java/org/simple/clinic/newentry/PatientEntryEffectHandlerTest.kt
```diff
@@ -114,6 +114,20 @@ class PatientEntryEffectHandlerTest {
     verifyNoMoreInteractions(ui)
   }
 
+  @Test
+  fun `when show colony or village list effect is received, then show them in the ui`() {
+    // given
+    val colonyOrVillages = listOf("Colony1, Colony2, Colony3, Colony 4")
```
msasikanth, line 120: I think you accidentally added the `,` in the String. Here it's a list of single string, it should be `listof("Colony1", "Colony2", "Colony3", "Colony4")`

app/src/main/java/org/simple/clinic/newentry/PatientEntryUpdate.kt
```diff
@@ -40,7 +40,7 @@ class PatientEntryUpdate(
       is SaveClicked -> onSaveClicked(model)
       is PatientEntrySaved -> next(model.buttonStateChanged(ButtonState.SAVED), OpenMedicalHistoryEntryScreen)
       is InputFieldsLoaded -> dispatch(SetupUi(event.inputFields))
-      is ColonyOrVillagesFetched -> next(model.colonyOrVillageListUpdated(event.colonyOrVillages))
+      is ColonyOrVillagesFetched -> next(model.colonyOrVillageListUpdated(event.colonyOrVillages), event.colonyOrVillages?.let { ShowColonyOrVillagesListEffect(it) } as PatientEntryEffect)
```
msasikanth, line 43: If we are saving it to the model already, why not use `ViewRenderer`? The effect way makes sense if we are not saving it to the model. Since you're saving it to the model, let's use the `ViewRenderer` for updating the autocomplete adapter, that way it will also be consistent across config changes and we don't have to make the DB call when the model already contains the names.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
```diff
@@ -284,7 +285,15 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
   }
 
   override fun showColonyOrVillagesList(colonyOrVillageList: List<String>) {
+    ArrayAdapter<String>(context, android.R.layout.simple_list_item_1, colonyOrVillageList).also { adapter ->
+      colonyOrVillageEditText.setAdapter(adapter)
+    }
 
+    colonyOrVillageEditText.threshold = 0
```
msasikanth, line 292: This doesn't make any difference, by default Android applies a threshold of 1 when the threshold is 0. So we can remove this.

https://developer.android.com/reference/android/widget/AutoCompleteTextView?authuser=1#setThreshold(int)

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
```diff
@@ -284,7 +285,15 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
...
+    colonyOrVillageEditText.setOnClickListener {
+      colonyOrVillageEditText.maxLines = 6
```
msasikanth, line 294: Any particular reason for setting `maxLines`? If not please remove this.

app/src/main/java/org/simple/clinic/editpatient/EditPatientModel.kt
```diff
@@ -29,7 +30,8 @@ data class EditPatientModel(
         phoneNumber: PatientPhoneNumber?,
         dateOfBirthFormatter: DateTimeFormatter,
         bangladeshNationalId: BusinessId?,
-        saveButtonState: EditPatientState?
+        saveButtonState: EditPatientState?,
+        colonyOrVillagesList: List<String>?
```
msasikanth, line 34: Since we won't be passing the list when creating the model using this factory function we can remove this from the function param, WDYT?

app/src/test/java/org/simple/clinic/editpatient/EditPatientEffectHandlerTest.kt
```diff
@@ -317,4 +317,18 @@ class EditPatientEffectHandlerTest {
     testCase.assertOutgoingEvents(ColonyOrVillagesFetched(colonyOrVillages))
     verifyNoMoreInteractions(ui)
   }
+
+  @Test
+  fun `when show colony or village list effect is received, then show them in the ui`() {
+    // given
+    val colonyOrVillages = listOf("Colony1, Colony2, Colony3, Colony 4")
```
msasikanth, line 324: I think you accidentally added the `,` in the String. Here it's a list of single string, it should be `listof("Colony1", "Colony2", "Colony3", "Colony4")`

msasikanth (changes requested):
Left few comments, PTAL

msasikanth (changes requested):
Unit tests and Android test are failing, PTAL

app/src/main/java/org/simple/clinic/editpatient/EditPatientModel.kt
```diff
@@ -29,7 +30,7 @@ data class EditPatientModel(
         phoneNumber: PatientPhoneNumber?,
         dateOfBirthFormatter: DateTimeFormatter,
         bangladeshNationalId: BusinessId?,
-        saveButtonState: EditPatientState?
+        saveButtonState: EditPatientState?,
```
msasikanth, line 33: Not necessary to remove it in this PR. But seems like there is an extra comma here.

app/src/test/java/org/simple/clinic/editpatient/EditPatientEffectHandlerTest.kt
```diff
@@ -304,4 +304,17 @@ class EditPatientEffectHandlerTest {
...
+  fun `when fetch colony or village names effect is received, then fetch them from the database`() {
+    // given
+    val colonyOrVillages = listOf("colony1", "colony2", "colony3", "colony4")
+    whenever(patientRepository.allColoniesOrVillagesInPatientAddress()).thenReturn(colonyOrVillages)
+    // when
+    testCase.dispatch(FetchColonyOrVillagesEffect)
+
+    //then
+    testCase.assertOutgoingEvents(ColonyOrVillagesFetched(colonyOrVillages))
+    verifyNoMoreInteractions(ui)
```
msasikanth, line 318: This should be `verifyZeroInteractions`. When we are asserting to make sure no UI interactions are happening we use `verifyZeroInteractions`. We use `verifyNoMoreInteractions` if want to make that after certain interactions there are no more interactions for the given class.

app/src/main/java/org/simple/clinic/editpatient/EditPatientViewRenderer.kt
```diff
@@ -21,6 +21,8 @@ class EditPatientViewRenderer(private val ui: EditPatientUi) : ViewRenderer<Edit
     ui.setDateOfBirthAndAgeVisibility(fieldToShow)
 
     manageButtonState(model)
+
+    model.colonyOrVillagesList?.let { ui.showColonyOrVillagesList(it) }
```
msasikanth, line 25: Can we instead use an if condition along with computed properties in the model. Please refer to other ViewRenderer files

```kt
if (model.hasColonyOrVillagesList) {
  ui.showColonyOrVillagesList(model.colonyOrVillagesList!!)
}
```

app/src/test/java/org/simple/clinic/editpatient/EditPatientViewRendererTest.kt
```diff
@@ -0,0 +1,50 @@
...
+    val colonyOrVillagesFetchedState = model.updateColonyOrVillagesList(colonyOrVillages).updateDateOfBirth("2018-03-09")
+
+    // when
+    uiRenderer.render(colonyOrVillagesFetchedState)
+
+    // then
+    verify(ui).setDateOfBirthAndAgeVisibility(DATE_OF_BIRTH_VISIBLE)
+    verify(ui).showColonyOrVillagesList(colonyOrVillages)
+    verify(ui).showProgress()
+    verifyZeroInteractions(ui)
```
msasikanth, line 48: This should be `verifyNoMoreInteractions(ui)`.

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -274,7 +275,13 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
   }
 
   override fun showColonyOrVillagesList(colonyOrVillageList: List<String>) {
+    ArrayAdapter<String>(context, android.R.layout.simple_list_item_1, colonyOrVillageList).also { adapter ->
```
msasikanth, line 278: Instead of using an Android layout file, let's create one in our project. We don't want to rely on that layout being the same across different OEMs.

app/src/main/java/org/simple/clinic/newentry/PatientEntryInit.kt
```diff
@@ -6,5 +6,5 @@ import com.spotify.mobius.Init
 
 class PatientEntryInit : Init<PatientEntryModel, PatientEntryEffect> {
   override fun init(model: PatientEntryModel): First<PatientEntryModel, PatientEntryEffect> =
-      first(model, setOf(FetchPatientEntry, LoadInputFields))
+      first(model, setOf(FetchPatientEntry, LoadInputFields, FetchColonyOrVillagesEffect))
```
msasikanth, line 9: Let's add a check to make sure we are not fetching the list again when village or colony names are already present in the model.

app/src/test/java/org/simple/clinic/newentry/PatientEntryEffectHandlerTest.kt
```diff
@@ -100,6 +100,20 @@ class PatientEntryEffectHandlerTest {
...
+    // given
+    val colonyOrVillages = listOf("colony1", "colony2", "colony3", "colony4")
+    whenever(patientRepository.allColoniesOrVillagesInPatientAddress()).thenReturn(colonyOrVillages)
+    // when
+    setupTestCase()
+    testCase.dispatch(FetchColonyOrVillagesEffect)
+
+    //then
+    testCase.assertOutgoingEvents(ColonyOrVillagesFetched(colonyOrVillages))
+    verifyNoMoreInteractions(ui)
```
msasikanth, line 114: This should be `verifyZeroInteractions`. When we are asserting to make sure no UI interactions are happening we use `verifyZeroInteractions`. We use `verifyNoMoreInteractions` if want to make that after certain interactions there are no more interactions for the given class.

app/src/main/java/org/simple/clinic/newentry/PatientEntryUiRenderer.kt
```diff
@@ -19,6 +19,8 @@ class PatientEntryUiRenderer(val ui: PatientEntryUi) : ViewRenderer<PatientEntry
     val patientEntry = model.patientEntry
     identifierValueChangedCallback.pass(patientEntry.identifier.toOptional()) { renderIdentifier(patientEntry.identifier) }
 
+    model.colonyOrVillagesList?.let { ui.showColonyOrVillagesList(it) }
```
msasikanth, line 22: Can we instead use an if condition along with computed properties in the model. Please refer to other ViewRenderer files

```kt
if (model.hasColonyOrVillagesList) {
  ui.showColonyOrVillagesList(model.colonyOrVillagesList!!)
}
```

app/src/test/java/org/simple/clinic/newentry/PatientEntryUiRendererTest.kt
```diff
@@ -0,0 +1,28 @@
...
+    val colonyOrVillages = listOf("colony1", "colony2", "colony3", "colony4")
+    val colonyOrVillagesFetchedState = model.colonyOrVillageListUpdated(colonyOrVillages)
+
+    // when
+    uiRenderer.render(colonyOrVillagesFetchedState)
+
+    // then
+    verify(ui).showColonyOrVillagesList(colonyOrVillages)
+    verify(ui).hideIdentifierSection()
+    verifyZeroInteractions(ui)
```
msasikanth, line 26: We should be using `verifyNoMoreInteractions` here.

app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
```diff
@@ -284,7 +285,13 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
   }
 
   override fun showColonyOrVillagesList(colonyOrVillageList: List<String>) {
+    ArrayAdapter<String>(context, android.R.layout.simple_list_item_1, colonyOrVillageList).also { adapter ->
```
msasikanth, line 288: Let's create a layout file in our project instead of using the Android one, since we don't it will be the same across different versions or OEMs

msasikanth (changes requested):
Overall LGTM. Left few comments PTAL.

app/src/main/java/org/simple/clinic/newentry/PatientEntryInit.kt
janhavisinghh: Can you suggest a better way to achieve this?
 ```
    val effects = setOf(
        PrefillFormEffect(patient, address, phoneNumber, bangladeshNationalId),
        FetchBpPassportsEffect(patient.uuid),
        LoadInputFields
    )
    
    val effect1 = mutableSetOf(
        PrefillFormEffect(patient, address, phoneNumber, bangladeshNationalId),
        FetchBpPassportsEffect(patient.uuid),
        LoadInputFields,
        FetchColonyOrVillagesEffect
    )

    return if (model.hasColonyOrVillagesList) {
      first(model, effects)
    } else {
      first(model, effect1)
    }
```

Comment (msasikanth) [reply app/src/main/java/org/simple/clinic/newentry/PatientEntryInit.kt:9]: We can use a mutable set, so we can do something like this

```kt
val effects = mutableSetOf(Effect1, Effect2)

if (model.hasColonyOrVillagesList) {
  effects.add(FetchEffect)
}

first(model, effects)
```

app/src/test/java/org/simple/clinic/editpatient/EditPatientInitTest.kt
```diff
@@ -0,0 +1,52 @@
...
+  private val patientUuid = UUID.fromString("e40f42f4-0867-4891-ac77-95df5fe1fdef")
+  private val patient = TestData.patient(patientUuid, fullName = "TestName")
+  private val patientAddress = TestData.patientAddress(UUID.fromString("0d19e592-ae4d-4d4f-8881-8764709a17dd"), streetAddress = "Street Address", colonyOrVillage = "Colony")
+  private val patientPhoneNumber = TestData.patientPhoneNumber(UUID.fromString("f204b770-83c5-4145-9ca7-c2273be2bbdc"), number = "9999999999")
+  private val bangladeshNationalId = TestData.businessId(UUID.fromString("c9be5e5d-770c-4f37-a9de-2c304dfebfcd"))
+  private val dateOfBirthFormat = DateTimeFormatter.ofPattern("dd/MM/yyyy", Locale.ENGLISH)
+  private val initSpec = InitSpec(EditPatientInit(patient, patientAddress, patientPhoneNumber, bangladeshNationalId))
+
+  @Test
+  fun `when edit patient screen is created and colony or villages list is empty, then fetch colony or village list`() {
```
msasikanth, line 23: We should not be including the details like if the color on the village list empty or not empty in the test. Instead, can we rename it to something like this?

`when screen is created, then load initial data`

app/src/test/java/org/simple/clinic/editpatient/EditPatientInitTest.kt
```diff
@@ -0,0 +1,52 @@
...
+        hasEffects(PrefillFormEffect(patient, patientAddress, patientPhoneNumber, bangladeshNationalId),
+            FetchBpPassportsEffect(patientUuid),
+            LoadInputFields,
+            FetchColonyOrVillagesEffect
+        )
+    ))
+  }
+
+  @Test
+  fun `when edit patient screen is created and has colony or villages list, then don't fetch colony or village list`() {
```
msasikanth, line 38: We should not be including the details like if the color on the village list empty or not empty in the test. Instead, can we rename it to something like this?

`when screen is restored, then don't fetch colony or villages`

app/src/test/java/org/simple/clinic/editpatient/EditPatientEffectHandlerTest.kt
```diff
@@ -304,4 +304,17 @@ class EditPatientEffectHandlerTest {
     verify(ui).setupUi(inputFields)
     verifyNoMoreInteractions(ui)
   }
+
+  @Test
+  fun `when fetch colony or village names effect is received, then fetch them from the database`() {
```
msasikanth, line 309: Maybe instead of saying `then fetch them from the database`, we can say `then fetch colony or village names` or `then load colony or village names` so that the test is not tied to the implementation details. Like if we were to replace the implementation of getting to the database with something different, then the test will lose context.

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -273,6 +273,10 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
     deletePatient.setOnClickListener { router.push(DeletePatientScreenKey(screenKey.patient.uuid).wrap()) }
```
msasikanth, line 274: The commit message says `Fetch the colonyOrVillage list when the EditPatientScreen is created`, but here we are making changes to the ViewRenderer, can you update the commit message to reflect that, please?

app/src/main/res/layout/village_typeahead_list_item.xml
```diff
@@ -0,0 +1,24 @@
...
+    android:minHeight="?android:attr/listPreferredItemHeightSmall"
+    android:padding="@dimen/spacing_12"
+    android:textAppearance="?android:attr/textAppearanceListItemSmall"
+    tools:text="001 Charni Road" />
+
+  <View
+    android:layout_width="match_parent"
+    android:layout_height="1dp"
+    android:layout_marginHorizontal="@dimen/spacing_12"
+    android:background="#DADDE0" />
```
msasikanth, line 23: We have a divider drawable, can we use that, please?

```xml
  <View
    android:layout_height="1dp"
    android:layout_marginHorizontal="@dimen/spacing_12"
    android:background="@drawable/divider" />
```

app/src/main/res/layout/village_typeahead_list_item.xml
```diff
@@ -0,0 +1,24 @@
...
+  android:orientation="vertical">
+
+  <TextView
+    android:id="@+id/villageTypeAheadItemTextView"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:gravity="center_vertical"
+    android:minHeight="?android:attr/listPreferredItemHeightSmall"
+    android:padding="@dimen/spacing_12"
+    android:textAppearance="?android:attr/textAppearanceListItemSmall"
```
msasikanth, line 16: `textAppearance` for this TextView should be `textAppearanceBody1` based on the designs.

app/src/main/res/layout/village_typeahead_list_item.xml
```diff
@@ -0,0 +1,24 @@
...
+    tools:text="001 Charni Road" />
+
+  <View
+    android:layout_width="match_parent"
+    android:layout_height="1dp"
+    android:layout_marginHorizontal="@dimen/spacing_12"
```
msasikanth, line 22: When you use `android:layout_marginHorizontal` it applies margin on `start` and `end`. In the design, the margin is only applied to the start. Can we resolve this, please?

app/src/test/java/org/simple/clinic/newentry/PatientEntryInitTest.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.spotify.mobius.test.InitSpec
+import com.spotify.mobius.test.InitSpec.assertThatFirst
+import org.junit.Test
+
+class PatientEntryInitTest {
+  private val initSpec = InitSpec(PatientEntryInit())
+  private val defaultModel = PatientEntryModel.DEFAULT
+
+  @Test
+  fun `when patient entry screen is created and colony or villages list is empty, then fetch colony or village list`() {
```
msasikanth, line 14: We should not be including the details like if the color on the village list empty or not empty in the test. Instead, can we rename it to something like this?

`when screen is created, then load initial data`

app/src/test/java/org/simple/clinic/newentry/PatientEntryInitTest.kt
```diff
@@ -0,0 +1,36 @@
...
+    initSpec.whenInit(defaultModel).then(assertThatFirst(
+        hasModel(defaultModel),
+        hasEffects(
+            FetchPatientEntry, LoadInputFields, FetchColonyOrVillagesEffect
+        )
+    ))
+  }
+
+  @Test
+  fun `when patient entry screen is created and has colony or villages list, then don't fetch colony or village list`() {
```
msasikanth, line 24: We should not be including the details like if the color on the village list empty or not empty in the test. Instead, can we rename it to something like this?

`when screen is restored, then don't fetch colony or villages`

app/src/test/java/org/simple/clinic/newentry/PatientEntryEffectHandlerTest.kt
```diff
@@ -100,6 +100,20 @@ class PatientEntryEffectHandlerTest {
     verifyZeroInteractions(validationActions)
   }
 
+  @Test
+  fun `when fetch colony or village names effect is received, then fetch them from the database`() {
```
msasikanth, line 104: Maybe instead of saying `then fetch them from the database`, we can say `then fetch colony or village names` or `then load colony or village names` so that the test is not tied to the implementation details. Like if we were to replace the implementation of getting to the database with something different, then the test will lose context.

msasikanth (changes requested):
Left few comments, PTAL

app/src/main/res/layout/village_typeahead_list_item.xml
janhavisinghh: I didn't find this specified in the designs can you point me where it was later? Thanks.

msasikanth: In Figma the design team didn't seem to attach the text style, in those cases you should leave a comment on Figma tagging Pragati asking for the style or you can check the text styles in the app and see which matches the Figma values.

janhavisinghh: 👍🏻

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
msasikanth: We are technical not rendering it when screen is created, we don't have to fix this. But just an FYI

app/src/main/java/org/simple/clinic/editpatient/EditPatientUi.kt
```diff
@@ -29,4 +29,5 @@ interface EditPatientUi {
   fun showProgress()
   fun hideProgress()
   fun setupUi(inputFields: InputFields)
+  fun showColonyOrVillagesList(colonyOrVillageList: List<String>)
```
msasikanth, line 32: Not needed in this PR, but let's rename this method to `setColonyOrVillagesList` or `setColonyOrVillagesAutoComplete` since we are not actually showing once they are loaded.WDYT?

app/src/main/res/layout/village_typeahead_list_item.xml
msasikanth: Looks like you missed resolving this, can you resolve this in the next PR, please?

app/src/main/java/org/simple/clinic/newentry/PatientEntryUi.kt
```diff
@@ -16,4 +16,5 @@ interface PatientEntryUi {
   fun nextButtonShowInProgress()
   fun enableNextButton()
   fun setupUi(inputFields: InputFields)
+  fun showColonyOrVillagesList(colonyOrVillageList: List<String>)
```
msasikanth, line 19: Similar comment as before, maybe we start calling it `setColonyOrVillageAutoComplete` or something like that. Let's do this in a different PR

msasikanth (approved):
LGTM, there are a couple of changes that need to be resolved in a different PR.

PR #2437, opened 2021-03-01
Migrate `LinkIdWithPatientView` to `LinkIdWithPatientSheet`

msasikanth:


app/src/main/java/org/simple/clinic/summary/linkId/LinkIdWithPatientSheetKey.kt
```diff
@@ -0,0 +1,22 @@
+package org.simple.clinic.summary.linkId
+
+import androidx.fragment.app.Fragment
+import kotlinx.android.parcel.Parcelize
+import org.simple.clinic.navigation.v2.ScreenKey
+import org.simple.clinic.patient.PatientUuid
+import org.simple.clinic.patient.businessid.Identifier
+
+@Parcelize
+data class LinkIdWithPatientSheetKey(
```
vinaysshenoy, line 11: I generally prefer to make the keys for the new screens nested classes in the same screen. Can we follow the same pattern here, please?

app/src/main/java/org/simple/clinic/summary/linkId/LinkIdWithPatientView.kt
```diff
@@ -185,7 +168,7 @@ class LinkIdWithPatientView :
   private fun viewShows(): Observable<UiEvent> {
     return downstreamUiEvents
         .ofType<LinkIdWithPatientViewShown>()
-        .map { it as UiEvent }
+        .map { it }
```
vinaysshenoy: Is this necessary anymore since we are not actually converting it to anything else?

app/src/main/java/org/simple/clinic/summary/linkId/LinkIdWithPatientView.kt
```diff
@@ -139,37 +139,6 @@ class LinkIdWithPatientView :
...
-  override fun onFinishInflate() {
-    super.onFinishInflate()
-
-    if (isInEditMode) {
-      return
-    }
-
-    backgroundView.setOnClickListener {
-      // Intentionally done to swallow click events.
-    }
```
vinaysshenoy, line 151: This is being handled automatically by `BaseBottomSheet`, right? I don't remember if we did this. 😅

app/src/test/java/org/simple/clinic/summary/linkId/LinkIdWithPatientUpdateTest.kt
```diff
@@ -60,13 +60,12 @@ class LinkIdWithPatientUpdateTest {
 
   @Test
   fun `when add identifier is clicked, then update UI`() {
-    val patientFetchedModel = defaultModel.linkIdWithPatientViewShown(patientUuid, identifier)
     updateSpec
-        .given(patientFetchedModel)
+        .given(defaultModel)
```
vinaysshenoy, line 51: Shouldn't this model be the patient fetched model instead of the `defaultModel`. We might have used the wrong model when adding this screen initially, but I think we can correct this now.

app/src/main/java/org/simple/clinic/summary/linkId/LinkIdWithPatientEvent.kt
```diff
@@ -1,15 +1,9 @@
 package org.simple.clinic.summary.linkId
 
-import org.simple.clinic.patient.businessid.Identifier
 import org.simple.clinic.widgets.UiEvent
-import java.util.UUID
 
 sealed class LinkIdWithPatientEvent : UiEvent
 
-data class LinkIdWithPatientViewShown(val patientUuid: UUID, val identifier: Identifier) : LinkIdWithPatientEvent() {
-  override val analyticsName: String = "LinkIdWithPatient:Sheet Created"
```
vinaysshenoy, line 10: If I am not mistaken, we might be using this event in some Mixpanel reports to see how the `Link Id With Patient` sheet is used. We might have to update those reports to also take the `ScreenChange` event into account as well.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/summary/linkId/LinkIdWithPatientView.kt
vinaysshenoy: Please disregard this since this event stream has been removed in a later commit.

app/src/test/java/org/simple/clinic/summary/linkId/LinkIdWithPatientUpdateTest.kt
msasikanth: Ah right, Good catch. Will fix that 👍🏾

app/src/main/java/org/simple/clinic/summary/linkId/LinkIdWithPatientView.kt
msasikanth: Yep, we are not allowing cancelation from background click in `BaseBottomSheet` at all. But I did notice we are allowing the swipe to dismiss, which is causing issues with this flow. Should we disable the swipe behavior for the entire bottom sheet or keep it?

app/src/main/java/org/simple/clinic/summary/linkId/LinkIdWithPatientEvent.kt
msasikanth: I have overriden the analytics name for the key, so we should still know when the sheet is opened. But the analytics name is not the same as this, what should I be doing here?

Analytics name is `Link ID With Patient Sheet`

app/src/main/java/org/simple/clinic/summary/linkId/LinkIdWithPatientView.kt
msasikanth: I ended up disabling it since it's crashing the app when we dismiss the sheet and hit back button. My guess is probably because we disabled the cancelation of the dialog and then handling the back press ourselves.

app/src/main/java/org/simple/clinic/summary/linkId/LinkIdWithPatientEvent.kt
vinaysshenoy: Earlier, this would have been reported as a `UserInteraction` event. However, now it would be a `ScreenChange` event with the `incoming` set to `Link ID With Patient Sheet`.

We should be looking for reports which deal with the link sheet and update these, basically.

vinaysshenoy (approved):

PR #2442, opened 2021-03-02
Improve performance of the Instant Search queries

vinaysshenoy:


app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -3938,11 +3938,12 @@ class PatientRepositoryAndroidTest {
...
-        patient2WithCurrentFacilityAsAssignedFacility,
-        patientWithCurrentFacilityAsAssignedFacility,
-        patientWithAnotherFacilityAsAssignedFacility
-    )
+    assertThat(searchResults)
+        .containsExactly(
+            patientWithCurrentFacilityAsAssignedFacility,
+            patient2WithCurrentFacilityAsAssignedFacility,
+            patientWithAnotherFacilityAsAssignedFacility
+        ).inOrder()
```
msasikanth, line 3946: Correct me if I am wrong, will `contains` doesn't verify the order but passes even if there are more items than expected, `containsExactly` verifies that it has the same items as the expected but doesn't verify the order right?

app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
```diff
@@ -158,17 +158,17 @@ data class PatientSearchResult(
...
-            WHERE P.deletedAt IS NULL AND P.assignedFacilityId = :facilityId AND namePosition > 0
-            UNION
-            SELECT searchResult.*, 1 priority, INSTR(lower(P.fullName), lower(:name)) namePosition FROM 
-            PatientSearchResult searchResult
-            LEFT JOIN Patient P ON P.uuid = searchResult.uuid
-            WHERE P.deletedAt IS NULL AND P.assignedFacilityId != :facilityId AND namePosition > 0
-        )
+        SELECT 
+            searchResult.*, 
+            (
```
msasikanth, line 163: OMG, the easiest approach. Can't believe I missed it when implementing this feature 😅

msasikanth (approved):
LGTM 🚀

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
vinaysshenoy: Yup, that is correct.

vinaysshenoy: The clue is in the assertion name itself, `contains`.

PR #2445, opened 2021-03-03
Fix cannot exit patient summary screen after scheduling an appointment

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2805/fix-cannot-exit-patient-summary-screen-after-scheduling-an-appointment


janhavisinghh (approved):
LGTM

PR #2458, opened 2021-03-05
Update play core library version

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2600/update-play-core-library-version

### DIFF
--- a/app/src/main/java/org/simple/clinic/appupdate/UpdateManager.kt
+++ b/app/src/main/java/org/simple/clinic/appupdate/UpdateManager.kt
@@ -1,0 +1,40 @@
+package org.simple.clinic.appupdate
+
+import com.google.android.play.core.appupdate.AppUpdateInfo
+import com.google.android.play.core.appupdate.AppUpdateManager
+import com.google.android.play.core.install.model.AppUpdateType.FLEXIBLE
+import com.google.android.play.core.install.model.UpdateAvailability.UPDATE_AVAILABLE
+import io.reactivex.Observable
+import javax.inject.Inject
+
+interface UpdateManager {
+  fun updateInfo(): Observable<UpdateInfo>
+}
+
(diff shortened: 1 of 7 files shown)


vinaysshenoy:
@msasikanth There seems to be a CodeClimate issue. Would you be able to resolve this before I start reviewing?

app/src/main/java/org/simple/clinic/appupdate/UpdateManager.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.google.android.play.core.install.model.AppUpdateType.FLEXIBLE
+import com.google.android.play.core.install.model.UpdateAvailability.UPDATE_AVAILABLE
+import io.reactivex.Observable
+import javax.inject.Inject
+
+interface UpdateManager {
+  fun updateInfo(): Observable<UpdateInfo>
+}
+
+class PlayUpdateManager @Inject constructor(
```
vinaysshenoy, line 14: Is there a reason this implementation is in the same file? Can we break it out into another?

vinaysshenoy (approved):
Overall PR looks good. Left a comment that can be resolved later.

app/src/main/java/org/simple/clinic/appupdate/UpdateManager.kt
msasikanth: No particular reason, I think it was because there was no other implementation for `UpdateManager` I have decided to keep both of them in the same file.

vinaysshenoy: Right, for interfaces and their class implementations I generally prefer to have them as separate files even if there's only one implementation. But that's a personal preference, maybe we can decide on a convention later.

vinaysshenoy (approved):

PR #2461, opened 2021-03-08
Enable running GH actions manually for ci_checks & code_formatting_checks

msasikanth:
Just a quality of life change, for manually running GH Actions for a given branch from the Actions tab on GitHub.

### DIFF
--- a/.github/workflows/ci_checks.yml
+++ b/.github/workflows/ci_checks.yml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - master
+  workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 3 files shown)


vinaysshenoy:
Oh, this looks interesting. What do we need to do to manually trigger workflows?

msasikanth:
> Oh, this looks interesting. What do we need to do to manually trigger workflows?

We just have to go to the Actions tab, select a workflow that has a manual trigger and it will show an option to trigger the workflow like this.

Screenshot 2021-03-09 122650

vinaysshenoy (approved):

PR #2464, opened 2021-03-09
Show link id sheet after patient summary profile is loaded and link id is present

msasikanth:
Bug explanation: https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1615270894171400

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Bump Play Core to v1.10.0
 - Add option to manually trigger GH Actions
 - Migrate `SetupActivity` to the new navigation framework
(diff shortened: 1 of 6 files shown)


Honey14 (changes requested):
Assigning BP passport to a patient is not working. Please take a look.

msasikanth:
> Assigning BP passport to a patient is not working. Please take a look.

Ah yes, found the issue. I forgot to place `!` for the identifier boolean check. Also please include testing steps you have tried when something doesn't work as expected, it helps the developer narrow down where to look.

Honey14 (dismissed):
Looks great

Honey14 (approved):

PR #2467, opened 2021-03-10
BP passport scanner improvements

msasikanth:


app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -81,6 +86,7 @@ constructor(
     }, ContextCompat.getMainExecutor(context))
   }
 
+  @SuppressLint("ClickableViewAccessibility")
```
vinaysshenoy, line 94: What exactly is this suppressing? 🤔

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -59,6 +61,9 @@ constructor(
   @Inject
   lateinit var bitmapUtils: BitmapUtils
 
+  @Inject
+  lateinit var googleApiAvailability: GoogleApiAvailability
```
vinaysshenoy, line 65: Correct me if I am wrong, but this will only tell us if Play Services is available, right? Not whether the MLKit dependencies are downloaded or not.

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -106,23 +106,24 @@ constructor(
...
+        .build()
+
     val googlePlayServicesAvailability = googleApiAvailability.isGooglePlayServicesAvailable(context)
     val isGooglePlayServicesAvailable = googlePlayServicesAvailability == ConnectionResult.SUCCESS
 
     val isMLKitQrCodeScannerEnabled = features.isEnabled(Feature.MLKitQrCodeScanner)
     val qrCodeAnalyzer = if (isMLKitQrCodeScannerEnabled && isGooglePlayServicesAvailable) {
-      MLKitQrCodeAnalyzer(bitmapUtils, scans::onNext)
+      MLKitQrCodeAnalyzer(bitmapUtils, scans::onNext, mlKitUnavailable = {
+        setQrCodeAnalyzer(analyzer, ZxingQrCodeAnalyzer(scans::onNext))
```
vinaysshenoy, line 120: Ah, I see what's happening now.

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
msasikanth: When we implement `onTouchListener` without a `performClick` action Android platform shows it as a warning since it can have an on-click listener separately. Since in this scenario we are sure preview view won't have an on-click listener I have suppressed that warning.

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
msasikanth: Yep, we check if the ML Kit is available from the crash in the `MlKitQrCodeAnalyzer`

msasikanth:
Assigning @janhavisinghh as a reviewer since @Honey14 is on leave.

janhavisinghh (changes requested):
LGTM. Left one small comment PTAL. Also, can you answer these two questions? Or point me to a blog/documentation where I can learn more about this?
1. Why are we switching to zxing analyzer when gplay services or ml kit is unavailable? Why are we not using only zxing analyzer?
2. As far as I understand we are using ML Kit for recognising qr code and zxing analyzer is a library that does that as well. Can you tell me what is the need of google play services here?

app/src/main/java/org/simple/clinic/widgets/qrcodescanner/QrCodeScannerView.kt
```diff
@@ -25,6 +29,7 @@ import kotlin.math.abs
 import kotlin.math.max
 import kotlin.math.min
 
+
```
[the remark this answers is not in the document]
msasikanth: Since this is not a breaking implementation, can we get this merged in. This is blocking a release.

janhavisinghh (approved):

msasikanth:
> LGTM. Left one small comment PTAL. Also, can you answer these two questions? Or point me to a blog/documentation where I can learn more about this?
> 
> 1. Why are we switching to zxing analyzer when gplay services or ml kit is unavailable? Why are we not using only zxing analyzer?
> 2. As far as I understand we are using ML Kit for recognising qr code and zxing analyzer is a library that does that as well. Can you tell me what is the need of google play services here?

1. Zxing analyzer doesn't properly work on some devices and we previously had issues with it, that's why we primarily using ML Kit and only if that fails we switch to Zxing.

2. ML Kit is a Google service, so it needs to download the machine learning model from Google Play to decoding the QR code. In order to do that, the device should have Google Play Services.

More info [here](https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1615285875196700).

PR #2471, opened 2021-03-12
Use personal access token for GitHub rebase command

msasikanth:
Reference: https://github.com/cirrus-actions/rebase/pull/70

### DIFF
--- a/.github/workflows/ci_check_pr_comment.yml
+++ b/.github/workflows/ci_check_pr_comment.yml
@@ -37,6 +37,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           token: ${{ secrets.GITHUB_ACCESS_TOKEN  }}
+          persist-credentials: false
           fetch-depth: 0
       - name: Rebase
         uses: cirrus-actions/rebase@1.4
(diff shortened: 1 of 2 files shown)


msasikanth:
Currently, I have created a PAT and added it to secrets. Feel free to replace that if needed.

.github/workflows/ci_check_pr_comment.yml
```diff
@@ -32,15 +32,17 @@ jobs:
...
       - uses: actions/checkout@v2
         with:
+          token: ${{ secrets.PAT }}
+          persist-credentials: false
           fetch-depth: 0
       - name: Rebase
         uses: cirrus-actions/rebase@1.4
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_ACCESS_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.PAT }}
```
vinaysshenoy, line 45: We already have a personal access token secret `GITHUB_ACCESS_TOKEN`, which we added for the code formatting checks (iirc). Is that not usable for this check?

msasikanth: Oh, if it's already a personal access token then it should work. I thought it was a different one. Let me replace it with that.

vinaysshenoy (approved):

.github/workflows/ci_check_pr_comment.yml
vinaysshenoy: Also, please avoid using abbreviations like `PAT` and use the full name like `PERSONAL_ACCESS_TOKEN`. Someone looking at the list of secrets a month from now will have no idea what `PAT` actually is.

PR #2474, opened 2021-03-12
Add different author associations conditions when running PR comment commands

msasikanth:


msasikanth:
Adding one author_association check seems to have caused the commands to get skipped. Since we didn't satisfy with the `MEMBER` role. So I have added few other roles as well. (I have tested the action on a new repo and it's working)

vinaysshenoy (approved):

PR #2477, opened 2021-03-15
Show 'Or' between Age and DOB fields in Patient entry and edit

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2932/show-or-between-age-and-dob-fields-in-patient-entry-and-edit

### DIFF
--- a/app/src/main/res/layout/screen_manual_patient_entry.xml
+++ b/app/src/main/res/layout/screen_manual_patient_entry.xml
@@ -55,7 +55,8 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_marginTop="@dimen/spacing_16"
-        android:paddingStart="@dimen/spacing_4">
+        android:paddingStart="@dimen/spacing_4"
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/screen_manual_patient_entry.xml
```diff
@@ -55,7 +55,8 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_marginTop="@dimen/spacing_16"
-        android:paddingStart="@dimen/spacing_4">
```
janhavisinghh, line 58: I see that the separator was already there. What exactly was the reason for the separator to not be visible?

janhavisinghh (approved):
LGTM

app/src/main/res/layout/screen_manual_patient_entry.xml
msasikanth: It's because of the margin top and gravity of the container.

PR #2492, opened 2021-03-18
Show 'CHANGE' button in `SetAppointmentReminderView` for changing appointment date

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2933/show-change-button-in-setappointmentreminderview-for-changing-appointment-date


app/src/main/res/values/strings.xml
```diff
@@ -788,4 +788,4 @@
   <string name="patientsearch_error_no_search_results">No patients match</string>
   <string name="patientsearch_error_no_patients_in_facility_heading">No patients in %1$s</string>
 
-</resources>
+</resources>
```
janhavisinghh: Can you remove this useless change?

janhavisinghh (changes requested):
Left one comment PTAL. LGTM otherwise

app/src/main/res/values/strings.xml
msasikanth: It's not redundant. It's happening because of our [GitHub Actions](https://github.com/simpledotorg/simple-android/blob/a660b76d60fc090adecfe9b1bdf27e76088ff591/.github/workflows/ci_checks.yml#L155) that fixes the string ellipsis. It changes the line feed for the file, so that's why we see this change on almost all of our PRs. 

I guess it's something that needs to be fixed a the GH Action level.

janhavisinghh (approved):

PR #2493, opened 2021-03-18
Improvements to instant search

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2891/improvements-to-instant-search

### DIFF
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreen.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreen.kt
@@ -287,7 +287,13 @@ class InstantSearchScreen :
         .textChanges()
         .skipInitialValue()
         .debounce(500, TimeUnit.MILLISECONDS)
-        .map { SearchQueryChanged(it.toString()) }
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreen.kt
```diff
@@ -287,7 +287,13 @@ class InstantSearchScreen :
         .textChanges()
         .skipInitialValue()
         .debounce(500, TimeUnit.MILLISECONDS)
-        .map { SearchQueryChanged(it.toString()) }
+        .map { searchQuery ->
+          val trimmedSearchQuery = searchQuery
+              .trim()
+              .toString()
```
janhavisinghh, line 293: Do we need to convert this to toString? Isn't searchQuery a string already?

msasikanth: It's a charsequence, so we need to convert it to string in the end.

janhavisinghh (approved):

PR #2494, opened 2021-03-21
Make entire prescribed drugs list scrollable

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/2934/make-entire-prescribed-drugs-list-scrollable-instead-of-list-scrolling-inside-a-fixed-card-view

### DIFF
--- a/app/src/main/java/org/simple/clinic/drugs/selection/DrugListItem.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/selection/DrugListItem.kt
@@ -1,6 +1,8 @@
 package org.simple.clinic.drugs.selection
 
 import androidx.recyclerview.widget.DiffUtil
+import com.google.android.material.shape.CornerFamily
+import com.google.android.material.shape.ShapeAppearanceModel
 import io.reactivex.subjects.Subject
 import org.simple.clinic.R
 import org.simple.clinic.databinding.ListPrescribeddrugsCustomDrugBinding
@@ -8,6 +10,7 @@ import org.simple.clinic.databinding.ListPrescribeddrugsNewCustomDrugBinding
 import org.simple.clinic.databinding.ListPrescribeddrugsProtocolDrugBinding
 import org.simple.clinic.drugs.PrescribedDrug
 import org.simple.clinic.widgets.ItemAdapter
+import org.simple.clinic.widgets.dp
 import org.simple.clinic.widgets.recyclerview.BindingViewHolder
 import org.simple.clinic.widgets.visibleOrGone
 
@@ -41,10 +44,24 @@ data class ProtocolDrugListItem(
     viewBinding.protocoldrugItemDosage.visibleOrGone(prescribedDrug != null)
     viewBinding.protocoldrugItemDosage.text = prescribedDrug?.dosage
 
+    viewBinding.prescribeddrugItemProtocoldrugRootlayout.shapeAppearanceModel = topMostCardShapeAppearanceModel(isTopMostCard = holder.adapterPosition == 0)
+
     viewBinding.root.setOnClickListener {
       subject.onNext(DrugListItemClicked.PrescribedDrugClicked(drugName, prescribedDrug))
     }
   }
+}
+
+private fun topMostCardShapeAppearanceModel(isTopMostCard: Boolean) = if (isTopMostCard) {
+  ShapeAppearanceModel()
+      .toBuilder()
+      .setTopLeftCorner(CornerFamily.ROUNDED, 4.dp.toFloat())
+      .setTopRightCorner(CornerFamily.ROUNDED, 4.dp.toFloat())
+      .build()
+} else {
+  ShapeAppearanceModel()
+      .toBuilder()
+      .build()
 }
 
 data class CustomPrescribedDrugListItem(
@@ -64,6 +81,9 @@ data class CustomPrescribedDrugListItem(
     viewBinding.root.setOnClickListener {
       subject.onNext(DrugListItemClicked.CustomPrescriptionClicked(prescribedDrug))
     }
+
+    viewBinding.prescribeddrugItemCustomdrugRootlayout.shapeAppearanceModel = topMostCardShapeAppearanceModel(isTopMostCard = holder.adapterPosition == 0)
+
   }
 }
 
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/simple/clinic/drugs/selection/DrugListItem.kt
```diff
@@ -41,12 +44,26 @@ data class ProtocolDrugListItem(
...
 
+    viewBinding.prescribeddrugItemProtocoldrugRootlayout.shapeAppearanceModel = topMostCardShapeAppearanceModel(holder)
+
     viewBinding.root.setOnClickListener {
       subject.onNext(DrugListItemClicked.PrescribedDrugClicked(drugName, prescribedDrug))
     }
   }
 }
 
+private fun topMostCardShapeAppearanceModel(holder: BindingViewHolder) = if (holder.adapterPosition == 0) {
```
vinaysshenoy, line 55: Instead of passing in the entire holder, this parameter should take in a boolean named `isTopMostCard` since that is the piece of information that is actually relevant to this method.

vinaysshenoy (changes requested):
Overall PR looks good. Left a comment about improving a method definition. Can we also update the CHANGELOG, please?

app/src/main/res/values/shape.xml
```diff
@@ -47,4 +47,12 @@
     <item name="cornerRadius">2dp</item>
   </style>
 
+  <style name="ShapeAppearanceOverlay.Simple.AddNewCustomDrugButton" parent="ShapeAppearance.MaterialComponents.MediumComponent">
```
msasikanth, line 50: Just an FYI, we don't have to extend the parents for overlays. Whether it might be for `ThemeOverlays` or `ShapeAppearanceOverlay` we simple extend an empty parent.

vinaysshenoy:
The divider items seems to have lost the margin ([Figma](https://www.figma.com/file/vmZlmlFSCOrEYzTMkqsNB8/Simple-Production?node-id=14958%3A53)) on the sides. Please take a look?

device-2021-03-26-091318

janhavisinghh:
> The divider items seem to have lost the margin ([Figma](https://www.figma.com/file/vmZlmlFSCOrEYzTMkqsNB8/Simple-Production?node-id=14958%3A53)) on the sides. Please take a look?
>

@vinaysshenoy I added margin to the Item Decoration but it doesn't seem to be working in this screen. I time boxed it to figure it out but not able to yet. This is what I did: 

```val dividerMargin = 8.dp

    recyclerView.addItemDecoration(DividerItemDecorator(requireContext(), marginStart = dividerMargin, marginEnd = dividerMargin))```
    
    Can you suggest what the issue might be here? Or, should we leave this as is for now?

vinaysshenoy:
> > The divider items seem to have lost the margin ([Figma](https://www.figma.com/file/vmZlmlFSCOrEYzTMkqsNB8/Simple-Production?node-id=14958%3A53)) on the sides. Please take a look?
> 
> @vinaysshenoy I added margin to the Item Decoration but it doesn't seem to be working in this screen. I time boxed it to figure it out but not able to yet. This is what I did:
> 
> ```
> val dividerMargin = 8.dp
> 
>  recyclerView.addItemDecoration(DividerItemDecorator(requireContext(), marginStart = dividerMargin, marginEnd = dividerMargin))
> ```
> 
> Can you suggest what the issue might be here? Or, should we leave this as is for now?

Let me check.

vinaysshenoy:
> > The divider items seem to have lost the margin ([Figma](https://www.figma.com/file/vmZlmlFSCOrEYzTMkqsNB8/Simple-Production?node-id=14958%3A53)) on the sides. Please take a look?
> 
> @vinaysshenoy I added margin to the Item Decoration but it doesn't seem to be working in this screen. I time boxed it to figure it out but not able to yet. This is what I did:
> 
> ```
> val dividerMargin = 8.dp
> 
>  recyclerView.addItemDecoration(DividerItemDecorator(requireContext(), marginStart = dividerMargin, marginEnd = dividerMargin))
> ```
> 
> Can you suggest what the issue might be here? Or, should we leave this as is for now?

I think the divider itself is working fine (see the change in divider colour below the 'lm' part of 'Telmisartan' (I used a divider margin of 64 dp to be sure). 

What is happening (I guess) is that the individual card items are having an elevation or something which casts a shadow below them.

Screenshot from 2021-03-29 15-16-16

msasikanth:
> > > The divider items seem to have lost the margin ([Figma](https://www.figma.com/file/vmZlmlFSCOrEYzTMkqsNB8/Simple-Production?node-id=14958%3A53)) on the sides. Please take a look?
> > 
> > 
> > @vinaysshenoy I added margin to the Item Decoration but it doesn't seem to be working in this screen. I time boxed it to figure it out but not able to yet. This is what I did:
> > ```
> > val dividerMargin = 8.dp
> > 
> >  recyclerView.addItemDecoration(DividerItemDecorator(requireContext(), marginStart = dividerMargin, marginEnd = dividerMargin))
> > ```
> > 
> > 
> > Can you suggest what the issue might be here? Or, should we leave this as is for now?
> 
> I think the divider itself is working fine (see the change in divider colour below the 'lm' part of 'Telmisartan' (I used a divider margin of 64 dp to be sure).
> 
> What is happening (I guess) is that the individual card items are having an elevation or something which casts a shadow below them.
> 
> Screenshot from 2021-03-29 15-16-16

Ah, it's because we are using an item decorator so it's adding a space between the card items. Move the divider inside the card and manually set the visibility of it, which should resolve the issue.

app/src/main/res/values/shape.xml
janhavisinghh: I was following the pattern that was followed in the above Shape Appearance styles.

msasikanth: Those were shape appearance styles, here we are creating an overlay. You can take a look at `ShapeAppearanceOverlay.Simple.BottomSheet` for example.

msasikanth: One of the reasons why this can cause issues is, it will inherit fields that we are not overriding from the parent. Instead what a overlay should do is, override from our theme shape appearances.

janhavisinghh: @vinaysshenoy give me a moment. Pushing this change as well.

janhavisinghh: @vinaysshenoy done

vinaysshenoy (approved):

PR #2495, opened 2021-03-22
Show custom error screen for drug stock web view when there is no internet connection

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2980/show-custom-error-screen-for-drug-stock-web-view-when-there-is-no-internet-connection

### DIFF
--- a/app/src/main/assets/drug_stock_no_connection.html
+++ b/app/src/main/assets/drug_stock_no_connection.html
@@ -1,0 +1,7 @@
+<html>
+  <body style="background: #F0F2F5; font-family: Roboto, sans-serif; text-align: center;">
+    <div style="font-size: 24px; font-weight: bold; padding: 24px 32px 8px 32px; color: #2F363D;">Not available</div>
+    <div style="padding: 0 32px 24px 32px; color: #6C737A;">An internet connection is needed to use this feature.</div>
+    <div style="padding: 0 32px 24px 32px;"><a href="simple://progress-tab" style="color: #0075EB; text-decoration: none; font-weight: bold;">BACK</a></div>
+  </body>
+</html>
--- a/app/src/main/java/org/simple/clinic/home/report/ReportsWebViewClient.kt
+++ b/app/src/main/java/org/simple/clinic/home/report/ReportsWebViewClient.kt
@@ -1,8 +1,12 @@
 package org.simple.clinic.home.report
 
+import android.os.Build
+import android.webkit.WebResourceError
 import android.webkit.WebResourceRequest
 import android.webkit.WebView
 import android.webkit.WebViewClient
+import androidx.annotation.RequiresApi
+import timber.log.Timber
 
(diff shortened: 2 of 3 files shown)


msasikanth:
Since Janhavi already has a bunch of PRs to review, adding @vinaysshenoy as a reviewer.

app/src/main/assets/drug_stock_no_connection.html
```diff
@@ -0,0 +1,8 @@
+<html>
```
vinaysshenoy, line 1: Can we format this HTML, please?

app/src/main/assets/drug_stock_no_connection.html
```diff
@@ -0,0 +1,8 @@
...
+<body style="background: #F0F2F5; font-family: Roboto, sans-serif; text-align: center;">
+<div style="font-size: 24px; font-weight: bold; padding: 24px 32px 8px 32px; color: #2F363D;">Not available</div>
+<div style="padding: 0 32px 24px 32px; color: #6C737A;">An internet connection is needed to use this feature.</div>
```
vinaysshenoy, line 4: These user visible strings need to be localised, no?

app/src/main/java/org/simple/clinic/home/report/ReportsWebViewClient.kt
```diff
@@ -11,10 +15,29 @@ class ReportsWebViewClient(
   override fun shouldOverrideUrlLoading(view: WebView?, request: WebResourceRequest?): Boolean {
     val url = request?.url?.toString() ?: return super.shouldOverrideUrlLoading(view, request)
 
+    Timber.d(url)
```
vinaysshenoy, line 18: Leftover log?

vinaysshenoy (changes requested):

app/src/main/assets/drug_stock_no_connection.html
msasikanth: I did, Android Studio and Idea is formatting like this 🤷🏾

app/src/main/assets/drug_stock_no_connection.html
msasikanth: Ideally yes, but I am unsure how translations would work for HTML files. Is there a specific format we should follow here? 🤔

app/src/main/assets/drug_stock_no_connection.html
msasikanth: I will manually format the file and push the changes.

app/src/main/assets/drug_stock_no_connection.html
msasikanth: Yeah. Let's go with a later option, we can add it now and improve it later if someone complains regarding the translations. For now, doing it during runtime seems like a bit of work to implement 😅

vinaysshenoy (approved):

PR #2508, opened 2021-03-25
Remove `DateTimeFormatter` from `PatientRepository`

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3010/remove-datetimeformatter-from-patientrepository

PR #2509, opened 2021-03-26
Render Ethiopian dates in the app

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3023/render-ethiopian-dates-in-the-app

### DIFF
--- a/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
@@ -2,10 +2,10 @@ package org.simple.clinic.bp.entry
 
 import com.google.firebase.perf.metrics.AddTrace
 import com.spotify.mobius.rx2.RxMobius
+import dagger.Lazy
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
-import dagger.Lazy
 import io.reactivex.ObservableTransformer
 import io.reactivex.Scheduler
 import io.reactivex.rxkotlin.cast
@@ -85,8 +85,8 @@ class BloodPressureEntryEffectHandler @AssistedInject constructor(
       prefillDates
           .map(::convertToLocalDate)
           .observeOn(scheduler)
-          .doOnNext { ui.setDateOnInputFields(it) }
-          .doOnNext { ui.showDateOnDateButton(it) }
+          .doOnNext(ui::setDateOnInputFields)
+          .doOnNext(ui::showDateOnDateButton)
           .map { DatePrefilled(it) }
     }
   }
--- a/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
@@ -85,7 +85,7 @@ class BloodPressureEntryEffectHandler @AssistedInject constructor(
       prefillDates
           .map(::convertToLocalDate)
           .observeOn(scheduler)
-          .doOnNext { setDateOnInputFields(it) }
+          .doOnNext { ui.setDateOnInputFields(it) }
           .doOnNext { ui.showDateOnDateButton(it) }
           .map { DatePrefilled(it) }
     }
@@ -94,14 +94,6 @@ class BloodPressureEntryEffectHandler @AssistedInject constructor(
   private fun convertToLocalDate(prefillDate: PrefillDate): LocalDate {
     val instant = if (prefillDate is PrefillSpecificDate) prefillDate.date else Instant.now(userClock)
     return instant.toLocalDateAtZone(userClock.zone)
-  }
-
-  private fun setDateOnInputFields(dateToSet: LocalDate) {
-    ui.setDateOnInputFields(
-        dateToSet.dayOfMonth.toString(),
-        dateToSet.monthValue.toString(),
-        dateToSet.year.toString()
-    )
   }
 
   private fun fetchBloodPressureMeasurement(
--- a/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
+++ b/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
@@ -457,12 +457,6 @@ class BloodPressureEntrySheet : BottomSheetActivity(), BloodPressureEntryUi, Rem
     dateErrorTextView.visibility = View.GONE
   }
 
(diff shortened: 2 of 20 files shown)


msasikanth:
Blocked by #2508

app/src/main/java/org/simple/clinic/di/DateFormatterModule.kt
```diff
@@ -2,35 +2,65 @@ package org.simple.clinic.di
...
 
 @Module
 class DateFormatterModule {
 
+  @Provides
+  fun providesChronology(country: Country): Chronology {
+    return when (country.isoCountryCode) {
+      Country.ETHIOPIA -> EthiopicChronology.INSTANCE
+      else -> IsoChronology.INSTANCE
+    }
```
vinaysshenoy: Should we move this `when` clause to a computed property for the `Country` type, similar to what we have done for the `alternativeIdentifierType`?

app/src/main/java/org/simple/clinic/di/DateFormatterModule.kt
```diff
@@ -24,8 +24,16 @@ class DateFormatterModule {
...
-    return DateTimeFormatter.ofPattern("d-MMM-yyyy", locale)
+  fun provideDateFormatterForFullDate(
+      locale: Locale,
+      chronology: Chronology,
+      country: Country
+  ): DateTimeFormatter {
+    val pattern = when (country.isoCountryCode) {
+      Country.ETHIOPIA -> "d-MM-yyyy"
+      else -> "d-MMM-yyyy"
+    }
```
vinaysshenoy, line 43: Similar feedback as the earlier comment whether we should move this `when` clause to a computed property of the `Country` type. There's a strong case given that this block is being repeated in multiple places.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
```diff
@@ -447,6 +463,12 @@ class BloodPressureEntrySheet : BottomSheetActivity(), BloodPressureEntryUi, Rem
     yearEditText.setTextAndCursor(fourDigitYear)
   }
 
+  override fun setDateOnInputFields(date: LocalDate) {
+    dayEditText.setTextAndCursor(dayDateFormatter.format(date))
+    monthEditText.setTextAndCursor(monthDateFormatter.format(date))
+    yearEditText.setTextAndCursor(fullYearDateFormatter.format(date))
```
vinaysshenoy, line 463: Instead of three different `DateTimeFormatters`, what do you think of injecting a single formatter that generates it in the `dd/MM/yyyy` format and then splitting it on the `/` character?

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryEffectHandler.kt
```diff
@@ -85,7 +85,7 @@ class BloodPressureEntryEffectHandler @AssistedInject constructor(
       prefillDates
           .map(::convertToLocalDate)
           .observeOn(scheduler)
-          .doOnNext { setDateOnInputFields(it) }
+          .doOnNext { ui.setDateOnInputFields(it) }
           .doOnNext { ui.showDateOnDateButton(it) }
```
vinaysshenoy, line 89: Personal preference, but these two could be simplified using method references.

app/src/main/java/org/simple/clinic/di/DateFormatterModule.kt
```diff
@@ -18,9 +20,12 @@ import javax.inject.Named
...
-  fun providesChronology(country: Country): Chronology {
+  fun providesChronology(country: Country, features: Features): Chronology {
+    if (!features.isEnabled(Feature.EthiopianCalendar))
+      return IsoChronology.INSTANCE
+
     return when (country.isoCountryCode) {
-      Country.ETHIOPIA -> EthiopicChronology.INSTANCE
+      Country.ETHIOPIA -> return EthiopicChronology.INSTANCE
       else -> IsoChronology.INSTANCE
     }
```
vinaysshenoy: Can we try and clean this up so that there's only a single `return` statement? Seems like there are too many at the moment.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/di/DateFormatterModule.kt
msasikanth: Yeah, good point. Will move it inside `Country` as a computed property.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
msasikanth: Yeah, that can work. Although I feel like this is much more explicit than the splitting because then we have to get the item based on position.

app/src/main/java/org/simple/clinic/di/DateFormatterModule.kt
msasikanth: In this case there are different patterns we will be handling. So we would have to move all the patterns (most of them), so the computed properties would like.

```kt
val fullDatePatterns: String
val userInputDatePattern: String
```

Is that fine?

vinaysshenoy: Yeah, I think that looks good.

app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntrySheet.kt
vinaysshenoy: That's a good point. My intention was to reduce the number of injections we are doing purely for extracting date strings out of a date instance.

This could be a potential enhancement for the future, maybe we could consider creating something like a `DateInformationExtractor` class that abstracts this away so that it can reused between the two screens. What do you think?

msasikanth: Yep, that sounds good. Tracking it [here](https://app.clubhouse.io/simpledotorg/story/3060/add-dateinformationextractor).

app/src/main/java/org/simple/clinic/di/DateFormatter.kt
```diff
@@ -6,6 +6,9 @@ import javax.inject.Qualifier
 annotation class DateFormatter(val value: Type) {
```
Honey14, line 6: There's been a spelling mistake in the commit message. Should have been `FullYear` and not `FullyYear`

msasikanth: Good catch. Resolved it 👍🏾

vinaysshenoy (approved):

PR #2510, opened 2021-03-29
Add feature flag for village type ahead search

msasikanth:


app/src/main/java/org/simple/clinic/feature/Feature.kt
```diff
@@ -23,5 +23,6 @@ enum class Feature(
   ChangeLanguage(Build.VERSION.SDK_INT >= Build.VERSION_CODES.M),
   MLKitQrCodeScanner(true, "ml_kit_qr_code_scanner"),
   InstantSearch(true, "instant_search"),
-  OverdueCount(true, "overdue_count")
+  OverdueCount(true, "overdue_count"),
+  VillageTypeAhead(true, "village_type_ahead")
```
vinaysshenoy, line 27: Should the default value for this be set to `false` instead, given that we'll need to wait for the remote config sync to complete.

app/src/main/java/org/simple/clinic/editpatient/EditPatientInit.kt
```diff
@@ -21,7 +24,7 @@ class EditPatientInit(
         LoadInputFields,
     )
 
-    if (!model.hasColonyOrVillagesList) {
+    if (!model.hasColonyOrVillagesList && features.isEnabled(Feature.VillageTypeAhead)) {
```
vinaysshenoy, line 27: We are introducing an impurity here. Can we instead pass the flag in as a `Boolean` constructor property instead?

app/src/test/java/org/simple/clinic/editpatient/EditPatientInitTest.kt
```diff
@@ -17,7 +21,16 @@ class EditPatientInitTest {
   private val patientPhoneNumber = TestData.patientPhoneNumber(UUID.fromString("f204b770-83c5-4145-9ca7-c2273be2bbdc"), number = "9999999999")
   private val bangladeshNationalId = TestData.businessId(UUID.fromString("c9be5e5d-770c-4f37-a9de-2c304dfebfcd"))
   private val dateOfBirthFormat = DateTimeFormatter.ofPattern("dd/MM/yyyy", Locale.ENGLISH)
-  private val initSpec = InitSpec(EditPatientInit(patient, patientAddress, patientPhoneNumber, bangladeshNationalId))
+  private val features = Features(
+      remoteConfigService = NoOpRemoteConfigService(DefaultValueConfigReader()),
+      overrides = mapOf(
+          Feature.VillageTypeAhead to true
+      ))
```
vinaysshenoy, line 28: Yeah, as mentioned earlier, this information is unnecessary for testing business logic. Passing in the value as a `Boolean` property to `EditPatientInitTest` should simplify this a lot.

app/src/main/java/org/simple/clinic/newentry/PatientEntryInit.kt
```diff
@@ -3,12 +3,14 @@ package org.simple.clinic.newentry
 import com.spotify.mobius.First
 import com.spotify.mobius.First.first
 import com.spotify.mobius.Init
+import org.simple.clinic.feature.Feature
+import org.simple.clinic.feature.Features
 
-class PatientEntryInit : Init<PatientEntryModel, PatientEntryEffect> {
+class PatientEntryInit(private val features: Features) : Init<PatientEntryModel, PatientEntryEffect> {
```
vinaysshenoy, line 9: Same comments as earlier about the `EditPatientScreen`.

app/src/main/java/org/simple/clinic/newentry/PatientEntryInit.kt
```diff
@@ -3,12 +3,14 @@ package org.simple.clinic.newentry
...
   override fun init(model: PatientEntryModel): First<PatientEntryModel, PatientEntryEffect> {
     val effects = mutableSetOf(FetchPatientEntry, LoadInputFields)
 
-    if (!model.hasColonyOrVillagesList) {
+    if (!model.hasColonyOrVillagesList && features.isEnabled(Feature.VillageTypeAhead)) {
```
vinaysshenoy, line 13: Can we also verify this behaviour in tests?

app/src/main/java/org/simple/clinic/editpatient/EditPatientInit.kt
vinaysshenoy, line 27: Can we verify this behaviour in tests?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/feature/Feature.kt
msasikanth: Ah, good point. Will make the change.

app/src/main/java/org/simple/clinic/editpatient/EditPatientInit.kt
msasikanth: We test feature flags in our tests as well? 🤔

vinaysshenoy: Depends on the feature flag itself, I guess.

msasikanth: I pushed other changes, but if you can confirm if this is something we need for this feature flag I am gonna add this case and push again or else the rest of the changes are done.

vinaysshenoy: Let's skip this for now and add the test in another PR.

vinaysshenoy (approved):

PR #2516, opened 2021-03-31
Migrate `ContactPatientBottomSheet` to use `BaseBottomSheet`

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3078/migrate-scheduleappointmentsheet-to-use-basebottomsheet


msasikanth:
Once this is merged in, I want to split out the 3 UI modes into individual sheets/screens so that it's easier to maintain those going forward.

app/src/main/res/layout/sheet_contact_patient.xml
```diff
@@ -3,9 +3,8 @@
   xmlns:tools="http://schemas.android.com/tools"
   android:id="@+id/contentFlipper"
   android:layout_width="match_parent"
-  android:layout_height="match_parent"
-  android:animateLayoutChanges="true"
```
vinaysshenoy, line 7: `animateLayoutChanges=true` has been removed. Is this intentional?

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
```diff
@@ -130,6 +134,14 @@ class ContactPatientBottomSheet : BaseBottomSheet<
     context.injector<Injector>().inject(this)
   }
 
+  override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+    val dialog = super.onCreateDialog(savedInstanceState) as BottomSheetDialog
+
+    dialog.overrideCancellation(::backPressed)
+
+    return dialog
```
vinaysshenoy: We already are doing this in `BaseBottomSheet`. 🤔 

If you need access to the `Back Press` event, I would suggest making `onBackPressed` a protected method instead and overriding it (while enforcing a call to the super method.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
```diff
@@ -240,6 +252,10 @@ class ContactPatientBottomSheet : BaseBottomSheet<
     removeAppointmentView.disableRemoveAppointmentDoneButton()
   }
 
+  private fun backPressed() {
+    hotEvents.onNext(BackClicked)
```
vinaysshenoy, line 256: We might be able to remove the need for this, I think it was originally added because of the activity and screen lifecycle interactions which should not be a problem anymore.

vinaysshenoy (commented):
Overall PR looks good. Left a few comments, we can merge the PR after discussing them.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
msasikanth: I thought about that. But I felt like this was much more explicit and we don't end up easily missing the super call in `onBackPressed`. Also when we call `backPressed` we are doing a back press-on activity, so it can cause problems down the line when someone doesn't know to make the super call for it? WDYT?

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
msasikanth: For now, it's needed, because we are handling back clicks in `ContactPatientUpdate#backClicks` to either update UI or go back. Once we separate out the 3 UI modes into their own sheets/screens, we can then remove this.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
msasikanth: Another reason is, it seemed much more in line with the `OnBackPressedDispatcher` implementation Android has.

app/src/main/res/layout/sheet_contact_patient.xml
vinaysshenoy: Cool, could we consider showing something manually instead? This is a database call, so some form of transition might be useful.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
vinaysshenoy: One reason we wanted to implement this in the base class is because we need the `Router` to know that this screen is going to be removed from the hierarchy. Is this not required in this case?

msasikanth: It is required in certain scenarios. Currently, we are overriding the back click to handle the UI changes in this sheet, so in certain cases, we won't remove the sheet from the back stack but simply update the UI to go back to the previous UI mode. Once we separate the 3 UI modes into their own sheets/screens, we don't have to override this.

app/src/main/res/layout/sheet_contact_patient.xml
msasikanth: Even with or without the `animateLayoutChanges` we are simply showing the sheet and controls inside it. The only difference is a slight fade-in for the overdue controls. For that, I can look into `TransitionManager` usage in a different PR. Is that fine?

vinaysshenoy: Yeah, sounds good.

vinaysshenoy (approved):

app/src/main/res/layout/sheet_contact_patient.xml
msasikanth: Resolved it in #2531

PR #2523, opened 2021-04-01
Fix prescribed drugs item corner radius not being updated when order is changed

msasikanth:
Without this we are getting a weird UI glitch where the top corner radius are being preserved if a new item is pushed to top.

*Reproducing steps*
- Add a dosage to protocol/custom drug to move it to top of the list
- The item that is pushed down now has top corner radius even though it's not longer the first item

### DIFF
--- a/app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt
@@ -36,6 +36,14 @@ class EditMedicinesUiRenderer(private val ui: EditMedicinesUi) : ViewRenderer<Ed
     val customPrescribedDrugItems = customPrescribedDrugItems(prescribedCustomDrugs)
     val drugsList = (protocolDrugSelectionItems + customPrescribedDrugItems)
         .sortedByDescending { it.prescribedDrug?.updatedAt }
+        .mapIndexed { index, drugListItem ->
+          val isTopItem = index == 0
+          when {
+            isTopItem && drugListItem is ProtocolDrugListItem -> drugListItem.withTopCorners()
+            isTopItem && drugListItem is CustomPrescribedDrugListItem -> drugListItem.withTopCorners()
+            else -> drugListItem
+          }
+        }
 
     ui.populateDrugsList(drugsList)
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt
```diff
@@ -36,6 +36,20 @@ class EditMedicinesUiRenderer(private val ui: EditMedicinesUi) : ViewRenderer<Ed
...
+          when {
+            isTopItem && drugListItem is ProtocolDrugListItem -> {
+              drugListItem.copy(hasTopCorners = true)
+            }
+            isTopItem && drugListItem is CustomPrescribedDrugListItem -> {
+              drugListItem.copy(hasTopCorners = true)
+            }
+            else -> {
+              drugListItem
+            }
```
vinaysshenoy, line 50: This logic has a lot of cases, can we update the tests to verify that they are working as expected?

vinaysshenoy (approved):

app/src/main/java/org/simple/clinic/drugs/EditMedicinesUiRenderer.kt
msasikanth: I think the only case we haven't covered in the tests is `CustomPrescribedDrugListItem ` as the top item. Can do that. For the rest of them, I have verified they output the correct `hasTopCorners` value.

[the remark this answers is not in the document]
msasikanth: Yeah, good point. Will make that change.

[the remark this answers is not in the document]
msasikanth: Just to confirm in both of the cases we apply top corners. By default, it doesn't have top corners.

msasikanth: Done. Added the missing case of verifying `CustomPrescribedDrugListItem` as the top item.

vinaysshenoy (approved):

PR #2532, opened 2021-04-02
Add Somali language option to language selection

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3098/add-somali-language-option-to-language-selection


app/src/main/java/org/simple/clinic/settings/SettingsModule.kt
```diff
@@ -35,7 +35,8 @@ class SettingsModule {
         ProvidedLanguage(displayName = "ትግርኛ", languageCode = "ti-ET"),
         ProvidedLanguage(displayName = "Afan Oromo", languageCode = "om-ET"),
         ProvidedLanguage(displayName = "አማርኛ", languageCode = "am-ET"),
-        ProvidedLanguage(displayName = "தமிழ்", languageCode = "ta-IN")
+        ProvidedLanguage(displayName = "தமிழ்", languageCode = "ta-IN"),
+        ProvidedLanguage(displayName = "Soomaali", languageCode = "so")
```
vinaysshenoy, line 39: Can we also be explicit about the region code? This should be done for the resources and transifex as well.

msasikanth: Yeah, I was basing it off the transifex. Who should I talk to, to add that region code?

vinaysshenoy: Maybe Vineet?

vinaysshenoy: It's not something that's critical to do, but it will become harder to fix later if we need to have a translation for the same language in a different country.

msasikanth: Yep, I have added the region code manually in this PR and asked Vineet to check with transifex folks to start adding it.

Honey14 (approved):
Looks good to me

PR #2533, opened 2021-04-05
Migrate `PatientsTabScreen` to a fragment

msasikanth:
ViewBinding migration: https://app.clubhouse.io/simpledotorg/story/1828/migrate-app-to-use-view-binding-instead-of-kotlin-synthetic-view-accessors


app/src/main/java/org/simple/clinic/home/patients/PatientsTabScreen.kt
```diff
@@ -299,4 +303,12 @@ class PatientsTabScreen(context: Context, attrs: AttributeSet) : RelativeLayout(
   interface Injector {
     fun inject(target: PatientsTabScreen)
   }
+
+  @Parcelize
+  object Key : ScreenKey() {
```
vinaysshenoy, line 331: Using `object` is risky if we need to use multiple instances of the screen as part of the navigation graph because there are places where we check if this key already exists in the backstack. We should prefer using standard class instances unless we specifically require this behaviour.

vinaysshenoy (commented):
Overall PR looks good, left a comment about the type of the patient tab screen key. Please take a look?

app/src/main/java/org/simple/clinic/home/patients/PatientsTabScreen.kt
msasikanth: In this scenario does that matter? 🤔 . Since we are using it in a ViewPager we are not actually using the key directly. We are simply calling the `createFragment` method from the `ScreenKey`. But I can change it to class if you think this would be problematic in the future.

vinaysshenoy: Hmm, makes sense. Still something that could be risky if we start using this screen elsewhere since the person implementing that change may not realize the potential issues that can happen.

msasikanth: Okay. Converting it to a class 👍🏾

vinaysshenoy (approved):

PR #2535, opened 2021-04-05
Migrate `ReportsScreen` to a fragment

msasikanth:
Viewbinding migration: https://app.clubhouse.io/simpledotorg/story/1828/migrate-app-to-use-view-binding-instead-of-kotlin-synthetic-view-accessors


msasikanth:
Blocked by #2533

app/src/main/java/org/simple/clinic/home/report/ReportsScreen.kt
```diff
@@ -45,6 +47,23 @@ class ReportsScreen : BaseScreen<
...
 
+  override fun defaultModel() = ReportsModel.create()
+
+  override fun bindView(layoutInflater: LayoutInflater, container: ViewGroup?) =
+      ScreenReportBinding.inflate(layoutInflater, container, false)
+
+  override fun uiRenderer() = ReportsUiRenderer(this)
+
+  override fun events() = Observable
+      .mergeArray(webViewBackClicks)
```
vinaysshenoy, line 46: Can we also report events to analytics here? I think we might have forgotten to do that in the original screen implementation?

vinaysshenoy (changes requested):
Overall PR looks good, needs a small change and we should be good to go.

app/src/main/java/org/simple/clinic/home/report/ReportsEvent.kt
```diff
@@ -1,8 +1,9 @@
 package org.simple.clinic.home.report
 
 import org.simple.clinic.util.Optional
+import org.simple.clinic.widgets.UiEvent
 
-sealed class ReportsEvent
+sealed class ReportsEvent : UiEvent
```
vinaysshenoy, line 6: We'll also need to override `analyticsName` on the events themselves in order to actually report them.

msasikanth: I completely forgot about that 🤦🏾

app/src/main/java/org/simple/clinic/home/report/ReportsEvent.kt
```diff
@@ -1,9 +1,13 @@
...
 
 data class ReportsLoaded(val reportsContent: Optional<String>) : ReportsEvent()
 
-object WebBackClicked : ReportsEvent()
+object WebBackClicked : ReportsEvent() {
+
+  override val analyticsName = "Reports Back clicked"
```
vinaysshenoy, line 12: Can we follow the same convention as the rest of the screens: "<Screen Name>:<Event name>" please?

vinaysshenoy, line 12: Can we follow the same convention as the other screen events please?

`<Screen Name>:<Event name>`

msasikanth: Done ✅

vinaysshenoy (approved):

PR #2538, opened 2021-04-06
Support delegating SQL query operations

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/3011/automatically-report-sql-query-performance-to-firebase

### DIFF
--- a/app/src/main/java/org/simple/clinic/storage/monitoring/DelegateableSQLiteOpenHelper.kt
+++ b/app/src/main/java/org/simple/clinic/storage/monitoring/DelegateableSQLiteOpenHelper.kt
@@ -1,0 +1,42 @@
+package org.simple.clinic.storage.monitoring
+
+import androidx.sqlite.db.SupportSQLiteOpenHelper
+
+class DelegateableSQLiteOpenHelper(
+    private val delegate: SQLiteDatabaseSqlDelegate,
+    private val openHelper: SupportSQLiteOpenHelper
+) : SupportSQLiteOpenHelper {
+
+  override fun close() {
+    openHelper.close()
+  }
+
+  override fun getDatabaseName() = openHelper.databaseName
+
+  override fun setWriteAheadLoggingEnabled(enabled: Boolean) {
+    openHelper.setWriteAheadLoggingEnabled(enabled)
+  }
+
+  override fun getWritableDatabase() = DelegateableSQLiteDatabase(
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/storage/monitoring/SQLiteDatabaseSqlDelegate.kt
```diff
@@ -0,0 +1,66 @@
+package org.simple.clinic.storage.monitoring
+
+import android.content.ContentValues
+import android.database.Cursor
+import android.os.CancellationSignal
+import androidx.sqlite.db.SupportSQLiteDatabase
+import androidx.sqlite.db.SupportSQLiteQuery
+
+interface SQLiteDatabaseSqlDelegate {
```
vinaysshenoy, line 9: The methods here are mostly copied from [SupportSQLiteDatabase](https://developer.android.com/reference/androidx/sqlite/db/SupportSQLiteDatabase).

app/src/main/java/org/simple/clinic/storage/monitoring/DelegateableSQLiteOpenHelper.kt
```diff
@@ -0,0 +1,42 @@
+package org.simple.clinic.storage.monitoring
+
+import androidx.sqlite.db.SupportSQLiteOpenHelper
+
+class DelegateableSQLiteOpenHelper(
```
msasikanth, line 5: Should we mark this as a private constructor so that we only use the `Factory` function for creating the helper?

msasikanth (commented):
LGTM. Correct me if I am wrong, currently, we are just delegating some of the CRUD methods and eventually use those to report analytics and write to DB right?

msasikanth (dismissed):

vinaysshenoy:
> LGTM. Correct me if I am wrong, currently, we are just delegating some of the CRUD methods and eventually use those to report analytics and write to DB right?

Yes, that is correct.

app/src/main/java/org/simple/clinic/storage/monitoring/DelegateableSQLiteOpenHelper.kt
vinaysshenoy: Good point, I'll make this change in the next PR.

msasikanth (approved):

PR #2539, opened 2021-04-06
Add `TextInputDatePicker`

Honey14:
https://app.clubhouse.io/simpledotorg/story/3029/create-textinputdatepicker-sheet


app/src/test/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerInitTest.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.spotify.mobius.test.InitSpec
+import com.spotify.mobius.test.InitSpec.assertThatFirst
+import org.junit.Test
+import org.simple.clinic.textInputdatepicker.TextInputDatePickerEffect.PrefilledDate
+import java.time.LocalDate
+
+class TextInputDatePickerInitTest {
+
+  @Test
+  fun `when prefilled date is received, then upda`() {
```
msasikanth, line 14: The test name seems to be not fully completed.

app/src/test/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerUpdateTest.kt
```diff
@@ -27,7 +27,8 @@ class TextInputDatePickerUpdateTest {
       LocalDate.now(localClock),
       UserInputDatePaddingCharacter.ZERO
   ))
-  private val defaultModel = TextInputDatePickerModel.create()
+  private val date = LocalDate.now()
```
msasikanth, line 30: Let's use an actual date here instead of using `now`

app/src/main/java/org/simple/clinic/main/TheActivityComponent.kt
```diff
@@ -73,6 +73,7 @@ import org.simple.clinic.teleconsultlog.shareprescription.TeleconsultSharePrescr
 import org.simple.clinic.teleconsultlog.success.TeleConsultSuccessScreen
```
msasikanth, line 73: In the commit message change `theActivityComponent` to `TheActivityComponent`

msasikanth (changes requested):
The approach looks good overall. I have left few comments PTAL

msasikanth:
@Honey14 There also seems to be a unit test that's failing, please take a look at that.

msasikanth:
@Honey14 Since `DatePickerKeyFactory` PR is merged in (#2540 ). Add the `TextInputDatePicker` key there, so that we can switch the date pickers based on the country. Please take a look at that PR implementation to see how we can pass data back.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEvent.kt
```diff
@@ -13,3 +13,7 @@ data class DayChanged(val day: String) : TextInputDatePickerEvent()
 data class MonthChanged(val month: String): TextInputDatePickerEvent()
```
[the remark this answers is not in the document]
Honey14: The commit does not show the error messages, neither it's validating.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEffectHandler.kt
```diff
@@ -18,6 +25,25 @@ class TextInputDatePickerEffectHandler @AssistedInject constructor(
...
         .addAction(HideDateErrorMessage::class.java, { uiActions.hideErrorMessage() }, schedulersProvider.ui())
+        .addConsumer(ShowDateValidationError::class.java, { showDateValidationErrors(it.dateValidation) }, schedulersProvider.ui())
         .build()
   }
+
+  private fun showDateValidationErrors(dateValidation: Result) {
+    when (dateValidation) {
+      is InvalidPattern -> {
+        uiActions.showErrorMessage()
+        uiActions.showInvalidDateError()
```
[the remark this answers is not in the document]
Honey14: Yes we can but since these two have different actions, keeping them different for readability.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEffectHandler.kt
```diff
@@ -28,9 +36,26 @@ class TextInputDatePickerEffectHandler @AssistedInject constructor(
...
         .addConsumer(ShowDateValidationError::class.java, { showDateValidationErrors(it.dateValidation) }, schedulersProvider.ui())
         .addConsumer(SendUserEnteredDate::class.java, { uiActions.sendUserEnteredDate(it.userEnteredDate) }, schedulersProvider.ui())
+        .addTransformer(PrefilledDate::class.java, prefilledDate(schedulersProvider.ui()))
         .build()
   }
 
+  private fun prefilledDate(scheduler: Scheduler): ObservableTransformer<PrefilledDate, TextInputDatePickerEvent> {
+    return ObservableTransformer { prefilledDate ->
+      prefilledDate
+          .map(::convertToLocalDate)
```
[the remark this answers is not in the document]
Honey14: Sorry, my bad. I changed it from Instant to LocalDate and forgot to remove the conversion from the handler. Removing it.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEvent.kt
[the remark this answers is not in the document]
msasikanth: WDYM? When the done button is clicked we are validating the date and depending on that we are showing errors no?. Currently, the commit message implies we are showing error messages whenever done is clicked.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEffectHandler.kt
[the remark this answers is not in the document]
msasikanth: In this scenario, it's fine because we won't be calling `showInvalidDateError` when the view is hidden right? So we can merge these 2 methods into one.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEvent.kt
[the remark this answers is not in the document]
Honey14: So we have many errors, this commit message seems to be very vague and we're not showing errors in this commit. We're hiding errors when field is changed.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEffectHandler.kt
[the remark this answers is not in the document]
Honey14: Yes that is true but we'll be writing the same visibility property in all three functions if we use one method. Anyhow, I'll merge this into one method.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerModel.kt
```diff
@@ -0,0 +1,12 @@
+package org.simple.clinic.textInputdatepicker
```
msasikanth, line 1: Instead of calling it `Extend BaseBottomSheet in TextInputDatePicker` we can rename this to `Create TextInpuDatePicker` or something like that. Because we are creating a new sheet and not just extending the `BaseBottomSheet` for already existing screen right?. 

Also just an FYI, usually we create layout files and class files in a single commit either at the start or at the end when wiring up the loop with UI.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerSheet.kt
```diff
@@ -20,6 +23,21 @@ class TextInputDatePickerSheet : BaseBottomSheet<
...
+
+  private val imageTextInputSheetClose
+    get() = binding.imageTextInputSheetClose
+
+  private fun sheetCloseClicks() = imageTextInputSheetClose
+      .clicks()
+      .map { DismissSheetClicked }
+
+  override fun dismissSheet() {
+    dismiss()
```
msasikanth, line 134: Currently, dismiss doesn't remove the sheet from the back stack. Instead, we have to use `router.pop()`

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerSheet.kt
```diff
@@ -20,6 +23,21 @@ class TextInputDatePickerSheet : BaseBottomSheet<
 
   override fun bindView(inflater: LayoutInflater, container: ViewGroup?) = SheetTextInputDatePickerBinding.inflate(inflater, container, false)
 
+  override fun events() = sheetCloseClicks()
+      .compose(ReportAnalyticsEvents())
+      .cast<TextInputDatePickerEvent>()
+
+  private val imageTextInputSheetClose
```
msasikanth, line 93: Can we move these view binding variable declarations above the function calls? Currently, these are in between function calls.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerValidator.kt
```diff
@@ -0,0 +1,49 @@
...
+import java.time.LocalDate
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
+import java.time.format.DateTimeParseException
+import javax.inject.Inject
+import javax.inject.Named
+
+class TextInputDatePickerValidator @Inject constructor(
+    private val userClock: UserClock,
+    @Named("exact_date") private val dateOfBirthFormat: DateTimeFormatter
```
msasikanth, line 19: Can we rename this variable to something else? We are not validating the date of birth here.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEffectHandler.kt
```diff
@@ -18,6 +25,20 @@ class TextInputDatePickerEffectHandler @AssistedInject constructor(
...
+  }
+
+  private fun showDateValidationErrors(dateValidation: Result) {
+    when (dateValidation) {
+      is InvalidPattern -> uiActions.showInvalidDateError()
+      is DateIsInPast -> uiActions.showDateIsInPastError()
+      is MaximumAllowedDateRange -> uiActions.showMaximumDateRangeError()
+      is Valid -> throw IllegalStateException("Date validation error cannot be $dateValidation")
+    }.exhaustive()
+  }
```
msasikanth: Can this be done in a single method instead?

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEvent.kt
```diff
@@ -13,3 +14,7 @@ data class DayChanged(val day: String) : TextInputDatePickerEvent()
...
 
 data class YearChanged(val year: String): TextInputDatePickerEvent()
+
+data class DoneClicked(val minDate: LocalDate, val maxDate: LocalDate) : TextInputDatePickerEvent() {
```
msasikanth, line 18: Instead of trying to pass the min and max dates like this using an event, do you think we can pass these as constructor arguments for the validator instead? Since these don't change and are passed by screen key I think that should make it easy to validate dates or we can even add them to the model when creating to avoid passing these dates like this. WDYT?

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerSheet.kt
```diff
@@ -71,6 +76,12 @@ class TextInputDatePickerSheet : BaseBottomSheet<
       .map(CharSequence::toString)
       .map(::YearChanged)
 
+  private fun imeDoneClicks() = listOf(dayEditText, monthEditText, yearEditText)
+      .map { it.editorActions { actionID -> actionID == EditorInfo.IME_ACTION_DONE } }
+      .toObservable()
+      .flatMap { it }
```
msasikanth, line 130: Do we need to map the action keys here? Couldn't we directly map it to the `DoneClicked` instead? Like
```kt
private fun imeDoneClicks() = listOf(dayEditText, monthEditText, yearEditText)
    .map { it.editorActions { actionID -> actionID == EditorInfo.IME_ACTION_DONE } }
    .map { DoneClicked(screenKey.minDate, screenKey.maxDate) }
```

app/src/test/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerUpdateTest.kt
```diff
@@ -71,4 +83,24 @@ class TextInputDatePickerUpdateTest {
...
         )
   }
+
+  @Test
+  fun `when done button is clicked, then show date validation errors`() {
+    val model = defaultModel
+        .yearChanged("2019")
+        .dayChanged("04")
+        .monthChanged("34")
+    val minDate = LocalDate.of(2019, 4, 4)
```
msasikanth: FYI, we usually use `parse` to define local dates. Let's replace `LocalDate.of` with `LocalDate.parse` in this PR
```kt
LocalDate.parse("2018-01-01")
```

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEffectHandler.kt
```diff
@@ -28,13 +36,30 @@ class TextInputDatePickerEffectHandler @AssistedInject constructor(
...
+          .map(::getCurrentDateOrPrefilledDate)
+          .observeOn(scheduler)
+          .doOnNext(uiActions::setDateOnInputFields)
+          .map { DatePrefilled(it) }
+    }
+  }
+
+  private fun getCurrentDateOrPrefilledDate(prefillDate: PrefilledDate?): LocalDate {
+    val nowDate = Instant.now(userClock)
+    val nowLocalDate = LocalDateTime.ofInstant(nowDate, ZoneOffset.UTC).toLocalDate()
```
msasikanth, line 65: We can use the `toLocalDateAtZone` `Instant` extension function for creating the date.
```kt
val instant = Instant.now(userClock)
val currentDate = instant.toLocalDateAtZone(userClock.zone)
```

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEffectHandler.kt
```diff
@@ -28,13 +36,30 @@ class TextInputDatePickerEffectHandler @AssistedInject constructor(
...
+    return prefillDate?.date ?: nowLocalDate
```
msasikanth, line 66: It seems like irrespective of whether or not the prefill date is present we are creating the current date. Instead of elvis operator can we use if/else condition to make sure we are only creating current date object when prefill date is not present.
```kt
private fun getCurrentDateOrPrefilledDate(prefillDate: PrefilledDate): LocalDate {
  return if (prefillDate.date != null) {
    prefillDate.date
  } else {
    val instant = Instant.now(userClock)
    instant.toLocalDateAtZone(userClock.zone)
  }
}
```

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerValidator.kt
```diff
@@ -0,0 +1,49 @@
...
+        parsedDate.isBefore(minDate) -> DateIsInPast
+        else -> Result.Valid(parsedDate)
+      }
+    } catch (dte: DateTimeParseException) {
+      InvalidPattern
+    }
+  }
+
+  @VisibleForTesting
+  fun dateInUserTimeZone(): LocalDate {
```
msasikanth, line 46: Are we using this somewhere, If not we can remove it?

msasikanth (changes requested):
Left few comments PTAL

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEvent.kt
Honey14: I think adding them to the model constructor makes more sense.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerSheet.kt
Honey14: Yes we can.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEffectHandler.kt
Honey14: I don't understand. What do you mean in a single method?

msasikanth: Current the only thing that's happening in `applyValidationErrors` is calling `showDateValidationErrors(effect.dateValidation)`. Can we instead just use one method?

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerSheet.kt
msasikanth: Turns out we do need it, without it the events are not being sent to the Mobius loop.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerSheet.kt
```diff
@@ -0,0 +1,32 @@
+package org.simple.clinic.textInputdatepicker
+
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import kotlinx.android.parcel.Parcelize
```
msasikanth, line 14: We need to use the new `Parcelize` import. `kotlinx.parcelize.Parcelize`

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerValidator.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.simple.clinic.textInputdatepicker.TextInputDatePickerValidator.Result.Notvalid.InvalidPattern
+import org.simple.clinic.textInputdatepicker.TextInputDatePickerValidator.Result.Notvalid.MaximumAllowedDateRange
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
+import java.time.format.DateTimeParseException
+import javax.inject.Inject
+import javax.inject.Named
+
+class TextInputDatePickerValidator @Inject constructor(
+    @Named("exact_date") private val userEnteredDateFormat: DateTimeFormatter
```
msasikanth, line 13: Let's use `date_for_user_input`, one of the reasons is the dateformatter that we get with `exact_date` is dependent on the `Country`. So we decided to show this date picker for some other `Country` the date format will not match.

app/src/main/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerEffectHandler.kt
msasikanth: We haven't combined these functions yet?

app/src/test/java/org/simple/clinic/textInputdatepicker/TextInputDatePickerValidatorTest.kt
```diff
@@ -0,0 +1,49 @@
...
+import java.time.format.DateTimeFormatter
+import java.util.Locale
+
+class TextInputDatePickerValidatorTest {
+
+  private val userEnteredDateFormat = DateTimeFormatter.ofPattern("dd/MM/yyyy", Locale.ENGLISH)
+  private val textInputDatePickerValidator = TextInputDatePickerValidator(
+      userEnteredDateFormat = userEnteredDateFormat
+  )
+  private val minDate = LocalDate.of(2021, 3, 1)
```
msasikanth, line 19: I think you forgot to use `LocalDate.parse` here.

app/src/main/res/layout/sheet_text_input_date_picker.xml
```diff
@@ -0,0 +1,176 @@
+<?xml version="1.0" encoding="utf-8"?>
```
msasikanth, line 1: The UI doesn't look the same as in designs
Screenshot_20210416_075022

Screenshot_20210416_071946

msasikanth (changes requested):

msasikanth (approved):

PR #2541, opened 2021-04-06
Implement Search by patient identifiers in Instant Search 

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/2225/search-by-patient-identifiers-bangladesh-national-id-etc

When user searches from a numeric criteria and wants to register a new patient/update an existing entry then they add it as a phone number. Currently this is how I've implemented this flow. Lmk if you think this is not how it should work. @vinaysshenoy

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
@@ -34,6 +34,7 @@ import org.simple.clinic.overdue.Appointment.Status.Scheduled
 import org.simple.clinic.overdue.Appointment.Status.Visited
 import org.simple.clinic.overdue.AppointmentRepository
 import org.simple.clinic.patient.PatientSearchCriteria.Name
+import org.simple.clinic.patient.PatientSearchCriteria.NumericCriteria
 import org.simple.clinic.patient.PatientSearchCriteria.PhoneNumber
 import org.simple.clinic.patient.PatientSearchResult.LastSeen
 import org.simple.clinic.patient.PatientStatus.Active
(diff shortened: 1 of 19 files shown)


janhavisinghh:
@vinaysshenoy it would be great if you could review this on priority tomorrow as I want to wrap up this task by Friday.

app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
```diff
@@ -194,6 +194,23 @@ data class PatientSearchResult(
...
+                    WHEN P.assignedFacilityId = :facilityId THEN 0
+                    ELSE 1
+                END
+            ) AS priority,
+            INSTR(phoneNumber, :numericCriteria) phoneNumberPosition, 
+            INSTR(id_identifier, :numericCriteria) identifierPosition FROM PatientSearchResult searchResult
+        LEFT JOIN Patient P ON P.uuid = searchResult.uuid
+        WHERE P.deletedAt IS NULL AND phoneNumberPosition > 0 OR identifierPosition > 0
+        ORDER BY priority ASC, phoneNumberPosition ASC, identifierPosition ASC
+        """)
```
vinaysshenoy, line 221: This query can possible result in duplicate rows if either:
- A patient has multiple business identifiers that have a common sequence of numbers
- If a patient's phone number and their business ID have a common sequence of numbers

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchUpdate.kt
```diff
@@ -13,8 +14,11 @@ import org.simple.clinic.mobius.next
 import org.simple.clinic.patient.OngoingNewPatientEntry
 import org.simple.clinic.patient.PatientSearchCriteria
 import org.simple.clinic.patient.businessid.Identifier
+import javax.inject.Inject
 
-class InstantSearchUpdate : Update<InstantSearchModel, InstantSearchEvent, InstantSearchEffect> {
+class InstantSearchUpdate @Inject constructor(
+    private val features: Features
```
vinaysshenoy, line 24: We should not be passing the `Features` type to any `Update` function because it makes the `Update` function impure. What we should be doing instead is only taking the feature flag we need (in this case, the `InstantSearchByPatientIdentifier` feature and passing it as a `Boolean` property to the `Update` function.

Then we use the `Features` type in the screen at the time of instantiating the `Update` class to get the actual flag value.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -12,6 +12,7 @@ import org.simple.clinic.facility.Facility
 import org.simple.clinic.overdue.Appointment.AppointmentType.Manual
```
vinaysshenoy, line 12: Can you please add/update tests for repositories whenever you make changes to the data layer to ensure that they are working as expected?

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchUpdate.kt
```diff
@@ -67,9 +71,10 @@ class InstantSearchUpdate @Inject constructor(
   }
 
   private fun registerNewPatient(model: InstantSearchModel): Next<InstantSearchModel, InstantSearchEffect> {
-    var ongoingPatientEntry = when (val searchCriteria = searchCriteriaFromInput(model.searchQuery.orEmpty(), model.additionalIdentifier)) {
-      is PatientSearchCriteria.Name -> OngoingNewPatientEntry.fromFullName(searchCriteria.patientName)
-      is PatientSearchCriteria.PhoneNumber -> OngoingNewPatientEntry.fromPhoneNumber(searchCriteria.phoneNumber)
+    var ongoingPatientEntry = when (val searchCriteria = searchCriteriaFromInput(model.searchQuery.orEmpty(), model.additionalIdentifier, features.isEnabled(InstantSearchByPatientIdentifier))) {
```
vinaysshenoy, line 74: I see a lot has changed in the business logic, but there are no tests added to verify this. Please add the tests.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/feature/Feature.kt
```diff
@@ -27,5 +27,6 @@ enum class Feature(
   OverdueCount(true, "overdue_count"),
   VillageTypeAhead(false, "village_type_ahead"),
   InstantSearchQrCode(true, "instant_search_qr_code"),
-  EthiopianCalendar(true, "ethiopian_calendar")
+  EthiopianCalendar(true, "ethiopian_calendar"),
+  InstantSearchByPatientIdentifier(true, "instant_search_by_patient_identifier")
```
vinaysshenoy, line 31: Can we disable this by default until the feature is complete?

app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
vinaysshenoy: I see that you added a `DISTINCT` clause to the SQL query. Was this meant to resolve this comment? This will not solve the issue because the `DISTINCT` clause operates on the entire row, what you might be looking for is the `GROUP BY` clause.

Please verify these changes using the instrumented tests when you make them.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -4133,4 +4134,185 @@ class PatientRepositoryAndroidTest {
...
+              patientAssignedFacilityId = assignedFacilityId,
+              businessId = businessId
+          )
+
+      patientRepository.save(listOf(patientProfile)).blockingAwait()
+    }
+
+    // given
+    val currentFacility = TestData.facility(uuid = UUID.fromString("c0056c11-105d-4079-80fa-c745128f7fc5"), name = "CHC Bucho")
+    val facility2 = TestData.facility(uuid = UUID.fromString("532264e6-b358-445e-8b48-c6253677db24"), name = "Test")
```
vinaysshenoy, line 4160: Please try to follow consistency when writing tests. This means that if you are naming a facility `currentFacility`, give it an appropriate name like `otherFacility` instead of `facility2`. The same goes for the test data itself, "CHC Bucho" and "Test" do not go well together.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -4133,4 +4134,185 @@ class PatientRepositoryAndroidTest {
...
+        .search2(NumericCriteria("12"), currentFacility.uuid)
+        .map { it.uuid }
+
+    //then
+    assertThat(searchResults)
+        .containsExactly(
+            patientWithCurrentFacilityAsAssignedFacility,
+            patient2WithCurrentFacilityAsAssignedFacility,
+            patientWithAnotherFacilityAsAssignedFacility
+        ).inOrder()
```
vinaysshenoy, line 4208: Please change the test data for some of the patients to verify that identifiers that do not match the criteria do not show up in the list. Here, since all the identifiers match the search term, there is no guarantee that the query is actually functioning correctly.

app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
vinaysshenoy: This comment also applies to all the queries you have added, not just this one specific one.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -4133,4 +4134,185 @@ class PatientRepositoryAndroidTest {
...
+            patientUuid = patient2WithCurrentFacilityAsAssignedFacility)
+    )
+
+    //when
+    val searchResults = patientRepository
+        .search2(NumericCriteria("12"), currentFacility.uuid)
+        .map { it.uuid }
+
+    //then
+    assertThat(searchResults)
```
vinaysshenoy, line 4203: This test seems incomplete. As far as I understand it , this query is supposed to test for both phone numbers _and_ business ID identifiers. This seems to only verify the business ID queries without checking the phone number part.

app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
vinaysshenoy: I see there are checks in the test for this kind of data. However, the test is passing there because we are extracting the `UUID` instances from the search results and then converting it to a `Set`, which will remove duplicate `UUID` instances. If you convert it to a `List` instead, you'll see the duplicates showing up.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchUpdate.kt
```diff
@@ -110,11 +114,24 @@ class InstantSearchUpdate @Inject constructor(
...
   ): PatientSearchCriteria {
     return when {
-      digitsRegex.matches(inputString) -> PatientSearchCriteria.PhoneNumber(inputString.filterNot { it.isWhitespace() }, additionalIdentifier)
-      else -> PatientSearchCriteria.Name(inputString, additionalIdentifier)
+      digitsRegex.matches(inputString) -> numericPatientSearchCriteriaBasedOnFeatureFlag(isInstantSearchByIdentifierEnabled, inputString.filterNot { it.isWhitespace() }, additionalIdentifier)
+      else -> Name(inputString, additionalIdentifier)
+    }
+  }
+
+  private fun numericPatientSearchCriteriaBasedOnFeatureFlag(
```
vinaysshenoy, line 126: Not necessary to change this, but passing the feature flag as a method parameter here is unnecessary since the flag is already a property of `InstantSearchUpdate`.

vinaysshenoy (changes requested):
Left a few more comments regarding the correctness of the tests and the queries. Please take a look?

vinaysshenoy:
Closing this for now since @janhavisinghh is on leave for a week. We can reopen this once she's back.

app/src/main/java/org/simple/clinic/patient/PatientSearchResult.kt
janhavisinghh: I have added the `GROUPBY` clause. I have also added some duplicate values in the test to see if it's working properly. Firstly, I tested without `GROUPBY` and then with it. So, it is working correctly.

vinaysshenoy (approved):

PR #2542, opened 2021-04-07
Request camera permissions when opening QR code scanner from instant search screen

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3129/request-camera-permissions-when-opening-qr-code-scanner-from-instant-search-screen


Honey14 (approved):
Looks good

PR #2544, opened 2021-04-12
Show remove overdue appointment reasons in a separate screen - 1

msasikanth:


app/src/main/java/org/simple/clinic/removeoverdueappointment/RemoveOverdueUiRenderer.kt
```diff
@@ -0,0 +1,20 @@
+package org.simple.clinic.removeoverdueappointment
+
+import org.simple.clinic.contactpatient.RemoveAppointmentReason
+import org.simple.clinic.mobius.ViewRenderer
+
+class RemoveOverdueUiRenderer(private val ui: RemoveOverdueUi) : ViewRenderer<RemoveOverdueModel> {
+
+  override fun render(model: RemoveOverdueModel) {
+    val allRemoveAppointmentReasons = RemoveAppointmentReason.values().toList()
```
vinaysshenoy, line 9: Since this list will never change, should we make it a class property instead of a local variable?

app/src/test/java/org/simple/clinic/removeoverdueappointment/RemoveOverdueUiRendererTest.kt
```diff
@@ -28,4 +28,22 @@ class RemoveOverdueUiRendererTest {
     verify(ui).disableDoneButton()
     verifyNoMoreInteractions(ui)
   }
+
+  @Test
+  fun `render selected remove appointment reason`() {
```
vinaysshenoy, line 34: Can we rename this test method to better reflect the use case? Ex: "when a cancel reason is selected, the 'done' button must be enabled".

This comment also applies to the test name in the previous commit.

vinaysshenoy (changes requested):

app/src/test/java/org/simple/clinic/removeoverdueappointment/RemoveOverdueUpdateTest.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.simple.clinic.contactpatient.RemoveAppointmentReason
+
+class RemoveOverdueUpdateTest {
+
+  private val updateSpec = UpdateSpec(RemoveOverdueUpdate())
+  private val defaultModel = RemoveOverdueModel.create()
+
+  @Test
+  fun `when remove appointment reason is clicked, then update the UI`() {
+    val selectedReason = RemoveAppointmentReason.values().random()
```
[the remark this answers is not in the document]
msasikanth: But in this scenario won't we have to make sure that no matter any reason is selected the UI should update the selected state? 🤔

app/src/main/java/org/simple/clinic/removeoverdueappointment/RemoveOverdueUiRenderer.kt
msasikanth: Yeah, good point. Will make the change.

app/src/test/java/org/simple/clinic/removeoverdueappointment/RemoveOverdueUpdateTest.kt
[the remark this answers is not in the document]
msasikanth: NVM, since it's an enum checking with one item should automatically work with others as well.

vinaysshenoy (approved):

PR #2548, opened 2021-04-15
Move QR code scanner preview view inside `ScanSimpleIdScreen`

msasikanth:
Source message : https://simpledotorg.slack.com/archives/C01Q7S69AGL/p1618460636000500?thread_ts=1618460527.000400&cid=C01Q7S69AGL


app/src/main/java/org/simple/clinic/widgets/qrcodescanner/OnQrCodeDetected.kt
```diff
@@ -1,3 +0,0 @@
-package org.simple.clinic.widgets.qrcodescanner
-
-typealias OnQrCodeDetected = (String) -> Unit
```
Honey14, line 3: Did we delete this file and created the same in a different package? Want to understand why the typealias is red in color if we're using the same thing.

Honey14 (approved):
Looks good

PR #2549, opened 2021-04-15
Remove dialog & sheet from navigation back stack when dismissing

msasikanth:


app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseBottomSheet.kt
```diff
@@ -138,4 +138,9 @@ abstract class BaseBottomSheet<K : ScreenKey, B : ViewBinding, M : Parcelable, E
     super.onSaveInstanceState(outState)
     outState.putParcelable(KEY_MODEL, controller.model)
   }
+
+  override fun dismiss() {
+    requireActivity().onBackPressed()
```
Honey14, line 143: Can we remove our custom `backPressed` function now? Since we are overriding dismiss as a function, we could just call dismiss instead.

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseDialog.kt
```diff
@@ -102,4 +102,9 @@ abstract class BaseDialog<K : ScreenKey, M : Parcelable, E, F, R : ViewRenderer<
     super.onSaveInstanceState(outState)
     outState.putParcelable(KEY_MODEL, controller.model)
   }
+
+  override fun dismiss() {
+    requireActivity().onBackPressed()
```
Honey14, line 107: Same comment as above.

Honey14 (changes requested):
Looks good. Left few comments. PTAL

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseBottomSheet.kt
msasikanth: Good question, the back-pressed function is a private function so we cannot override it for child classes. But one of the reasons we have that is as a callback for `overrideCancellation` extension function.

We will still keep calling `router.pop` to dismiss the sheet when we are creating bottom sheets. This change is mainly to avoid any issues with the system or accidental calls to the `dismiss` function.

msasikanth: WDYT @vinaysshenoy ?

Honey14: Do we really need backPressed as a different function? Can we use dismiss and it'll look something like `dialog.overrideCancellation(::dismiss)`

msasikanth: `dismiss` has a super call that does more than just calling `requireActivity().onBackPressed()`. Like I mention the reason I added this was not that we can start using `dismiss` but instead just to handle cases where system calls `dismiss` and we don't end up with a  weird navigation state.

Honey14 (approved):
Looks good

PR #2553, opened 2021-04-20
Add `IndiaNationalHealthId` as an identifier in `IdentifierType`

Honey14:
https://app.clubhouse.io/simpledotorg/story/3196/add-new-identifiertype-for-indiannationalhealthid


app/src/main/java/org/simple/clinic/patient/businessid/Identifier.kt
```diff
@@ -39,6 +40,7 @@ data class Identifier(
       }
       BangladeshNationalId -> value
       EthiopiaMedicalRecordNumber -> value
+      IndiaNationalHealthId -> value
```
msasikanth, line 43: NHID should have spaces when displaying similar to BP passport, can you check with the design team regarding the spaces?

msasikanth (changes requested):
LGTM. Left one comment PTAL

app/src/main/java/org/simple/clinic/patient/businessid/Identifier.kt
Honey14: We're waiting for the design team to confirm this.

app/src/main/java/org/simple/clinic/patient/businessid/Identifier.kt
```diff
@@ -79,7 +87,19 @@ data class Identifier(
     object EthiopiaMedicalRecordNumber : IdentifierType()
 
     @Parcelize
-    object IndiaNationalHealthId : IdentifierType()
+    object IndiaNationalHealthId : IdentifierType() {
+
+      @IgnoredOnParcel
+      const val ENTERED_CODE_LENGTH = 14
+
+      fun enteredCodeLength(identifier: Identifier): String {
```
msasikanth, line 95: I don't think we need this. It was added for short code, because we need to get short code from the UUID. But since we are directly saving the 14 digit NHID we don't have to do this here.

app/src/main/java/org/simple/clinic/patient/businessid/Identifier.kt
```diff
@@ -40,7 +40,15 @@ data class Identifier(
       }
       BangladeshNationalId -> value
       EthiopiaMedicalRecordNumber -> value
-      IndiaNationalHealthId -> value
+      IndiaNationalHealthId -> {
+        val enteredCode = IndiaNationalHealthId.enteredCodeLength(this)
```
msasikanth, line 44: We can directly use `value` here, since we are saving 14 digit NHID in the DB.

app/src/main/java/org/simple/clinic/patient/businessid/Identifier.kt
```diff
@@ -40,7 +40,15 @@ data class Identifier(
...
+        val prefix = enteredCode.substring(0, 2)
+        val subString1 = enteredCode.substring(2,6)
+        val subString2 = enteredCode.substring(6,10)
+        val suffix = enteredCode.substring(10)
+
+        "$prefix${Unicode.nonBreakingSpace}$subString1${Unicode.nonBreakingSpace}$subString2${Unicode.nonBreakingSpace}$suffix"
```
msasikanth, line 50: Probably not for this PR, but maybe we can look into how we can extract out adding spacing to IDs into a common function or something like that. A factory perhaps.

msasikanth, line 50: Did we update the `IdentifierDisplayTest` to verify that we are getting the correct string here?

msasikanth (changes requested):

app/src/main/java/org/simple/clinic/patient/businessid/Identifier.kt
msasikanth: That unit test is failing, so do update that.

Honey14: Yes will take a look at it

msasikanth (approved):

PR #2555, opened 2021-04-21
Migrate Sentry Android to v4.3.0

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3251/migrate-sentry-android-to-v4-3-0

### DIFF
--- a/app/src/main/java/org/simple/clinic/crash/SentryCrashReporter.kt
+++ b/app/src/main/java/org/simple/clinic/crash/SentryCrashReporter.kt
@@ -4,8 +4,9 @@ import android.annotation.SuppressLint
 import android.app.Application
 import io.reactivex.schedulers.Schedulers.io
 import io.sentry.Sentry
-import io.sentry.android.AndroidSentryClientFactory
-import io.sentry.event.BreadcrumbBuilder
+import org.simple.clinic.BuildConfig.SENTRY_DSN
+import org.simple.clinic.BuildConfig.SENTRY_ENVIRONMENT
+import org.simple.clinic.BuildConfig.SENTRY_SAMPLE_RATE
 import org.simple.clinic.appconfig.AppConfigRepository
 import org.simple.clinic.facility.FacilityRepository
 import org.simple.clinic.platform.crash.Breadcrumb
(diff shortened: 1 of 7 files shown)


msasikanth:
Sentry migration guide: https://docs.sentry.io/platforms/android/migration/

gradle.properties
```diff
@@ -61,3 +61,21 @@ securityFallbackApiEndpoint=do_not_change_here
...
 # strips debug symbols from the APK.
 androidNdkVersion=21.3.6528147
+
+# Sentry Config
+## DSN
+qaSentryDsn=do_not_change_here
+sandboxSentryDsn=do_not_change_here
+stagingSentryDsn=do_not_change_here
+productionSentryDsn=do_not_change_here
+securitySentryDsn=do_not_change_here
```
vinaysshenoy, line 71: The DSN is also same for all the builds. We only need one.

app/src/main/java/org/simple/clinic/crash/SentryCrashReporter.kt
```diff
@@ -30,7 +31,11 @@ class SentryCrashReporter @Inject constructor(
 ) : CrashReporter {
 
   override fun init(appContext: Application) {
-    Sentry.init(AndroidSentryClientFactory(appContext))
+    Sentry.init { options ->
+      options.dsn = SENTRY_DSN
+      options.sampleRate = SENTRY_SAMPLE_RATE
+      options.environment = SENTRY_ENVIRONMENT
```
vinaysshenoy, line 37: The `sentry.properties` file also has an auth token which should be a secret. Where is that getting passed?

app/src/main/java/org/simple/clinic/crash/SentryCrashReporter.kt
```diff
@@ -64,19 +69,19 @@ class SentryCrashReporter @Inject constructor(
...
-        .setLevel(priorityToLevel(breadcrumb.priority))
-        .setCategory(breadcrumb.tag)
-        .setMessage(breadcrumb.message)
-        .setTimestamp(Date(System.currentTimeMillis()))
-        .build()
-    Sentry.getContext().recordBreadcrumb(sentryBreadcrumb)
+    val sentryBreadCrumb = SentryBreadcrumb().apply {
+      level = priorityToLevel(breadcrumb.priority)
+      category = breadcrumb.tag
+      message = breadcrumb.message
```
vinaysshenoy, line 81: Is the timestamp set automatically by Sentry now?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/crash/SentryCrashReporter.kt
msasikanth: Yep

app/src/main/java/org/simple/clinic/crash/SentryCrashReporter.kt
msasikanth: Do we really need it? The API doesn't expose a way to set that. As far as I can tell from the docs, I don't find it as needed for the Android implementation?

gradle.properties
msasikanth: Oh okay, will change it

app/src/main/java/org/simple/clinic/crash/SentryCrashReporter.kt
msasikanth: Found the use case for it Gradle plugin. Adding it there
https://docs.sentry.io/platforms/android/proguard/

vinaysshenoy (approved):

vinaysshenoy (approved):

PR #2557, opened 2021-04-21
Show patient results if they have same BP passport assigned

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3284/show-patient-results-if-they-have-same-bp-passport-assigned

### DIFF
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
@@ -5,6 +5,7 @@ import com.spotify.mobius.Next.noChange
 import com.spotify.mobius.Update
 import org.simple.clinic.mobius.dispatch
 import org.simple.clinic.mobius.next
+import org.simple.clinic.patient.Patient
 import org.simple.clinic.patient.businessid.Identifier
 import org.simple.clinic.patient.businessid.Identifier.IdentifierType.BpPassport
 import org.simple.clinic.platform.crash.CrashReporter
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -33,15 +34,23 @@ class ScanSimpleIdUpdate @Inject constructor(
...
     return next(model = model.notSearching(), SendScannedIdentifierResult(scanResult))
   }
 
+  private fun patientFoundByIdentifierSearch(patients: List<Patient>, identifier: Identifier): ScanResult {
+    return if (patients.size > 1) {
+      EnteredShortCode(BpPassport.shortCode(identifier))
+    } else {
+      val patientId = patients.first().uuid
+      PatientFound(patientId)
+    }
```
vinaysshenoy: It took me a while to realise that the only reason we are doing this is because the `ShortCodeSearchResultScreen` expects a short code as input.

This works for now, but I would recommend a better way to do it would be to change `ShortCodeSearchResultScreen` to accept multiple types of input (short codes/full identifiers).

vinaysshenoy: That way, it can also handle types of identifiers that may not have the notion of a short code (like the upcoming NHID).

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUpdateTest.kt
```diff
@@ -101,4 +101,25 @@ class ScanSimpleIdUpdateTest {
...
+
+  @Test
+  fun `when identifier is scanned and more than 1 patient is found, then send the short code to parent screen`() {
+    val patientId1 = UUID.fromString("60822507-9151-4836-944b-9cbbd1530c0b")
+    val patientId2 = UUID.fromString("de90d491-29ab-4bb7-938c-d436815794c6")
+
+    val patient1 = TestData.patient(uuid = patientId1)
+    val patient2 = TestData.patient(uuid = patientId2)
+
+    val identifier = Identifier("123456", BpPassport)
```
vinaysshenoy, line 113: This is a nitpick, but since a `BpPassport` identifier will be a UUID string in reality, we should try to use realistic looking data in tests.

vinaysshenoy (changes requested):
Overall PR looks good. Left a couple of comments that need not be addressed immediately. Will approve once you respond.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
msasikanth: Yeah, good point. Will look into the improvements we can make.

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUpdateTest.kt
msasikanth: Ah, my bad. I think I just copied from another test and forgot to update them. Will keep that in mind in the future.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
msasikanth: One of the reasons why I was fine with leaving it as the string is, we can just pass the identifier value and compare it across the different identifiers to show the results.

vinaysshenoy (approved):

PR #2560, opened 2021-04-26
Fix dead patients showing in overdue list

msasikanth:
Source: https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1619405180375500

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
@@ -75,6 +75,7 @@ import java.util.UUID
           
           WHERE 
             P.deletedAt IS NULL
+            AND P.status != 'dead'
             AND A.deletedAt IS NULL
             AND A.status = 'scheduled'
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
```diff
@@ -75,6 +75,7 @@ import java.util.UUID
           
           WHERE 
             P.deletedAt IS NULL
+            AND P.status != 'dead'
```
vinaysshenoy, line 78: Can we add a test to verify this?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
msasikanth: Good point.

vinaysshenoy (dismissed):

vinaysshenoy (approved):

PR #2565, opened 2021-04-28
Show national health id in patient summary screen

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3230/show-national-health-id-in-patient-summary-screen


app/src/main/res/values/strings.xml
```diff
@@ -28,7 +28,7 @@
 
   <!--Business id Identifier -->
   <string name="identifiertype_bp_passport">BP Passport</string>
-  <string name="identifiertype_bangladesh_national_id">National ID number</string>
```
vinaysshenoy, line 31: I'm wondering if this should just be a "Heath ID" label instead of national ID. Not necessary to wait for this to merge the PR, but might be worth opening that conversation with the design folks.

vinaysshenoy (approved):

app/src/main/res/values/strings.xml
msasikanth: We will provide different strings based on the countries. Discussed with designers [here](https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1619593931412000?thread_ts=1619588985.411500&cid=CFK2PHJ1L)

PR #2567, opened 2021-04-30
Show national health id in patient edit screen

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3335/show-national-health-id-in-patient-edit-screen

### DIFF
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
@@ -703,8 +703,8 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
     saveButton.setButtonState(Enabled)
   }
 
-  override fun setBangladeshNationalId(nationalId: String) {
-    alternativeIdInputEditText.setTextAndCursor(nationalId)
+  override fun setAlternateId(alternateId: String) {
+    alternativeIdInputEditText.setTextAndCursor(alternateId)
   }
 
   override fun onBackPressed(): Boolean {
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -705,9 +712,34 @@ class EditPatientScreen(context: Context, attributeSet: AttributeSet) : Relative
...
   override fun setAlternateId(alternateId: Identifier) {
+    when (alternateId.type) {
+      Identifier.IdentifierType.BangladeshNationalId -> setAlternateIdTextField(alternateId)
+      Identifier.IdentifierType.EthiopiaMedicalRecordNumber -> setAlternateIdTextField(alternateId)
+      Identifier.IdentifierType.IndiaNationalHealthId -> setAlternateIdContainer(alternateId)
+      else -> throw IllegalArgumentException("Unknown alternate id: $alternateId")
+    }
+  }
+
+  private fun setAlternateIdTextField(alternateId: Identifier) {
```
Honey14, line 723: We're not hiding the visibility of `alternateIdLabel` and it's container. Can we hide them in `setAlternateIdTextField` function?

Honey14 (changes requested):

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
Honey14: Ah I see. This makes sense

Honey14 (approved):

PR #2572, opened 2021-05-03
Add NHID support in scanning screen #1

Honey14:
https://app.clubhouse.io/simpledotorg/story/3211/add-support-for-indian-national-health-id-in-scansimpleidscreen


app/src/main/java/org/simple/clinic/scanid/PatientPrefillInfo.kt
```diff
@@ -0,0 +1,32 @@
+package org.simple.clinic.scanid
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+data class PatientPrefillInfo(
```
msasikanth, line 7: Aren't these variables supposed to be nullable?

app/src/main/java/org/simple/clinic/di/network/NetworkModule.kt
```diff
@@ -19,18 +19,18 @@ import org.simple.clinic.patient.businessid.BusinessId
 import org.simple.clinic.patient.businessid.Identifier
```
msasikanth, line 19: Just an FYI, we should not include a random format to a file that is not part of the PR.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -3,19 +3,22 @@ package org.simple.clinic.scanid
...
 import org.simple.clinic.platform.crash.CrashReporter
 import org.simple.clinic.scanid.ShortCodeValidationResult.Failure
 import org.simple.clinic.scanid.ShortCodeValidationResult.Success
 import java.util.UUID
 import javax.inject.Inject
 
 class ScanSimpleIdUpdate @Inject constructor(
-    private val crashReporter: CrashReporter
+    private val crashReporter: CrashReporter,
+    private val moshi: Moshi
```
msasikanth, line 21: Passing `Moshi`  directly to the Update function is not recommended, it causes the update function to be impure. Instead of this, we can verify whether the scanned information is JSON in effect handler and return the appropriate result back to the Update function. Similar to how we handle validation results.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -55,12 +58,39 @@ class ScanSimpleIdUpdate @Inject constructor(
     if (model.isSearching) return noChange()
 
     return try {
-      val bpPassportCode = UUID.fromString(event.text)
-      val identifier = Identifier(bpPassportCode.toString(), BpPassport)
-      next(model = model.searching(), SearchPatientByIdentifier(identifier))
-    } catch (e: IllegalArgumentException) {
-      crashReporter.report(e)
-      noChange()
+      val patientPrefillInfoDetail = parseJsonIntoObject(event.text)
```
msasikanth, line 61: We are nesting try/catch blocks. Can we think of a way to simplify this, please?

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -55,12 +58,39 @@ class ScanSimpleIdUpdate @Inject constructor(
...
+        crashReporter.report(e)
+        noChange()
+      }
+    }
+  }
+
+  private fun parseJsonIntoObject(text: String): PatientPrefillInfo? {
+    val adapter = moshi.adapter(PatientPrefillInfo::class.java)
+    val patientPrefillInfoDetail: PatientPrefillInfo? = adapter.fromJson(text)
+    return patientPrefillInfoDetail?.let {
```
msasikanth, line 83: We already have `PatientPrefillInfo?` above, so is there a reason we need to construct it again?

app/src/test/java/org/simple/clinic/scanid/ShortCodeInputTest.kt
```diff
@@ -56,4 +56,18 @@ class ShortCodeInputTest {
     assertThat(result)
         .isEqualTo(Empty)
   }
+
+  @Test
+  fun `when entered code length is equal to 14 then validation should succeed`() {
```
msasikanth: Can we also verify the error condition when the user enters a code that is more than 14 characters.

app/src/main/java/org/simple/clinic/scanid/EnteredCodeInput.kt
```diff
@@ -9,17 +9,17 @@ import org.simple.clinic.scanid.EnteredCodeValidationResult.Failure.NotEqualToRe
...
   fun validate(): EnteredCodeValidationResult {
     return when {
-      shortCodeText.isEmpty() -> Empty
-      shortCodeText.length != SHORT_CODE_LENGTH && shortCodeText.length != longCodeLength -> NotEqualToRequiredLength
-      shortCodeText.length == SHORT_CODE_LENGTH || shortCodeText.length == longCodeLength -> Success
-      else -> throw UnsupportedOperationException("Unknown situation for short code input: $shortCodeText")
+      enteredCodeText.isEmpty() -> Empty
+      enteredCodeText.length != SHORT_CODE_LENGTH && enteredCodeText.length != longCodeLength -> NotEqualToRequiredLength
+      enteredCodeText.length == SHORT_CODE_LENGTH || enteredCodeText.length == longCodeLength -> Success
+      else -> throw UnsupportedOperationException("Unknown situation for short code input: $enteredCodeText")
```
msasikanth, line 22: We should update the error message as well here.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEvent.kt
```diff
@@ -16,16 +16,16 @@ object HideKeyboard : ScanSimpleIdEvent() {
     get() = "Scan Simple Card:Hide keyboard"
 }
 
-object ShortCodeChanged : ScanSimpleIdEvent() {
+object EnteredCodeChanged : ScanSimpleIdEvent() {
   override val analyticsName: String
-    get() = "Scan Simple Card:Short code changed"
+    get() = "Scan Simple Card:Entered code changed"
```
msasikanth, line 21: FYI, once we changed the analytics name we need to update the reports on Mixpanel to track the new analytics names as well.

msasikanth (changes requested):
LGTM, left few comments PTAL

app/src/main/java/org/simple/clinic/scanid/PatientPrefillInfo.kt
Honey14: We are getting all the fields from the NHID. Yes, making state and district to be nullable.

msasikanth: Have checked the documentation to make sure the rest of the fields won't be nullable in any scenario?

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEvent.kt
Honey14: Will do this. Asked how to do this here : https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1620124416023800

app/src/main/java/org/simple/clinic/scanid/PatientPrefillInfo.kt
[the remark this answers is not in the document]
msasikanth: So in the end `PatientPrefillInfo` class is only used as a data class for `PatientEntryScreen` to prefill.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
```diff
@@ -28,8 +30,21 @@ class ScanSimpleIdEffectHandler @AssistedInject constructor(
...
 
+  private fun parseJsonIntoObject(): ObservableTransformer<ParseJsonIntoPatientPrefillInfoObject, ScanSimpleIdEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .map {
+            val adapter = moshi.adapter(PatientPrefillInfo::class.java)
+            adapter.fromJson(it.text)
+          }
+          .map(::EnteredCodeJsonParsed)
```
msasikanth, line 44: Json is from the scanned QR code and not entered code, can we rename this data class to reflect that, please?

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffect.kt
```diff
@@ -17,3 +17,5 @@ data class ValidateShortCode(val shortCode: ShortCodeInput) : ScanSimpleIdEffect
 data class SendScannedIdentifierResult(val scannedId: ScanResult) : ScanSimpleIdEffect()
 
 data class SearchPatientByIdentifier(val identifier: Identifier) : ScanSimpleIdEffect()
+
+data class ParseJsonIntoPatientPrefillInfoObject(val text: String) : ScanSimpleIdEffect()
```
msasikanth, line 21: Let's just call it `ParseScannedJson` or something like that. Instead of adding a specific class name at the end.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
```diff
@@ -11,7 +12,8 @@ import org.simple.clinic.util.scheduler.SchedulersProvider
 class ScanSimpleIdEffectHandler @AssistedInject constructor(
     private val schedulersProvider: SchedulersProvider,
     private val patientRepository: PatientRepository,
-    @Assisted private val uiActions: ScanSimpleIdUiActions
+    @Assisted private val uiActions: ScanSimpleIdUiActions,
+    private val moshi: Moshi
```
msasikanth, line 16: Instead of passing the `Moshi` object directly, can we instead have a wrapper that parses the data? That way we don't have to build an actual `Moshi` instance in tests and instead just use mocks. WDYT?

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandlerTest.kt
```diff
@@ -52,4 +55,32 @@ class ScanSimpleIdEffectHandlerTest {
...
+    "gender":"M",
+    "statelgd":"34",
+    "distlgd":"600",
+    "dob":"25/6/2012",
+    "address":"No.42,
+     eswaran kovil street"
+     }
+     """
+    val adapter = moshi.adapter(PatientPrefillInfo::class.java)
+    val deserialized: PatientPrefillInfo? = adapter.fromJson(expectedJson)
```
msasikanth, line 76: Let's rename this to `patientPrefillInfo` or whatever the deserialized object name is.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -26,6 +27,16 @@ class ScanSimpleIdUpdate @Inject constructor(
       is ShortCodeSearched -> next(model.shortCodeChanged(event.shortCode), ValidateShortCode(event.shortCode))
       is ScanSimpleIdScreenQrCodeScanned -> simpleIdQrScanned(model, event)
       is PatientSearchByIdentifierCompleted -> patientSearchByIdentifierCompleted(model, event)
+      is EnteredCodeJsonParsed -> enteredCodeParsed(model, event)
+    }
+  }
+
+  private fun enteredCodeParsed(model: ScanSimpleIdModel, event: EnteredCodeJsonParsed): Next<ScanSimpleIdModel, ScanSimpleIdEffect> {
```
msasikanth, line 34: The same comment as renaming `EnteredCodeJsonParsed`, let's rename this method as well.

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUpdateTest.kt
```diff
@@ -36,6 +51,50 @@ class ScanSimpleIdUpdateTest {
...
+        .whenEvent(ScanSimpleIdScreenQrCodeScanned(expectedJson))
+        .then(assertThatNext(
+            hasModel(defaultModel.searching()),
+            hasEffects(ParseJsonIntoPatientPrefillInfoObject(expectedJson))
+        ))
+  }
+
+  @Test
+  fun `when json is parsed then search patient with NHID`() {
+    val patientPrefillInfo: PatientPrefillInfo? = adapter.fromJson(expectedJson)
```
msasikanth, line 67: We should not be doing that here, instead, we can construct the parsed JSON object.

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUpdateTest.kt
```diff
@@ -36,6 +51,50 @@ class ScanSimpleIdUpdateTest {
...
+        .given(defaultModel)
+        .whenEvent(EnteredCodeJsonParsed(patientPrefillInfo))
+        .then(assertThatNext(
+            hasModel(defaultModel.searching()),
+            hasEffects(SearchPatientByIdentifier(identifier))
+        ))
+  }
+
+  @Test
+  fun `when searching for patient, then ignore newly scanned india national health Id identifier`() {
```
msasikanth, line 82: I don't understand this test, can you explain what we are trying to do here, please?

msasikanth (changes requested):

app/src/main/java/org/simple/clinic/scanid/PatientPrefillInfo.kt
Honey14: Yes i have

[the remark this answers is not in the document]
Honey14: We can still add the different JSON object class for other QR codes if needed, right? I'm not clear about why do we need it? Currently `PatientPrefillInfo` class can also be used as a data class to prefill, right?

app/src/test/java/org/simple/clinic/scanid/ShortCodeInputTest.kt
```diff
@@ -56,4 +56,45 @@ class ShortCodeInputTest {
...
+    "345678",
+    "34567890",
+    "345678900",
+    "3456789001",
+    "34567890012",
+    "345678900123",
+    "3456789001234",
+    "345678900123435",
+  ])
+  @Test
```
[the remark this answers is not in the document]
Honey14: We have spaces in between the tests. This is the parameter value for the same Test above the annotation. We don't need space here.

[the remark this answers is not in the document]
msasikanth: Ah, my bad.

app/src/main/java/org/simple/clinic/scanid/PatientPrefillInfo.kt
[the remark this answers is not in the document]
msasikanth: So for first question, we would create different JSON object classes for parsing and we would need to pass those to patient entry right? Because currently with prefill info class we are doing that. We are parsing directly with prefill info class and passing that to patient entry.

Instead what we should be doing is, we need to parse it with a different JSON object class, and then map it to `PatientPrefillInfo` class. That way even if we add new JSON object class in future, we would still end up mapping it to `PatientPrefillInfo` and sending it to the `PatientEntryScreen`. Does that make sense?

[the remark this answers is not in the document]
msasikanth: For the second question, yes, we are using `PatientPrefillInfo` class to prefill. But at the same time, we are using it to parse as well. So currently the variables/implementation is tied to NHID. So in future if we want to support other countries JSON we would end up creating a different prefill info class and using that for parsing different country.

Instead of doing all that we will use a different class for parsing and map the results to `PatientPrefillInfo`.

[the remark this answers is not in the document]
msasikanth: For a better example, consider our current data. We have a payload object and then we have our Room entities and then we have data classes that are specific to our UI rendering needs, right? So, we will be following a similar pattern, where we don't expose the JSON data directly to the entry screen and instead just provide `PatientPrefillInfo` class.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
Honey14: There seem to be no problem in creating Moshi instances in tests. Just checked their Github. I don't think we need this then

msasikanth: Hmm, @vinaysshenoy do you think it's fine for us to use Moshi directly in this scenario?

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUpdateTest.kt
Honey14: The scenario where we've scanned an ID and are searching for patient, while searching if we scan another then we ignore that ID.

app/src/main/java/org/simple/clinic/scanid/PatientPrefillInfo.kt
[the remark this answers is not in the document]
Honey14: The example made sense, understood.

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUpdateTest.kt
msasikanth: Hmm, that scenario won't happen, right? For example, when an ID is scanned and we are searching we show a progress bar and then navigate to a different screen based on the results. WDYT?

Honey14: Right. Checked ✅

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUpdateTest.kt
Honey14: Instead create a function with testdata in `TestData`.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
vinaysshenoy: Instead of Moshi, we can pass the adapter, maybe? This is what I generally do in this scenario. Apart from that, I think this is fine.

vinaysshenoy: A potential improvement can be have something like an `NHIDDataParser` which abstracts away the implementation detail that the scanned string is a JSON string, but I don't think it is important enough to hold up merging this PR in. Might be a future maintenance effort.

app/src/main/java/org/simple/clinic/scanid/PatientPrefillInfoPayload.kt
```diff
@@ -0,0 +1,32 @@
+package org.simple.clinic.scanid
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+data class PatientPrefillInfoPayload(
```
msasikanth, line 7: Let's rename this to `IndiaNHIDPayload` because this payload object is specific to India. That way in the future we can add more countries if needed. WDYT?

msasikanth, line 7: Also, correct me if I am wrong, we seem to be doing this in reverse order? We use the payload object for parsing then map it to `PatientPrefillInfo`. Currently, we are doing it another way around.

msasikanth (changes requested):
The overall approach seems good and it's ready to be merged in. Left one comment PTAL

app/src/main/java/org/simple/clinic/scanid/PatientPrefillInfoPayload.kt
Honey14: Oh yaa,right. thanks for bringing this up.

app/src/main/java/org/simple/clinic/scanid/IndiaNHIDInfo.kt
```diff
@@ -0,0 +1,12 @@
+package org.simple.clinic.scanid
+
+data class IndiaNHIDInfo(
```
msasikanth, line 3: Why did we rename `PatientPrefillInfo`? It's fine. We just had to rename the payload.

app/src/main/java/org/simple/clinic/scanid/IndiaNHIDInfoPayload.kt
```diff
@@ -29,4 +29,17 @@ data class IndiaNHIDInfoPayload(
 
     @Json(name = "address")
     val address: String
-)
+) {
+  fun fromPayload(): IndiaNHIDInfo {
```
msasikanth, line 33: A more appropriate name would be `toPatientPrefillInfo` wouldn't it? We want to say that the payload object is being converted to whatever the local model is.

msasikanth (changes requested):

msasikanth:
I am merging in the changes,  please do those renaming tasks in next PR

msasikanth (approved):

PR #2574, opened 2021-05-04
Show error dialog when assigning a National health id to a patient with an existing ID

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/3219/show-error-dialog-when-assigning-a-national-health-id-to-a-patient-with-an-existing-id


app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffectHandler.kt
```diff
@@ -73,6 +77,29 @@ class InstantSearchEffectHandler @AssistedInject constructor(
     }
   }
 
+  private fun checkIfPatientAlreadyHasAnExistingNHID(scheduler: Scheduler): ObservableTransformer<CheckIfPatientAlreadyHasAnExistingNHID, InstantSearchEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulers.computation())
+          .flatMap { patientRepository.patientProfile(it.patientId).subscribeOn(scheduler) }
```
vinaysshenoy, line 85: Why do we need to subscribe to this on the UI scheduler?

vinaysshenoy, line 85: Why do we need to subscribe to patient profile changes here?

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffectHandler.kt
```diff
@@ -73,6 +77,29 @@ class InstantSearchEffectHandler @AssistedInject constructor(
...
+          .map { patientProfile ->
+            val patientUUID = patientProfile.get().patientUuid
```
vinaysshenoy, line 87: We can use an additional `.map` operator to move the unwrapping of the `Optional` outside this block.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffectHandler.kt
```diff
@@ -73,6 +77,29 @@ class InstantSearchEffectHandler @AssistedInject constructor(
...
+              PatientAlreadyHasAnExistingNHID(patientUUID)
+            else
+              PatientDoesNotHaveAnExistingNHID(patientUUID)
+          }
+    }
+  }
+
+  private fun hasNationalHealthID(patientProfile: PatientProfile): Boolean {
+    return patientProfile
+        .withoutDeletedBusinessIds()
```
vinaysshenoy, line 99: We could do this removal of delete `BusinessId` instances earlier in the Rx chain.

```kotlin
.map { it.get().withoutDeletedBusinessIds() }
```

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffectHandler.kt
```diff
@@ -73,6 +77,29 @@ class InstantSearchEffectHandler @AssistedInject constructor(
...
+        .businessIds
+        .any { it.identifier.type == IndiaNationalHealthId }
```
vinaysshenoy, line 101: We could move this check to a computed property within the `PatientProfile` type.

app/src/test/java/org/simple/clinic/instantsearch/InstantSearchEffectHandlerTest.kt
```diff
@@ -335,4 +339,29 @@ class InstantSearchEffectHandlerTest {
     verify(uiActions).openQrCodeScanner()
     verifyNoMoreInteractions(uiActions)
   }
+
+  @Test
+  fun `when check if patient has an existing NHID effect is received, then check check if patient has an existing NHID`() {
```
vinaysshenoy, line 344: There seems to be an extra "check" in the test name.

vinaysshenoy, line 344: We have two branches added to the effect handler, but only one test. Are you not following the TDD process?

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchUpdate.kt
```diff
@@ -43,6 +43,8 @@ class InstantSearchUpdate @Inject constructor(
       } else {
         searchResultClicked(model, event.patientId)
       }
+      is PatientAlreadyHasAnExistingNHID -> dispatch(ShowNHIDErrorDialog)
+      is PatientDoesNotHaveAnExistingNHID -> dispatch(OpenLinkIdWithPatientScreen(event.patientId, model.additionalIdentifier!!))
```
vinaysshenoy, line 47: The handling of this change should be a separate commit and not part of this one. What we should do is use `noChange()` to let the tests pass and then handle this in another commit.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEvent.kt
```diff
@@ -71,3 +71,7 @@ data class OpenQrCodeScannerClicked(
 ) : InstantSearchEvent(), RequiresPermission {
   override val analyticsName = "Instant Search Screen:Open QR code scanner"
 }
+
+data class PatientAlreadyHasAnExistingNHID(val patientId: UUID): InstantSearchEvent()
```
vinaysshenoy, line 75: It does not seem like we are using the `patientId` property anywhere.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffectHandler.kt
```diff
@@ -48,6 +48,7 @@ class InstantSearchEffectHandler @AssistedInject constructor(
       .addConsumer(OpenShortCodeSearchScreen::class.java, ::openShortCodeSearch, schedulers.ui())
       .addAction(OpenQrCodeScanner::class.java, uiActions::openQrCodeScanner, schedulers.ui())
       .addTransformer(CheckIfPatientAlreadyHasAnExistingNHID::class.java, checkIfPatientAlreadyHasAnExistingNHID(schedulers.ui()))
+      .addConsumer(ShowNHIDErrorDialog::class.java, { uiActions.showNHIDErrorDialog() }, schedulers.ui())
```
vinaysshenoy, line 51: This should be an action, not a consumer, because we don't use any properties from the effect.

vinaysshenoy, line 51: A suggestion, we can make this easier to read by changing the lambda to a method reference.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/scanid/scannedqrcode/NationalHealthIDErrorDialog.kt
```diff
@@ -0,0 +1,28 @@
...
+    }
+  }
+
+  override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+    return MaterialAlertDialogBuilder(requireContext())
+        .setMessage(getString(R.string.nhid_error_cannot_assign_to_patient_with_existing_id_message))
+        .setPositiveButton(getString(R.string.nhid_error_cannot_assign_to_patient_with_existing_ok), null)
+        .create()
+  }
+}
```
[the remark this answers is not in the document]
msasikanth: Yeah, we can do that. I will move the code style from `.idea` folder to a `.editorconfig` file to keep it consistent instead of 2 separate files.

app/src/test/java/org/simple/clinic/instantsearch/InstantSearchEffectHandlerTest.kt
janhavisinghh: I am not handling the `ShowNHIDErrorDialog` effect here. I have written tests for the Update function where its being dispatched from. I handle this effect in the next commit.

janhavisinghh: Oh, you mean a test for asserting `PatientDoesNotHaveAnExistingNHID` event. I will do this.

vinaysshenoy: Yeah, I think a bigger concern is missing this test but having this branch present in the production code. This implies that you aren't following our TDD process, because otherwise you'd have added the test first before writing the production code.

Please make it a point to follow the process.

app/src/main/java/org/simple/clinic/scanid/scannedqrcode/NationalHealthIDErrorDialog.kt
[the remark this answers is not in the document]
vinaysshenoy: I think the ObviousX folks have set this up along with a commit hook that auto-formats on commit. We could look at incorporating this into the project.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchModel.kt
```diff
@@ -23,6 +24,9 @@ data class InstantSearchModel(
   val hasAdditionalIdentifier: Boolean
     get() = additionalIdentifier != null
 
+  val hasNHIDAdditionalIdentifier: Boolean
```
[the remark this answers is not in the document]
janhavisinghh: does `hasAdditionalIdentifierOfTheTypeNHID` sound good?

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchModel.kt
```diff
@@ -23,6 +24,9 @@ data class InstantSearchModel(
...
+  get() = additionalIdentifier?.type == IndiaNationalHealthId && hasAdditionalIdentifier
```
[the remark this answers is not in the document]
janhavisinghh: Isn't this something that should be taken care of by the code-styles file?

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchModel.kt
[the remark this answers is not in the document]
vinaysshenoy: No, it still doesn't convey the meaning. What we are looking for something like `isAdditionalIdentiderAnNHID` or something like that.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchModel.kt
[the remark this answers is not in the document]
vinaysshenoy: We're looking into making this automatic, but that doesn't mean it's fine for us to skip formatting our code when we write it.

vinaysshenoy:
@janhavisinghh There are conflicts and Code Climate issues. Can you resolve them before reviewing please?

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchModel.kt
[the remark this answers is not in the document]
janhavisinghh: I do try and refactor every time. Missed it out here, I think. Will be more careful from next time

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffectHandler.kt
janhavisinghh: Still trying to understand how to work with observables. I subscribed in order to read the data returned by the `patientRepository.patientProfile(it.patientId)` didn't know that the `.map` function on `.observeOn` takes care of it. Still learning RxJava better everyday.

janhavisinghh:
@vinaysshenoy resolved the issues. Can you please re review and merge?

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffectHandler.kt
vinaysshenoy: No, I mean there is no necessity to subscribe to changes at all here, right? We only need to query and check this once, for which `patientProfileImmediate` exists.

We should subscribe to change notifications only when we explicitly need to notified of changes to these records (for example, when displaying results in a list or something like that).

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffectHandler.kt
```diff
@@ -73,6 +76,22 @@ class InstantSearchEffectHandler @AssistedInject constructor(
...
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulers.computation())
+          .flatMap { patientRepository.patientProfile(it.patientId) }
+          .map { it.get().withoutDeletedBusinessIds() }
+          .map { checkIfPatientProfileHasNationalHealthId(it) }
+    }
+  }
+
+  private fun checkIfPatientProfileHasNationalHealthId(it: PatientProfile) = if (it.hasNationalHealthID) {
```
vinaysshenoy, line 89: Can you rename the method parameter please? When you use the IDE extract function, it does not do that good a job of selecting the name.

vinaysshenoy (changes requested):
The overall PR looks good to me, left a couple of small comments and we can merge once they are resolved.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffectHandler.kt
```diff
@@ -73,6 +74,20 @@ class InstantSearchEffectHandler @AssistedInject constructor(
...
+          .map { patientProfile -> patientProfile.get().withoutDeletedBusinessIds() }
+          .map { if (it.hasNationalHealthID) PatientAlreadyHasAnExistingNHID else PatientDoesNotHaveAnExistingNHID(it.patientUuid) }
+    }
+  }
+
+  private fun loadPatientProfileFromId(it: CheckIfPatientAlreadyHasAnExistingNHID) =
+      patientRepository
+          .patientProfile(it.patientId)
+          .subscribeOn(schedulers.io())
+
```
vinaysshenoy, line 91: Hmm, this hasn't changed anything from earlier, we are still subscribing to change notifications here.

What I meant was to just make it a synchronous call without needing to subscribe to changes, like this:

```kotlin
      effects
          .observeOn(schedulers.io())
          .map { patientRepository.patientProfileImmediate(it.patientId) }
          .map { patientProfile -> patientProfile.get().withoutDeletedBusinessIds() }
          .map { if (it.hasNationalHealthID) PatientAlreadyHasAnExistingNHID else PatientDoesNotHaveAnExistingNHID(it.patientUuid) }
    }
```

vinaysshenoy (changes requested):
Left a comment about the patient profile query.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchEffectHandler.kt
janhavisinghh: oh yes, sorry. Just pushed the branch with everything.

vinaysshenoy (approved):

PR #2575, opened 2021-05-04
Fix highlight color for instant search using phone number  

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/3360/if-you-do-an-instant-search-by-name-the-highlight-colour-is-yellow-but-it-s-blue-when-you-search-by-phone-number


Honey14 (approved):
Looks good

PR #2578, opened 2021-05-05
Bump AGP to v4.2.0

msasikanth:


app/build.gradle.kts
```diff
@@ -220,7 +220,6 @@ android {
   }
 
   lintOptions {
-    lintConfig = file("${project.rootDir}/quality/lint.xml")
```
vinaysshenoy, line 223: Moving this to the root makes it be picked up by default? So we don't need to explicitly set it anymore?

msasikanth: Yep

Honey14 (approved):
Looks good

PR #2579, opened 2021-05-05
Report SQL query performance to Mixpanel

vinaysshenoy:
Context: https://simpledotorg.slack.com/archives/CFKD42741/p1620193199258200


app/src/main/java/org/simple/clinic/storage/monitoring/AnalyticsReportingSqlDelegate.kt
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  private inline fun <reified R> runSqlOperationWithReporting(
+      daoInformation: DaoInformationExtractor.DaoMethodInformation,
+      operation: () -> R
+  ): R {
+    val operationStartTime = Instant.now(clock)
+    val result = operation()
+    val operationEndTime = Instant.now(clock)
+    val timeTakenForOperation = Duration.between(operationStartTime, operationEndTime).abs()
```
msasikanth: There is a standard library function we can use to get the time taken to run an operation.

```kt
val timeTakenForOperation = measureTimeMillis { operation() }
```

msasikanth (approved):

app/src/main/java/org/simple/clinic/storage/monitoring/AnalyticsReportingSqlDelegate.kt
vinaysshenoy: Interesting. Is this something new?

msasikanth: Nope, there since Kotlin 1.0

PR #2580, opened 2021-05-05
Bump Kotlin to v1.5.0

msasikanth:


msasikanth:
Moving to work in progress

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -13,9 +13,9 @@ object Versions {
   const val androidXTestExt = "1.1.1"
   const val androidXTest = "1.2.0"
   const val timber = "4.7.0"
-  const val dagger = "2.33"
+  const val dagger = "2.35.1"
   const val coreTesting = "2.0.0"
-  const val moshi = "1.11.0"
+  const val moshi = "1.12.0"
```
vinaysshenoy, line 19: Hmm, I would have preferred to do these in separate PRs instead of doing it in this one since errors can be caught sooner. This is fine for now, but lets try to make only one change per PR unless absolutely necessary.

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
```diff
@@ -142,8 +142,8 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
         patient = patientProfile.patient,
         address = patientProfile.address,
         phoneNumber = patientProfile.phoneNumbers.firstOrNull(),
-        bpPassport = patientProfile.businessIds.filter { it.identifier.type == BpPassport }.maxBy { it.createdAt },
-        alternativeId = patientProfile.businessIds.filter { it.identifier.type == country.alternativeIdentifierType }.maxBy { it.createdAt },
+        bpPassport = patientProfile.businessIds.filter { it.identifier.type == BpPassport }.maxByOrNull { it.createdAt },
+        alternativeId = patientProfile.businessIds.filter { it.identifier.type == country.alternativeIdentifierType }.maxByOrNull { it.createdAt },
```
vinaysshenoy, line 146: Are these just renames of the same method? The method behaviour has not changed, right?

build.gradle.kts
```diff
@@ -19,6 +19,7 @@ buildscript {
 
 plugins {
   id("com.diffplug.spotless") version versions.spotless
+  id("dev.zacsweers.kgp-150-leak-patcher") version versions.kgpLeakPatcher
```
vinaysshenoy, line 22: Can we add a comment on the `Kotlin` dependency version declaration to remind anyone who changes that version to check if this plugin will still be required for that version?

buildSrc/src/main/kotlin/Versions.kt
msasikanth: These are breaking changes, without these this PR would not compile. But yeah, I could have pulled out version bumps into a different PR and Kotlin change into another. Will remember for future such changes.

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
msasikanth: Yep

build.gradle.kts
msasikanth: Good point.

buildSrc/src/main/kotlin/Versions.kt
vinaysshenoy: Yeah, that's basically what I meant. If something breaks with these version changes, it is easier to make changes in a PR which only makes that change rather than one which also bundles other changes.

build.gradle.kts
```diff
@@ -9,6 +9,7 @@ buildscript {
 
   dependencies {
     classpath("com.android.tools.build:gradle:${versions.agp}")
+    // TODO: When bumping Kotlin version to 1.5.10. Remove the kgp-leak-patcher plugin
```
vinaysshenoy, line 12: This is the wrong place to add the comment because someone bumping up the version will not touch this section unless the version bump also requires a change to the maven coordinates. The right place to put this is at the version declaration site.

Does this make sense?

vinaysshenoy (changes requested):
Overall PR looks good, left a comment about moving the leak fix comment.

PR #2584, opened 2021-05-07
Remove explicit dependency on Trove4J from jcenter

vinaysshenoy:


msasikanth:
Why did we have this for `:lint` module? 🤔

vinaysshenoy:
> Why did we have this for `:lint` module? thinking

I think lint was depending on jcenter for this dependency. It's currently also available on Maven Central, was pushed in April 2021, probably because of the jcenter sunsetting.

msasikanth (approved):

PR #2586, opened 2021-05-07
Fix navigation issues in `Router`

msasikanth:


app/src/main/java/org/simple/clinic/navigation/v2/Router.kt
```diff
@@ -124,7 +124,7 @@ class Router(
       keyToPush: ScreenKey
   ) {
     val newHistory = history
-        .removeUntil { screenKey -> screenKey.matchesScreen(keyToPush) }
+        .removeUntil { screenKey -> !screenKey.matchesScreen(keyToPush) }
```
vinaysshenoy, line 127: Hmm, I just realised that the name `removeUntil` no longer makes sense for this specific method since it is going to pop the keys as long as the predicate keeps passing. Should we rename this to something like `removeWhile` to better reflect the change?

vinaysshenoy (changes requested):
Overall PR looks good, left one comment regarding method naming.

app/src/main/java/org/simple/clinic/navigation/v2/Router.kt
msasikanth: I think `removeUntil` makes sense as well. We are removing until the giving screen key, right? But if you think think the base method should be `removeWhile` instead of `removeUntil` I am fine with renaming.

vinaysshenoy: We aren't removing until the given key, there is a separate method which takes in a `ScreenKey` as a parameter for which this naming makes sense. The specific method I am referring to is the one which takes in ` Predicate` since we are actually removing keys until the predicate fails, while the name implies that we removing it until the predicate passes. This is exactly the opposite of what it is actually doing, so might be confusing.

vinaysshenoy (approved):

PR #2588, opened 2021-05-11
Show patients with national health id when searching with from scan screen

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3218/show-patients-with-national-health-id-when-searching-with-from-scan-screen

### DIFF
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
@@ -47,7 +47,7 @@ class ScanSimpleIdUpdate @Inject constructor(
       event: PatientSearchByIdentifierCompleted
   ): Next<ScanSimpleIdModel, ScanSimpleIdEffect> {
     val effect = if (event.patients.isEmpty()) {
-      OpenPatientSearch(event.identifier)
+      OpenPatientSearch(event.identifier, null)
     } else {
       patientFoundByIdentifierSearch(patients = event.patients, identifier = event.identifier)
     }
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -57,17 +57,20 @@ class ScanSimpleIdUpdate @Inject constructor(
...
     } else {
       val patientId = patients.first().uuid
       OpenPatientSummary(patientId)
     }
   }
 
+  private fun multiplePatientsWithId(identifier: Identifier): ScanSimpleIdEffect {
+    return when (identifier.type) {
+      BpPassport -> OpenShortCodeSearch(BpPassport.shortCode(identifier))
+      else -> OpenPatientSearch(null, initialSearchQuery = identifier.value)
```
Honey14, line 70: Why are we passing the identifier as null here and in the tests as well? Does this mean apart from BP passport just search with ID regardless of whatever identifier it is?

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUpdateTest.kt
```diff
@@ -162,4 +162,18 @@ class ScanSimpleIdUpdateTest {
             hasEffects(OpenPatientSearch(null, "12345612345612"))
         ))
   }
+
+  @Test
+  fun `when entered code is not a short code, then open patient search`() {
+    val shortCode = "12345678901234"
```
Honey14, line 168: ~Can we rename `shortCode` to `enteredCode` here?~ I have a renaming PR coming up for this package will include this as well.

Honey14 (approved):
Looks good.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
msasikanth: Yep, if we pass an identifier to instant search it tries to bring up the sheet that asks if we want to register or assign it to a patient. In this case, this is not a blank identifier, we know this is already assigned to a patient. I should probably switch to using factory constructors for this to make it much more apparent what it's doing 🤔

PR #2590, opened 2021-05-11
Change labels to indicate the ID search

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3451/change-labels-to-indicate-the-id-search


app/src/main/res/values/strings.xml
```diff
@@ -142,7 +142,7 @@
   <!--suppress UnusedResources -->
   <string name="tab_progress">Progress</string>
 
-  <string name="home_search_patient_name_phone">Enter patient\'s name</string>
+  <string name="home_search_patient_name_phone">Enter name, phone or ID</string>
```
janhavisinghh, line 146: Not that important to change, but the string name should also include id now, I think. `home_search_patient_name_phone_id`

msasikanth: Yeah, it's fine. In this scenario, I didn't want to remove all the old strings and add a new key. the `home_search` aspect is still true and we don't have to keep adding the supported searches to the key.

PR #2591, opened 2021-05-11
Bump Mobius dependency version

msasikanth:


app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityEffectHandler.kt
```diff
@@ -8,24 +8,23 @@ import org.simple.clinic.facility.FacilityRepository
 import org.simple.clinic.patient.PatientRepository
 import org.simple.clinic.patient.PatientSearchResult
 import org.simple.clinic.util.scheduler.SchedulersProvider
+import javax.inject.Inject
```
Honey14, line 11: We did not need this commit in the current PR. We should've created a different PR for maintaining this. Merging this now.

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -54,7 +54,7 @@ object Versions {
   const val playCore = "1.10.0"
   const val fastScroll = "2.0.0"
   const val mixpanel = "5.6.4"
-  const val mobius = "1.3.4"
+  const val mobius = "1.5.3"
```
Honey14, line 57: Was changing the version twice approach to make sure there are no major changes in the versions and nothing is broke?

Honey14 (approved):
Looks good. Left 2 questions

buildSrc/src/main/kotlin/Versions.kt
msasikanth: Yep

app/src/main/java/org/simple/clinic/allpatientsinfacility/AllPatientsInFacilityEffectHandler.kt
msasikanth: Actually, we had for this 😅 . With Mobius update, the older approach was not working for this, so I had to change it to our latest implementations.

Honey14 (approved):

PR #2611, opened 2021-05-11
Use `PatientPrefillInfo` for prefilling patient entry information

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/3215/use-indianhidinfo-for-prefilling-patient-entry-information\

Currently you will see that the nhid number is being shown where the bp passport number is shown. The handling of this is being done in [another story](https://app.clubhouse.io/simpledotorg/story/3535/add-support-for-nhid-in-patiententryscreen)

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientsTabScreen.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientsTabScreen.kt
@@ -214,7 +214,8 @@ class PatientsTabScreen : BaseScreen<
   override fun openPatientSearchScreen(additionalIdentifier: Identifier?) {
     val screenKey = InstantSearchScreenKey(
         additionalIdentifier = additionalIdentifier,
-        initialSearchQuery = null)
+        initialSearchQuery = null,
+        patientPrefillInfo = null)
 
     router.push(screenKey)
   }
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchModel.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchModel.kt
@@ -74,4 +74,8 @@ data class InstantSearchModel(
   fun additionalIdentifierUpdated(additionalIdentifier: Identifier): InstantSearchModel {
     return copy(additionalIdentifier = additionalIdentifier)
   }
+
+  fun patientPrefillInfoUpdated(patientPrefillInfo: PatientPrefillInfo): InstantSearchModel {
+    return copy(patientPrefillInfo = patientPrefillInfo)
+  }
 }
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchModel.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchModel.kt
@@ -5,6 +5,7 @@ import kotlinx.parcelize.Parcelize
 import org.simple.clinic.facility.Facility
 import org.simple.clinic.patient.businessid.Identifier
 import org.simple.clinic.patient.businessid.Identifier.IdentifierType.IndiaNationalHealthId
+import org.simple.clinic.scanid.PatientPrefillInfo
 
 @Parcelize
 data class InstantSearchModel(
@@ -12,7 +13,8 @@ data class InstantSearchModel(
     val searchQuery: String?,
     val additionalIdentifier: Identifier?,
     val instantSearchProgressState: InstantSearchProgressState?,
-    val scannedQrCodeSheetAlreadyOpened: Boolean
+    val scannedQrCodeSheetAlreadyOpened: Boolean,
+    val patientPrefillInfo: PatientPrefillInfo?
 ) : Parcelable {
 
   val hasFacility: Boolean
@@ -27,13 +29,17 @@ data class InstantSearchModel(
   val isAdditionalIdentifierAnNHID: Boolean
     get() = additionalIdentifier?.type == IndiaNationalHealthId && hasAdditionalIdentifier
 
+  val canBePrefilled: Boolean
+    get() = patientPrefillInfo != null && isAdditionalIdentifierAnNHID
+
   companion object {
     fun create(additionalIdentifier: Identifier?) = InstantSearchModel(
         facility = null,
         searchQuery = null,
         additionalIdentifier = additionalIdentifier,
         instantSearchProgressState = null,
-        scannedQrCodeSheetAlreadyOpened = false
+        scannedQrCodeSheetAlreadyOpened = false,
+        patientPrefillInfo = null
     )
   }
 
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchModel.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchModel.kt
@@ -33,13 +33,13 @@ data class InstantSearchModel(
     get() = patientPrefillInfo != null && isAdditionalIdentifierAnNHID
 
   companion object {
-    fun create(additionalIdentifier: Identifier?) = InstantSearchModel(
+    fun create(additionalIdentifier: Identifier?, patientPrefillInfo: PatientPrefillInfo?) = InstantSearchModel(
         facility = null,
         searchQuery = null,
         additionalIdentifier = additionalIdentifier,
         instantSearchProgressState = null,
         scannedQrCodeSheetAlreadyOpened = false,
-        patientPrefillInfo = null
+        patientPrefillInfo = patientPrefillInfo
     )
   }
 
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreenKey.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreenKey.kt
@@ -5,11 +5,13 @@ import kotlinx.android.parcel.IgnoredOnParcel
 import kotlinx.parcelize.Parcelize
 import org.simple.clinic.navigation.v2.ScreenKey
 import org.simple.clinic.patient.businessid.Identifier
+import org.simple.clinic.scanid.PatientPrefillInfo
 
 @Parcelize
 data class InstantSearchScreenKey(
     val additionalIdentifier: Identifier?,
-    val initialSearchQuery: String?
+    val initialSearchQuery: String?,
+    val patientPrefillInfo: PatientPrefillInfo?
 ) : ScreenKey(), Parcelable {
 
   @IgnoredOnParcel
--- a/app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
@@ -105,8 +105,13 @@ class PatientEntryEffectHandler @AssistedInject constructor(
           .flatMapSingle { getPatientEntryAndFacility }
           .subscribeOn(scheduler)
           .map { (entry, facility) ->
-            entry.takeIf { it.address != null }
-                ?: entry.withAddress(Address.withDistrictAndState(facility.district, facility.state))
+            if (entry.address != null && entry.address.doesNotHaveDistrictAndState) {
+              entry.withDistrict(facility.district).withState(facility.state)
+            } else if (entry.address != null) {
+              entry
+            } else {
+              entry.withAddress(Address.withDistrictAndState(facility.district, facility.state))
+            }
           }
           .map { OngoingEntryFetched(it) }
     }
--- a/app/src/main/java/org/simple/clinic/patient/Gender.kt
+++ b/app/src/main/java/org/simple/clinic/patient/Gender.kt
@@ -11,6 +11,7 @@ import org.simple.clinic.patient.Gender.Female
 import org.simple.clinic.patient.Gender.Male
 import org.simple.clinic.patient.Gender.Transgender
 import org.simple.clinic.patient.Gender.Unknown
+import org.simple.clinic.scanid.IndiaNHIDGender
 import org.simple.clinic.util.room.SafeEnumTypeAdapter
 
 sealed class Gender : Parcelable {
@@ -54,6 +55,16 @@ sealed class Gender : Parcelable {
 
     @ToJson
     fun toJson(answer: Gender): String? = TypeAdapter.fromEnum(answer)
+  }
+
+  companion object {
+    fun fromIndiaNHIDToGender(value: IndiaNHIDGender): Gender {
+      return when (value) {
+        IndiaNHIDGender.MALE -> Male
+        IndiaNHIDGender.FEMALE -> Female
+        else -> Unknown("U")
+      }
+    }
   }
 }
 
--- a/app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
+++ b/app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
@@ -28,6 +28,7 @@ import org.simple.clinic.registration.phone.PhoneNumberValidator.Result.Blank
 import org.simple.clinic.registration.phone.PhoneNumberValidator.Result.LengthTooLong
 import org.simple.clinic.registration.phone.PhoneNumberValidator.Result.LengthTooShort
 import org.simple.clinic.registration.phone.PhoneNumberValidator.Type.LANDLINE_OR_MOBILE
+import org.simple.clinic.scanid.PatientPrefillInfo
 import org.simple.clinic.util.Optional
 import org.simple.clinic.util.toNullable
 import org.simple.clinic.widgets.ageanddateofbirth.UserInputAgeValidator
@@ -116,6 +117,17 @@ data class OngoingNewPatientEntry(
 
   fun withZone(zone: String): OngoingNewPatientEntry =
       copy(address = addressOrBlank().copy(zone = zone))
+
+  fun withPatientPrefillInfo(patientProfileInfo: PatientPrefillInfo, identifier: Identifier): OngoingNewPatientEntry =
+      copy(
+          personalDetails = PersonalDetails(
+              fullName = patientProfileInfo.fullName,
+              dateOfBirth = null,
+              gender = Gender.Unknown(patientProfileInfo.gender),
+              age = null),
+          address = addressOrBlank().withColonyOrVillage(patientProfileInfo.address),
+          identifier = identifier)
+
 
   fun validationErrors(
       dobValidator: UserInputDateValidator,
--- a/app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
+++ b/app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
@@ -119,11 +119,11 @@ data class OngoingNewPatientEntry(
   fun withZone(zone: String): OngoingNewPatientEntry =
       copy(address = addressOrBlank().copy(zone = zone))
 
-  fun withPatientPrefillInfo(patientProfileInfo: PatientPrefillInfo, identifier: Identifier): OngoingNewPatientEntry =
+  fun withPatientPrefillInfo(patientProfileInfo: PatientPrefillInfo, identifier: Identifier, dateTimeFormatter: DateTimeFormatter): OngoingNewPatientEntry =
       copy(
           personalDetails = PersonalDetails(
               fullName = patientProfileInfo.fullName,
-              dateOfBirth = patientProfileInfo.dateOfBirth.format(DateTimeFormatter.ofPattern("dd/MM/yyyy")),
+              dateOfBirth = patientProfileInfo.dateOfBirth.format(dateTimeFormatter),
               gender = patientProfileInfo.gender,
               age = null),
           address = addressOrBlank().withColonyOrVillage(patientProfileInfo.address),
--- a/app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
+++ b/app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
@@ -289,6 +289,9 @@ data class OngoingNewPatientEntry(
           Address("", district, state, "", "")
     }
 
+    val doesNotHaveDistrictAndState: Boolean
+      get() = district.isBlank() && state.isBlank()
+
     fun withColonyOrVillage(colonyOrVillage: String): Address =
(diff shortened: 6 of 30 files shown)


vinaysshenoy:
@janhavisinghh This still seems to be a `work in progress` PR. Is this intentional?

janhavisinghh:
@vinaysshenoy yeah, working on to add few more things in this PR which we figured out later. This should be up for review today. ([slack message](https://simpledotorg.slack.com/archives/C01T2MYJB97/p1620835412007700))

janhavisinghh:
@vinaysshenoy this is ready for review. There is on small thing while rebasing I messed up a bit while merging the branches so I've made a [commit](https://github.com/simpledotorg/simple-android/pull/2611/commits/bd403d9db57741302aca96da99635a2c96550927) to fix that. Please, let me know if there is a better way to handle this. Other than that everything should be good to go.

app/src/main/java/org/simple/clinic/util/moshi/LocalDateMoshiAdapter.kt
```diff
@@ -4,22 +4,34 @@ import com.squareup.moshi.FromJson
...
   }
 
   @ToJson
   fun fromLocalDate(date: LocalDate?): String? {
-    return date?.format(formatter)
+    return try {
+      date?.format(formatter)
+    } catch (e: DateTimeParseException) {
+      date?.format(indiaNHIDFormatter)
+    }
```
vinaysshenoy, line 35: This is the wrong way to support alternate adapters for a specific type based on the source. The right way to do this is to use Moshi's `@JsonQualifier` annotation.

Please look at the relevant documentation here: https://github.com/square/moshi#alternate-type-adapters-with-jsonqualifier

app/src/main/java/org/simple/clinic/util/moshi/LocalDateMoshiAdapter.kt
```diff
@@ -4,22 +4,34 @@ import com.squareup.moshi.FromJson
...
 import java.time.LocalDate
 import java.time.format.DateTimeFormatter
+import java.time.format.DateTimeParseException
+import java.util.Date
 
 class LocalDateMoshiAdapter {
 
   companion object {
     private val formatter = DateTimeFormatter.ISO_DATE!!
+    private val indiaNHIDFormatter = DateTimeFormatter.ofPattern("d/M/yyyy")
```
vinaysshenoy, line 14: When defining this date formatter, we also should be explicitly setting the locale to `Locale.ENGLISH` here. Otherwise, we might run into issues if users have the device locale set to something other than non-english.

app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
```diff
@@ -118,7 +119,7 @@ data class OngoingNewPatientEntry(
       copy(
           personalDetails = PersonalDetails(
               fullName = patientProfileInfo.fullName,
-              dateOfBirth = null,
+              dateOfBirth = patientProfileInfo.dateOfBirth.format(DateTimeFormatter.ofPattern("dd/MM/yyyy")),
```
vinaysshenoy, line 122: Please define date formatters in `DateFormatterModule` and pass them to the method as a parameter instead of hardcoding them here.

In fact, we already have a formatter for this specific format already defined.

app/src/sharedTest/java/org/simple/clinic/TestData.kt
```diff
@@ -1226,7 +1227,7 @@ object TestData {
       gender: String = "M",
       state: String = "Maharashtra",
       district: String = "Thane",
-      dateOfBirth: String = "12/12/1997",
+      dateOfBirth: LocalDate = LocalDate.parse("12/12/1997", DateTimeFormatter.ofPattern("d/M/yyyy")),
```
vinaysshenoy, line 1231: What's the need to create a formatter here? Using `LocalDate.parse` should be sufficient, right?

app/src/test/java/org/simple/clinic/instantsearch/InstantSearchUpdateTest.kt
```diff
@@ -315,14 +316,53 @@ class InstantSearchUpdateTest {
...
+  }
+
+  @Test
+  fun `when register new patient is clicked and patient prefill info is not empty and dob is in wrong format, then save it in ongoing patient entry with dob in correct format`(){
+    val facility = TestData.facility()
+    val searchQueryModel = defaultModel
+        .facilityLoaded(facility)
+        .searchQueryChanged("Pat")
+
+    val dateOfBirthWithWrongFormat = LocalDate.parse("2/3/2012", DateTimeFormatter.ofPattern("d/M/yyyy"))
```
vinaysshenoy, line 351: I am not understanding how this is in the wrong format? If we expect the date in the NHID payload to follow a certain pattern, then yes, that can be wrong.

However, here we are creating a valid `LocalDate` and it can never be in the wrong format at all.

vinaysshenoy: Here, what you are actually testing is whether `LocalDate` is using the `DateTimeFormatter` passed to the constructor which we should not be testing since we don't own it.

app/src/main/java/org/simple/clinic/patient/Gender.kt
```diff
@@ -65,6 +65,14 @@ val Gender.displayLetterRes: Int
       is Unknown -> R.string.gender_unknown_letter
     }
 
+val Gender.displayChar : String
+  get() = when(this) {
+    Male -> "M"
+    Female -> "F"
+    Transgender -> "T"
+    is Unknown -> "U"
+  }
```
vinaysshenoy, line 74: We already have a `displayLetterRes` property which handles different languages. Why are we introducing a hard-coded english character string here?

app/src/main/java/org/simple/clinic/scanid/IndiaNHIDGender.kt
```diff
@@ -0,0 +1,54 @@
...
+}
+
+val IndiaNHIDGender.toGender: Gender
+  get() = when (this) {
+    IndiaNHIDGender.Male -> Gender.Male
+    IndiaNHIDGender.Female -> Gender.Female
+    else -> Gender.Unknown(this.toChar)
+  }
+
+val IndiaNHIDGender.toChar: String
```
vinaysshenoy, line 49: This should be `private` since we don't want it to be used anywhere else.

app/src/main/java/org/simple/clinic/scanid/IndiaNHIDGender.kt
```diff
@@ -0,0 +1,54 @@
...
+  get() = when (this) {
+    IndiaNHIDGender.Male -> "M"
+    IndiaNHIDGender.Female -> "F"
```
vinaysshenoy, line 52: Since we technically don't need to handle `Male` and `Female` gender letters, we can just remove this property and use `U` directly in the other property `when` clause, right?

app/src/main/java/org/simple/clinic/scanid/IndiaNHIDGender.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    @FromJson
+    fun fromJson(value: String?): IndiaNHIDGender? = TypeAdapter.toEnum(value)
+
+    @ToJson
+    fun toJson(answer: IndiaNHIDGender): String? = TypeAdapter.fromEnum(answer)
+  }
+}
+
+val IndiaNHIDGender.toGender: Gender
```
vinaysshenoy, line 42: The `toGender` name implies that this is an action that can be performed, but this has been defined as a property instead of a method. Please either rename the property or convert it to a method so it makes sense.

vinaysshenoy: An alternate suggestion would be to add a static factory method `fromIndiaNHIDGender` to the `Gender` type.

app/src/main/java/org/simple/clinic/scanid/IndiaNHIDGender.kt
```diff
@@ -0,0 +1,54 @@
...
+
+import android.os.Parcelable
+import androidx.annotation.VisibleForTesting
+import com.squareup.moshi.FromJson
+import com.squareup.moshi.ToJson
+import kotlinx.parcelize.Parcelize
+import org.simple.clinic.patient.Gender
+import org.simple.clinic.util.room.SafeEnumTypeAdapter
+
+sealed class IndiaNHIDGender : Parcelable {
```
vinaysshenoy, line 11: This does not need to be a sealed class with a `SafeEnumTypeAdapter` since this is only necessary for types which are both present in the API *and* are saved in the local database. Since we do not store the `IndiaNHIDPayload` in the database and only consume it at runtime, we can use a regular enum type.

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
```diff
@@ -105,7 +105,13 @@ class PatientEntryEffectHandler @AssistedInject constructor(
           .flatMapSingle { getPatientEntryAndFacility }
           .subscribeOn(scheduler)
           .map { (entry, facility) ->
-            entry.takeIf { it.address != null } ?: entry.withAddress(Address.withDistrictAndState(facility.district, facility.state))
+            if (entry.address != null && entry.address.doesNotHaveDistrictAndState) {
```
vinaysshenoy, line 108: This change is unrelated to this PR, please extract it to a different PR.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
```diff
@@ -279,9 +279,13 @@ class ScanSimpleIdScreen : BaseScreen<
...
-  override fun openPatientSearch(additionalIdentifier: Identifier?, initialSearchQuery: String?) {
-    val keyToPush = InstantSearchScreenKey(additionalIdentifier = additionalIdentifier,
-        initialSearchQuery = null)
+  override fun openPatientSearch(additionalIdentifier: Identifier?, initialSearchQuery: String?, patientPrefillInfo: PatientPrefillInfo?) {
+    val keyToPush = if (features.isEnabled(Feature.InstantSearch)) {
+      InstantSearchScreenKey(additionalIdentifier = additionalIdentifier,
+          initialSearchQuery = null)
+    } else {
+      PatientSearchScreenKey(additionalIdentifier).wrap()
+    }
```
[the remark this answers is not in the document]
janhavisinghh: Ok, so this is where I messed up the rebase. I am fixing this. This commit only adds patientPrefillInfo as a parameter in the openPatientSearch method

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUpdateTest.kt
```diff
@@ -126,10 +126,32 @@ class ScanSimpleIdUpdateTest {
         .whenEvent(PatientSearchByIdentifierCompleted(patients, identifier))
         .then(assertThatNext(
             hasModel(defaultModel.notSearching()),
-            hasEffects(OpenPatientSearch(identifier, null))
+            hasEffects(OpenPatientSearch(identifier, null, null))
         ))
   }
 
+  @Test
+  fun `when national health identifier is scanned and patient is not found, then send the identifier to parent screen with patient prefill info`() {
```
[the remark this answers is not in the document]
janhavisinghh: it is referring to patient search screen. I will make it more explicit.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchModel.kt
```diff
@@ -27,13 +29,17 @@ data class InstantSearchModel(
   val isAdditionalIdentifierAnNHID: Boolean
     get() = additionalIdentifier?.type == IndiaNationalHealthId && hasAdditionalIdentifier
 
+  val hasPatientPrefillInfo: Boolean
```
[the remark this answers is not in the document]
janhavisinghh: `hasPatientPrefillInfoAndAdditionalIdentifierIsNHID` this sounds good?

app/src/main/java/org/simple/clinic/home/patients/PatientsTabScreen.kt
```diff
@@ -212,14 +212,10 @@ class PatientsTabScreen : BaseScreen<
       .map { SimpleVideoClicked }
 
   override fun openPatientSearchScreen(additionalIdentifier: Identifier?) {
-    val screenKey = if (features.isEnabled(Feature.InstantSearch)) {
```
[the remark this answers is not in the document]
janhavisinghh: I have resolved this issue so this commit will be dropped now. I found out the commit where the missed changes were.

app/src/main/java/org/simple/clinic/util/moshi/LocalDateMoshiAdapter.kt
janhavisinghh: @vinaysshenoy as far as I could understand the json qualifier annotation is used when we need different encodings for the same json type.. so for example, if there is a hex code and two integer in a json object we will use `@JsonQualifier` to differentiate the hex code in our code as colour is represented as integer type. I didn't understand how can I use this to put the condition if the date formatter is of a certain type do this and if it's not do something else.

janhavisinghh: Got it

app/src/main/java/org/simple/clinic/patient/Gender.kt
janhavisinghh: We don't. Removing it.

app/src/sharedTest/java/org/simple/clinic/TestData.kt
janhavisinghh: No. I was getting `DateTimeParseException` when I was going that as the default local date pattern is `yyyy-MM-dd`

app/src/test/java/org/simple/clinic/instantsearch/InstantSearchUpdateTest.kt
janhavisinghh: This is because we are saving the `dateOfBirth` that is being returned in the json response as `LocalDate` type. But, the format we want to prefill the `dateOfBirth` in `PatientEntryScreen` is in `dd/MM/yyyy` which is of `String` type. Which is what I'm trying to test here. If, the `LocalDate` is being converted into the correct format or not.

app/src/main/java/org/simple/clinic/newentry/PatientEntryEffectHandler.kt
janhavisinghh: I made this change part of this PR because when we prefilled the data then we added data to the `colonyOrVillage` field therefore making `entry.address != null` and the district and state fields to not being prefilled. Which kinda broke the previous flow.

app/src/main/java/org/simple/clinic/scanid/IndiaNHIDGender.kt
janhavisinghh: Ummm. Fair enough.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchModel.kt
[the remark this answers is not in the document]
vinaysshenoy: That name exposes the implementation details. Instead, we should name this based on usage. Example: `canBePrefilled`.

app/src/sharedTest/java/org/simple/clinic/TestData.kt
vinaysshenoy: Yeah, this means that you are passing the date in the wrong format to `LocalDate.parse`. This version of the factory method should only be used when the date string is coming from a source that you don't have control over, which is not the case here.

app/src/test/java/org/simple/clinic/instantsearch/InstantSearchUpdateTest.kt
vinaysshenoy: Yeah, but this is the wrong place to do it. The place this needs to be tested is at the time of scanning the QR code, because once you parse a `LocalDate`, it is the same regardless of what format the string originally was,

vinaysshenoy: As mentioned earlier, the only thing this test is doing is verifying the behaviour of `LocalDate.parse`, which is something we should not be testing.

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreenKey.kt
```diff
@@ -5,11 +5,13 @@ import kotlinx.android.parcel.IgnoredOnParcel
...
 import org.simple.clinic.navigation.v2.ScreenKey
 import org.simple.clinic.patient.businessid.Identifier
+import org.simple.clinic.scanid.PatientPrefillInfo
 
 @Parcelize
 data class InstantSearchScreenKey(
     val additionalIdentifier: Identifier?,
-    val initialSearchQuery: String?
+    val initialSearchQuery: String?,
+    val patientPrefillInfo: PatientPrefillInfo?
```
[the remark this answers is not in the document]
janhavisinghh: Created a [story](https://app.clubhouse.io/simpledotorg/story/3639/resolve-coupling-between-scanid-and-instantsearch-packages?vc_group_by=day&ct_workflow=all) for this will resolve in another PR.

app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
```diff
@@ -113,6 +114,17 @@ data class OngoingNewPatientEntry(
   fun withZone(zone: String): OngoingNewPatientEntry =
       copy(address = addressOrBlank().copy(zone = zone))
 
+  fun withPatientPrefillInfo(patientProfileInfo: PatientPrefillInfo, identifier: Identifier): OngoingNewPatientEntry =
```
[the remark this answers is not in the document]
janhavisinghh: Created a [story](https://app.clubhouse.io/simpledotorg/story/3639/resolve-coupling-between-scanid-and-instantsearch-packages?vc_group_by=day&ct_workflow=all) for this will resolve in another PR.

app/src/main/java/org/simple/clinic/shortcodesearchresult/ShortCodeSearchResultScreen.kt
```diff
@@ -173,7 +173,11 @@ class ShortCodeSearchResultScreen :
   }
 
   override fun openPatientSearch() {
-    val screenKey = InstantSearchScreenKey(additionalIdentifier = null, initialSearchQuery = null)
+    val screenKey = if (features.isEnabled(Feature.InstantSearch)) {
+      InstantSearchScreenKey(additionalIdentifier = null, initialSearchQuery = null, patientPrefillInfo = null)
+    } else {
+      PatientSearchScreenKey(additionalIdentifier = null).wrap()
+    }
```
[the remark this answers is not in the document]
janhavisinghh: Ok, so this is where I messed up the rebase. I am fixing this. This commit only adds patientPrefillInfo as a parameter in the openPatientSearch method

janhavisinghh:
@vinaysshenoy I have resolved all the issues. PTAL

app/src/main/java/org/simple/clinic/scanid/IndiaNHIDDateOfBirthMoshiAdapter.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.squareup.moshi.FromJson
+import com.squareup.moshi.ToJson
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
+import java.util.Locale
+
+class IndiaNHIDDateOfBirthMoshiAdapter {
+  companion object {
+    private val dateTimeFormatter = DateTimeFormatter.ofPattern("d/M/yyyy", Locale.ENGLISH)
+  }
```
vinaysshenoy, line 12: This should not be a static constant, can we make it a local constant please?

app/src/test/java/org/simple/clinic/instantsearch/InstantSearchUpdateTest.kt
```diff
@@ -290,6 +301,77 @@ class InstantSearchUpdateTest {
...
+        identifier = identifier)
+
+    updateSpec
+        .given(searchQueryModel.patientPrefillInfoUpdated(patientPrefillInfo).additionalIdentifierUpdated(identifier))
+        .whenEvent(RegisterNewPatientClicked)
+        .then(assertThatNext(
+            hasNoModel(),
+            hasEffects(SaveNewOngoingPatientEntry(ongoingNewPatientEntry))
+        ))
+  }
```
vinaysshenoy, line 373: I think I may not have been clear earlier. What I mean is that this test is unnecessary since a `LocalDate` itself does not have anything to do with the format once it is parsed.

Since the actual parsing here is handled by the `ScanSimpleIdScreen`, in the event that the DOB in the scanned QR code is a different format than we expect, the parsing will fail there and we will never end up in this case since the `LocalDate` will never get created in the first place.

Hence, this test is unnecessary and we can remove it.

Does this make sense?

app/src/main/java/org/simple/clinic/scanid/IndiaNHIDGender.kt
```diff
@@ -0,0 +1,39 @@
...
+  MALE, FEMALE, UNKNOWN;
+
+  companion object {
+    fun fromIndiaNHIDToGender(value: IndiaNHIDGender): Gender {
+      return when (value) {
+        MALE -> Gender.Male
+        FEMALE -> Gender.Female
+        else -> Gender.Unknown("U")
+      }
+    }
```
vinaysshenoy, line 17: What I meant was to add this factory function to the `Gender` type and call it `fromIndiaNHIDGender`. `IndiaNHIDGender` should not be responsible for converting itself to the `Gender` type, it should be the other way around.

vinaysshenoy (changes requested):
Overall PR looks to be in good shape, left a few comments and then we'll be good to go.

vinaysshenoy (approved):

PR #2647, opened 2021-05-20
Use `LifecycleCameraController` in `ScanSimpleIdScreen`

msasikanth:


app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
```diff
@@ -148,95 +136,41 @@ class ScanSimpleIdScreen : BaseScreen<
...
-   *  @return suitable aspect ratio
-   *
-   *  source: https://github.com/android/camera-samples/blob/bfc46b68ef51f337104d3487c2394b204cf0d453/CameraXBasic/app/src/main/java/com/android/example/cameraxbasic/fragments/CameraFragment.kt#L350
-   */
-  private fun aspectRatio(width: Int, height: Int): Int {
-    val previewRatio = max(width, height).toDouble() / min(width, height)
-    if (abs(previewRatio - RATIO_4_3_VALUE) <= abs(previewRatio - RATIO_16_9_VALUE)) {
-      return AspectRatio.RATIO_4_3
-    }
-    return AspectRatio.RATIO_16_9
```
vinaysshenoy, line 239: Huh, have they incorporated this into the new library now?

msasikanth: Not a new library, it's still in the same artifact. It's just they followed better SoC. One thing I am concerned about is there is no longer a way to set the aspect ratio. That's why I am on the fence about this implementation. 😅 

Requires thorough testing on all of our test devices.

msasikanth:
Closing this, for now, to be safe I will open a different PR with a feature flag to open a separate screen that has this implementation.

PR #2650, opened 2021-05-24
Redact sensitive information from logs

msasikanth:


app/src/main/java/org/simple/clinic/ReleaseClinicApp.kt
```diff
@@ -26,6 +27,8 @@ class ReleaseClinicApp : ClinicApp() {
     // for `ControllerStateBase` where we log those.
     val logger = LoggerFactory.getLogger("com.spotify.mobius.ControllerStateBase") as Logger
     logger.level = Level.OFF
+
+    MobiusHooks.setErrorHandler { /* No-op */ }
```
vinaysshenoy, line 31: What happens to the app in this scenario? Does it continue to run? This might be a bit problematic if the app just stops responding events.

msasikanth: If an error happens in the effect handler app will crash, but if it happens in the Update function, it's silently failing so the app will not respond to those events. 

Is there a way we can crash the app without sending the results back to Sentry in this scenario? 🤔

vinaysshenoy: Ok, that seems a bit problematic to me. For now, can we just override the `toString()` method of the event that contained PHI to stop it from getting serialized in production?

We can think of a better long term solution later, maybe something like https://mustafaali.net/2018/01/14/Kotlin-data-classes-and-sensitive-information/.

msasikanth: Done, for now, I have gone through Sentry issues and redacted the logs that had sensitive info. (only for these errors). Overall this issue seems to be related to some lifecycle-related issues. Will debug further.

app/src/main/java/org/simple/clinic/Globals.kt
```diff
@@ -4,3 +4,5 @@ const val MAX_ALLOWED_PATIENT_AGE: Int = 120
 const val MIN_ALLOWED_PATIENT_AGE: Int = 1
 const val SECURITY_PIN_LENGTH: Int = 4
 const val LOGIN_OTP_LENGTH = 6
+
+val REDACTED_STRING = "U+2588".repeat(5)
```
vinaysshenoy, line 8: We have a global object `org.simple.clinic.util.Unicode` which you can use for defining special unicode sequences like this. Can you move this there please?

app/src/main/java/org/simple/clinic/summary/PatientSummaryEvent.kt
```diff
@@ -9,7 +10,12 @@ import java.util.UUID
 
 sealed class PatientSummaryEvent : UiEvent
 
-data class PatientSummaryProfileLoaded(val patientSummaryProfile: PatientSummaryProfile) : PatientSummaryEvent()
+data class PatientSummaryProfileLoaded(val patientSummaryProfile: PatientSummaryProfile) : PatientSummaryEvent() {
+
+  override fun toString(): String {
+    return "PatientSummaryProfileLoaded(patientSummaryProfile=$REDACTED_STRING)"
+  }
+}
```
vinaysshenoy, line 18: Why don't we redact the `toString()` of `PatientSummaryProfile` and `MedicalOfficer` instead? That way, we don't need to do this at every usage site.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/Globals.kt
msasikanth: Ah, I was not aware of that. Moving it.

app/src/main/java/org/simple/clinic/summary/PatientSummaryEvent.kt
msasikanth: I have no idea why I didn't just do that, that makes a lot of sense 😅

vinaysshenoy (dismissed):

vinaysshenoy (approved):

PR #2656, opened 2021-05-25
Update approach for pre-filling NHID in `PatientEntryScreen`

Honey14:
https://app.clubhouse.io/simpledotorg/story/3621/update-approach-for-pre-filling-nhid-in-patient-entry-screen


app/src/main/java/org/simple/clinic/newentry/PatientEntryScreen.kt
```diff
@@ -477,6 +486,35 @@ class PatientEntryScreen(context: Context, attrs: AttributeSet) : RelativeLayout
...
+    val alternateIdView = PatientEntryAlternateIdViewBinding.inflate(layoutInflater, this, false)
+    alternateIdView.alternateIdentifier.text = identifier
+    alternateIdContainer.addView(alternateIdView.root)
+  }
+
+  private fun setAlternateIdTextField(alternateId: Identifier) {
+    alternativeIdInputEditText.setTextAndCursor(alternateId.displayValue())
+  }
+
   private fun prefillIdentifier(identifier: Identifier) {
```
janhavisinghh, line 518: Don't you think we should add a check here as well to only display identifier if it's a bp passport?

janhavisinghh: We should not be showing nhid or any other id value in bp passport section

Honey14: We will not receive an alternate ID in this field. Hence there is no need to check.

janhavisinghh: yeah, but, it's not that explicit I feel. Here from the code I am not able to make out that this will currently only display the bp passport. For example, even in `EditPatientScreen` we are defining the methods as `displayBPPassport`. I think we should atleast make the method names more explicit.

Honey14: Changing the `prefillIdentifier` function name to `prefillBpPassport`

janhavisinghh (dismissed):
LGTM

janhavisinghh (approved):

PR #2660, opened 2021-05-30
Prefill NHID as `alternateId` in `PatientEntryScreen`

janhavisinghh:


app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
```diff
@@ -119,15 +119,19 @@ data class OngoingNewPatientEntry(
...
+  ): OngoingNewPatientEntry =
       copy(
           personalDetails = PersonalDetails(
               fullName = patientProfileInfo.fullName,
               dateOfBirth = patientProfileInfo.dateOfBirth.format(dateTimeFormatter),
               gender = patientProfileInfo.gender,
               age = null),
           address = addressOrBlank().withColonyOrVillage(patientProfileInfo.address),
-          identifier = identifier)
+          alternativeId = identifier)
```
Honey14, line 134: Can we rename this `alternateId` to keep the naming convention consistent across for alternate id?

app/src/test/java/org/simple/clinic/instantsearch/InstantSearchUpdateTest.kt
```diff
@@ -315,7 +315,7 @@ class InstantSearchUpdateTest {
         healthIdNumber = indiaNationalHealthID
     ).toPatientPrefillInfo()
 
-    val identifier = Identifier(indiaNationalHealthID, IndiaNationalHealthId)
+    val alternativeId = Identifier(indiaNationalHealthID, IndiaNationalHealthId)
```
Honey14, line 318: Same comment as above

Honey14 (changes requested):
Looks good. Left few comments. PTAL

app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
janhavisinghh: `alternativeId` is the name given to the `OngoingNewPatientEntry` parameter. I'll change the places where I'm referring to it as `alternateId`. Sounds good?

Honey14: Actually in the `EditPatientScreen`, `PatientEntryScreen` and `PatientSummaryScreen` we're using `alternateId`. It would be inconsistent if we have `alternativeId` in `OngoingNewPatientEntry` for the additional identifier. 
@msasikanth What do you think about this?

Honey14 (changes requested):
LGTM. Left one comment

app/src/main/java/org/simple/clinic/patient/OngoingNewPatientEntry.kt
janhavisinghh: It's very inconsistent. Even in `PatientEntryScreen` it is being referred as `alternativeId` in some places. i'll make this change.

msasikanth: TBH not that big of a deal. But feel free to make the change if you want. Ideally, it should be named Alternative, since this is a secondary identifier option.

janhavisinghh: For now to maintain consistency renaming it to `alternateId`

janhavisinghh: Altho everywhere both of these names are used. I see a mixture of `alternative` and `alternate` in all the screens

app/src/test/java/org/simple/clinic/instantsearch/InstantSearchUpdateTest.kt
```diff
@@ -324,10 +324,11 @@ class InstantSearchUpdateTest {
             gender = Gender.Male,
             age = null),
         address = OngoingNewPatientEntry.Address.BLANK.withColonyOrVillage(patientPrefillInfo.address),
-        identifier = identifier)
+        alternativeId = alternateId,
```
Honey14: Now we have a mix of these? Maybe change in a different PR later?

Honey14 (approved):
Looks good.

app/src/test/java/org/simple/clinic/instantsearch/InstantSearchUpdateTest.kt
janhavisinghh: I've changed this in this PR in a later commit.

PR #2662, opened 2021-05-31
Highlight patient identifiers in search

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3154/instant-search-ui-changes

### DIFF
--- a/app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
+++ b/app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
@@ -1,12 +1,11 @@
 package org.simple.clinic.widgets
 
 import android.content.Context
-import android.text.Spannable
-import android.text.SpannableStringBuilder
-import android.text.style.BackgroundColorSpan
 import android.util.AttributeSet
 import android.view.LayoutInflater
 import android.view.View
+import androidx.core.text.backgroundColor
+import androidx.core.text.buildSpannedString
 import com.google.android.material.card.MaterialCardView
 import org.simple.clinic.R
 import org.simple.clinic.databinding.ViewPatientSearchResultBinding
@@ -17,6 +16,11 @@ import org.simple.clinic.patient.Gender
 import org.simple.clinic.patient.PatientAddress
 import org.simple.clinic.patient.PatientSearchResult
 import org.simple.clinic.patient.businessid.Identifier
+import org.simple.clinic.patient.businessid.Identifier.IdentifierType.BangladeshNationalId
+import org.simple.clinic.patient.businessid.Identifier.IdentifierType.BpPassport
+import org.simple.clinic.patient.businessid.Identifier.IdentifierType.EthiopiaMedicalRecordNumber
+import org.simple.clinic.patient.businessid.Identifier.IdentifierType.IndiaNationalHealthId
+import org.simple.clinic.patient.businessid.Identifier.IdentifierType.Unknown
 import org.simple.clinic.patient.displayIconRes
 import org.simple.clinic.patient.displayLetterRes
 import org.simple.clinic.router.util.resolveColor
@@ -68,6 +72,15 @@ class PatientSearchResultItemView(
   private val patientNameAgeGenderLabel
     get() = binding.patientNameAgeGenderLabel
 
+  private val identifierContainer
+    get() = binding.identifierContainer
+
+  private val identifierLabelTextView
+    get() = binding.identifierLabelTextView
+
+  private val identifierTextView
+    get() = binding.identifierTextView
+
   @Inject
   @Named("full_date")
   lateinit var dateTimeFormatter: DateTimeFormatter
@@ -93,6 +106,40 @@ class PatientSearchResultItemView(
     renderPatientPhoneNumber(searchQuery, model)
     renderVisited(model.lastSeen)
     renderLastSeen(model.lastSeen, currentFacilityId)
+    renderIdentifier(model.identifier, searchQuery)
+  }
+
+  private fun getId(identifier: Identifier, searchQuery: String): Id {
+    val id = when (identifier.type) {
+      BpPassport -> BpPassport.shortCode(identifier)
+      BangladeshNationalId, EthiopiaMedicalRecordNumber, IndiaNationalHealthId, is Unknown -> identifier.value
+    }
+    val indexOfIdentifier = id.indexOf(searchQuery, ignoreCase = true)
+    return if (indexOfIdentifier >= 0) {
+      Id.Highlighted(
+          value = id,
+          highlightStart = indexOfIdentifier,
+          highlightEnd = indexOfIdentifier + searchQuery.length
+      )
+    } else {
+      Id.Plain(value = id)
+    }
+  }
+
+  private fun renderIdentifier(identifier: Identifier?, searchQuery: String?) {
+    if (identifier == null || searchQuery == null) {
+      identifierContainer.visibility = View.GONE
+    } else {
+      identifierContainer.visibility = View.VISIBLE
+      val id = when (val id = getId(identifier, searchQuery)) {
+        is Id.Highlighted -> highlight(text = id.value,
+            startIndex = id.highlightStart,
+            endIndex = id.highlightEnd)
+        is Id.Plain -> id.value
+      }
+      identifierLabelTextView.text = identifier.displayType(resources)
+      identifierTextView.text = id
+    }
   }
 
   private fun getPatientPhoneNumber(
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -240,6 +240,16 @@ class PatientSearchResultItemView(
...
+  sealed class Id(open val value: String) {
+
+    data class Highlighted(
+        override val value: String,
+        val highlightStart: Int,
+        val highlightEnd: Int
+    ) : Id(value)
+
+    data class Plain(override val value: String) : Id(value)
+  }
```
vinaysshenoy, line 308: Hmm, it seems like we have repeated this similar structure of having a "search criteria" and whether it's highlighted or not in many places.

Maybe a better solution for the future (not necessary to change in this PR) would be to use composition over inheritance.'

```kotlin
sealed class SearchResultItem {
  data class Name(): SearchResultItem()
  data class Phone(): SearchResultItem()
  data class Id(): SearchResultItem()
}

sealed class SearchResultItemType(val result: SearchResultItem) {
  data class Plain(): SearchResultItemType()
  data class Highlighted(): SearchResultItemType()
}
```

That way, whenever we need to implement a new type of search, we will need to implement a highlight by default as well.

What do you think?

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -93,6 +107,48 @@ class PatientSearchResultItemView(
...
+    renderIdentifier(model.identifier, searchQuery)
+  }
+
+  private fun getId(identifier: Identifier, searchQuery: String): Id {
+    val id = when (identifier.type) {
+      BpPassport -> BpPassport.shortCode(identifier)
+      BangladeshNationalId -> identifier.value
+      EthiopiaMedicalRecordNumber -> identifier.value
+      IndiaNationalHealthId -> identifier.value
+      is Unknown -> throw IllegalArgumentException("Unknown identifier: $identifier")
```
vinaysshenoy, line 119: Won't throwing an error here cause a crash in older versions of the app if they don't know a particular type of ID which new app versions are creating?

vinaysshenoy, line 119: It seems like whenever we display a `BP Passport` anywhere in the UI, we always display the short code and not the full value. In this case, do you think it's worth extract this `when` clause into the `Identifier` type as a computed property, say `displayText` or something?

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -93,6 +107,48 @@ class PatientSearchResultItemView(
...
+    }
+    val canHighlightIdentifier = id.contains(searchQuery)
```
vinaysshenoy, line 121: Hmm, how would `contains` work in the scenario where we save characters like hyphens as part of the ID strings, but people may not necessarily add them when searching?

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -93,6 +107,48 @@ class PatientSearchResultItemView(
...
+
+    return if (canHighlightIdentifier) {
```
vinaysshenoy, line 123: Can't we skip the method call for `contains` and use the `indexOf` return value to check if the search query is present in the search result?

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -93,6 +107,48 @@ class PatientSearchResultItemView(
...
+      val indexOfIdentifier = id.indexOf(searchQuery)
```
vinaysshenoy, line 124: We might want to consider making these checks case-insensitive.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -93,6 +107,48 @@ class PatientSearchResultItemView(
...
+  private fun renderIdentifier(identifier: Identifier?, searchQuery: String?) {
+    if (identifier == null || searchQuery == null) {
+      identifierContainer.visibility = View.GONE
+    } else {
+      identifierContainer.visibility = View.VISIBLE
+      val id = when (val id = getId(identifier, searchQuery)) {
+        is Id.Highlighted -> {
+          val highlightId = SpannableStringBuilder(id.value)
+          val highlightColor = context.resolveColor(colorRes = R.color.search_query_highlight)
+          highlightId.setSpan(BackgroundColorSpan(highlightColor), id.highlightStart, id.highlightEnd, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
```
vinaysshenoy, line 144: What do you think of extracting the construction of the highlight span to a separate method for better readability?

```kotlin
val highlightId = highlight(id.value, id.highlightStart, id,highlightEnd, R.color.search_query_highlight)
```

If we move ahead with my earlier suggestion to use composition over inheritance for representing the search result types, we can even make this a method of the `Highlighted` type.

```kotlin
val highlightId = id.createHighlightedSpan(R.color.search_query_highlight)
```

vinaysshenoy (changes requested):
Overall PR looks good. Left some suggestions for improvements. Please take a look?

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
msasikanth: Yep, I was actually exploring the option of creating a specific type and simplifying this process. But that seemed like a different change and should be a different PR. So I will be creating a new change with those changes.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
msasikanth: Good point. Should we just show the actual value of the `Unknown` type? We use that for search as well. So I think that makes sense.

msasikanth: Yep, we already have a `displayValue` that does format the search results. But it adds specific Unicode characters depending on the identifier type. I felt like this display variant is only used here, so that's why I didn't move it to the `Identifier` class. Should I move it?

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
msasikanth: It wouldn't work. Even the search query fails, but when we are creating our identifier we are stripping the special characters (at least for the NHID that has been added recently).

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
vinaysshenoy: Ah, that makes sense. Let's leave this as is for now then.

vinaysshenoy: Yup, that should be the approach here.

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -93,6 +106,43 @@ class PatientSearchResultItemView(
...
     renderLastSeen(model.lastSeen, currentFacilityId)
+    renderIdentifier(model.identifier, searchQuery)
+  }
+
+  private fun getId(identifier: Identifier, searchQuery: String): Id {
+    val id = when (identifier.type) {
+      BpPassport -> BpPassport.shortCode(identifier)
+      BangladeshNationalId -> identifier.value
+      EthiopiaMedicalRecordNumber -> identifier.value
+      IndiaNationalHealthId -> identifier.value
```
vinaysshenoy, line 117: Not necessary to change this, but these clauses could be collapsed into a single one.

```kotlin
BangladeshNationalId, EthiopiaMedicalRecordNumber, IndiaNationalHealthId -> identifier.value
```

msasikanth: Omg, yes. Let me make this change.

vinaysshenoy (dismissed):

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
msasikanth: Changed it to 
```kt
BangladeshNationalId, EthiopiaMedicalRecordNumber, IndiaNationalHealthId, is Unknown -> identifier.value
```

vinaysshenoy (approved):

PR #2667, opened 2021-06-01
Update facility picker search edit text hint

janhavisinghh:
From [this](https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1622539666159700) slack message:


app/src/main/res/layout/view_facilitypicker.xml
```diff
@@ -33,7 +33,7 @@
             android:id="@+id/searchEditText"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
-            android:hint="@string/facilitypicker_search_hint"
+            android:hint="@string/facilitypicker_search_hint_new"
```
Honey14, line 36: We don't need to change the key to this string. Just the value for it. Can you undo this?

app/src/main/res/values-am-rET/strings.xml
```diff
@@ -580,8 +580,6 @@
   <string name="assigned_facility_view_title">የተመደቡበት ተቋም</string>
   <string name="assigned_facility_view_change">ለውጥ</string>
 
-  <!-- Facility Picker -->
-  <string name="facilitypicker_search_hint">ክሊኒክዎን ይፈልጉ</string>
```
Honey14, line 584: We don't need to remove this. Since its going to be the same key. The transifex will update the translations.

Honey14 (changes requested):
Looks good. Left 2 comments. PTAL.
FYI, just for future PR's we should've created a story for this to keep the track of stories that we're working on. Makes it easier for the team to go through.

Honey14 (approved):
Looks good

PR #2668, opened 2021-06-01
Fix user cannot see sync button bug

Honey14:
https://app.clubhouse.io/simpledotorg/story/3293/user-cannot-see-sync-button

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientsTabUi.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientsTabUi.kt
@@ -7,4 +7,5 @@ interface PatientsTabUi {
   fun showIllustration()
   fun showUserStatusAsPendingVerification()
   fun hideUserAccountStatus()
+  fun showUserStatusAsWaitingForApproval()
 }
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientsTabUi.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientsTabUi.kt
@@ -5,4 +5,6 @@ interface PatientsTabUi {
   fun showSyncIndicator()
   fun showSimpleVideo()
   fun showIllustration()
+  fun showUserStatusAsPendingVerification()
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/simple/clinic/home/patients/PatientsTabUi.kt
```diff
@@ -7,4 +7,5 @@ interface PatientsTabUi {
   fun showIllustration()
   fun showUserStatusAsPendingVerification()
   fun hideUserAccountStatus()
+  fun showUserStatusAsWaiting()
```
janhavisinghh, line 10: Maybe we should maintain consistency with `showUserStatusAsPendingVerification` and rename this to `showUserStatusAsWaitingForApproval`

app/src/test/java/org/simple/clinic/home/patients/PatientsLogicTest.kt
```diff
@@ -118,14 +119,13 @@ class PatientsLogicTest {
     uiEvents.onNext(NewPatientClicked)
 
     // then
-    verify(uiActions).hideUserAccountStatus()
+    verify(ui, times(2)).hideUserAccountStatus()
```
janhavisinghh, line 122: Why did we change this to `times(2)`?

janhavisinghh (changes requested):
Left a few comments PTAL.
There's also some code climate issues can you please fix them?
Also, I wanted to ask how did using ui renderer instead of effect handler fix this issue? @Honey14

app/src/test/java/org/simple/clinic/home/patients/PatientsLogicTest.kt
Honey14: If test had the occurrence of `hideUserAccountStatus` twice, then we add times(2) to it. There are 2 scenarios where this could occur 
* Where got it button is clicked on the approval of user
* Where It couldn't satisfy any other condition of account notification

janhavisinghh (approved):

PR #2670, opened 2021-06-02
Implement `QRCodeJsonParser` in `ScanSimpleIdEffectHandler`

Honey14:
https://app.clubhouse.io/simpledotorg/story/3561/implement-qr-code-json-parser-and-extract-direct-moshi-usage-in-scansimpleideffecthandler

### DIFF
--- a/app/src/main/java/org/simple/clinic/scanid/QRCodeJsonParser.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/QRCodeJsonParser.kt
@@ -1,0 +1,14 @@
+package org.simple.clinic.scanid
+
+import com.squareup.moshi.Moshi
+import javax.inject.Inject
+
+class QRCodeJsonParser @Inject constructor(
+    private val moshi: Moshi
+) {
+
+  fun parseQRCodeJson(text: String): IndiaNHIDInfoPayload? {
+    val adapter = moshi.adapter(IndiaNHIDInfoPayload::class.java)
+    return adapter.fromJson(text)
+  }
+}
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/scanid/QRCodeJsonParser.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import com.squareup.moshi.Moshi
+import javax.inject.Inject
+
+class QRCodeJsonParser @Inject constructor(
+    private val moshi: Moshi
+) {
+
+  fun parseQRCodeJson(text: String): IndiaNHIDInfoPayload? {
+    val adapter = moshi.adapter(IndiaNHIDInfoPayload::class.java)
```
vinaysshenoy, line 11: This adapter can be a class property instead of being fetched every time the method is called, right?

vinaysshenoy (changes requested):
Overall PR looks good, left one comment. Please take a look?

vinaysshenoy:
@Honey14 Some tests also seem to be failing/

app/src/main/java/org/simple/clinic/scanid/QRCodeJsonParser.kt
Honey14: Makes sense but If it were to fetch every time what would the consequences be?

msasikanth: It will still continue to work, it's just we end up creating a new adapter instance every time this method is called and not that memory efficient, since we can reuse an already available instance, we don't want to create it multiple times.

vinaysshenoy (approved):

PR #2672, opened 2021-06-03
Add training videos depending on language

Honey14:
https://app.clubhouse.io/simpledotorg/story/3194/change-training-videos-depending-on-country-language

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/Languages.kt
+++ b/app/src/main/java/org/simple/clinic/patient/Languages.kt
@@ -1,9 +1,0 @@
-package org.simple.clinic.patient
-
-import com.squareup.moshi.JsonClass
-
-@JsonClass(generateAdapter = true)
-data class Languages(
-    val languageLocaleCode: String,
-    val languageVideoUrl: String
-)
--- a/app/src/main/java/org/simple/clinic/patient/Languages.kt
+++ b/app/src/main/java/org/simple/clinic/patient/Languages.kt
@@ -1,0 +1,9 @@
+package org.simple.clinic.patient
+
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+data class Languages(
+    val languageLocaleCode: String,
+    val languageVideoUrl: String
+)
--- a/app/src/main/java/org/simple/clinic/patient/SimpleVideoModule.kt
+++ b/app/src/main/java/org/simple/clinic/patient/SimpleVideoModule.kt
@@ -22,6 +22,7 @@ class SimpleVideoModule {
     return when (locale.language) {
       "hi" -> "nHsQ06tiLzw"
       "am" -> "05dnOSyHXYs"
+      "om" -> "yca12Ec5Fg4"
       // Default to English
       else -> "YO3D1paAuqU"
     }
--- a/app/src/main/java/org/simple/clinic/patient/SimpleVideoModule.kt
+++ b/app/src/main/java/org/simple/clinic/patient/SimpleVideoModule.kt
@@ -21,6 +21,7 @@ class SimpleVideoModule {
   fun provideSimpleVideoUrlBasedOnLocale(locale: Locale): String {
     return when (locale.language) {
       "hi" -> "nHsQ06tiLzw"
+      "am" -> "05dnOSyHXYs"
       // Default to English
       else -> "YO3D1paAuqU"
     }
--- a/app/src/main/java/org/simple/clinic/patient/SimpleVideoModule.kt
+++ b/app/src/main/java/org/simple/clinic/patient/SimpleVideoModule.kt
@@ -20,7 +20,7 @@ class SimpleVideoModule {
   }
 
   @Provides
-  @Named("training_video_youtube_id")
+  @SimpleVideo(TrainingVideoYoutubeId)
   fun provideSimpleVideoUrlBasedOnLocale(
       remoteConfigReader: ConfigReader,
       locale: Locale,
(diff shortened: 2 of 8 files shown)


app/src/main/java/org/simple/clinic/patient/SimpleVideoModule.kt
```diff
@@ -21,6 +21,8 @@ class SimpleVideoModule {
   fun provideSimpleVideoUrlBasedOnLocale(locale: Locale): String {
     return when (locale.language) {
       "hi" -> "nHsQ06tiLzw"
+      "am" -> "05dnOSyHXYs"
+      "om" -> "yca12Ec5Fg4"
       // Default to English
       else -> "YO3D1paAuqU"
```
vinaysshenoy, line 27: Can we look into making this configurable via Firebase Remote Config?

app/src/main/java/org/simple/clinic/patient/SimpleVideoModule.kt
```diff
@@ -18,13 +22,24 @@ class SimpleVideoModule {
...
-      else -> "YO3D1paAuqU"
-    }
+  fun provideSimpleVideoUrlBasedOnLocale(
+      uuidSetJsonAdapter: JsonAdapter<Set<Languages>>,
+      remoteConfigService: RemoteConfigService,
+      locale: Locale
+  ): String {
+
+    val simpleVideoUrlsJsonArray: String = remoteConfigService.reader().string("simple_youtube_video_urls", "[]")
+    val simpleLanguageBasedVideoUrls: Set<Languages> = uuidSetJsonAdapter.fromJson(simpleVideoUrlsJsonArray)!!
```
vinaysshenoy, line 33: I see that the current configuration JSON for this is an array of dictionaries:

```json
[
  {
    "languageLocaleCode": "hi",
    "languageVideoUrl": "nHsQ06tiLzw"
  },
  {
    "languageLocaleCode": "am",
    "languageVideoUrl": "05dnOSyHXYs"
  },
  {
    "languageLocaleCode": "om",
    "languageVideoUrl": "yca12Ec5Fg4"
  },
  {
    "languageLocaleCode": "en",
    "languageVideoUrl": "YO3D1paAuqU"
  }
]
```

What do you think of representing this as a single dictionary instead?

```json
{
  "hi": "nHsQ06tiLzw",
  "en": "YO3D1paAuqU"
}
```

This would simplify the code required to read this a lot.

```kotlin
val stringMapAdapter = moshi.adapter<Map<String, String>>(type)
val defaultVideoJson = """
{
  "_default": "YO3D1paAuqU"
}
"""

val youtubeVideoIds: Map<String, String> = stringMapAdapter.parse(remoteConfigReader.string("simple_youtube_video_urls", defaultVideoJson))

return youtubeVideoIds[locale.language] ?: youtubeVideoIds["_default"]
```

vinaysshenoy: This would also mean that we need to add the `_default` as a key to the JSON on the remote config as well and it will handle cases where the current language is not present in the remote config as well as the app not being able to fetch the remote config,

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/patient/SimpleVideoModule.kt
```diff
@@ -2,27 +2,45 @@ package org.simple.clinic.patient
...
-  fun provideSimpleVideoUrlBasedOnLocale(locale: Locale): String {
-    return when (locale.language) {
-      "hi" -> "nHsQ06tiLzw"
-      // Default to English
-      else -> "YO3D1paAuqU"
-    }
+  @SimpleVideo(TrainingVideoYoutubeId)
+  fun provideSimpleVideoUrlBasedOnLocale(
+      uuidSetJsonAdapter: JsonAdapter<Set<Languages>>,
+      remoteConfigService: RemoteConfigService,
```
vinaysshenoy, line 28: We can also inject the remote config reader directly instead of injecting the service if the only thing we need is the reader.

app/src/main/java/org/simple/clinic/patient/SimpleVideoModule.kt
Honey14: Oh this works as well 👍
I'll update it to the list of single dictionary

vinaysshenoy: A nitpick: can we rename the remote config key to reflect that it contains YouTube video IDs and not the URLs?

Honey14: Hey defining a type with an array looked something like  `val type = Types.newParameterizedType(Set::class.java, Languages::class.java)`
With the single dictionary, what would the raw type be? I am not able to add Map<String,String> to it.

vinaysshenoy: It's a vararg parameter, so it should be `Types.newParameterizedType(Map::class.java, String::class.java, String::class.java)`.

app/src/main/java/org/simple/clinic/patient/Languages.kt
```diff
@@ -0,0 +1,9 @@
+package org.simple.clinic.patient
+
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+data class Languages(
```
vinaysshenoy, line 6: This class is no longer required now, right?

vinaysshenoy (changes requested):
Overall PR looks good to go. There's a class that's no longer required, we can merge this in once it's removed.

app/src/main/java/org/simple/clinic/patient/Languages.kt
Honey14: Oh right, I forgot to delete this.

vinaysshenoy (approved):

PR #2692, opened 2021-06-07
Fix progress view for manual OTP entry screen is not hidden when login fails

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/1012/progress-view-for-manual-otp-entry-screen-is-not-hidden-when-login-fails

### DIFF
--- a/app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
@@ -44,7 +44,7 @@ class EnterOtpUpdate(
     val updatedModel = model.loginFinished()
     return when (val result = event.result) {
       LoginResult.Success -> next(updatedModel, ClearLoginEntry, TriggerSync)
-      else -> next(model.loginFailed(AsyncOpError.from(result)), ClearPin as EnterOtpEffect)
+      else -> next(updatedModel.loginFailed(AsyncOpError.from(result)), ClearPin as EnterOtpEffect)
     }
   }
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
```diff
@@ -44,7 +44,7 @@ class EnterOtpUpdate(
     val updatedModel = model.loginFinished()
     return when (val result = event.result) {
       LoginResult.Success -> next(updatedModel, ClearLoginEntry, TriggerSync)
-      else -> next(model.loginFailed(AsyncOpError.from(result)), ClearPin as EnterOtpEffect)
```
vinaysshenoy, line 47: Can we add a failing test first that reproduces the bug first?

vinaysshenoy (changes requested):
Fix looks good, just need to add a test that reproduces the bug before we make the change and we are good to go.

vinaysshenoy (approved):

vinaysshenoy:
@janhavisinghh The tests are failing. Please take a look?

app/src/test/java/org/simple/clinic/enterotp/EnterOtpUpdateTest.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.spotify.mobius.test.UpdateSpec
+import com.spotify.mobius.test.UpdateSpec.assertThatNext
+import org.junit.Test
+import org.simple.clinic.TestData
+import org.simple.clinic.login.LoginResult.NetworkError
+import java.util.UUID
+
+class EnterOtpUpdateTest {
+  @Test
+  fun `when the login request is completed and is unsuccessful, then finish logging in and show error`() {
```
vinaysshenoy, line 14: For what it's worth, now that we have this test here, we should delete the tests from the regression test suite that were verifying this behaviour.

We can do this in another PR, I'm going to approve this for now.

vinaysshenoy (dismissed):

vinaysshenoy (approved):

PR #2703, opened 2021-06-10
Transform generated Room DAOs instead of using reflection to instrument performance

vinaysshenoy:


vinaysshenoy:
Adding @msasikanth as an additional reviewer on this since he has more context.

msasikanth (approved):

PR #2706, opened 2021-06-14
Paginate search results in `InstantSearchScreen`

msasikanth:
- Add query to load search by name results as paging source
- Add query to load search by number results as paging source
- Load search results as paging source
- Show patient search results using `PagingItemAdapter`
- When search results are loaded, then show search results
- Add load state listener to search results adapter
- Remove automatic scrolling to top when setting/updating adapter data
- Remove the default adapter set in `onViewCreated`
- Show and hide progress bar when the refresh load state is loading
- Remove `InstantSearchProgressState` from `InstantSearchModel`
- Remove effects to hide and show search no search results view
- Remove `InstantSearchResultsItemType_old`
- Remove `PatientRepository#search`
- Remove search by patient identifier feature flag and make it default
- Remove `PhoneNumberCriteria`
- Update CHANGELOG


app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -68,42 +57,12 @@ class PatientRepository @Inject constructor(
...
     return database
         .patientSearchDao()
         .searchByNamePagingSource(patientName, facilityId)
   }
 
-  private fun searchByPhoneNumber(
-      phoneNumber: String,
-      facilityId: UUID
-  ): List<PatientSearchResult> {
-    return reportTimeTaken(
```
vinaysshenoy, line 76: Can the `reportTimeTaken` method be removed now?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
msasikanth: Are we not using that anywhere else apart from `PatinetRepository`? If not, then yeah. We can remove the `reportTimeTaken` method from the project. Will open a different PR for it.

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -3135,4 +3135,82 @@ class PatientRepositoryAndroidTest {
...
+    val searchResults = searchResults(query = "Ramesh", facility = currentFacility)
+
+    //then
+    assertThat(searchResults)
+        .containsExactly(
+            "Ramesh Rao",
+            "Mohan Ramesh",
+            "Sai Ramesh"
+        )
+        .inOrder()
```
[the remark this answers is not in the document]
msasikanth: Yep, the last name in `containsExactly` is from other facility, and above two are from current facility.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
vinaysshenoy: Yeah, I see in your branch that it is unused. We can remove this and the corresponding method in `Analytics` as well.

vinaysshenoy (approved):

PR #2711, opened 2021-06-15
Migrate `OverdueAppointment` to add patients without phone number

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/3735/migrate-overdueappointment-to-add-patients-without-phone-number?vc_group_by=day&ct_workflow=all

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
@@ -372,7 +372,7 @@ class AppointmentRepositoryAndroidTest {
         .blockingAwait()
 
     // when
-    val overdueAppointments = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility(since = today,
+    val overdueAppointments = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility_old(since = today,
         facilityId = facility.uuid),
         loadSize = 10)
         .data
@@ -556,7 +556,7 @@ class AppointmentRepositoryAndroidTest {
         .blockingAwait()
 
     // when
-    val overdueAppointments = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility(since = today,
+    val overdueAppointments = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility_old(since = today,
         facilityId = facility.uuid),
         loadSize = 10)
         .data
@@ -922,7 +922,7 @@ class AppointmentRepositoryAndroidTest {
     )
 
     // when
-    val appointments = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility(since = LocalDate.now(clock),
+    val appointments = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility_old(since = LocalDate.now(clock),
         facilityId = facility.uuid),
         loadSize = 15)
         .data
@@ -990,7 +990,7 @@ class AppointmentRepositoryAndroidTest {
     createOverdueAppointment(patientWithOverAnYearDaysOverdue, now.minusDays(370), facilityUuid)
 
     //when
-    val overduePatientUuids = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility(since = now,
+    val overduePatientUuids = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility_old(since = now,
         facilityId = facilityUuid),
         loadSize = 10)
         .data
@@ -1182,7 +1182,7 @@ class AppointmentRepositoryAndroidTest {
     scheduleAppointment(appointmentUuidForSecondPatient, secondPatient).blockingGet()
 
     // when
-    val bloodPressuresByAppointmentUuid = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility(since = LocalDate.now(clock),
+    val bloodPressuresByAppointmentUuid = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility_old(since = LocalDate.now(clock),
         facilityId = facility.uuid),
         loadSize = 10)
         .data
@@ -1262,7 +1262,7 @@ class AppointmentRepositoryAndroidTest {
     scheduleAppointment(appointmentUuidForSecondPatient, secondPatient).blockingGet()
 
     // when
-    val bloodSugarByAppointmentUuid = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility(since = LocalDate.now(clock),
+    val bloodSugarByAppointmentUuid = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility_old(since = LocalDate.now(clock),
         facilityId = facility.uuid),
         loadSize = 10)
         .data
@@ -1339,7 +1339,7 @@ class AppointmentRepositoryAndroidTest {
     scheduleAppointment(appointmentUuidForFirstPatient, patient).blockingGet()
 
     // when
-    val bloodSugarByAppointmentUuid = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility(since = LocalDate.now(clock),
+    val bloodSugarByAppointmentUuid = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility_old(since = LocalDate.now(clock),
         facilityId = facility.uuid),
         loadSize = 10)
         .data
@@ -1415,7 +1415,7 @@ class AppointmentRepositoryAndroidTest {
     scheduleAppointment(appointmentUuidForFirstPatient, patient).blockingGet()
 
     // when
-    val bloodSugarByAppointmentUuid = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility(since = LocalDate.now(clock),
+    val bloodSugarByAppointmentUuid = PagingTestCase(pagingSource = appointmentRepository.overdueAppointmentsInFacility_old(since = LocalDate.now(clock),
         facilityId = facility.uuid),
         loadSize = 10)
         .data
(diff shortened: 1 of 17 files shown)


vinaysshenoy:
If we are going to show patients without phone numbers in the overdue list, wouldn't we also have to update the UI to indicate to nurses that this patient does not have a phone number and cannot be contacted?

msasikanth:
> 
> 
> If we are going to show patients without phone numbers in the overdue list, wouldn't we also have to update the UI to indicate to nurses that this patient does not have a phone number and cannot be contacted?

Yep, currently, we just hide the call icon when the phone number is `null`. But a good approach might be to create a different query that just loads overdue patients with phone number so that we can place this under a feature flag. wdyt?

vinaysshenoy:
> > If we are going to show patients without phone numbers in the overdue list, wouldn't we also have to update the UI to indicate to nurses that this patient does not have a phone number and cannot be contacted?
> 
> Yep, currently, we just hide the call icon when the phone number is `null`. But a good approach might be to create a different query that just loads overdue patients with phone number so that we can place this under a feature flag. wdyt?

Feature flagging this is a good idea. I just wanted to ask, however, if the change to hide the call button for patients without phone numbers has already been implemented or if it's being tracked somewhere. Because that will need to be in the same release where this change goes out.

msasikanth:
> > > If we are going to show patients without phone numbers in the overdue list, wouldn't we also have to update the UI to indicate to nurses that this patient does not have a phone number and cannot be contacted?
> > 
> > 
> > Yep, currently, we just hide the call icon when the phone number is `null`. But a good approach might be to create a different query that just loads overdue patients with phone number so that we can place this under a feature flag. wdyt?
> 
> Feature flagging this is a good idea. I just wanted to ask, however, if the change to hide the call button for patients without phone numbers has already been implemented or if it's being tracked somewhere. Because that will need to be in the same release where this change goes out.

There is a different story that tracks the overdue list UI changes. But currently, we are just hiding the call button, I think that has been the default behavior that was implemented before. So just having a feature flag for which query to load should be enough I think.

We can add the current default query to load patients with phone number present and add a new query to load even without phone numbers. That way we can feature flag it in the effect handler and once the UI changes are merged in, we can turn it on. wdyt?

janhavisinghh:
@msasikanth can we even feature flag a db migration?

msasikanth:
> 
> 
> @msasikanth can we even feature flag a db migration?

It's not the migration that we will be feature flagging. We will feature flag what query we will be using to load, so essentially we will have 2 queries. One to load patients who have the phone number, and the other one is to load patients with and without phone numbers.

Let's discuss about this tomorrow.

janhavisinghh:
@msasikanth this is ready for review. PTAL?

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
```diff
@@ -137,6 +138,21 @@ data class OverdueAppointment(
...
+    @Query("""
+      SELECT * FROM OverdueAppointment
+      WHERE 
+        IFNULL(patientAssignedFacilityUuid, appt_facilityUuid) = :facilityUuid 
+        AND (appt_scheduledDate < :scheduledBefore AND appt_scheduledDate > :scheduledAfter)
+        AND (appt_remindOn < :scheduledBefore OR appt_remindOn IS NULL)
+        GROUP BY appt_patientUuid
+        ORDER BY isAtHighRisk DESC, appt_scheduledDate DESC, appt_updatedAt ASC
+    """)
+    fun overdueInFacilityPagingSource2(
```
msasikanth, line 150: Can we rename the old DAO method that we gonna remove to `overdueInFacilityPagingSource_old` and this to `overdueInFacilityPagingSource`, please? 

We have started avoiding using numbers to indicate 2 different variants.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
```diff
@@ -77,8 +77,8 @@ import java.util.UUID
             P.deletedAt IS NULL
             AND P.status != 'dead'
             AND A.deletedAt IS NULL
+            AND PPN.deletedAt IS NULL
```
msasikanth, line 80: Are we sure we want to ignore patients whose phone number is deleted? The phone number might be deleted, but the patient might be still active. Can we verify this case once, please?

app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration89AndroidTest.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.simple.clinic.assertViewExists
+
+class Migration89AndroidTest : BaseDatabaseMigrationTest(88, 89) {
+
+  @Test
+  fun the_overdue_patient_list_should_be_added() {
+    before.assertViewExists("OverdueAppointment")
+    after.assertViewExists("OverdueAppointment")
+  }
+}
```
msasikanth, line 13: Please add a new line here

app/src/main/java/org/simple/clinic/storage/migrations/Migration_89.kt
```diff
@@ -0,0 +1,78 @@
...
+            P.deletedAt IS NULL
+            AND P.status != 'dead'
+            AND A.deletedAt IS NULL
+            AND PPN.deletedAt IS NULL
+            AND A.status = 'scheduled'
+            AND (BP.recordedAt IS NOT NULL OR BloodSugar.recordedAt IS NOT NULL)
+    """)
+    }
+  }
+}
```
msasikanth, line 78: Please add a new line here

app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration89AndroidTest.kt
```diff
@@ -0,0 +1,13 @@
+package org.simple.clinic.storage.migrations
+
+import org.junit.Test
+import org.simple.clinic.assertViewExists
+
+class Migration89AndroidTest : BaseDatabaseMigrationTest(88, 89) {
+
+  @Test
+  fun the_overdue_patient_list_should_be_added() {
```
msasikanth, line 9: Can we write another test verifying we receive the appointments with out phone number after the migration here?

app/src/main/java/org/simple/clinic/home/overdue/OverdueEffectHandler.kt
```diff
@@ -71,4 +72,25 @@ class OverdueEffectHandler @AssistedInject constructor(
...
+                sourceFactory = {
+                  appointmentRepository.overdueAppointmentsInFacility(
+                      since = overdueSince,
+                      facilityId = facility.uuid
+                  )
+                },
+                pageSize = overdueAppointmentsConfig.overdueAppointmentsLoadSize
+            )
+          }
+          .map { pagingData ->
```
msasikanth: We can use method reference here `.map(::OverdueAppointmentsLoaded)`

app/src/test/java/org/simple/clinic/home/overdue/OverdueUpdateTest.kt
```diff
@@ -54,4 +55,43 @@ class OverdueUpdateTest {
             hasEffects(ShowOverdueAppointments(overdueAppointments, isDiabetesManagementEnabled = true))
         ))
   }
+
+  @Test
+  fun `when current facility is loaded and overdue list changes feature is disabled, then load overdue appointments without patients with no phone numbers`() {
```
msasikanth, line 60: There seems to be a typo in the test name, we are loading patients **with phone number** when the overdue list changes feature is disabled right?

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -2824,7 +2904,7 @@ class AppointmentRepositoryAndroidTest {
             dateOfBirth = patientProfile.patient.dateOfBirth,
             age = patientProfile.patient.age,
             appointment = appointment,
-            phoneNumber = patientProfile.phoneNumbers.first(),
+            phoneNumber = if (patientProfile.phoneNumbers.isEmpty()) null else patientProfile.phoneNumbers.first(),
```
msasikanth, line 2907: Kotlin stdlib has a extension we can use called `firstOrNull` that does the same thing. Let's use that.

msasikanth (changes requested):
Overall approach LGTM. Left few comments, PTAL. Also, the code climate check seems to be failing with 1 issue.

app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration89AndroidTest.kt
msasikanth: @vinaysshenoy It always feels a bit weird writing a test for Database view, we are essentially just writing a test to verify a view is present before and present after migration. While a test that verifies the behaviour after the migration is present in repository tests, should we also include those tests in migration? 🤔

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
janhavisinghh: Yeah. The phone number can not be deleted once added unless all the app data is cleared.

app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration89AndroidTest.kt
janhavisinghh: I think we don't write another test here as we already wrote a test for this data base view change in `PatientRepositoryAndroidTest` which works fine. And if our data base view changes are different from the migration the app will crash anyways. Therefore it might not be useful to test it again here.

app/src/test/java/org/simple/clinic/home/overdue/OverdueUpdateTest.kt
janhavisinghh: Yeah, which is why it says - 'without patients with no phone numbers'

msasikanth: Alright, I think I lost my mind. Clearly, I did not see that on Friday 😂

app/src/androidTest/java/org/simple/clinic/storage/migrations/Migration89AndroidTest.kt
msasikanth: Yeah, but if we look at this migration test it doesn't actually help us determine what we are testing right? Basically in this test, we are testing the database view to exist before and after a migration. So it doesn't provide enough information on what exactly are we testing. At least that's what I feel like. But, yeah we don't have to hold back this PR for that. We can think about it later.

msasikanth (approved):

PR #2713, opened 2021-06-16
Adding analytics name to InvalidQrCode

Honey14:
https://app.clubhouse.io/simpledotorg/story/3813/adding-invalid-code-event

### DIFF
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEvent.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEvent.kt
@@ -43,4 +43,6 @@ data class ScannedQRCodeJsonParsed(
     val healthIdNumber: String?
 ) : ScanSimpleIdEvent()
 
-object InvalidQrCode : ScanSimpleIdEvent()
+object InvalidQrCode : ScanSimpleIdEvent() {
+  override val analyticsName = "Scan Simple Card:Invalid QR Code scanned"
+}


msasikanth:
I am not sure whether this would work 🤔 . As far as I remember we only report analytics names for events if they are UI events. Correct me if I am wrong @vinaysshenoy

vinaysshenoy:
> I am not sure whether this would work thinking . As far as I remember we only report analytics names for events if they are UI events. Correct me if I am wrong @vinaysshenoy

That is correct. However, `ScanSimpleIdEvent` itself implements `UiEvent`, so it should work. We can test this out by turning on logging for analytics events in a debug build and scanning an invalid QR code.

msasikanth:
> 
> 
> > I am not sure whether this would work thinking . As far as I remember we only report analytics names for events if they are UI events. Correct me if I am wrong @vinaysshenoy
> 
> That is correct. However, `ScanSimpleIdEvent` itself implements `UiEvent`, so it should work. We can test this out by turning on logging for analytics events in a debug build and scanning an invalid QR code.

Oh no, what I meant was we are currently reporting analytics for events that are triggered from UI where we are using `compose(ReportAnalyticsEvents())` to report to Mixpanel right? Similarly, are do we a mechanism in place to report events that are pushed from the effect handler? 🤔

vinaysshenoy:
> > > I am not sure whether this would work thinking . As far as I remember we only report analytics names for events if they are UI events. Correct me if I am wrong @vinaysshenoy
> > 
> > 
> > That is correct. However, `ScanSimpleIdEvent` itself implements `UiEvent`, so it should work. We can test this out by turning on logging for analytics events in a debug build and scanning an invalid QR code.
> 
> Oh no, what I meant was we are currently reporting analytics for events that are triggered from UI where we are using `compose(ReportAnalyticsEvents())` to report to Mixpanel right? Similarly, do we a mechanism in place to report events that are pushed from the effect handler? thinking

Ah, I see now. This is a good point, hadn't thought about it. Let's merge this in for now and then think on the mechanism we use to forward those events as well. Ideally, I'd like one way to push events to Analytics instead of having multiple, so maybe can we move this to middleware on the Mobius loop itself, which should work for all cases.

vinaysshenoy (approved):

PR #2718, opened 2021-06-18
Extract `CrashReporter` as a facade

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/3846/extract-crashreporter-out-as-a-facade


simple-platform/src/main/java/org/simple/clinic/platform/crash/CrashReporter.kt
```diff
@@ -1,16 +1,40 @@
...
+    }
   }
 
   override fun report(e: Throwable) {
+    sinks.forEach { sink ->
+      runSafely { sink.report(e) }
+    }
+  }
+
+  fun addSink(sink: Sink) {
```
janhavisinghh, line 21: What are sinks and breadcrumbs here? Can you please point me to some article I can read to learn more about this? @vinaysshenoy

vinaysshenoy: A `Sink` is just a term for something which we can send data to (crash events, in this scenario).

A breadcrumb is additional information we can add to a crash, so that when a crash occurs, we can see this additional information as part of the crash report.

You can take a look at this (https://sentry.io/organizations/resolve-to-save-lives/issues/2358857128/?environment=production&project=1212614&query=is%3Aunresolved) crash report to see a list of breadcrumbs that are associated with it.

janhavisinghh (approved):
LGTM

PR #2724, opened 2021-06-21
Fix `StringIndexOutOfBoundsException` while trying to display NHID number

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/3870/stringindexoutofboundsexception-while-trying-to-display-nhid-number

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Migrate `kotlinx.android.parcel` usage to `kotlinx.parcelize`
 - Change `CrashReporter` to be a facade instead of a discrete interface
 - Show download and share buttons in overdue list
+- Show invalid json error when `NHID` number is empty or less than 14 digits   
 
 ### Changes
 - Updated translations: `pa-IN`, `hi-IN`, `te-IN`, `kn-IN`, `mr-IN`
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
```diff
@@ -67,7 +67,13 @@ class ScanSimpleIdEffectHandler @AssistedInject constructor(
     return when (country.isoCountryCode) {
       Country.INDIA -> {
         val payload = qrCodeJsonParser.parseQRCodeJson(effect.text)
-        ScannedQRCodeJsonParsed(payload?.toPatientPrefillInfo(), payload?.healthIdNumber)
+        val healthIdNumber = payload?.healthIdNumber
+
+        if (!healthIdNumber.isNullOrBlank() && healthIdNumber.length < 14) {
```
msasikanth, line 72: These kinds of checks should be part of our business logic, so we should ideally have them in `Update` and then decide what to do.

msasikanth: Here we are referencing a magic number `14`. Ideally, we should have a variable that specifies this number so that we don't lose the context of why it should be less than 14 characters.

msasikanth: Also, what should happen when the health ID number has more than 14 characters?

msasikanth (changes requested):
There are small changes need to the approach, PTAL

msasikanth (approved):

PR #2734, opened 2021-06-22
Paginate recent patients list

msasikanth:
During this process we had to remove some unused code and add new Mobius tests. So I went ahead and removed the old logic tests code and added appropriate Mobius tests, since this is a straight forwad flow.


msasikanth:
waiting on #2725

app/src/androidTest/java/org/simple/clinic/patient/PatientRepositoryAndroidTest.kt
```diff
@@ -57,7 +57,6 @@ import org.simple.clinic.rules.LocalAuthenticationRule
 import org.simple.clinic.storage.text.TextStore
```
vinaysshenoy, line 57: There's an error in the commit message. "there" is being used when we should be using "their". Please correct it?

app/src/test/java/org/simple/clinic/recentpatient/AllRecentPatientsEffectHandlerTest.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.simple.clinic.TestData
+import org.simple.clinic.mobius.EffectHandlerTestCase
+import org.simple.clinic.patient.PatientRepository
+import org.simple.clinic.util.scheduler.TestSchedulersProvider
+import java.util.UUID
+
+class AllRecentPatientsEffectHandlerTest {
+
+  private val patientRepository = mock<PatientRepository>()
+  private val currentFacility = { TestData.facility() }
```
vinaysshenoy, line 27: Can we pin the facility ID we are using here?

vinaysshenoy: Unrelated, what do you think about changing the default values in `TestData` factory methods to not be random anymore? This would mean breaking some of the existing old tests, but new tests would automatically get deterministic data and consumers would need to opt into the random data usage explicitly when writing the tests.

app/src/main/java/org/simple/clinic/recentpatient/AllRecentPatientsEffectHandler.kt
```diff
@@ -40,10 +42,12 @@ class AllRecentPatientsEffectHandler @AssistedInject constructor(
       effects
           .observeOn(schedulersProvider.io())
           .switchMap {
-            patientRepository
-                .recentPatients_old(currentFacility.get().uuid)
-                .map(::RecentPatientsLoaded)
+            pagerFactory.createPager(
+                sourceFactory = { patientRepository.recentPatients(facilityUuid = currentFacility.get().uuid) },
```
vinaysshenoy, line 46: Curious, but what thread does the lambda passed as the `sourceFactory` argument run on? `currentFacility.get()` will perform a database call, so I just want to be sure that we are not falling into a main thread IO crash here.

app/src/main/java/org/simple/clinic/recentpatient/AllRecentPatientsEffectHandler.kt
```diff
@@ -40,10 +42,12 @@ class AllRecentPatientsEffectHandler @AssistedInject constructor(
...
+                pageSize = 20
```
vinaysshenoy, line 47: Hmm, are we planning to hardcode these page sizes everywhere? AFAIK, we had extracted those out to be injected via Dagger.

vinaysshenoy: Another issue is that we are duplicating these magic numbers across both the tests and the production code. Maybe once we are done migrating fully to paging 3, we can refactor this out so it's defined in a single place.

vinaysshenoy (changes requested):
Overall PR looks good. Left a couple of comments.

app/src/main/java/org/simple/clinic/recentpatient/AllRecentPatientsEffectHandler.kt
msasikanth: So everything that is run inside the `switchMap` should run on whatever thread we set in `observeOn` right? Since it's downstream. At least from my testing, I didn't receive any IO usage on main thread crashes. But I can move it outside the factory as a variable just to be safe.

app/src/test/java/org/simple/clinic/recentpatient/AllRecentPatientsEffectHandlerTest.kt
msasikanth: Yeah, that would be good. I think missed adding it here, usually, I add at least the IDs to avoid any issues. Will do the `TestData` changes in a different PR.

app/src/main/java/org/simple/clinic/recentpatient/AllRecentPatientsEffectHandler.kt
msasikanth: Usually, I am using firebase remote config for the page sizes. But the all recent patient screen seemed like it doesn't require that, so I hard-coded them. But we can extract out the page size to a config as we do with overdue and instant search. wdyt?

msasikanth: >Another issue is that we are duplicating these magic numbers across both the tests and the production code. Maybe once we are done migrating fully to paging 3, we can refactor this out so it's defined in a single place.

Usually, for effect handler tests I reference the page size in the config we define. But for instrumentation tests that might be hard, since the page size is defined per feature. So, for example all recent patients can have a different page size compared to recent patients in home, so that page size varies in effect handler. But when testing repo the query is still same right? So we can probably extract out the page size constants in tests. wdyt?

msasikanth: Another option we can use is, defining all the page sizes in `PagingModule` like so.
```kt
@Provides
@PagingSize(AllRecentPatients)
fun providesAllRecentPatientsPageSize(): Int {
  return 25
}
```

msasikanth: Implemented the above approach, this seems like a good plan for keeping all the paging sizes in a single file.

app/src/main/java/org/simple/clinic/recentpatient/AllRecentPatientsEffectHandler.kt
vinaysshenoy: Not really necessary, just wanted to confirm that the `Pager` class does not attempt to change the thread or something when invoking this.

app/src/main/java/org/simple/clinic/recentpatient/AllRecentPatientsEffectHandler.kt
vinaysshenoy: Yeah, this could work. Not really necessary to make this change in this PR, for what its worth. We could do this after migrating everything.

app/src/main/java/org/simple/clinic/recentpatient/AllRecentPatientsEffectHandler.kt
msasikanth: I made that change just to be safe anyway 😂

vinaysshenoy (approved):

PR #2745, opened 2021-06-23
Use macos version 11.0 to run instrumentation tests on CI

msasikanth:
- Use macos version 11.0 to run instrumentation tests
- Update CHANGELOG


janhavisinghh (approved):
LGTM

PR #2746, opened 2021-06-24
Show progress when loading overdue patient contact information

msasikanth:


app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUpdate.kt
```diff
@@ -46,6 +46,17 @@ class ContactPatientUpdate(
     }
   }
 
+  private fun overdueAppointmentLoaded(model: ContactPatientModel, event: OverdueAppointmentLoaded): Next<ContactPatientModel, ContactPatientEffect> {
+    val updatedModel = if (!model.hasLoadedPatientProfile) {
+      model.overdueAppointmentLoaded(event.overdueAppointment)
+    } else {
+      model.overdueAppointmentLoaded(event.overdueAppointment)
+          .contactPatientInfoLoaded()
```
Honey14, line 54: The commit message does not match with the update that we're doing in this commit. We're updating the model here and not the UI. Can you change the commit message to reflect that?

app/src/main/res/layout/sheet_contact_patient.xml
```diff
@@ -6,13 +6,24 @@
...
+
   <org.simple.clinic.contactpatient.views.CallPatientView
     android:id="@+id/callPatientView"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
     android:layout_gravity="bottom"
-    android:visibility="gone"
-    tools:visibility="visible" />
+    android:visibility="invisible"
+    tools:visibility="gone" />
```
Honey14, line 26: Can we update the visibility to show visible in tools because the whole point to adding tools visibility is for the preview.

Honey14 (dismissed):
Looks good overall. Left few comments PTAL

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientInfoProgressState.kt
```diff
@@ -0,0 +1,6 @@
+package org.simple.clinic.contactpatient
+
+enum class ContactPatientInfoProgressState {
+  IN_PROGRESS,
+  DONE
```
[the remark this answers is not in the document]
msasikanth: Here we are waiting for a combination of progress states. Like patient profile info and latest appointment, that makes the contact patient info, that's why it's named as `ContactPatientInfoProgressState`. So for our progress states have been mostly action driven, like on save or done button clicks. In this scenario, we are waiting for the initial data to load to hide the progress state.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUpdate.kt
msasikanth: That's intentional, I was keeping it consistent with the existing testing naming. Also when we update the model we update the UI anyway. For example, when the progress state gets changed we trigger the UI updates to show/hide the progress indicator here.

app/src/main/res/layout/sheet_contact_patient.xml
msasikanth: Oh, I meant to remove that entirely. We are showing a different view in the `tools` here. We are showing progress indicator in tools now. I must have forgotten to remove this bit. Will push this change.

msasikanth: Done, pushed the changes ✅

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUpdate.kt
Honey14: I know it does happen but that's not happening in that particular commit. I understand that only when I review the commits after that. So it wasn't very clear to me.

Honey14 (approved):
LGTM

PR #2751, opened 2021-06-24
Pin Android emulator build version to resolve emulator timeout error

msasikanth:
- Pin Android emulator build version to resolve emulator timeout error
- Update CHANGELOG


vinaysshenoy (approved):
Taking over review responsibility for this PR because it's urgent.

PR #2753, opened 2021-06-25
Increase tappable area of overdue call button

msasikanth:
- Add extension to increase tappable area of a view
- Increase tappable area of the overdue call button
- Update CHANGELOG

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentListItem.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentListItem.kt
@@ -2,6 +2,8 @@ package org.simple.clinic.home.overdue
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.graphics.Rect
+import android.view.TouchDelegate
 import android.view.View
 import androidx.paging.PagingData
 import androidx.paging.map
@@ -18,6 +20,8 @@ import org.simple.clinic.util.UserClock
 import org.simple.clinic.util.toLocalDateAtZone
 import org.simple.clinic.widgets.PagingItemAdapter
 import org.simple.clinic.widgets.UiEvent
+import org.simple.clinic.widgets.dp
+import org.simple.clinic.widgets.executeOnNextMeasure
 import org.simple.clinic.widgets.recyclerview.BindingViewHolder
 import org.simple.clinic.widgets.setCompoundDrawableStart
 import org.simple.clinic.widgets.visibleOrGone
@@ -225,6 +229,7 @@ sealed class OverdueAppointmentListItem : PagingItemAdapter.Item<UiEvent> {
         R.drawable.ic_overdue_call
       }
       binding.callButton.setImageResource(callButtonDrawable)
+      increaseCallButtonTapArea(callButton = binding.callButton)
 
       binding.isAtHighRiskTextView.visibility = if (isAtHighRisk) View.VISIBLE else View.GONE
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/widgets/Views.kt
```diff
@@ -280,6 +281,24 @@ fun ViewPropertyAnimator.setDuration(duration: Duration): ViewPropertyAnimator {
   return setDuration(duration.toMillis())
 }
 
+fun View.increaseTappableArea(verticalSpace: Int, horizontalSpace: Int) {
+  val parent = parent as View
+
+  parent.post {
```
vinaysshenoy, line 287: This assumes that the `View` has already been laid out and bounds are available when invoking this method. Should we consider using the `executeOnMeasure` extension instead of the `post` method?

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentListItem.kt
```diff
@@ -225,6 +227,7 @@ sealed class OverdueAppointmentListItem : PagingItemAdapter.Item<UiEvent> {
         R.drawable.ic_overdue_call
       }
       binding.callButton.setImageResource(callButtonDrawable)
+      binding.callButton.increaseTappableArea(verticalSpace = 40.dp, horizontalSpace = 24.dp)
```
vinaysshenoy, line 230: Should the value to increase by come from the resources and different depending on the screen size? As far as I know, large screens don't require the touch delegate, right?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/widgets/Views.kt
msasikanth: Umm, usually by the time `post` is called the view is laid out right? So it should be fine.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentListItem.kt
msasikanth: Atleast in this scenario we won't be changing this value based on screen size. Basically we want the call button to be clickable for the entire height of the card. According to the designs 40dp up and down below the button should cover it. So essentially we are making the entire right side of the card tap to show the contact sheet.

But we can move these dimensions to resources and refer them from there if you think that would be a better approach.

msasikanth: Instead of hardcoding or referencing the 40dp another option we can consider is getting the view height and doing the math to get top and bottom area. That way the space will fit the card height no matter what screen density it is. But I am guessing in most scenarios this should be more than enough.

app/src/main/java/org/simple/clinic/widgets/Views.kt
vinaysshenoy: Have we tested this change?

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentListItem.kt
vinaysshenoy: Understood. Then my only concern would be handling this correctly if the design of the card changes to become either smaller or larger.

app/src/main/java/org/simple/clinic/widgets/Views.kt
msasikanth: Yep. It works ✅, you can try it out but tapping at bottom of the right bottom and it should show the ripple for card button.

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentListItem.kt
msasikanth: Hmm, I guess than we have to do the math to fit the height of the card. Alright, removing the extension for this and adding this specifically to the `OverdueAppointmentListItem`.

vinaysshenoy (approved):

PR #2755, opened 2021-06-28
Revert "Pin Android emulator build version to resolve emulator timeout error (#2751)"

msasikanth:


CHANGELOG.md
```diff
@@ -14,7 +14,7 @@
 - [Reverted] Use macos version 11.0 to run instrumentation tests on CI
 - Migrate `RecentPatientsScreen` to a fragment
 - Paginate recent patients list
-- Pin Android emulator build version to resolve emulator timeout error
+- [Reverted] Pin Android emulator build version to resolve emulator timeout error
```
vinaysshenoy, line 17: If we reverted it in this release itself, then we could just delete the line instead of marking it as `[Reverted]`. What do you think?

vinaysshenoy (changes requested):

CHANGELOG.md
vinaysshenoy: Same comment for the older line related to the macos version.

msasikanth: Yeah, we can do that. I was just thinking of keeping track of something changes we made and reverted, that's all.

vinaysshenoy (approved):

PR #2756, opened 2021-06-28
Add support for computing a retain until timestamp from the record retention period

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/3807/add-support-for-online-patient-lookup-api

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/onlinelookup/api/RecordRetention.kt
+++ b/app/src/main/java/org/simple/clinic/patient/onlinelookup/api/RecordRetention.kt
@@ -2,7 +2,10 @@ package org.simple.clinic.patient.onlinelookup.api
 
 import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
+import org.simple.clinic.patient.onlinelookup.api.RetentionType.Permanent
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/simple/clinic/patient/onlinelookup/api/RecordRetention.kt
```diff
@@ -12,4 +15,12 @@ data class RecordRetention(
     @Json(name = "duration_seconds")
     @SecondsDuration
     val retainFor: Duration?
-)
+) {
+
+  fun computeRetainUntilTimestamp(instant: Instant): Instant? {
+    return when (type) {
+      Temporary -> instant + retainFor!!
```
msasikanth, line 22: Huh, didn't knew they had operators support. TIL

msasikanth (approved):

app/src/main/java/org/simple/clinic/patient/onlinelookup/api/RecordRetention.kt
vinaysshenoy: Yeah, this was new to me too. I think the Kotlin compiler handles this automatically if they see a `plus` or `minus` method name.

PR #2770, opened 2021-06-29
Add hypertension treatment question in medical history screen

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3918/add-hypertension-treatment-question-in-medical-history-screen


msasikanth:
Since @janhavisinghh already has couple of pending PRs to review and this is urgent. Assigning @vinaysshenoy as reviewer.

app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryUiRenderer.kt
```diff
@@ -35,13 +35,21 @@ class NewMedicalHistoryUiRenderer(
...
 
     ui.showDiagnosisRequiredError(model.showDiagnosisRequiredError)
 
     renderNextButton(model)
   }
 
+  private fun renderHypertensionTreatmentQuestion(model: NewMedicalHistoryModel) {
+    if (country.isoCountryCode == Country.INDIA && model.diagnosedWithHypertension) {
+      ui.showHypertensionTreatmentQuestion(model.ongoingMedicalHistoryEntry.isOnHypertensionTreatment)
+    }
```
vinaysshenoy: Deciding whether or not to show this question is business logic and must be handled in the update function which should dispatch an effect to show it.

I am approving this for now because this is urgent but we should correct this as soon as we have time.

app/src/test/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryUiRendererTest.kt
```diff
@@ -135,6 +136,50 @@ class NewMedicalHistoryUiRendererTest {
...
+    verify(ui).showDiagnosisRequiredError(false)
+    verify(ui).hideNextButtonProgress()
+    verify(ui).showHypertensionTreatmentQuestion(Unanswered)
+    verifyNoMoreInteractions(ui)
+  }
+
+  @Test
+  fun `when patient has hypertension and country is not from india, then show hypertension treatment question`() {
+    // given
+    val uiRenderer = NewMedicalHistoryUiRenderer(ui, TestData.country(isoCountryCode = Country.BANGLADESH))
```
vinaysshenoy, line 168: Needing to create a different UI renderer for this specific case is a symptom of the problem I mentioned earlier, that this decision should be a business logic decision and not a UI renderer decision.

vinaysshenoy (approved):
There are a couple of changes that need to be made, but we can schedule them for after the feature is done since we are on a timeline.

app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryUiRenderer.kt
msasikanth: I thought about this approach, but it made more sense to show/hide the question view from UI renderer to make sure it gets restored properly after a config change or something like that. So the other option would be to pass the country when creating the model so that we can have a computed property to check-in view render. This check will still be present in the renderer, but the property will now be computed. wdyt?

Ideally, I want to avoid using effects for these sorts of things to avoid UI issues.

vinaysshenoy: That works as well. This should be verified as part of the model, not in the UI renderer.

msasikanth: Yes, it will be present in the model. What I meant was we will use the computed property for the check in the UI renderer like so

```kt
if (model.showOngoingHypertensionTreatment) {
  // code
} else {
  // code
}
```

Sounds good?

vinaysshenoy: LGTM

PR #2786, opened 2021-06-30
Show hypertension treatment error dialog

msasikanth:
https://app.clubhouse.io/simpledotorg/story/3918/add-hypertension-treatment-question-in-medical-history-screen


Honey14 (approved):
Looks good overall. Just a clarification: No where in the app will the user know later about what they selected, right?

msasikanth:
> 
> 
> Looks good overall. Just a clarification: No where in the app will the user know later about what they selected, right?

Yep, this question is only supposed to be visible when creating a new patient. We will not show it in summary as we do with other diagnosis questions.

PR #2788, opened 2021-06-30
Update `ContactPatientBottomSheet` Behaviour - 1

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/3714/update-contactpatientbottomsheet-behaviour?vc_group_by=day&ct_workflow=all&cf_workflow=500000005

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
@@ -2444,13 +2444,13 @@ class AppointmentRepositoryAndroidTest {
     )).blockingAwait()
 
     // then
-    val latest_appointment_today = appointmentRepository.latestOverdueAppointmentForPatient(patientUuid, today.plusDays(1))
+    val latest_appointment_today = appointmentRepository.latestOverdueAppointmentForPatient_Old(patientUuid, today.plusDays(1))
     assertThat(latest_appointment_today.get().appointment).isEqualTo(appointment_scheduled_for_today)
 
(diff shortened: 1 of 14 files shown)


app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -2501,6 +2501,53 @@ class AppointmentRepositoryAndroidTest {
     assertThat(latest_appointment_a_week_later.get().appointment).isEqualTo(appointment_scheduled_for_today_with_reminder_a_week_in_the_future)
   }
 
+  @Suppress("LocalVariableName")
```
msasikanth, line 2504: What's this suppress annotation for? 🤔

msasikanth (commented):
@janhavisinghh Correct me if I am wrong, the query we have already loaded patient data with and without phone numbers right? Is there a reason we need a new query?

We already do this filtering when loading overdue appointments based on the feature flag, so do we need to have a separate query here again?

janhavisinghh:
@msasikanth we need this because we use a separate query to get overdue appointment of a patient while loading the bottomsheet. The databaseview has null phone numbers in it which is why we needed this change. In the last change we only made changes to the query which loads overdue appointments in the overdue list screen.

msasikanth:
> 
> 
> @msasikanth we need this because we use a separate query to get overdue appointment of a patient while loading the bottomsheet. The databaseview has null phone numbers in it which is why we needed this change. In the last change we only made changes to the query which loads overdue appointments in the overdue list screen.

I mean we are already loading the overdue list without phone numbers right (by default, when feature flag is off)? What are the chances that the phone number will be null when we just get a single patient? If the phone number is null they won't be visible in the list in the first place right? We can get on a call if you want to discuss this.

janhavisinghh:
Oooohhh. Fair enough. Got it. We don't need this change🤦🏻‍♀️. We can close this PR.

PR #2797, opened 2021-07-05
Delete patients when the retention time has passed

Honey14:
https://app.clubhouse.io/simpledotorg/story/3804/delete-patient-data-if-it-does-not-comply-to-the-regular-sync-rules

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Add `Drug`s table
 - Migrate `BloodPressureHistoryScreen` to a fragment
 - Migrate `BloodSugarHistoryScreen` to a fragment
+- Delete patients and it's complete medical records when the retention time has passed
 
 ### Features
 - [In Progress: 06 Jul 2021] Add support for finding a patient online from ID scan
--- a/app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
@@ -447,4 +447,45 @@ class PurgeDatabaseAndroidTest {
     assertThat(patientDao.patientProfileImmediate(patientWithPassedRetentionTime.patientUuid)).isNull()
     assertThat(patientDao.patientProfileImmediate(patientWithNotPassedRetentionTime.patientUuid)).isEqualTo(patientWithNotPassedRetentionTime)
   }
+
+  @Test
+  fun purging_the_database_should_hard_delete_blood_pressure_measurements_when_patient_is_null() {
+    // given
+    val deletedPatientUuid = UUID.fromString("1f86c321-5539-44f8-8708-3bcc0e44feed")
+    val syncedPatientProfile = TestData.patientProfile(
+        patientUuid = UUID.fromString("73f22e49-669e-4959-843e-7684622c4184"),
+        generateBusinessId = false,
+        syncStatus = SyncStatus.DONE
+    )
+    val notSyncedPatientProfile = TestData.patientProfile(
+        patientUuid = UUID.fromString("ee60db3e-9ecb-49aa-829c-474f707b3739"),
+        generateBusinessId = false,
+        syncStatus = SyncStatus.PENDING
+    )
+    val deletedBloodPressureMeasurement = TestData.bloodPressureMeasurement(
+        uuid = UUID.fromString("aa2c81ce-236c-4e4a-9ebd-f88a31afb15d"),
+        patientUuid = deletedPatientUuid,
+        syncStatus = SyncStatus.DONE
+    )
+    val notDeletedBloodPressureMeasurement = TestData.bloodPressureMeasurement(
+        uuid = UUID.fromString("34ab2f90-9332-4743-925e-3088a23a933d"),
+        patientUuid = syncedPatientProfile.patientUuid,
+        syncStatus = SyncStatus.DONE,
+        deletedAt = null
+    )
+
+    patientAddressDao.save(listOf(syncedPatientProfile.address, notSyncedPatientProfile.address))
+    patientDao.save(listOf(syncedPatientProfile.patient, notSyncedPatientProfile.patient))
+    bloodPressureDao.save(listOf(deletedBloodPressureMeasurement, notDeletedBloodPressureMeasurement))
+
+    assertThat(bloodPressureDao.getOne(deletedBloodPressureMeasurement.uuid)).isEqualTo(deletedBloodPressureMeasurement)
+    assertThat(bloodPressureDao.getOne(notDeletedBloodPressureMeasurement.uuid)).isEqualTo(notDeletedBloodPressureMeasurement)
+
+    // when
+    appDatabase.purge()
+
+    // then
+    assertThat(bloodPressureDao.getOne(deletedBloodPressureMeasurement.uuid)).isNull()
+    assertThat(bloodPressureDao.getOne(notDeletedBloodPressureMeasurement.uuid)).isEqualTo(notDeletedBloodPressureMeasurement)
+  }
 }
--- a/app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
@@ -488,4 +488,43 @@ class PurgeDatabaseAndroidTest {
     assertThat(bloodPressureDao.getOne(deletedBloodPressureMeasurement.uuid)).isNull()
     assertThat(bloodPressureDao.getOne(notDeletedBloodPressureMeasurement.uuid)).isEqualTo(notDeletedBloodPressureMeasurement)
   }
+
+  @Test
+  fun purging_the_database_should_hard_delete_blood_sugar_measurements_when_patient_is_null() {
+    // given
+    val deletedPatientUuid = UUID.fromString("1f86c321-5539-44f8-8708-3bcc0e44feed")
+    val syncedPatientProfile = TestData.patientProfile(
+        patientUuid = UUID.fromString("b5ae0427-4513-41f9-87ad-127166f35916"),
+        syncStatus = SyncStatus.DONE
+    )
+    val notSyncedPatientProfile = TestData.patientProfile(
+        patientUuid = UUID.fromString("1bd41323-9abc-4c3a-908f-6902a19fe26b"),
+        syncStatus = SyncStatus.PENDING
+    )
+    val deletedBloodSugarMeasurement = TestData.bloodSugarMeasurement(
+        uuid = UUID.fromString("d6530f41-c4be-455c-b5c3-c388e451d3c4"),
+        patientUuid = deletedPatientUuid,
+        deletedAt = Instant.parse("2021-07-01T00:00:00Z"),
+        syncStatus = SyncStatus.DONE
+    )
+    val notDeletedBloodSugarMeasurement = TestData.bloodSugarMeasurement(
+        uuid = UUID.fromString("489b6439-b424-487e-a01a-7ebd48628421"),
+        patientUuid = syncedPatientProfile.patientUuid,
+        syncStatus = SyncStatus.PENDING
+    )
+
+    patientAddressDao.save(listOf(syncedPatientProfile.address, notSyncedPatientProfile.address))
+    patientDao.save(listOf(syncedPatientProfile.patient, notSyncedPatientProfile.patient))
+    bloodSugarDao.save(listOf(deletedBloodSugarMeasurement, notDeletedBloodSugarMeasurement))
+
+    assertThat(bloodSugarDao.getOne(deletedBloodSugarMeasurement.uuid)).isEqualTo(deletedBloodSugarMeasurement)
+    assertThat(bloodSugarDao.getOne(notDeletedBloodSugarMeasurement.uuid)).isEqualTo(notDeletedBloodSugarMeasurement)
+
+    // when
+    appDatabase.purge()
+
+    // then
+    assertThat(bloodSugarDao.getOne(deletedBloodSugarMeasurement.uuid)).isNull()
+    assertThat(bloodSugarDao.getOne(notDeletedBloodSugarMeasurement.uuid)).isEqualTo(notDeletedBloodSugarMeasurement)
+  }
 }
--- a/app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
@@ -412,7 +412,7 @@ class PurgeDatabaseAndroidTest {
   }
 
   @Test
-  fun purging_the_database_should_delete_soft_patients_when_retention_time_has_passed() {
+  fun purging_the_database_should_hard_delete_patients_when_retention_time_has_passed() {
     // given
     val patientWithPassedRetentionTime = TestData.patientProfile(
         patientUuid = UUID.fromString("c3c84410-02f5-430e-a067-1830683a4d27"),
--- a/app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
@@ -527,4 +527,43 @@ class PurgeDatabaseAndroidTest {
     assertThat(bloodSugarDao.getOne(deletedBloodSugarMeasurement.uuid)).isNull()
     assertThat(bloodSugarDao.getOne(notDeletedBloodSugarMeasurement.uuid)).isEqualTo(notDeletedBloodSugarMeasurement)
   }
+
+  @Test
+  fun purging_the_database_should_hard_delete_appointments_when_patient_is_null() {
+    // given
+    val deletedPatientUuid = UUID.fromString("1f86c321-5539-44f8-8708-3bcc0e44feed")
+    val syncedPatientProfile = TestData.patientProfile(
+        patientUuid = UUID.fromString("4913a16c-9ae8-4f20-9c0b-f13c8b61d8ed"),
+        syncStatus = SyncStatus.DONE
+    )
+    val notSyncedPatientProfile = TestData.patientProfile(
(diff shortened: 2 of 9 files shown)


app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -433,5 +433,11 @@ data class Patient(
       )
     """)
     abstract fun deletePatientsNotInFacilities(facilityIds: List<UUID>)
+
+    @Query("""
+        DELETE FROM Patient
```
msasikanth, line 438: Are we sure we want to hard delete patients after retention time even if the sync status is not done? 🤔

CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ## Next Release
 ### Features
-- [In Progress: 29 Jun 2021] Add support for finding a patient online from ID scan
+- [In Progress: 29 Jun 2021]
+  - Add support for finding a patient online from ID scan
+  - Delete patients when the retention time has passed
```
msasikanth, line 7: We can move this to internal, `Add support for finding a patient online from ID scan` is the feature. So any other change that is not visible to the user that is not covered by the above mentioned changelog doesn't have to go here.

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -195,6 +196,7 @@ abstract class AppDatabase : RoomDatabase() {
         purgeDeleted()
         purgeDeletedPhoneNumbers()
         purgeDeletedBusinessIds()
+        purgeDeletedPatientAfterRetentionTime(Instant.now())
```
msasikanth, line 199: Let's add a test to verify this purge. Please, refer to `PurgeDatabaseAndroidTest`

msasikanth (changes requested):
LGTM, left a couple of comments PTAL.

app/src/main/java/org/simple/clinic/patient/Patient.kt
msasikanth: Ah, NVM. `purge` doesn't have to worry about sync.

msasikanth: @Honey14 I have checked the code again, looks like we do need to make sure it's synced before hard deleting the record. Other purge queries have that check `syncStatus = 'DONE'`. Can you confirm if this is the expected behaviour or not once, please?

Honey14: Yes, will confirm this

vinaysshenoy: Yes, in the scenario where a patient was looked up via the API and then edited, we should retain them until the changes are synced to the server even if the retention period passed. The point of purge is to only do it when we are _absolutely_ sure the record is no longer required, which includes changes being synced to the server.

app/src/main/java/org/simple/clinic/AppDatabase.kt
vinaysshenoy: Can we also verify that the associated records (Blood pressures and sugars, Appointments, Prescribed drugs, Medical history) are also deleted when a patient record fetched from online lookup is deleted?

Honey14: Yes, realised this when reviewing a different PR. Will update to deleting the records attached to it as well.

CHANGELOG.md
```diff
@@ -13,7 +13,9 @@
 - Migrate `BloodPressureHistoryScreen` to a fragment
 
 ### Features
-- [In Progress: 06 Jul 2021] Add support for finding a patient online from ID scan
+- [In Progress: 06 Jul 2021]
+  - Add support for finding a patient online from ID scan
+  - Delete patients and it's complete medical records when the retention time has passed
```
msasikanth, line 18: We can just leave the previous changelog "Add support for finding a patient online from ID scan". As far as the users are aware, they don't care about retention times and automatic removal. We can move that change to internal.

app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
```diff
@@ -410,4 +410,52 @@ class PurgeDatabaseAndroidTest {
     assertThat(appointmentDao.getOne(visitedButUnsyncedAppointment.uuid)).isEqualTo(visitedButUnsyncedAppointment)
     assertThat(appointmentDao.getOne(appointmentWithUnknownStatus.uuid)).isEqualTo(appointmentWithUnknownStatus)
   }
+
+  @Test
+  fun purging_the_database_should_delete_soft_patients_when_retention_time_has_passed() {
```
msasikanth, line 415: wdym by `soft_patients`? Was it supposed to be `soft_delete`? 🤔

app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -436,7 +436,8 @@ data class Patient(
 
     @Query("""
         DELETE FROM Patient
-        WHERE deletedAt IS NOT NULL AND retainUntil > :now
+        WHERE deletedAt IS NOT NULL AND retainUntil < :now
```
msasikanth, line 439: So, unless the user is soft deleted on the device. We don't remove them, even if the retention time is completed?

For example, let's say a "Patient X" is synced by the online patient lookup with a 1 day retention time. Some measurements were updated, and then synced. Once the retention period is completed, I don't think this query will remove it since the `deleatedAt` for "Patient X" is null. Please, confirm the behaviour.

app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -440,5 +440,16 @@ data class Patient(
         AND syncStatus == 'DONE'
     """)
     abstract fun purgeDeletedPatientAfterRetentionTime(now: Instant)
+
+    @Query("""
```
msasikanth, line 444: I feel like this query should be part of `BloodPressureMeasurement` DAO and not `Patient` DAO.

app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
```diff
@@ -458,4 +458,52 @@ class PurgeDatabaseAndroidTest {
     assertThat(patientDao.patientProfileImmediate(patientWithNotPassedRetentionTime.patientUuid)).isEqualTo(patientWithNotPassedRetentionTime)
     assertThat(patientDao.patientProfileImmediate(patientWithNotPassedRetentionTimeAndDeleteAtNull.patientUuid)).isEqualTo(patientWithNotPassedRetentionTimeAndDeleteAtNull)
   }
+
+  @Test
+  fun purging_the_database_should_delete_soft_deleted_blood_pressure_meansurements_when_patient_is_null() {
```
msasikanth, line 463: Same comment as before, are we going to only remove data that is synced and soft deleted? You might have carried over conditions from other purge queries, please confirm that once.

app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
```diff
@@ -506,4 +506,50 @@ class PurgeDatabaseAndroidTest {
     assertThat(bloodPressureDao.getOne(notDeletedBloodPressureMeasurement.uuid)).isEqualTo(notDeletedBloodPressureMeasurement)
     assertThat(bloodPressureDao.getOne(deletedButUnsyncedBloodPressureMeasurement.uuid)).isEqualTo(deletedButUnsyncedBloodPressureMeasurement)
   }
+
+  @Test
+  fun purging_the_database_should_delete_soft_deleted_blood_sugar_meansurements_when_patient_is_null() {
```
msasikanth, line 511: Same comment as before about the soft delete.

app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -451,5 +451,16 @@ data class Patient(
 		    )
     """)
     abstract fun purgeDeletedBloodPressureMeasurementWhenPatientIsNull()
+
+    @Query("""
```
msasikanth, line 455: I feel like this should be part of `BloodSugarMeasurementDao` and not `Patient`

app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
```diff
@@ -552,4 +552,50 @@ class PurgeDatabaseAndroidTest {
     assertThat(bloodSugarDao.getOne(notDeletedBloodSugarMeasurement.uuid)).isEqualTo(notDeletedBloodSugarMeasurement)
     assertThat(bloodSugarDao.getOne(deletedButUnsyncedBloodSugarMeasurement.uuid)).isEqualTo(deletedButUnsyncedBloodSugarMeasurement)
   }
+
+  @Test
+  fun purging_the_database_should_delete_soft_deleted_appointments_when_patient_is_null() {
```
msasikanth, line 557: Same comment as above about the soft delete

app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -462,5 +462,16 @@ data class Patient(
 		    )
     """)
     abstract fun purgeDeletedBloodSugarMeasurementWhenPatientIsNull()
+
+    @Query("""
```
msasikanth, line 466: Same comment as above, moving this query to appropriate table DAO

app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
```diff
@@ -598,4 +598,50 @@ class PurgeDatabaseAndroidTest {
     assertThat(appointmentDao.getOne(notDeletedAppointment.uuid)).isEqualTo(notDeletedAppointment)
     assertThat(appointmentDao.getOne(deletedButUnsyncedAppointment.uuid)).isEqualTo(deletedButUnsyncedAppointment)
   }
+
+  @Test
+  fun purging_the_database_should_delete_soft_deleted_medical_history_when_patient_is_null() {
```
msasikanth, line 603: Same comment as above, about the soft delete

app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -473,5 +473,16 @@ data class Patient(
 		    )
     """)
     abstract fun purgeDeletedAppointmentsWhenPatientIsNull()
+
+    @Query("""
```
msasikanth, line 477: Same comment as above, about moving this to appropriate table DAO

app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
```diff
@@ -644,4 +644,50 @@ class PurgeDatabaseAndroidTest {
     assertThat(medicalHistoryDao.getOne(notDeletedMedicalHistory.uuid)).isEqualTo(notDeletedMedicalHistory)
     assertThat(medicalHistoryDao.getOne(deletedButUnsyncedMedicalHistory.uuid)).isEqualTo(deletedButUnsyncedMedicalHistory)
   }
+
+  @Test
+  fun purging_the_database_should_delete_soft_deleted_prescribed_drug_when_patient_is_null() {
```
msasikanth, line 649: Same comment as above, about the soft delete.

app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -484,5 +484,16 @@ data class Patient(
 		    )
     """)
     abstract fun purgeDeletedMedicalHistoryWhenPatientIsNull()
+
+    @Query("""
```
msasikanth, line 488: Same comment as above, about moving this to appropriate table DAO

msasikanth (changes requested):
Left few comments PTAL

app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
vinaysshenoy: It should be a hard delete. Was this just a copy-paste error?

app/src/main/java/org/simple/clinic/patient/Patient.kt
vinaysshenoy: The `deletedAt` check should only be used for soft deletes. This property should not factor into records fetched for online lookup.

vinaysshenoy: Also, `retainUntil` will be `NULL` for most patients. Have we verified how this query behaves in this case?

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -201,6 +202,12 @@ abstract class AppDatabase : RoomDatabase() {
...
```
vinaysshenoy, line 205: We are creating an `Instant` without a `Clock` here. Please use the `Clock` to create an `Instant`, always.

vinaysshenoy: @msasikanth We should consider adding a lint check for creating date and time instances without the `Clock`.

msasikanth: Yeah, makes sense 👍🏾

msasikanth: Also @Honey14 we can probably move this param to the `purge` method so that we can override the current time in tests as well.

app/src/main/java/org/simple/clinic/patient/Patient.kt
vinaysshenoy: ^ Similar feedback for the other comments as well.

app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
Honey14: Yes it should be a hard delete, sorry. I got confused why all other tests were using soft delete. Should've asked, my bad.

msasikanth (approved):
LGTM. There is one change that can be made in the next PR. The DAO methods names are named as `purgeDeleted`. We are not purging deleted values, rather we are purging data that's is past retain until value or data that doesn't have the patient in DB anymore. So we just have to rename those methods. 

@Honey14 There is a code climate issue, can you resolve that. Feel free to make the above-mentioned change along with it or else you can do it in a different PR.

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -196,13 +197,13 @@ abstract class AppDatabase : RoomDatabase() {
   }
 
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-  fun purge() {
+  fun purge(utcClock: UtcClock) {
```
msasikanth, line 200: I completely missed this before. We are supposed to use `UserClock` here and not `UtcClock`. The `retainUntil` value we save is based on the user clock, not a UTC value.

msasikanth (changes requested):

app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -164,6 +166,9 @@ abstract class AppDatabase : RoomDatabase() {
 
   abstract fun teleconsultRecordDao(): TeleconsultRecord.RoomDao
 
+  @Inject
+  lateinit var userClock: UserClock
```
msasikanth, line 170: Ideally, we should not inject it into this `AppDatabase` abstract class, instead, we can pass the `UserClock` to the `prune` from the call site. You can make that change in later PR.

msasikanth: I am amazed that this didn't throw in dagger build errors 🤔

msasikanth (approved):

app/src/main/java/org/simple/clinic/AppDatabase.kt
vinaysshenoy: Ideally, the `prune` method should just take an `Instant` as a parameter. Not even the clock directly.

msasikanth: +1

Honey14:
@msasikanth Hey Sasi, about the comment about renaming. The Dao methods are not named as purgeDeleted. Please check again. The naming convention is perfect we don’t need to change. Purge deleted is basically deleting unwanted data and something like `purgeDeletedAppointmentsWhenPatientIsNull` means deleting unwanted appointments when patient is null and the same for the rest of the medical records.

msasikanth:
> 
> 
> @msasikanth Hey Sasi, about the comment about renaming. The Dao methods are not named as purgeDeleted. Please check again. The naming convention is perfect we don’t need to change. Purge deleted is basically deleting unwanted data and something like `purgeDeletedAppointmentsWhenPatientIsNull` means deleting unwanted appointments when patient is null and the same for the rest of the medical records.

I am not sure about that. While reading that method name it seemed like we are purging deleted data. `purgeDeletedAppointmentsWhenPatientIsNull` is the perfect example for it, here the method name starts with purge deleted. Instead, it should be something like `purgeAppointmentsWhenPatientIsNull`. Deleted is in the past tense, so it seems like we are purging data that has been already deleted. purge & deleted have a similar meaning, so we don't have to add both of them in the name to indicate we are deleting data.

Honey14:
Understood so I think the only problem is the past tense of delete and `purgeAppointmentsWhenPatientIsNull` wouldn't make sense. I'll update it to something like `purgeDeleteAppointmentsWhenPatientIsNull`

msasikanth:
> 
> 
> Understood so I think the only problem is the past tense of delete and `purgeAppointmentsWhenPatientIsNull` wouldn't make sense. I'll update it to something like `purgeDeleteAppointmentsWhenPatientIsNull`

Like I said, purge and delete have the same meaning, just removing something. So does it make sense to have both? wdyt? If you look at other purge methods, you can see we are just adding purge to the condition we are running that purge on. For example `purgeDeletedPhoneNumbers` or `purgeDeletedBusinessIds`. So, we don't have to add `delete` in the name, purge means deleting, so it's fine.

PR #2801, opened 2021-07-06
Add fallback behaviour for unknown record retention

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/3927/add-a-fallback-behaviour-if-the-retention-type-enum-is-not-known-to-the-app

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next Release
 ### Features
-- [In Progress: 29 Jun 2021] Add support for finding a patient online from ID scan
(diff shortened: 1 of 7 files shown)


msasikanth (approved):
LGTM. Just one comment, should we consider the default fallback duration to be at least a week or so instead of 1 day? 🤔

vinaysshenoy:
> LGTM. Just one comment, should we consider the default fallback duration to be at least a week or so instead of 1 day? thinking

We can configure this remotely if needed. The server controlled retention is actually one hour so I didn't want to retain the record for much longer than that.

PR #2802, opened 2021-07-06
Clean up `gradle.properties`

msasikanth:
- Add `mixpanelToken` to `gradle.properties`
- Add `manifestEndpoint` in `gradle.properties`
- Add `fallbackApiEndpoint` in `gradle.properties`
- Remove unused code

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -135,6 +135,7 @@ android {
 
   productFlavors {
     val mixpanelToken: String by project
+    val manifestEndpoint: String by project
 
     create("qa") {
       val qaManifestEndpoint: String by project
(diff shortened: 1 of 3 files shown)


Honey14 (approved):
Looks good to me

app/build.gradle.kts
```diff
@@ -134,75 +134,59 @@ android {
   flavorDimensions("track")
 
   productFlavors {
-    create("qa") {
-      val qaManifestEndpoint: String by project
-      val qaFallbackApiEndpoint: String by project
-      val qaMixpanelToken: String by project
+    val mixpanelToken: String by project
+    val manifestEndpoint: String by project
+    val fallbackApiEndpoint: String by project
```
vinaysshenoy, line 139: Correct me if I am wrong, but now I think we can also move the build config declaration fields into the default config instead of duplicating them for every flavour, right?

msasikanth: We can move them to `android` scope but not `defaultConfig` since that has a different scope than `productFlavors`. I generally prefer to keep these definitions close to where they are being used. But we can move all the variables to the top of `android` scope and then reference them.

Another reason is, these variables are only being used in `productFlavors` scope, so I didn't see the reason to extract them out for other scopes.

vinaysshenoy: I think I left my comment in the wrong place. I meant to add it for the `buildConfigField` declarations. Specifically

```kotlin
buildConfigField("String", "MANIFEST_ENDPOINT", "\"$manifestEndpoint\"")
buildConfigField("String", "FALLBACK_ENDPOINT", "\"$fallbackApiEndpoint\"")
buildConfigField("String", "MIXPANEL_TOKEN", "\"$mixpanelToken\"")
```

We were declaring them in `productFlavours` scope earlier because we needed to reference different gradle properties in each flavour, but that is no longer the case.

msasikanth: Ah, yes these can be extracted out. Will raise a PR for that change.

PR #2806, opened 2021-07-07
Add ADRs for SQL performance profiling

vinaysshenoy:


doc/arch/013-sql-performance-profiling-v2.md
```diff
@@ -0,0 +1,104 @@
...
+    });
+}
+```
+
+Adding this step to the build allowed us to automatically capture SQL performance profiling at whatever level of granularity we need.
+
+### Implementation
+The implementation of the build tool that transforms the generated Room DAOs is available [here](https://github.com/simpledotorg/room-metadata-generator). The integration with the `simple-android` build process and the reporting of the events to our analytics platform is available [here](https://github.com/simpledotorg/simple-android/pull/2703).
+
+## Consequences
```
msasikanth, line 97: Should we add limitations like paging again here?

msasikanth (approved):

doc/arch/013-sql-performance-profiling-v2.md
vinaysshenoy: This is no longer a limitation since we can add support for it. The older method made it impossible.

PR #2815, opened 2021-07-08
Fix `ContactPatientBottomSheet` not going back to call patient view on back click in call later mode

msasikanth:
**Testing steps**:
- Go to overdue list
- Open contact patient bottom sheet
- Select "Remind to call later..."
- Click Android back button
- It should go back to call patient view

### DIFF
--- a/app/src/main/java/org/simple/clinic/util/AndroidExtensions.kt
+++ b/app/src/main/java/org/simple/clinic/util/AndroidExtensions.kt
@@ -64,3 +64,14 @@ private fun Configuration.isLocaleAlreadyOverriden(): Boolean {
     else -> false
   }
 }
+
+inline fun Dialog.onBackPressed(crossinline backPressed: () -> Unit) {
+  setOnKeyListener { _, keyCode, event ->
+    if (event.action == KeyEvent.ACTION_UP && keyCode == KeyEvent.KEYCODE_BACK) {
+      backPressed()
+      true
+    } else {
+      false
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


msasikanth:
Unfortunately [`OnBackPressedCallback`](https://developer.android.com/reference/androidx/activity/OnBackPressedCallback), doesn't work in this scenario (for dialogs). So we would have to use `setOnKeyListener`.

app/src/main/java/org/simple/clinic/util/AndroidExtensions.kt
```diff
@@ -64,3 +64,14 @@ private fun Configuration.isLocaleAlreadyOverriden(): Boolean {
     else -> false
   }
 }
+
+inline fun Dialog.onBackPressed(crossinline backPressed: () -> Unit) {
+  setOnKeyListener { _, keyCode, event ->
+    if (event.action == KeyEvent.ACTION_UP && keyCode == KeyEvent.KEYCODE_BACK) {
```
vinaysshenoy, line 70: I feel like this condition `event.action == KeyEvent.ACTION_UP && keyCode == KeyEvent.KEYCODE_BACK` is being used in multiple places. Should we move this to an extension property, say, `KeyEvent.isBackPressEvent` or something?

vinaysshenoy (changes requested):
Overall PR looks good, left one comment.

app/src/main/java/org/simple/clinic/util/AndroidExtensions.kt
msasikanth: Actually, this is the only place we are using it now, I have removed the previous `onOverrideKey` extension. So for sheets that want to handle the back clicks manually, they need to use this extension.

vinaysshenoy (approved):

PR #2817, opened 2021-07-09
Support batching sync push requests

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/4085/support-batching-of-sync-push-requests

### DIFF
--- a/app/src/main/java/org/simple/clinic/bloodsugar/BloodSugarMeasurement.kt
+++ b/app/src/main/java/org/simple/clinic/bloodsugar/BloodSugarMeasurement.kt
@@ -91,6 +91,17 @@ data class BloodSugarMeasurement(
     @Query("SELECT * FROM BloodSugarMeasurements WHERE syncStatus = :status")
     fun withSyncStatus(status: SyncStatus): List<BloodSugarMeasurement>
 
+    @Query("""
+      SELECT * FROM bloodsugarmeasurements
+      WHERE syncStatus = :syncStatus
+      LIMIT :limit OFFSET :offset
+    """)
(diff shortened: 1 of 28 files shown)


app/src/main/java/org/simple/clinic/bloodsugar/BloodSugarMeasurement.kt
```diff
@@ -91,6 +91,17 @@ data class BloodSugarMeasurement(
     @Query("SELECT * FROM BloodSugarMeasurements WHERE syncStatus = :status")
     fun withSyncStatus(status: SyncStatus): List<BloodSugarMeasurement>
 
+    @Query("""
+      SELECT * FROM bloodsugarmeasurements
+      WHERE syncStatus = :syncStatus
+      LIMIT :limit OFFSET :offset
+    """)
+    fun withSyncStatusBatched(
```
Honey14, line 99: The naming convention for other syncStatus batched is `dao#recordsWithSyncStatusBatched`. Can you add the same here as well. It'll be consistent throughout

Honey14 (changes requested):
Looks good overall. Left one comment PTAL

app/src/main/java/org/simple/clinic/bloodsugar/BloodSugarMeasurement.kt
vinaysshenoy: Good point. Will do this.

vinaysshenoy: Done.

Honey14 (approved):
LGTM

PR #2810, opened 2021-07-07
Update `ContactPatientBottomSheet` Behaviour - 1

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/3714/update-contactpatientbottomsheet-behaviour?vc_group_by=day&ct_workflow=all&cf_workflow=500000005 #2810

PR #2818, opened 2021-07-09
Update `ContactPatientBottomSheet` Behaviour to use the new `CallPatientView` 

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/3714/update-contactpatientbottomsheet-behaviour

Merge after #2810

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Enqueue `UpdateRemoteConfigWorker` on app cold starts
 - Use `SyncConfigType` qualifier for daily and frequent sync configs
 - Bump Kotlin to v1.5.21
+- Updated `ContactPatientBottomSheet` behaviour for supporting patients with and without phone number
 
 ### Features
 - [In Progress: 06 Jul 2021] Add support for finding a patient online from ID scan
--- a/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
+++ b/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
@@ -209,6 +209,14 @@ class ContactPatientBottomSheet : BaseBottomSheet<
     callPatientView.showPatientWithNoPhoneNumberLayout = true
   }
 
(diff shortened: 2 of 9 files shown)


janhavisinghh:
@msasikanth this is ready for review

vinaysshenoy:
Can we please avoid vague PR names like "Update `ContactPatientBottomSheet `-Final" and be specific about what changes are happening as part this PR?

msasikanth (changes requested):
Left few comments PTAL. Also, it looks like we are not showing transferred from the label in the contact sheet. Please refer to the PRD for more explanation on it.

msasikanth:
Also, it looks like the UI doesn't match the specs from Figma, PTAL at a couple of screenshots attached below.

screenshot-1626062914516

screenshot-1626062778453

janhavisinghh:
Removed `switchToSetAppointmentReminderView_Old` and `switchToSetAppointmentReminderView` changes as we needed to hide both `callPatientView` and `callPatientView_Old` when that method was called. Hence, just added that as a new commit.

janhavisinghh:
@msasikanth To display the transferred from facility name I will have to write a new migration to add the `patientAssignedFacilityName` in the `OverdueAppointment` database view. I'm planning to raise another small PR for it. Currently we are displaying the `assignedFacilityName` for both transferred from and registered at labels.

msasikanth:
> @msasikanth To display the transferred from facility name I will have to write a new migration to add the `patientAssignedFacilityName` in the `OverdueAppointment` database view. I'm planning to raise another small PR for it. Currently we are displaying the `assignedFacilityName` for both transferred from and registered at labels.

We don't have to add assigned facility name column, we have to add registered facility name. Let's get on a short call, I just want to make sure we are all on same page about this.

janhavisinghh:
As discussed offline I will raise another PR for transferred from facility implementation.

msasikanth (approved):
LGTM. There are still some layout margins and paddings that need to be adjusted, we can do that in the next PR.

PR #2819, opened 2021-07-12
Fix overdue list not changing when switching facility from overdue screen

msasikanth:
- Inject current facility stream in `OverdueEffectHandler`
- Subscribe to `currentFacilityStream` for getting current facility
- Rename `isLoading` to `isLoadingInitialData` in `overdueListAdapterLoadStateListener`
- Update `overdueListAdapterLoadStateListener` to display progress when loading data after switching facility
- Update CHANGELOG

**Story:** https://app.clubhouse.io/simpledotorg/story/4108/overdue-list-not-changing-when-switching-facility-from-the-overdue-screen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 ### Fixes
 - Fix invalid qr code error when scanning a valid Indian NHID
 - Fix `ContactPatientBottomSheet` not going back to call patient view on back click in call later mode
+- Fix overdue list not changing when switching facility from overdue screen
   
 ## On Demo
 ### Internal
--- a/app/src/main/java/org/simple/clinic/home/overdue/OverdueEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/OverdueEffectHandler.kt
@@ -4,17 +4,17 @@ import com.spotify.mobius.rx2.RxMobius
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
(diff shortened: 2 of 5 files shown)


vinaysshenoy:
@msasikanth Seems like there's a code climate issue. Please take a look?

msasikanth:
> 
> 
> @msasikanth Seems like there's a code climate issue. Please take a look?

I cannot think of a logical name for combining those 2 boolean variables 😅 . Do we need to combine the conditional logic if it's just 2 boolean flags?

vinaysshenoy:
> > @msasikanth Seems like there's a code climate issue. Please take a look?
> 
> I cannot think of a logical name for combining those 2 boolean variables sweat_smile . Do we need to combine the conditional logic if it's just 2 boolean flags?

From a product perspective, I feel that the notion we are trying to represent here is "Has the overdue list fully loaded?". If I am right here, how about we try to create a new variable based on this notion, say `hasOverdueListFullyLoaded`?

msasikanth:
> 
> 
> > > @msasikanth Seems like there's a code climate issue. Please take a look?
> > 
> > 
> > I cannot think of a logical name for combining those 2 boolean variables sweat_smile . Do we need to combine the conditional logic if it's just 2 boolean flags?
> 
> From a product perspective, I feel that the notion we are trying to represent here is "Has the overdue list fully loaded?". If I am right here, how about we try to create a new variable based on this notion, say `hasOverdueListFullyLoaded`?

Done ✅

vinaysshenoy (approved):

PR #2820, opened 2021-07-12
Update sync status of pending records when a batch completes

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/4105/update-sync-status-on-a-batch-level


vinaysshenoy:
Closing this because the implementation was more complex than I expected. Will revisit after the sync refactoring work is done.

PR #2822, opened 2021-07-12
Update `Instant` to be passed from prune()

Honey14:
Refer : https://github.com/simpledotorg/simple-android/pull/2797#discussion_r667745078
Refer : https://github.com/simpledotorg/simple-android/pull/2797#issuecomment-878143897

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
@@ -8,8 +8,6 @@ import org.simple.clinic.TestClinicApp
 import org.simple.clinic.TestData
 import org.simple.clinic.overdue.Appointment
 import org.simple.clinic.patient.SyncStatus
-import org.simple.clinic.util.TestUserClock
-import org.simple.clinic.util.TestUtcClock
 import java.time.Instant
 import java.util.UUID
 import javax.inject.Inject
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/simple/clinic/AppDatabase.kt
```diff
@@ -182,9 +182,9 @@ abstract class AppDatabase : RoomDatabase() {
     }
   }
 
-  fun prune() {
+  fun prune(currentInstant: Instant) {
```
msasikanth, line 185: We can call it `now` instead of `currentInstant` for consistency

app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
```diff
@@ -79,7 +79,7 @@ class PurgeDatabaseAndroidTest {
     assertThat(patientDao.patientProfileImmediate(deletedButUnsyncedPatientProfile.patientUuid)).isEqualTo(deletedButUnsyncedPatientProfile)
 
     // when
-    appDatabase.purge(testUserClock)
+    appDatabase.purge(Instant.now(testUserClock))
```
msasikanth, line 82: If we do this, tests are not deterministic anymore. For example, let's say we have set the `retainUntil` to sometime in future in a test verifying it doesn't get deleted. Once `Instant.now` exceeds that, those tests will start failing. Instead, we should use `Instant.parse("2018-01-01T00:00:00Z")` to set a more deterministic value that we can control and stays consistent.

msasikanth (changes requested):

app/src/androidTest/java/org/simple/clinic/storage/PurgeDatabaseAndroidTest.kt
Honey14: Oh yes, my mistake. Will update this.

msasikanth (approved):

msasikanth (approved):

msasikanth (approved):

PR #2823, opened 2021-07-12
Implement `Drug` sync

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2895/implement-drug-sync

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/di/TestAppComponent.kt
+++ b/app/src/androidTest/java/org/simple/clinic/di/TestAppComponent.kt
@@ -12,6 +12,7 @@ import org.simple.clinic.bp.BloodPressureHistoryListItemDataSourceTest
 import org.simple.clinic.bp.BloodPressureRepositoryAndroidTest
 import org.simple.clinic.drugs.PrescriptionRepositoryAndroidTest
 import org.simple.clinic.drugs.search.DrugRepositoryAndroidTest
+import org.simple.clinic.drugs.search.sync.DrugSyncIntegrationTest
 import org.simple.clinic.facility.FacilityRepositoryAndroidTest
 import org.simple.clinic.login.LoginUserWithOtpServerIntegrationTest
 import org.simple.clinic.medicalhistory.MedicalHistoryRepositoryAndroidTest
@@ -116,4 +117,5 @@ interface TestAppComponent {
   fun inject(target: ServerRegistrationAtFacilityRule)
   fun inject(target: LookupPatientOnlineApiIntegrationTest)
   fun inject(target: DrugRepositoryAndroidTest)
+  fun inject(target: DrugSyncIntegrationTest)
 }
--- a/app/src/androidTest/java/org/simple/clinic/di/TestAppComponent.kt
+++ b/app/src/androidTest/java/org/simple/clinic/di/TestAppComponent.kt
@@ -11,6 +11,7 @@ import org.simple.clinic.bloodsugar.BloodSugarRepositoryAndroidTest
 import org.simple.clinic.bp.BloodPressureHistoryListItemDataSourceTest
 import org.simple.clinic.bp.BloodPressureRepositoryAndroidTest
 import org.simple.clinic.drugs.PrescriptionRepositoryAndroidTest
+import org.simple.clinic.drugs.search.DrugRepositoryAndroidTest
 import org.simple.clinic.facility.FacilityRepositoryAndroidTest
 import org.simple.clinic.login.LoginUserWithOtpServerIntegrationTest
 import org.simple.clinic.medicalhistory.MedicalHistoryRepositoryAndroidTest
@@ -114,4 +115,5 @@ interface TestAppComponent {
   fun inject(target: DeleteSyncGroupDatabaseAndroidTest)
   fun inject(target: ServerRegistrationAtFacilityRule)
   fun inject(target: LookupPatientOnlineApiIntegrationTest)
+  fun inject(target: DrugRepositoryAndroidTest)
 }
--- a/app/src/androidTest/java/org/simple/clinic/drugs/search/DrugRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/drugs/search/DrugRepositoryAndroidTest.kt
@@ -1,0 +1,55 @@
+package org.simple.clinic.drugs.search
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+import org.simple.clinic.AppDatabase
+import org.simple.clinic.TestClinicApp
+import org.simple.clinic.TestData
+import java.util.UUID
+import javax.inject.Inject
+
+class DrugRepositoryAndroidTest {
+
+  @Inject
+  lateinit var database: AppDatabase
+
+  @Inject
+  lateinit var drugRepository: DrugRepository
+
+  @Before
+  fun setUp() {
+    TestClinicApp.appComponent().inject(this)
+  }
+
+  @After
+  fun tearDown() {
+    database.clearAllTables()
+  }
+
+  @Test
+  fun saving_drugs_should_work_correctly() {
+    // given
+    val drugs = listOf(
+        TestData.drug(
+            id = UUID.fromString("4e64d256-ea7e-4b90-8177-1eb1485630c3"),
+            name = "Amlodipine",
+            dosage = "10 mg"
+        ),
+        TestData.drug(
+            id = UUID.fromString("0864635d-fd63-4005-9cea-ba843ea804e6"),
+            name = "Amlodipine",
+            dosage = "20 mg"
+        )
+    )
+
+    // when
+    drugRepository.save(drugs).blockingAwait()
+
+    // then
+    val savedDrugs = drugRepository.drugs()
+
+    assertThat(savedDrugs).isEqualTo(drugs)
+  }
+}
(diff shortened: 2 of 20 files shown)


msasikanth:
Conversation about batching support in medications API: https://simpledotorg.slack.com/archives/C01P131R1GR/p1626094192001300

app/src/main/java/org/simple/clinic/drugs/search/sync/DrugPayload.kt
```diff
@@ -0,0 +1,71 @@
...
+        protocol = protocol,
+        common = common,
+        timestamps = Timestamps(
+            createdAt = createdAt,
+            updatedAt = updatedAt,
+            deletedAt = deletedAt
+        )
+    )
+  }
+}
```
vinaysshenoy, line 71: There seems to be a missing line feed here. Please correct?

app/src/androidTest/java/org/simple/clinic/di/TestAppComponent.kt
```diff
@@ -11,6 +11,7 @@ import org.simple.clinic.bloodsugar.BloodSugarRepositoryAndroidTest
 import org.simple.clinic.bp.BloodPressureHistoryListItemDataSourceTest
```
vinaysshenoy, line 11: Can we please reword the commit message to follow the convention of "If applied, this commit will <commit message>"?

app/src/androidTest/java/org/simple/clinic/drugs/search/DrugRepositoryAndroidTest.kt
```diff
@@ -0,0 +1,57 @@
...
+            name = "Amlodipine",
+            dosage = "20 mg"
+        )
+    )
+
+    // when
+    drugRepository.save(drugs).blockingAwait()
+
+    // then
+    val expectedDrugs = drugs.map { it.id }
```
vinaysshenoy, line 51: Can we verify the complete record by default instead of just the IDs? We have followed this pattern in some places because of some issues with testability, but we should fix those instead of following that pattern here.

app/src/main/java/org/simple/clinic/drugs/search/DrugRepository.kt
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  private fun saveRecords(records: List<Drug>) {
+    drugDao.save(records)
+  }
+
+  override fun pendingSyncRecords(limit: Int, offset: Int): List<Drug> {
+    return emptyList()
+  }
+}
```
vinaysshenoy, line 53: Missing line feed here. I think you might have forgotten to add this setting when you started using your Mac for development again. 😅

app/src/main/java/org/simple/clinic/sync/SyncConfigModule.kt
```diff
@@ -29,6 +30,27 @@ class SyncConfigModule {
     )
   }
 
+  /**
+   * Drug sync doesn't support batch size at API level yet. So using the normal batch size can cause
+   * API loop if the payloads size is greater than the batch size. So, to avoid that issue and also
+   * support batching in case server implements batching support for this API in future, we will have
+   * a separate `SyncConfig` for the `DrugSync`.
+   * 
+   * We can remove this once medications API has added the batching support and replace this usage
```
vinaysshenoy, line 39: Can we create tickets in Clubhouse to track this instead of leaving it as a comment here? Also, is this expected to be done before this feature goes to `PROD`? I'm concerned about leaving a potential infinite sync loop case in `PROD`,

vinaysshenoy: The reason is that the remote config sync currently happens _after_ all the other syncs are done. So in the event that we do run into this case in `PROD`, we might end up in a situation where bumping the batch size remotely has no effect in `PROD` because the remote config does not update because it is waiting for the drug sync to complete.

app/src/main/java/org/simple/clinic/sync/SyncConfigType.kt
```diff
@@ -0,0 +1,11 @@
+package org.simple.clinic.sync
+
+import javax.inject.Qualifier
+
+@Qualifier
+annotation class SyncConfigType(val type: Type) {
+
+  enum class Type {
+    Drugs
```
vinaysshenoy, line 9: I'm guessing we would also move `Daily` and `Frequent` here, correct?

app/src/androidTest/java/org/simple/clinic/drugs/search/sync/DrugSyncIntegrationTest.kt
```diff
@@ -0,0 +1,91 @@
...
+
+  private fun resetLocalData() {
+    appDatabase.clearAllTables()
+    lastPullToken.delete()
+  }
+
+  @Test
+  fun syncing_records_should_work_as_expected() {
+    // when
+    Truth.assertThat(repository.recordCount().blockingFirst()).isEqualTo(0)
```
vinaysshenoy, line 83: Can we import the `Truth.assertThat` method, please?

app/src/main/java/org/simple/clinic/drugs/search/DrugModule.kt
```diff
@@ -11,4 +21,15 @@ object DrugModule {
...
     return database.drugDao()
   }
+
+  @Provides
+  fun drugSyncApi(@Named("for_country") retrofit: Retrofit): DrugSyncApi {
+    return retrofit.create(DrugSyncApi::class.java)
+  }
+
+  @Provides
+  @TypedPreference(LastDrugPullToken)
```
vinaysshenoy, line 31: I've been considering creating resource scoped modules for DI. An example is the `@ForPatientSync` qualifier, which is currently being used only for getting the record retention fallback duration, but I am planning to move all the DI qualifiers related to patient sync to this and refactoring the DI setup around resource scope,

What do you think of this approach?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/drugs/search/DrugRepository.kt
msasikanth: Ah, I forgot to enable it again after install AS again on Windows 😅

app/src/androidTest/java/org/simple/clinic/drugs/search/DrugRepositoryAndroidTest.kt
msasikanth: Ah, that makes sense. I was keeping it consistent with the other tests.

app/src/main/java/org/simple/clinic/sync/SyncConfigModule.kt
msasikanth: It's not, according to Prabhanshu we are going to add the batching support if the medications cross 1000 or something like that.

https://simpledotorg.slack.com/archives/C01P131R1GR/p1626094192001300


>The reason is that the remote config sync currently happens after all the other syncs are done. So in the event that we do run into this case in PROD, we might end up in a situation where bumping the batch size remotely has no effect in PROD because the remote config does not update because it is waiting for the drug sync to complete.

That can be an issue, currently, we want to make sure the older versions also has support for batching. So when we eventually end up adding batching support we should first update the remote config and also the app (to go back to normal sync config) and then roll the server changes to `PROD`. wdyt?

msasikanth: > Can we create tickets in Clubhouse to track this instead of leaving it as a comment here? Also, is this expected to be done before this feature goes to PROD? I'm concerned about leaving a potential infinite sync loop case in PROD,

That comment was mainly for developers to let them know why there is a new sync config. Added the CH story for removing the sync config [here](https://app.clubhouse.io/simpledotorg/story/4125/remove-drugs-sync-config-after-batching-support-is-added-to-medications-api)

app/src/main/java/org/simple/clinic/sync/SyncConfigType.kt
msasikanth: Yep, will be done in a different PR eventually

app/src/main/java/org/simple/clinic/drugs/search/DrugModule.kt
msasikanth: Yeah, that would be really useful and easy to understand

app/src/main/java/org/simple/clinic/sync/SyncConfigModule.kt
vinaysshenoy: Hmm, one thing we should do then before pushing this to `PROD` is to decouple updating the remote config from the sync resources. This is something I've been meaning to do for a while because there is no reason for remote config to be synced in the same way.

This will allow us to change the batch size to support older versions since the remote config will get updated regardless of whether or not the drug sync gets into one. What do you think?

app/src/main/java/org/simple/clinic/drugs/search/DrugModule.kt
vinaysshenoy: https://app.clubhouse.io/simpledotorg/story/4126/refactor-sync-related-di-providers-into-resource-scoped-modules

app/src/main/java/org/simple/clinic/sync/SyncConfigModule.kt
msasikanth: Yep, that sounds good. We can move this into draft until that PR is merged in if you want.

vinaysshenoy: Not necessary, we can merge this in and then pick that up on priority.

vinaysshenoy: Would you be able to take a crack at this? We essentially need to add a new work manager job to keep the remote config updated.

vinaysshenoy (approved):

app/src/main/java/org/simple/clinic/sync/SyncConfigModule.kt
msasikanth: Okay, will do that 👍🏾

PR #2826, opened 2021-07-13
Extract remote config sync from the regular sync resources

msasikanth:
https://app.clubhouse.io/simpledotorg/story/4139/extract-remote-config-sync-from-the-regular-sync-resources


msasikanth:
I haven't added the remote config sync to enqueue when sync indicator is clicked. So currently it only runs after cold start, do let me know if I need to run it on sync indicator clicks as well.

vinaysshenoy:
> I haven't added the remote config sync to enqueue when sync indicator is clicked. So currently it only runs after cold start, do let me know if I need to run it on sync indicator clicks as well.

I think this should be fine for now.

app/src/main/java/org/simple/clinic/sync/RemoteConfigSyncWorker.kt
```diff
@@ -0,0 +1,50 @@
...
+import androidx.work.Constraints
+import androidx.work.NetworkType
+import androidx.work.OneTimeWorkRequest
+import androidx.work.Worker
+import androidx.work.WorkerParameters
+import org.simple.clinic.ClinicApp
+import org.simple.clinic.remoteconfig.RemoteConfigService
+import javax.inject.Inject
+
+class RemoteConfigSyncWorker(
```
vinaysshenoy, line 13: Since we already use the "Sync" to refer to the domain specific sync resources, can we use a different name for this so that it is clear the two are not related? Something like `UpdateRemoteConfigWorker` or something similar?

Similar thoughts on the package naming as well.

app/src/main/java/org/simple/clinic/sync/RemoteConfigSyncWorker.kt
```diff
@@ -0,0 +1,50 @@
...
+
+  override fun doWork(): Result {
+    ClinicApp.appComponent.inject(this)
+
+    try {
+      remoteConfigService.update()
+    } catch (e: Exception) {
+      // Individual syncs report their errors internally so we can just
+      // ignore this caught error. This is a good place for future
+      // improvements like attempting a backoff based retry.
```
vinaysshenoy, line 45: This comment is irrelevant here, was this a copy-pasta? 😅

vinaysshenoy: Should we report this as an error to Sentry, just for information?

app/src/main/java/org/simple/clinic/sync/RemoteConfigSyncWorker.kt
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    return Result.success()
```
vinaysshenoy, line 48: Just a thought, should we return `Result.failure()` if there's an actual failure to sync. This will cause the work manager to attempt an exponential backoff retry, so might be a nice to have.

vinaysshenoy (changes requested):
Overall PR looks good, left a couple of minor comments.

app/src/main/java/org/simple/clinic/sync/RemoteConfigSyncWorker.kt
msasikanth: Yep, I just copied the current sync worker and forget to remove this comment 🤦🏾

app/src/main/java/org/simple/clinic/sync/RemoteConfigSyncWorker.kt
msasikanth: Why are we not doing that for `SyncWorker`, doesn't it also have an exponential backoff setup? 🤔

vinaysshenoy (dismissed):

vinaysshenoy (approved):

PR #2831, opened 2021-07-14
Implement `DrugsSearchScreen`

msasikanth:
**Story**: https://app.clubhouse.io/simpledotorg/story/2899/implement-drugssearchscreen

- Add query to search for drugs
- Add page size for drug search results
- Add effect to search drugs
- Handle `SetDrugsSearchResults` effect
- When drug search results are loaded then show search results
- When search query is changed, then update model and search drugs
- When search query is empty, then update model
- When search query is empty, then hide search results
- When search query is not empty, then show search results

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/drugs/search/DrugRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/drugs/search/DrugRepositoryAndroidTest.kt
@@ -8,6 +8,8 @@ import org.simple.clinic.AppDatabase
 import org.simple.clinic.PagingTestCase
 import org.simple.clinic.TestClinicApp
 import org.simple.clinic.TestData
+import org.simple.clinic.protocol.ProtocolAndProtocolDrugs
+import org.simple.clinic.protocol.ProtocolRepository
 import java.util.UUID
 import javax.inject.Inject
 
@@ -18,6 +20,9 @@ class DrugRepositoryAndroidTest {
 
   @Inject
   lateinit var drugRepository: DrugRepository
+
+  @Inject
+  lateinit var protocolRepository: ProtocolRepository
 
   @Before
   fun setUp() {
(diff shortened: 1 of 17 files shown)


app/src/androidTest/java/org/simple/clinic/drugs/search/DrugRepositoryAndroidTest.kt
```diff
@@ -52,4 +53,38 @@ class DrugRepositoryAndroidTest {
...
+        id = UUID.fromString("5db739bb-1fa6-470c-b230-4fead4e92543"),
+        name = "Losartan",
+        dosage = "25mg"
+    )
+    val drugs = listOf(amlodipine10, amlodipine20, losartan25)
+
+    drugRepository.save(drugs).blockingAwait()
+
+    // when
+    val expectedSearchResults = PagingTestCase(pagingSource = drugRepository.search(searchQuery),
```
vinaysshenoy, line 111: Can we pick the search query such that the case-insensitivity of the SQL query is also verified?

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchEffectHandler.kt
```diff
@@ -0,0 +1,38 @@
...
+        .subtypeEffectHandler<DrugSearchEffect, DrugSearchEvent>()
+        .addTransformer(SearchDrugs::class.java, searchDrugs())
+        .build()
+  }
+
+  private fun searchDrugs(): ObservableTransformer<SearchDrugs, DrugSearchEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .flatMap {
```
vinaysshenoy, line 42: This should be a `.switchMap`, right?

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchEvent.kt
```diff
@@ -5,3 +5,5 @@ import androidx.paging.PagingData
 sealed class DrugSearchEvent
 
 data class DrugsSearchResultsLoaded(val searchResults: PagingData<Drug>) : DrugSearchEvent()
+
+data class SearchQueryChanged(val searchQuery: String?) : DrugSearchEvent()
```
vinaysshenoy, line 9: Why does this need to be nullable? 🤔

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchModel.kt
```diff
@@ -1,9 +1,17 @@
 package org.simple.clinic.drugs.search
 
-class DrugSearchModel {
+data class DrugSearchModel(
+    val searchQuery: String?
```
vinaysshenoy, line 4: Instead of a nullable string, should it be non-nullable and the default value be a blank string instead?

app/src/test/java/org/simple/clinic/drugs/search/DrugSearchUiRendererTest.kt
```diff
@@ -0,0 +1,29 @@
...
+
+  private val ui = mock<DrugSearchUi>()
+  private val uiRenderer = DrugSearchUiRenderer(ui)
+
+  private val defaultModel = DrugSearchModel.create()
+
+  @Test
+  fun `when search query is empty, then hide search results`() {
+    // given
+    val searchQuery = ""
```
vinaysshenoy, line 18: We can inline this variable, right?

app/src/test/java/org/simple/clinic/drugs/search/DrugSearchUiRendererTest.kt
```diff
@@ -26,4 +26,19 @@ class DrugSearchUiRendererTest {
     verify(ui).hideSearchResults()
     verifyNoMoreInteractions(ui)
   }
+
+  @Test
+  fun `when search query is not empty, then show search results`() {
+    // given
+    val searchQuery = "Amlodipine"
```
vinaysshenoy, line 33: Can we inline this variable?

app/src/androidTest/java/org/simple/clinic/drugs/search/DrugRepositoryAndroidTest.kt
```diff
@@ -85,6 +114,6 @@ class DrugRepositoryAndroidTest {
         .blockingFirst()
 
     // then
-    assertThat(expectedSearchResults).containsExactly(amlodipine10, amlodipine20).inOrder()
+    assertThat(expectedSearchResults).containsExactly(amlodipine10).inOrder()
```
vinaysshenoy, line 117: Can we also rename the `DrugRepository#search` method to make it explicit that this excludes drugs found in protocols?

app/src/main/java/org/simple/clinic/drugs/search/Drug.kt
```diff
@@ -52,7 +52,8 @@ data class Drug(
 
     @Query("""
       SELECT drugs.*, INSTR(lower(drugs.name), lower(:query)) namePosition FROM Drug drugs
-      WHERE namePosition > 0
+      LEFT JOIN ProtocolDrug PD ON PD.rxNormCode == drugs.rxNormCode
+      WHERE PD.uuid IS NULL AND namePosition > 0
```
vinaysshenoy, line 56: This will exclude all protocol drugs. Don't we only want to exclude protocol drugs from the _current_ facility protocol instead?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/drugs/search/Drug.kt
msasikanth: I did not know that protocols had a facility link 😅 . Will make that change.

app/src/test/java/org/simple/clinic/drugs/search/DrugSearchUiRendererTest.kt
msasikanth: Yep, I will use the function params names. I just wanted to have a name for the variable

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchModel.kt
msasikanth: Yeah, good point. We can do the `orEmpty` extension at the UI level rather than at `Update`. Currently, I am doing that in `Update` function to decide search or do nothing.

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchEvent.kt
msasikanth: I was trying a different approach for handling the nullable text changes. But we can filter them out at UI level, will make this change.

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchEffectHandler.kt
msasikanth: Yep ✅

app/src/main/java/org/simple/clinic/drugs/search/Drug.kt
```diff
@@ -48,5 +49,11 @@ data class Drug(
 
     @Query("SELECT COUNT(uuid) FROM Protocol")
     fun count(): Observable<Int>
+
+    @Query("""
+      SELECT drugs.*, INSTR(lower(drugs.name), lower(:query)) namePosition FROM Drug drugs
+      WHERE namePosition > 0
```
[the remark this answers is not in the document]
msasikanth: We are already using `INSTR` in the project and that seems to be the latest implementation in sqlite. So I think it makes sense to use this 🤔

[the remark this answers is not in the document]
vinaysshenoy: The only reason we used that was because we wanted the number position of the search query to contribute to the sort ordering. Here, we don't need to do that so we can use the `LIKE` clause which is simpler.

What do you think?

vinaysshenoy (changes requested):
Overall PR looks good, left one comment about the SQL query,

app/src/main/java/org/simple/clinic/drugs/search/Drug.kt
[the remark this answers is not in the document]
msasikanth: Makes sense. Making that change in this PR

vinaysshenoy (approved):

PR #2834, opened 2021-07-14
Load online lookup API if patient not found locally #2

Honey14:
https://app.clubhouse.io/simpledotorg/story/3794/load-online-lookup-api-if-patient-not-found-locally

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
+++ b/app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
@@ -712,16 +712,18 @@ class PatientRepository @Inject constructor(
     return database.patientDao().allPatientProfiles()
   }
 
-  fun saveCompleteMedicalRecord(medicalRecord: CompleteMedicalRecord) {
-    database.runInTransaction {
-      saveRecords(listOf(medicalRecord.patient))
-      if (medicalRecord.medicalHistory != null) {
-        database.medicalHistoryDao().save(medicalRecord.medicalHistory)
+  fun saveCompleteMedicalRecord(medicalRecords: List<CompleteMedicalRecord>) {
+    medicalRecords.forEach { medicalRecord ->
+      database.runInTransaction {
+        saveRecords(listOf(medicalRecord.patient))
+        if (medicalRecord.medicalHistory != null) {
+          database.medicalHistoryDao().save(medicalRecord.medicalHistory)
+        }
+        database.bloodPressureDao().save(medicalRecord.bloodPressures)
+        database.bloodSugarDao().save(medicalRecord.bloodSugars)
+        database.appointmentDao().save(medicalRecord.appointments)
+        database.prescriptionDao().save(medicalRecord.prescribedDrugs)
       }
-      database.bloodPressureDao().save(medicalRecord.bloodPressures)
-      database.bloodSugarDao().save(medicalRecord.bloodSugars)
-      database.appointmentDao().save(medicalRecord.appointments)
-      database.prescriptionDao().save(medicalRecord.prescribedDrugs)
     }
   }
 
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
@@ -38,7 +38,19 @@ class ScanSimpleIdEffectHandler @AssistedInject constructor(
       .addConsumer(OpenPatientSummary::class.java, ::openPatientSummary, schedulersProvider.ui())
       .addConsumer(OpenPatientSearch::class.java, ::openPatientSearch, schedulersProvider.ui())
       .addTransformer(OnlinePatientLookupWithIdentifier::class.java, onlinePatientLookupWithIdentifier())
+      .addTransformer(SaveCompleteMedicalRecords::class.java, saveCompleteMedicalRecords())
       .build()
+
+  private fun saveCompleteMedicalRecords(): ObservableTransformer<SaveCompleteMedicalRecords, ScanSimpleIdEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .doOnNext {
+            patientRepository.saveCompleteMedicalRecord(it.completeMedicalRecords)
+          }
+          .map { CompleteMedicalRecordsSaved(it.completeMedicalRecords) }
+    }
+  }
 
   private fun onlinePatientLookupWithIdentifier(): ObservableTransformer<OnlinePatientLookupWithIdentifier, ScanSimpleIdEvent> {
     return ObservableTransformer { effects ->
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
@@ -6,12 +6,12 @@ import com.spotify.mobius.Update
 import org.simple.clinic.INDIA_NHID_LENGTH
 import org.simple.clinic.mobius.dispatch
 import org.simple.clinic.mobius.next
+import org.simple.clinic.patient.CompleteMedicalRecord
 import org.simple.clinic.patient.Patient
 import org.simple.clinic.patient.PatientPrefillInfo
 import org.simple.clinic.patient.businessid.Identifier
 import org.simple.clinic.patient.businessid.Identifier.IdentifierType.BpPassport
 import org.simple.clinic.patient.businessid.Identifier.IdentifierType.IndiaNationalHealthId
-import org.simple.clinic.patient.onlinelookup.api.LookupPatientOnline
 import org.simple.clinic.patient.onlinelookup.api.LookupPatientOnline.Result.Found
 import org.simple.clinic.patient.onlinelookup.api.LookupPatientOnline.Result.NotFound
 import org.simple.clinic.patient.onlinelookup.api.LookupPatientOnline.Result.OtherError
@@ -40,6 +40,7 @@ class ScanSimpleIdUpdate @Inject constructor(
       is ScannedQRCodeJsonParsed -> scannedQRCodeParsed(model, event)
       InvalidQrCode -> next(model.notSearching().invalidQrCode())
       is OnlinePatientLookupWithIdentifierCompleted -> onlinePatientLookupWithIdentifierCompleted(model, event)
+      is CompleteMedicalRecordsSaved -> patientsFoundByOnlineLookup(event.completeMedicalRecords)
     }
   }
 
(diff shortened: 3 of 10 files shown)


app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandlerTest.kt
```diff
@@ -241,4 +241,26 @@ class ScanSimpleIdEffectHandlerTest {
     testCase.assertOutgoingEvents(OnlinePatientLookupWithIdentifierCompleted(results, identifier))
     verifyNoMoreInteractions(uiActions)
   }
+
+  @Test
+  fun `when save complete medical records effect is received, then save the patient medical records received`() {
```
janhavisinghh, line 246: Can you rename the test to.. `when save complete medical records effect is received, then save the patient medical records`

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
```diff
@@ -38,8 +38,25 @@ class ScanSimpleIdEffectHandler @AssistedInject constructor(
...
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .map(SaveCompleteMedicalRecords::completeMedicalRecords)
+          .map { completeMedicalRecords ->
+            completeMedicalRecords to completeMedicalRecords.map { medicalRecord ->
+              patientRepository.saveCompleteMedicalRecord(medicalRecord)
+            }
+          }
+          .map { (completeMedicalRecords, Unit)  ->
```
janhavisinghh, line 54: Can you format this line, please? It has an extra space between `(completeMedicalRecords, Unit)` and `->`

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUpdateTest.kt
```diff
@@ -276,4 +276,20 @@ class ScanSimpleIdUpdateTest {
             )
         )
   }
+
+  @Test
+  fun `When complete medical records are saved and one patient is found, open patient summary`() {
```
janhavisinghh, line 282: Not that important I think.. We follow `when this, then do that` format in update tests. Can we rename this test to follow the same?
`When complete medical records are saved and one patient is found, then open patient summary`

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -40,19 +41,26 @@ class ScanSimpleIdUpdate @Inject constructor(
...
 
   private fun onlinePatientLookupWithIdentifierCompleted(
       model: ScanSimpleIdModel,
       event: OnlinePatientLookupWithIdentifierCompleted
   ): Next<ScanSimpleIdModel, ScanSimpleIdEffect> {
-    return when(event.result) {
-      is NotFound, is OtherError -> next(model.notSearching(), OpenPatientSearch(event.identifier, null, model.patientPrefillInfo))
-      else -> noChange() // update the list of patients or patient received here
+    return when (event.result) {
+      is NotFound -> next(model.notSearching(), OpenPatientSearch(event.identifier, null, model.patientPrefillInfo))
```
janhavisinghh, line 53: Afaig aren't we still doing the same thing in `NotFound` and `OtherError`? Why did we separate these out?

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUpdateTest.kt
janhavisinghh: Also, we are changing two update branches here for `OnlinePatientLookupWithIdentifierCompleted` and `CompleteMedicalRecordsSaved` can we separate these out into two commits and write update tests for both the changes? Currently we've only written for when `CompleteMedicalRecordsSaved`

janhavisinghh (changes requested):
Left a few small comments PTAL.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
Honey14: Yes I had. I think while merging this went back. Will do this.

app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUpdateTest.kt
Honey14: * I think comma separation does the job for then but I'll still add this here. 
* No we're not doing two branches here. Completed and saved is written for the better context readability. There are tests already present for `OnlinePatientLookupWithIdentifierCompleted` in the previous PR. Can you PTAL?

janhavisinghh: We added `is Found -> next(model.notSearching(), SaveCompleteMedicalRecords(event.result.medicalRecords))` instead of `else -> noChange() // update the list of patients or patient received here` in `onlinePatientLookupWithIdentifierCompleted` which is a part of `is OnlinePatientLookupWithIdentifierCompleted` branch. Which is why I said we might need one more test as we are doing something new here.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
```diff
@@ -38,8 +38,25 @@ class ScanSimpleIdEffectHandler @AssistedInject constructor(
...
+  private fun saveCompleteMedicalRecords(): ObservableTransformer<SaveCompleteMedicalRecords, ScanSimpleIdEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .map(SaveCompleteMedicalRecords::completeMedicalRecords)
+          .map { completeMedicalRecords ->
+            completeMedicalRecords to completeMedicalRecords.map { medicalRecord ->
+              patientRepository.saveCompleteMedicalRecord(medicalRecord)
+            }
+          }
```
vinaysshenoy, line 53: `.map` is the wrong operator to use here since we are not converting the medical records to anything. What we should be using is a `.doOnNext` instead.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
```diff
@@ -38,8 +38,25 @@ class ScanSimpleIdEffectHandler @AssistedInject constructor(
...
 
+  private fun saveCompleteMedicalRecords(): ObservableTransformer<SaveCompleteMedicalRecords, ScanSimpleIdEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .map(SaveCompleteMedicalRecords::completeMedicalRecords)
+          .map { completeMedicalRecords ->
+            completeMedicalRecords to completeMedicalRecords.map { medicalRecord ->
+              patientRepository.saveCompleteMedicalRecord(medicalRecord)
+            }
```
vinaysshenoy, line 52: Maybe in another PR, we can change the signature of `PatientRepository#saveCompleteMedicalRecord` to accept a `List<CompleteMedicalRecord>` so we don't need this additional iteration. Would be more efficient as well since we can merge all the records in a single transaction.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -57,8 +57,13 @@ class ScanSimpleIdUpdate @Inject constructor(
   }
 
   private fun patientsFoundByOnlineLookup(completeMedicalRecords: List<CompleteMedicalRecord>): Next<ScanSimpleIdModel, ScanSimpleIdEffect> {
-    val patientId = completeMedicalRecords.first().patient.patientUuid
-    return dispatch(OpenPatientSummary(patientId))
+    val effect = if (completeMedicalRecords.isNotEmpty() && completeMedicalRecords.size > 1) {
```
vinaysshenoy, line 60: I prefer to move compound boolean logic like this into a separate local variable for better readability, i.e something like `atLeastOneMedicalRecordFound`.

vinaysshenoy: Also, the `.isNotEmpty()` check is redundant since the `size > 1` check should include that.

vinaysshenoy (changes requested):

janhavisinghh (approved):
LGTM

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -712,16 +712,18 @@ class PatientRepository @Inject constructor(
     return database.patientDao().allPatientProfiles()
   }
 
-  fun saveCompleteMedicalRecord(medicalRecord: CompleteMedicalRecord) {
-    database.runInTransaction {
-      saveRecords(listOf(medicalRecord.patient))
-      if (medicalRecord.medicalHistory != null) {
-        database.medicalHistoryDao().save(medicalRecord.medicalHistory)
+  fun saveCompleteMedicalRecord(medicalRecords: List<CompleteMedicalRecord>) {
+    medicalRecords.map { medicalRecord ->
```
vinaysshenoy, line 716: This should be a `.forEach` instead of a `.map` since none of the records are being transformed in any way.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdEffectHandler.kt
```diff
@@ -38,8 +38,20 @@ class ScanSimpleIdEffectHandler @AssistedInject constructor(
...
       .addTransformer(OnlinePatientLookupWithIdentifier::class.java, onlinePatientLookupWithIdentifier())
+      .addTransformer(SaveCompleteMedicalRecords::class.java, saveCompleteMedicalRecords())
       .build()
 
+  private fun saveCompleteMedicalRecords(): ObservableTransformer<SaveCompleteMedicalRecords, ScanSimpleIdEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .map {
+            patientRepository.saveCompleteMedicalRecord(it.completeMedicalRecords)
```
vinaysshenoy, line 49: The saving of medical records should be done in a `.doOnNext` block outside of the `.map` block since the saving of these records does not factor into the construction of the `CompleteMedicalRecordsSaved` event.

Honey14: Ah I see what you meant now. Updating this

janhavisinghh (approved):

janhavisinghh (approved):

vinaysshenoy (approved):

PR #2835, opened 2021-07-16
Implement `DrugsSearchScreen` - Final

msasikanth:
https://app.clubhouse.io/simpledotorg/story/2899/implement-drugssearchscreen

Enable `CustomDrugSearchScreen` feature to try this out. Once that is enabled, click on "Add another medicine" in edit medicines screen.


app/src/main/java/org/simple/clinic/drugs/search/DrugSearchListItem.kt
```diff
@@ -0,0 +1,89 @@
...
+      return buildString {
+        append(drug.name)
+
+        if (!drug.dosage.isNullOrBlank()) {
+          append(", ${drug.dosage}")
+        }
+
+        if (drug.frequency != null) {
+          append(", ${drug.frequency}")
+        }
```
vinaysshenoy, line 63: Can we clarify if appending commas to display strings is applicable to all the languages that we support?

app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
```diff
@@ -0,0 +1,149 @@
...
+  private fun drugSearchLoadStateListener(combinedLoadStates: CombinedLoadStates) {
+    val isLoading = combinedLoadStates.refresh is LoadState.Loading
+    progressIndicator.visibleOrGone(isLoading)
+  }
+
+  private fun searchQueryChanges(): Observable<UiEvent> {
+    return searchQueryEditText
+        .textChanges()
+        .skipInitialValue()
+        .debounce(500, TimeUnit.MILLISECONDS)
```
vinaysshenoy, line 132: Hmm, what do you think of adding a `.debounce(java.time.Duration)` extension to make this kind of thing more readable?

app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
```diff
@@ -0,0 +1,149 @@
...
+        .map { searchQuery ->
+          val trimmedSearchQuery = searchQuery
+              .trim()
+              .toString()
+
+          SearchQueryChanged(trimmedSearchQuery)
+        }
```
vinaysshenoy, line 139: This could be cleaner  with an additional `.map` operation:

```kotlin
.map(String::trim)
.map(::SearchQueryChanged)
```

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchListItem.kt
```diff
@@ -75,6 +99,7 @@ sealed class DrugSearchListItem : PagingItemAdapter.Item<DrugSearchListItem.Drug
       return when {
         oldItem is DrugSearchResult && newItem is DrugSearchResult -> oldItem.drug.id == newItem.drug.id
         oldItem is Divider && newItem is Divider -> false
+        oldItem is NewCustomDrug && newItem is NewCustomDrug -> true
```
vinaysshenoy, line 137: Do we need to compare the names of the new custom drug in this case to see if they refer to the same item?

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchListItem.kt
```diff
@@ -0,0 +1,149 @@
...
+            } else if (oldItem == null && newItem == null) {
+              NewCustomDrug(name = searchQuery)
+            } else if (oldItem == null && newItem != null) {
+              TopCornerCapItem
+            } else if (oldItem != null && newItem == null) {
+              BottomCornerCapItem
+            } else {
+              null
+            }
+          }
```
vinaysshenoy, line 37: This logic seems quite complex. Can we add a test to verify that the items are being generated correctly?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchListItem.kt
msasikanth: Yeah, I thought about moving the "," into strings for translation. But currently the languages we support don't have a different comma notation. So hard coded it, but we can move it to strings just to be safe.

app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
msasikanth: Sure, personally I think this is more readable. But using `Duration` will be consistent with other usages.

app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
msasikanth: This wouldn't work, we require another `map` function to convert charsequence into a string. Making that change.

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchListItem.kt
vinaysshenoy: It's not about whether they have a comma notation or not, it's about whether the notion of a comma exists in those languages as well. I feel this is something that should be left up to the localizers to decide.

app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
vinaysshenoy: More than readable, it is also safer since changing the time unit to something like `SECONDS` without changing the actual value will cause issues.

app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
msasikanth: Finally went with this approach
```kt
.map(CharSequence::trim)
.map { searchQuery ->
  SearchQueryChanged(searchQuery.toString())
}
```

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchListItem.kt
msasikanth: The UI transition were not smooth when the name changes. So left it as true to prevent the item from updating, we anyway get the search query from the edit text directly. So we don't need to worry about item data not getting updated.

app/src/main/res/values/strings.xml
```diff
@@ -813,5 +813,6 @@
 
   <!-- Custom drugs search -->
   <string name="customdrugssearch_hint">Enter medicine name</string>
+  <string name="customdrugssearch_item_name_separator">, </string>
```
vinaysshenoy, line 816: Just wondering if it would be better for the localisation team to have two different string resources, `drug_with_dosage` and `drug_with_dosage_and_frequency`, so they can make a better decision when adding the translations.

msasikanth: I initially went with that, but dosage and frequency is nullable. So sometimes there is a white space before the separator. That's why I switch to a string builder.

vinaysshenoy (approved):

PR #2836, opened 2021-07-16
Fix memory leaks in `HomeScreen`

msasikanth:
https://app.clubhouse.io/simpledotorg/story/4181/fix-memory-leaks-in-homescreen


Honey14 (approved):
Looks good

PR #2837, opened 2021-07-16
Implement `transferred to` and `registered at` patient facility in `ContactPatientBottomSheet`

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/4185/implement-transferred-to-and-registered-at-patient-facility-in-contactpatientbottomsheet

### DIFF
--- a/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientEffectHandler.kt
@@ -1,11 +1,13 @@
 package org.simple.clinic.contactpatient
 
 import com.spotify.mobius.rx2.RxMobius
+import dagger.Lazy
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
 import io.reactivex.ObservableTransformer
 import io.reactivex.Scheduler
+import org.simple.clinic.facility.Facility
 import org.simple.clinic.overdue.AppointmentRepository
 import org.simple.clinic.patient.PatientRepository
 import org.simple.clinic.phone.Dialer
@@ -19,6 +21,7 @@ class ContactPatientEffectHandler @AssistedInject constructor(
     private val appointmentRepository: AppointmentRepository,
     private val clock: UserClock,
     private val schedulers: SchedulersProvider,
+    private val currentFacility: Lazy<Facility>,
     @Assisted private val uiActions: ContactPatientUiActions
 ) {
 
@@ -41,7 +44,17 @@ class ContactPatientEffectHandler @AssistedInject constructor(
         .addConsumer(ShowManualDatePicker::class.java, { uiActions.showManualDatePicker(it.preselectedDate, it.datePickerBounds) }, schedulers.ui())
         .addTransformer(SetReminderForAppointment::class.java, setReminderForAppointment(schedulers.io()))
         .addConsumer(OpenRemoveOverdueAppointmentScreen::class.java, ::openRemoveOverdueAppointmentScreen, schedulers.ui())
+        .addTransformer(LoadCurrentFacility::class.java, loadCurrentFacility())
         .build()
+  }
+
+  private fun loadCurrentFacility(): ObservableTransformer<LoadCurrentFacility, ContactPatientEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulers.io())
+          .map { currentFacility.get() }
+          .map(::CurrentFacilityLoaded)
+    }
   }
 
   private fun openRemoveOverdueAppointmentScreen(effect: OpenRemoveOverdueAppointmentScreen) {
--- a/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientModel.kt
+++ b/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientModel.kt
@@ -103,4 +103,8 @@ data class ContactPatientModel(
   fun contactPatientInfoLoading(): ContactPatientModel {
     return copy(contactPatientInfoProgressState = IN_PROGRESS)
   }
+
+  fun currentFacilityLoaded(currentFacility: Facility): ContactPatientModel {
+    return copy(currentFacility = currentFacility)
+  }
 }
--- a/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientModel.kt
+++ b/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientModel.kt
@@ -4,6 +4,7 @@ import android.os.Parcelable
 import kotlinx.parcelize.Parcelize
 import org.simple.clinic.contactpatient.ContactPatientInfoProgressState.DONE
 import org.simple.clinic.contactpatient.ContactPatientInfoProgressState.IN_PROGRESS
+import org.simple.clinic.facility.Facility
 import org.simple.clinic.home.overdue.OverdueAppointment
 import org.simple.clinic.overdue.AppointmentConfig
 import org.simple.clinic.overdue.PotentialAppointmentDate
@@ -26,7 +27,8 @@ data class ContactPatientModel(
     val selectedAppointmentDate: LocalDate,
     val selectedRemoveAppointmentReason: RemoveAppointmentReason?,
     val contactPatientInfoProgressState: ContactPatientInfoProgressState?,
-    val overdueListChangesFeatureEnabled: Boolean
+    val overdueListChangesFeatureEnabled: Boolean,
+    val currentFacility: Facility? = null
 ) : Parcelable {
 
   companion object {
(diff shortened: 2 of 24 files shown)


app/src/main/java/org/simple/clinic/contactpatient/ContactPatientEffectHandler.kt
```diff
@@ -19,6 +21,7 @@ class ContactPatientEffectHandler @AssistedInject constructor(
     private val appointmentRepository: AppointmentRepository,
     private val clock: UserClock,
     private val schedulers: SchedulersProvider,
+    private val currentFacilityStream: Observable<Facility>,
```
msasikanth, line 24: We don't have to listen to facility changes here. So we can just inject `dagger.Lazy<Facility>` here.

app/src/test/java/org/simple/clinic/contactpatient/ContactPatientInitTest.kt
```diff
@@ -37,7 +37,7 @@ class ContactPatientInitTest {
         .whenInit(defaultModel)
         .then(assertThatFirst(
             hasModel(defaultModel.contactPatientInfoLoading()),
-            hasEffects(LoadPatientProfile(patientUuid), LoadLatestOverdueAppointment(patientUuid))
+            hasEffects(LoadPatientProfile(patientUuid), LoadLatestOverdueAppointment(patientUuid), LoadCurrentFacility)
```
msasikanth, line 41: Can we update the test name to indicate that we are testing loading current facility as well, please?

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientModel.kt
```diff
@@ -103,4 +103,8 @@ data class ContactPatientModel(
   fun contactPatientInfoLoading(): ContactPatientModel {
```
msasikanth, line 106: Can we update the commit message to "When current facility is loaded, then update the model"? The current commit message doesn't properly indicate that.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientModel.kt
```diff
@@ -67,6 +67,9 @@ data class ContactPatientModel(
   val hasLoadedAppointment: Boolean
     get() = appointment != null
 
+  val hasPatientRegistrationIDInOverdueAppointment: Boolean
```
msasikanth, line 70: We can simply call is `hasRegisteredFacility`

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
```diff
@@ -85,6 +91,12 @@ class ContactPatientUiRenderer(
     ui.switchToCallPatientView()
   }
 
+  private fun renderPatientFacilityLabel(patientRegisteredFacilityID: UUID, currentFacility: Facility) {
+    if (patientRegisteredFacilityID == currentFacility.uuid) {
```
msasikanth, line 95: Here we should compare assigned facility no?

msasikanth: We can also move this condition as computed variable in the model, considering we have both facility and overdue appointment there.

app/src/test/java/org/simple/clinic/contactpatient/CallPatientUiRendererTest.kt
```diff
@@ -371,10 +372,44 @@ class CallPatientUiRendererTest {
     verifyNoMoreInteractions(ui)
   }
 
+  @Test
+  fun `display registered at facility label text when patient's registered facility is the same as user's current facility`() {
```
msasikanth, line 376: As far as I remember, we need to show the "Registered label" when the assigned facility ID is same as the current facility ID. If not, then we show transferred from label. In both scenarios we show the registered facility name, just the label changes.

app/src/test/java/org/simple/clinic/contactpatient/CallPatientUiRendererTest.kt
```diff
@@ -405,6 +405,39 @@ class CallPatientUiRendererTest {
     verifyNoMoreInteractions(ui)
   }
 
+  @Test
+  fun `display transferred from facility label text when patient's registered facility is not the same as user's current facility`() {
```
msasikanth, line 409: Same comment as before, we need to compare assigned facility ID.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
```diff
@@ -70,6 +72,10 @@ class ContactPatientUiRenderer(
       renderPatientProfile(model.patientProfile!!, model.appointment!!)
     }
 
+    if (model.hasPatientRegistrationIDInOverdueAppointment && model.hasCurrentFacility) {
```
msasikanth, line 75: When this is false, i.e., when the user doesn't have a registered facility. Should we specifically call a hide method for the label and name text? 🤔

msasikanth (changes requested):
Left few comments PTAL

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
janhavisinghh: As this is a case that will never really happen.. Do we really need to handle this as well?

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
janhavisinghh: A patient will only be shown in a facility when they are assigned to that facility, right? So the assigned facility and current user facility will never be different.. Also we load overdue appointments based on the current user facility..

app/src/test/java/org/simple/clinic/contactpatient/CallPatientUiRendererTest.kt
janhavisinghh: Yes this update branch only changes the label text name

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
janhavisinghh: So a patient will never be in the overdue appointment list if their assigned facility is not the current user facility. Also we are displaying the registered facility in the text view.

msasikanth: Good point, let's go with registered facility. I have left comment for Mahima to update PRD as well. Let's move this as a computed property then.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
msasikanth: Hmm, true. I guess I we can leave this for now, but let's keep an eye out for any potential issues related to this.

app/src/test/java/org/simple/clinic/contactpatient/ContactPatientEffectHandlerTest.kt
```diff
@@ -33,12 +34,22 @@ class ContactPatientEffectHandlerTest {
...
+  )
+
   private val effectHandler = ContactPatientEffectHandler(
       patientRepository = patientRepository,
       appointmentRepository = appointmentRepository,
       clock = clock,
       schedulers = TrampolineSchedulersProvider(),
-      uiActions = uiActions
+      uiActions = uiActions,
+      currentFacility = { facility }
```
msasikanth, line 52: Ideally this change should be done in the commit where we are changing `Init` block

msasikanth (changes requested):

app/src/test/java/org/simple/clinic/contactpatient/ContactPatientInitTest.kt
```diff
@@ -43,9 +44,19 @@ class ContactPatientInitTest {
...
+        facilityConfig = FacilityConfig(
+            diabetesManagementEnabled = true,
+            teleconsultationEnabled = false
+        )
+    )
+
     val model = defaultModel()
         .patientProfileLoaded(TestData.patientProfile(patientUuid = patientUuid))
         .overdueAppointmentLoaded(Optional.empty())
+        .currentFacilityLoaded(currentFacility = facility)
```
msasikanth, line 59: Ideally this change should be done in the commit where we are changing Init block

msasikanth: Just an FYI, we haven't added this method yet before this commit. So this commit alone wouldn't compile. But since we are squashing commits, it's fine now. But something to be careful about.

msasikanth (approved):

PR #2839, opened 2021-07-16
Remove the `DAILY` enum value from `SyncGroup`

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/4148/remove-separation-between-daily-and-frequent-syncs

### DIFF
--- a/app/src/main/java/org/simple/clinic/sync/DataSync.kt
+++ b/app/src/main/java/org/simple/clinic/sync/DataSync.kt
@@ -63,16 +63,6 @@ class DataSync(
 
   private val syncErrors = PublishSubject.create<ResolvedError>()
 
-  private fun allSyncs(): Single<List<SyncResult>> {
-    val syncAllGroups = SyncGroup
-        .values()
-        .map { executeAllSyncs() }
-
-    return Single
-        .merge(syncAllGroups)
-        .reduce(listOf(), { list, results -> list + results })
-  }
-
   private fun executeAllSyncs(): Single<List<SyncResult>> {
     return Single
         .fromCallable { userSession.loggedInUserImmediate().toOptional() }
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/simple/clinic/sync/DataSync.kt
```diff
@@ -66,20 +66,18 @@ class DataSync(
...
 
-  private fun syncsForGroup(syncGroup: SyncGroup): Single<List<SyncResult>> {
-    val syncsInGroup = modelSyncs.filter { it.syncConfig().syncGroup == syncGroup }
-
+  private fun syncsForGroup(): Single<List<SyncResult>> {
     return Single
         .fromCallable { userSession.loggedInUserImmediate().toOptional() }
         .subscribeOn(schedulersProvider.io())
-        .compose(filterSyncsThatRequireAuthentication(syncsInGroup))
+        .compose(filterSyncsThatRequireAuthentication(modelSyncs))
```
Honey14, line 70: the `filterSyncsThatRequireAuthentication` function parameter name is not changed in the body. Can you update?

app/src/test/java/org/simple/clinic/sync/DataSyncTest.kt
```diff
@@ -38,26 +38,16 @@ class DataSyncTest {
...
-      syncGroup = SyncGroup.DAILY
-  )
-
   @Test
-  fun `when syncing everything, all the model syncs should be invoked`() {
+  fun `when syncing, all the model syncs should be invoked`() {
     val modelSync1 = FakeModelSync(
         _name = "sync1",
-        config = dailySyncConfig
+        config = frequentSyncConfig
```
Honey14, line 45: So we'll be removing `SyncInterval` and `SyncGroup` from the FakeModelSync when we finish removing frequent type as well, right?

Honey14 (changes requested):
Looks good overall. Requested one change. PTAL

app/src/test/java/org/simple/clinic/sync/DataSyncTest.kt
vinaysshenoy: Yup, that is correct.

app/src/main/java/org/simple/clinic/sync/DataSync.kt
vinaysshenoy: Good catch, fixing this.

Honey14 (approved):
Looks good

PR #2840, opened 2021-07-16
Don't show progress if overdue list already has items

msasikanth:
- Don't show progress if overdue adapter already has items
- Update CHANGELOG


msasikanth:
This ensures that the we don't show progress every time user swipes to overdue page even if the items are already present. The loading will be done silently.

vinaysshenoy (approved):

PR #2842, opened 2021-07-19
Use the `Room` @Relation` annotation for representing `PatientProfile`

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/4195/use-room-s-relation-annotation-for-patientprofile

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/Patient.kt
+++ b/app/src/main/java/org/simple/clinic/patient/Patient.kt
@@ -261,21 +261,6 @@ data class Patient(
         offset: Int
     ): List<PatientProfile>
 
-    protected data class PatientQueryModel(
-
-        @Embedded(prefix = "patient_")
-        val patient: Patient,
-
-        @Embedded(prefix = "addr_")
-        val address: PatientAddress,
-
-        @Embedded(prefix = "phone_")
-        val phoneNumber: PatientPhoneNumber?,
-
-        @Embedded(prefix = "businessid_")
-        val businessId: BusinessId?
-    )
-
     @Query("""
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -258,11 +256,12 @@ data class Patient(
...
-      return loadPatientQueryModelsForPatientUuid(patientUuid)
-          .map { queryModelsToPatientProfiles(it) }
-          .toObservable()
-    }
+    @Transaction
+    @Query("""
+      SELECT * FROM Patient
+      WHERE uuid = :patientUuid
+    """)
+    abstract fun patientProfile(patientUuid: UUID): Observable<List<PatientProfile>>
```
msasikanth, line 239: Couldn't we just return `Observable<PatientProfile?>` here instead of nullable? Reason I ask is, we are returning `PatientProfile?` for `patientProfileImmediate` in later commits. So I want to check if we need to keep it consistent across both the usages.

msasikanth (commented):
LGTM. Left one comment PTAL

app/src/main/java/org/simple/clinic/patient/Patient.kt
vinaysshenoy: Yeah, we could. I would prefer to make this change in another PR because this would require changing downstream consumers of this method which is not related to the change in this PR.

vinaysshenoy:
@msasikanth Can you approve this PR again, please?

msasikanth (approved):

PR #2848, opened 2021-07-20
Use full date formatter when rendering last visited date in contact patient sheet

msasikanth:
- Use full date formatter in contact patient sheet
- Update CHANGELOG


Honey14 (approved):
LGTM

PR #2851, opened 2021-07-20
Bump AppCompat to v1.3.0

msasikanth:
- Bump AppCompat version to v1.3.0
- Use `app:tint` instead of platform tint
- Update CHANGELOG

### DIFF
--- a/router/build.gradle.kts
+++ b/router/build.gradle.kts
@@ -22,7 +22,7 @@ android {
 }
 
 dependencies {
-  implementation("androidx.appcompat:appcompat:${versions.supportLib}")
+  implementation("androidx.appcompat:appcompat:${versions.appcompat}")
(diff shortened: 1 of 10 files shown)


router/build.gradle.kts
```diff
@@ -22,7 +22,7 @@ android {
 }
 
 dependencies {
-  implementation("androidx.appcompat:appcompat:${versions.supportLib}")
+  implementation("androidx.appcompat:appcompat:${versions.appcompat}")
```
janhavisinghh, line 25: Why were we using supportLib before?

janhavisinghh (approved):

router/build.gradle.kts
msasikanth: No particular reason, it was there before and we must have forgot to switch to `appcompat` version

PR #2857, opened 2021-07-21
Use `MobiusLoopViewModel` for sending view effects that are lifecycle aware

msasikanth:
- Add view effects generic type for `BaseScreen`
- Pass `Any` as view effect type when extending `BaseScreen`
- Add `MobiusLoopViewModel` in `BaseScreen`
- Remove Mobius controller usage from `BaseScreen`
- Forward UI events to ViewModel
- When ViewModel model updates, then render UI
- Change `BaseScreen#createEffectHandler` signature to accept view effect consumer
- Add `NoopViewEffectHandler`
- Handle view effects in `BaseScreen`
- Update CHANGELOG

**Story**: https://app.clubhouse.io/simpledotorg/story/4074/use-mobiusloopviewmodel-for-sending-view-effects-that-are-lifecycle-aware


app/src/main/java/org/simple/clinic/bloodsugar/history/BloodSugarHistoryScreen.kt
```diff
@@ -59,7 +59,8 @@ class BloodSugarHistoryScreen : BaseScreen<
     ScreenBloodSugarHistoryBinding,
     BloodSugarHistoryScreenModel,
     BloodSugarHistoryScreenEvent,
-    BloodSugarHistoryScreenEffect>(), BloodSugarHistoryScreenUi, BloodSugarHistoryScreenUiActions {
+    BloodSugarHistoryScreenEffect,
+    Any>(), BloodSugarHistoryScreenUi, BloodSugarHistoryScreenUiActions {
```
vinaysshenoy, line 64: Instead of `Any`, won't `None` or `Unit` be a better option here since there are no effects?

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
```diff
@@ -77,6 +84,23 @@ abstract class BaseScreen<K : ScreenKey, B : ViewBinding, M : Parcelable, E, F,
...
+    _viewModel = ViewModelProvider(viewModelStore, object : ViewModelProvider.Factory {
+      private fun loop(viewEffectsConsumer: Consumer<V>) = RxMobius
+          .loop(createUpdate(), createEffectHandler())
+          .eventSources(additionalEventSources())
+
+      @Suppress("UNCHECKED_CAST")
+      override fun <T : ViewModel?> create(modelClass: Class<T>): T {
+        return MobiusLoopViewModel.create<M, E, F, V>(
+            ::loop,
+            startModel ?: defaultModel(),
```
vinaysshenoy, line 89: Can we move the `?: defaultModel()` declaration to where we are reading the model from the saved state. This would be a more readable approach since we are defining the model to start the loop in one statement.

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
```diff
@@ -92,6 +92,8 @@ abstract class BaseScreen<K : ScreenKey, B : ViewBinding, M : Parcelable, E, F,
     disposable.add(events().subscribe { event ->
       _viewModel.dispatchEvent(event!!)
     })
+    
+    _viewModel.models.observe(viewLifecycleOwner, uiRenderer()::render)
```
vinaysshenoy, line 95: When it comes to method references, I think this pattern of calling a function and then using a reference on that seems a little unreadable. In this specific case, I'd personally prefer to extract the creation of the UI renderer to a separate variable.

```kotlin
val renderer = uiRenderer()
_viewModel.models.observe(viewLifecycleOwner, renderer::render)
```

What do you think?

vinaysshenoy, line 95: Won't this current way of instantiating the renderer create a new instance for every model update?

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
```diff
@@ -28,6 +33,8 @@ abstract class BaseScreen<K : ScreenKey, B : ViewBinding, M : Parcelable, E, F,
     private const val KEY_MODEL = "org.simple.clinic.navigation.v2.fragments.BaseScreen.KEY_MODEL"
   }
 
+  private lateinit var _viewModel: MobiusLoopViewModel<M, E, F, V>
```
vinaysshenoy, line 36: Is there a reason we are prefixing this with `_`?

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
```diff
@@ -76,7 +76,7 @@ abstract class BaseScreen<K : ScreenKey, B : ViewBinding, M : Parcelable, E, F,
 
     _viewModel = ViewModelProvider(viewModelStore, object : ViewModelProvider.Factory {
       private fun loop(viewEffectsConsumer: Consumer<V>) = RxMobius
-          .loop(createUpdate(), createEffectHandler())
+          .loop(createUpdate(), createEffectHandler(viewEffectsConsumer))
```
vinaysshenoy, line 82: As I understand it, we'll create a new kind of effect for representing view effects and instead of invoking UI actions in the current effect handler, we will pass the view effects to the view effect handler and invoke the UI actions from there instead?

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
```diff
@@ -94,6 +97,14 @@ abstract class BaseScreen<K : ScreenKey, B : ViewBinding, M : Parcelable, E, F,
     })
 
     _viewModel.models.observe(viewLifecycleOwner, uiRenderer()::render)
+
+    _viewModel.viewEffects.setObserver(
+        viewLifecycleOwner,
+        viewEffectHandler()::handle,
+        { viewEffects ->
+          viewEffects.forEach(viewEffectHandler()::handle)
+        }
```
vinaysshenoy, line 102: Can we extract this to a named function (like `runQueuedViewEffects` or something) so that the intention is more clear?

vinaysshenoy: If we can use Kotlin named parameters, that should work as well.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
msasikanth: My initial plan was to use this as a private reference and then expose the `viewModel` using a getter similar to binding. So I left it in case we decided to do that in future. But we can remove this if you think we don't want to expose the view model.

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
msasikanth: OMG yes, for some reason I thought this was a reference 😂

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
msasikanth: Yep, instead of calling `uiActions.methodName` in effect handler. We will just call `viewEffectsConsumer.accept` and send a view effect to Mobius loop. And that will be passed to view effects handler for triggering the UI actions.

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
msasikanth: This is a Java method, so we cannot use named params. Will extract this into a named function.

app/src/main/java/org/simple/clinic/bloodsugar/history/BloodSugarHistoryScreen.kt
msasikanth: I tried using `Nothing`, but it showed some compile errors trying to infer type. Let me give `Unit` a try.

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
msasikanth: Updated it. Moved the handling into separate methods, we have to handle 2 types of view effects. One is live view effect, when the app is in foreground. When app is in background Mobius will add the view effects in to queue and trigger them once app is in foreground again. That will happen in paused view effects observer.
image

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
msasikanth: I renamed it, we can make this change when needed

app/src/main/java/org/simple/clinic/navigation/v2/fragments/BaseScreen.kt
vinaysshenoy: Hmm, can't we just delegate to `runLiveViewEffect` for the paused effects instead of creating a new method?

```kotlin
pausedViewEffects.forEach { runLiveViewEffect(viewEffectHandler, it) }
```

vinaysshenoy: Nitpick: `runLiveViewEffects` should be singular since we will only execute one effect.

msasikanth: I didn't want to use `runLiveViewEffect` inside the pause view effects loop, so created a new method. We can just go back to using the `viewEffectHandler` method references directly and have named params for the observer result like we do now?

msasikanth: This
```kt
val viewEffectHandler = viewEffectHandler()
_viewModel.viewEffects.setObserver(
    viewLifecycleOwner,
    { liveViewEffect -> viewEffectHandler.handle(liveViewEffect) },
    { pausedViewEffects -> pausedViewEffects.forEach(viewEffectHandler::handle) }
)
```

vinaysshenoy: I like this approach.

app/src/main/java/org/simple/clinic/bloodsugar/history/BloodSugarHistoryScreen.kt
vinaysshenoy: Did you make this change? I don't see it. 🤔

msasikanth: Forgot to make this change, pushed it now

vinaysshenoy (approved):

PR #2868, opened 2021-07-23
Fix race condition in `onScreenResults` in`ScheduleAppointmentSheet`

Honey14:
https://app.clubhouse.io/simpledotorg/story/3964/fix-race-condition-for-screen-results
This is a possible solution, since we cannot reproduce this, we'll keep it open in clubhouse for a while just in case. If we don't receive any sentry problem like this again then we can remove it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
   
 ### Fixes
 - Fix `ContactPatientBottomSheet` UI spacing and styling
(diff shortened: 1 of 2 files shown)


msasikanth:
While this solution helps with checking the type for this screen, it doesn't actually fix the underlying problem. For example we have other screens where this issue happens, so we should try to identify the core issue and resolve that.

PR #2869, opened 2021-07-26
Remove the `SyncGroup` enum

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/4148/remove-separation-between-daily-and-frequent-syncs

### DIFF
--- a/app/src/main/java/org/simple/clinic/sync/SyncScheduler.kt
+++ b/app/src/main/java/org/simple/clinic/sync/SyncScheduler.kt
@@ -21,7 +21,7 @@ class SyncScheduler @Inject constructor(
   fun schedule(): Completable {
     return Single.just(syncConfig)
         .map { config -> createWorkRequest(config.syncInterval) }
-        .doOnSuccess { request -> workManager.enqueueUniquePeriodicWork("sync-patient-resources", REPLACE, request) }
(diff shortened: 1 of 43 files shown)


app/src/main/java/org/simple/clinic/sync/SyncScheduler.kt
```diff
@@ -21,7 +21,7 @@ class SyncScheduler @Inject constructor(
   fun schedule(): Completable {
     return Single.just(syncConfig)
         .map { config -> createWorkRequest(config.syncInterval) }
-        .doOnSuccess { request -> workManager.enqueueUniquePeriodicWork("sync-patient-resources", REPLACE, request) }
+        .doOnSuccess { request -> workManager.enqueueUniquePeriodicWork(syncConfig.name, REPLACE, request) }
```
msasikanth, line 24: If we are going to use the `SyncConfig#name` as only for the unique work name, can't we just a use a constant defined in `SyncConfig` instead of as a class param 🤔

msasikanth (commented):
LGTM. Left one comment PTAL

app/src/main/java/org/simple/clinic/sync/SyncScheduler.kt
vinaysshenoy: We can, but we shouldn't. Each `SyncConfig` is mapped to one kind of unique work, each of which requires a name. So I made it a property.

msasikanth (approved):

PR #2870, opened 2021-07-26
Support GZIP compression for HTTP request bodies

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/4078/support-gzip-for-http-request-bodies

### DIFF
--- a/app/src/main/java/org/simple/clinic/di/network/CompressRequestInterceptor.kt
+++ b/app/src/main/java/org/simple/clinic/di/network/CompressRequestInterceptor.kt
@@ -3,6 +3,8 @@ package org.simple.clinic.di.network
 import okhttp3.Interceptor
 import okhttp3.Request
 import okhttp3.Response
+import org.simple.clinic.feature.Feature.HttpRequestBodyCompression
+import org.simple.clinic.feature.Features
 import javax.inject.Inject
 
 /**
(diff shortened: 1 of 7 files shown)


app/src/test/java/org/simple/clinic/di/network/CompressRequestInterceptorTest.kt
```diff
@@ -0,0 +1,4 @@
+package org.simple.clinic.di.network
+
+class CompressRequestInterceptorTest {
```
janhavisinghh, line 3: Shouldn't this be a part of the next commit?

app/src/main/java/org/simple/clinic/di/network/CompressRequestInterceptor.kt
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Copied from https://github.com/square/okhttp/blob/8bb58332dbacdd2bb9beb8e5c813f22508ebeb27/samples/guide/src/main/java/okhttp3/recipes/RequestBodyCompression.java
+ **/
+class CompressRequestInterceptor @Inject constructor(
+    private val requestCompression: RequestCompression
+) : Interceptor {
+
+  override fun intercept(chain: Interceptor.Chain): Response {
+    val originalRequest = chain.request()
+
```
janhavisinghh, line 31: We have not added any tests for this class and we have an empty test class for this. Is there a reason for this?

janhavisinghh (changes requested):
LGTM. Left a few comments PTAL?

app/src/main/java/org/simple/clinic/di/network/CompressRequestInterceptor.kt
vinaysshenoy: Ah, good catch. I wanted to test the interceptor, but that required a bit of setup since it was dependent on some OkHttp specific components, which was irrelevant to this PR. So I instead used the `RequestCompression` to implement the relevant bits and tested that instead. But I think I forgot to delete the empty test file I created for testing the interceptor.

Will delete this.

app/src/test/java/org/simple/clinic/di/network/CompressRequestInterceptorTest.kt
vinaysshenoy: This file being checked in was an accident, as described [here](https://github.com/simpledotorg/simple-android/pull/2870#discussion_r677313713). Will remove it.

app/src/main/java/org/simple/clinic/di/network/CompressRequestInterceptor.kt
vinaysshenoy: Done.

janhavisinghh (approved):
LGTM

PR #2873, opened 2021-07-26
Update Simple video view to be translatable

msasikanth:
https://app.clubhouse.io/simpledotorg/story/4285/update-simple-video-view-to-be-translatable


msasikanth:
Check with Mahima about configuring video duration as well, will move this to remote config for dynamically adding new vide durations.

vinaysshenoy (approved):

PR #2883, opened 2021-07-27
Restart `TheActivity` after changing language

msasikanth:
https://app.clubhouse.io/simpledotorg/story/4286/restart-theactivity-after-changing-language


Honey14 (approved):
LGTM

PR #2887, opened 2021-07-28
Don't push to new screen if the top screen is same as the new screen

msasikanth:
- Don't push to new screen if the top screen is same as the new screen
- Update CHANGELOG

Fixes [issue](https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1627448041022400)


janhavisinghh (approved):
LGTM

PR #2903, opened 2021-07-30
Run chmod for git hooks only if the OS is Linux or MacOS

msasikanth:


buildSrc/src/main/kotlin/plugins/git/install-hooks.gradle.kts
```diff
@@ -1,9 +1,14 @@
...
+import org.gradle.internal.os.OperatingSystem
+
 // Git hooks have to be manually copied and made executable. This task automates that.
 val gitExecutableHooks: Task = tasks.create("gitExecutableHooks") {
   doLast {
-    Runtime.getRuntime().exec("chmod -R +x .git/hooks/")
+    val operatingSystem = OperatingSystem.current()
+    if (operatingSystem.isMacOsX || operatingSystem.isLinux) {
+      Runtime.getRuntime().exec("chmod -R +x .git/hooks/")
+    }
```
vinaysshenoy, line 11: I'm wondering if we should be setting the executable flags at all automatically since it can be a security risk. Since setting the executable flag is both an OS specific task as well as something that only needs to be done _once_ (since the flag is set on the file), I think we can remove this altogether and leave it up to the developers to set the executable flag.

Even better, can we check if the executable bit on the source files are retained when they are copied into `.git/hooks`? That should take care of this problem for us.

vinaysshenoy (changes requested):

buildSrc/src/main/kotlin/plugins/git/install-hooks.gradle.kts
msasikanth: Yeah fair enough, closing this PR then.

PR #2904, opened 2021-07-30
Support contact patient sheet with no appointment- #2

Honey14:
https://app.clubhouse.io/simpledotorg/story/4298/support-contact-patient-sheet-with-no-appointment

### DIFF
--- a/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientModel.kt
+++ b/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientModel.kt
@@ -66,6 +66,9 @@ data class ContactPatientModel(
   val hasLoadedAppointment: Boolean
     get() = appointment != null
 
+  val isAppointmentPresent : Boolean
+  get() = appointment?.isPresent() == true
+
   val hasRegisteredFacility: Boolean
     get() = appointment?.isPresent() == true && appointment.get().patientRegisteredFacilityID != null
 
--- a/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientModel.kt
+++ b/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientModel.kt
@@ -8,7 +8,6 @@ import org.simple.clinic.facility.Facility
 import org.simple.clinic.home.overdue.OverdueAppointment
 import org.simple.clinic.overdue.AppointmentConfig
 import org.simple.clinic.overdue.PotentialAppointmentDate
-import org.simple.clinic.patient.PatientProfile
 import org.simple.clinic.util.ParcelableOptional
 import org.simple.clinic.util.UserClock
 import org.simple.clinic.util.parcelable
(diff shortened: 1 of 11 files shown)


Honey14:
Kindly ignore this [commit](https://github.com/simpledotorg/simple-android/pull/2904/commits/adbff73db0c1846a7b2d111782290174777f5625). Have dropped the commit, will push the change with the feedback, if any.

app/src/main/java/org/simple/clinic/contactpatient/views/CallPatientView.kt
```diff
@@ -147,7 +150,17 @@ class CallPatientView(
     patientAddressTextView.text = patientDetails.patientAddress
     registeredFacilityTextView.text = patientDetails.registeredFacility
     diagnosisTextView.text = diagnosis
-    lastVisitedTextView.text = dateTimeFormatter.format(patientDetails.lastVisited?.toLocalDateAtZone(userClock.zone))
+    isLastSeenAvailable(dateTimeFormatter, patientDetails, userClock)
+  }
+
+  private fun isLastSeenAvailable(dateTimeFormatter: DateTimeFormatter, patientDetails: PatientDetails, userClock: UserClock) {
```
vinaysshenoy, line 165: The prefix `is` for the method implies that the method returns a `Boolean`, but here we are changing the state of the UI based on some conditions. Can we update the method name to better reflect this, please?

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
```diff
@@ -78,6 +78,13 @@ class ContactPatientUiRenderer(
       ui.hidePatientWithNoPhoneNumberUi()
       ui.setResultOfCallLabelText()
 
+      loadSecureCallingUi(model)
+    } else if (model.patientProfileHasPhoneNumber && model.hasLoadedAppointment) {
```
vinaysshenoy: This `else if` clause seems to be identical to the parent `if` clause. Please take a look?

vinaysshenoy: Never mind: resolved in a later commit.

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientModel.kt
```diff
@@ -68,7 +67,7 @@ data class ContactPatientModel(
     get() = appointment != null
 
   val hasRegisteredFacility: Boolean
-    get() = appointment?.get()?.patientRegisteredFacilityID != null
+    get() = appointment?.isPresent() == true && appointment.get().patientRegisteredFacilityID != null
```
vinaysshenoy, line 73: Is there a need for the `== true` here since `isPresent()` will return `Boolean`?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientModel.kt
Honey14: So isPresent will only return a boolean if the appointment value is not null, since this is a ParcelableOptional object so if it's null we would wanna send false. 
`appointment?.isPresent() == false` if it returns null.

vinaysshenoy: No, what I meant was `appointment?.isPresent()` _should be_ functionally the same as `appointment?,isPresent() == true`. 

However, the reason it seems to be required here is because `appointment` is nullable, so the type of the expression `appointment?.isPresent()` is inferred to be `Boolean?` and not `Boolean`.

vinaysshenoy (approved):

PR #2907, opened 2021-08-02
Fix incorrect test in `TheActivityControllerTest`

vinaysshenoy:
There was a test where the setup was being done incorrectly (using the current timestamp instead of the test stub). However, the test was always passing because we were not setting up a mock for the logged in user, so the code path that works with the timestamp was never getting triggered.


Honey14 (approved):
LGTM

PR #2911, opened 2021-08-03
Show keyboard and prefill search query when instant search screen is shown

msasikanth:
https://app.clubhouse.io/simpledotorg/story/4401/show-keyboard-and-prefill-search-query-when-instant-search-screen-is-shown

### DIFF
--- a/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreen.kt
+++ b/app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreen.kt
@@ -108,6 +108,7 @@ class InstantSearchScreen :
 
   private val subscriptions = CompositeDisposable()
   private val loadStateChanges = PublishSubject.create<SearchResultsLoadStateChanged>()
+  private val lifecycleEvents = PublishSubject.create<UiEvent>()
 
   private val instantSearchToolbar
     get() = binding.instantSearchToolbar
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreen.kt
```diff
@@ -199,6 +201,11 @@ class InstantSearchScreen :
     searchResultsAdapter.addLoadStateListener(::searchResultsAdapterLoadStateListener)
   }
 
+  override fun onStart() {
+    super.onStart()
+    lifecycleEvents.onNext(InstantSearchScreenShown)
```
Honey14, line 206: ~Why did we choose onstart, when the same thing was happening in Init?~ Okay so first we wanna load the facility and then check the prefill and keyboard, is that right?

Honey14 (changes requested):
Looks good. Left one comment and there's a conflicting Update class. PTAL

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchScreen.kt
msasikanth: Yeah, partly. But more importantly it has to do with lifecycle of Fragments and ViewModel. Fragments have a tricky lifecycle. Here we are trying to do certain actions when an instant search screen is shown (showing keyboard and prefilling search). 

When we are opening a search result we are simply destroying the view in the fragment and not the entire fragments, so viewmodel is not cleared. So when we revisit the search screen the keyboard and prefill doesn't work if these are present in `Init`.

For that reason since these actions are tied to lifecycle of fragment rather than ViewModel, we are sending a UiEvent to do that action everytime the instant search screen is started.

This article explains the lifecycles well : https://zsmb.co/fragment-lifecycles-in-the-age-of-jetpack/

Honey14: Ah now it makes much more sense.

Honey14 (approved):
LGTM

PR #2912, opened 2021-08-03
Don't load search results if search query is not changed from previous search query

msasikanth:
https://app.clubhouse.io/simpledotorg/story/4404/don-t-load-search-results-if-search-query-is-not-changed-from-previous-search-query


app/src/main/java/org/simple/clinic/instantsearch/InstantSearchUpdate.kt
```diff
@@ -50,6 +50,17 @@ class InstantSearchUpdate @Inject constructor(
     }
   }
 
+  private fun searchQueryChanged(
+      model: InstantSearchModel,
+      event: SearchQueryChanged
+  ): Next<InstantSearchModel, InstantSearchEffect> {
+    return if (event.searchQuery != model.searchQuery) {
+      next(model.searchQueryChanged(event.searchQuery), ValidateSearchQuery(event.searchQuery))
```
vinaysshenoy, line 58: I'm curious, what scenario will this be triggered? We emit text change notifications only when the text actually changes, right?

vinaysshenoy (approved):

app/src/main/java/org/simple/clinic/instantsearch/InstantSearchUpdate.kt
msasikanth: Ah, I forgot to include testing steps. So once we entered the search query and open the search results, if we go back to search again it will prefill the search query and loads search results again.

PR #2919, opened 2021-08-04
Add Sri Lanka national id business identifier

msasikanth:
- Add Sri Lanka identifier type
- Add Sri Lanka national ID meta data version in `BusinessId`


msasikanth:
Didn't update the CHANGELOG, since there is already an in-progress entry when this PR is opened.

vinaysshenoy (approved):

PR #2943, opened 2021-08-06
Set Sri Lanka display name in country selection screen

msasikanth:
- Set Sri Lanka display name
- Update CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,15 +11,15 @@
 - Move logic for deciding the initial screen to the Mobius loop in `TheActivity`
 - Update OverdueListChanges feature flag to support `ContactPatientSheet` changes
 - Change registered facility as nullable in `ContactPatientProfile`
+- Add home screen illustrations for Sri Lanka
 - Change Sri Lanka iso country code to `LK`
 - Bump asm to v9.2
-- [In Progress: 04 Aug 2021] Add support for Sri Lanka 
+- Set Sri Lanka display name in country selection screen  
 
 ### Changes
 - Don't load search results if search query is not changed from previous search query
 - Show keyboard and prefill search query if present when instant search screen is shown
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Migrate to Gradle Version Catalog  
 - Move logic for deciding the initial screen to the Mobius loop in `TheActivity`
 - Update OverdueListChanges feature flag to support `ContactPatientSheet` changes
+- Set Sri Lanka display name in country selection screen
```
Honey14, line 13: Shouldn't the Sri Lanka be a pointer under `Add support for Sri Lanka`?

Honey14 (changes requested):
LGTM. Left one comment. PTAL

CHANGELOG.md
msasikanth: It's not a public change yet. We haven't enabled the support for Sri Lanka right now.

Honey14: Right, and the changelog says, it's in progress. 
Wouldn't this be like :

```suggestion
- in progres Add support for Sri Lanka
   - Set Sri Lanka display name in country selection screen ```

msasikanth: Nope, it's just an internal change.

msasikanth: So basically even after we add this country support, users will not be able to see it yet. The server has to start sending us the manifest with the new country. So most of these changes will be internal.

Honey14: So Add support for Sri Lanka is also added as internal change, right? 
I know it's internal, just meant to say, this should be a pointer inside the feature. Like we started doing.

msasikanth: We are only doing that for features that require more explanation or have a vague changelog. For example things like overdue list improvements, we have added what the improvements are. In this situation we are adding support for a new country, I don't think it needs any more explanation. Once the server has finished the Sri Lanka support work, we have to coordinate the release changelog.

msasikanth: If you look at our existing changelog on master, you can see we are not adding points for every new feature. We are adding them in internal and we are only adding one-liners most of the time that are user-facing and are useful for them.

msasikanth: I have updated the changelog. I have kept the changes in internal but moved the in-progress of entry from internal to changes.

Honey14 (approved):
LGTM

PR #2954, opened 2021-08-09
Migrate `PatientEntryScreen` to a fragment

msasikanth:
- Use `RelativeLayout` as parent in `screen_manual_patient_entry`
- Add `ScreenKey` in `PatientEntryScreen`
- Use `PatientEntryScreen.Key` to open `PatientEntryScreen`
- Extend `BaseScreen` in `PatientEntryScreen`
- Replace `context` with `requireContext`
- Override lifecycle method of the fragment
- Remove unused code
- Update CHANGELOG

**Story**: https://app.clubhouse.io/simpledotorg/story/4460/migrate-patiententryscreen-to-a-fragment

### DIFF
--- a/app/src/main/res/layout/screen_manual_patient_entry.xml
+++ b/app/src/main/res/layout/screen_manual_patient_entry.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<org.simple.clinic.newentry.PatientEntryScreen xmlns:android="http://schemas.android.com/apk/res/android"
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
   xmlns:app="http://schemas.android.com/apk/res-auto"
   xmlns:tools="http://schemas.android.com/tools"
   android:id="@+id/patientEntryRoot"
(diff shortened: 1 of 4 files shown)


app/src/main/res/layout/screen_manual_patient_entry.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<org.simple.clinic.newentry.PatientEntryScreen xmlns:android="http://schemas.android.com/apk/res/android"
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
Honey14, line 2: Hey why are we not using ConstraintLayout here?

Honey14 (changes requested):
LGTM. Left one comment. PTAL

app/src/main/res/layout/screen_manual_patient_entry.xml
msasikanth: The goal of this PR is not to change the root layout to `ConstraintLayout`. We are just migrating the view based screen to a fragment. So there is no need to change the root layout, that can be done later if needed.

Honey14 (approved):

PR #2956, opened 2021-08-10
Unify patient age details into a single embedded Room model

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/4426/change-dateofbirth-to-an-embedded-room-model


app/src/main/java/org/simple/clinic/patient/Patient.kt
```diff
@@ -83,14 +83,9 @@ data class Patient(
   fun withNameAndGender(fullName: String, gender: Gender): Patient =
       copy(fullName = fullName, gender = gender)
 
-  fun withoutAgeAndDateOfBirth(): Patient =
```
msasikanth, line 77: I am guessing we removed this because there is no use case where we want a patient with age or dob?

msasikanth (dismissed):
LGTM

msasikanth (dismissed):

msasikanth (approved):

app/src/main/java/org/simple/clinic/patient/Patient.kt
vinaysshenoy: Yup.

PR #2957, opened 2021-08-10
Migrate `EditPatientScreen` to a fragment

msasikanth:
- Use `RelativeLayout` as parent in `screen_edit_patient`
- Add `ScreenKey` in `EditPatientScreen`
- Use `EditPatientScreen.Key` to open `EditPatientScreen`
- Extend `BaseScreen` in `EditPatientScreen`
- Fix `context` usage in `EditPatientScreen`
- Override lifecycle methods of the fragment
- Use root view for child layout inflation and transitions
- Open discard changes dialog using `childFragmentManager`
- Remove unused code
- Update CHANGELOG

**Story**: https://app.clubhouse.io/simpledotorg/story/4483/migrate-editpatientscreen-to-a-fragment


msasikanth:
There is a weird issue after migrating this screen to a fragment. The reason this happens is we are setting the input fields after the `setDateOfBirthAndAgeVisibility` is called. The reason being one of them is a UI Action and the other is called in UI renderer. I think the ideal solution here would be to move the input fields visibility call into the UI renderer as well to call them sequentially. But I don't think that would be possible (one of the fields require a date formatter, so we cannot mark it parcelize. Need to look into it more).

image


For now, one immediate solution would just check if those 2 fields are already hidden or not and let the `setDateOfBirthAndAgeVisibility` handle the visibility. (I know this is not perfect, but we currently are not hiding any one of these fields in a different country)

```kt
allTypesOfInputFields.forEach { (clazz, view) ->
  when {
      clazz == AgeField::class.java && view.isGone -> {
        return@forEach
      }
      clazz == DateOfBirthField::class.java && view.isGone -> {
        return@forEach
      }
      else -> view.visibleOrGone(clazz in receivedTypesOfInputFields)
  }
}
```

But I think the showing or hiding input fields should definitely be moved to UI renderer, or else these kinds of race conditions will occur. wdyt?

vinaysshenoy:
> There is a weird issue after migrating this screen to a fragment. The reason this happens is we are setting the input fields after the `setDateOfBirthAndAgeVisibility` is called. The reason being one of them is a UI Action and the other is called in UI renderer. I think the ideal solution here would be to move the input fields visibility call into the UI renderer as well to call them sequentially. But I don't think that would be possible (one of the fields require a date formatter, so we cannot mark it parcelize. Need to look into it more).
> 
> image
> 
> For now, one immediate solution would just check if those 2 fields are already hidden or not and let the `setDateOfBirthAndAgeVisibility` handle the visibility. (I know this is not perfect, but we currently are not hiding any one of these fields in a different country)
> 
> ```kotlin
> allTypesOfInputFields.forEach { (clazz, view) ->
>   when {
>       clazz == AgeField::class.java && view.isGone -> {
>         return@forEach
>       }
>       clazz == DateOfBirthField::class.java && view.isGone -> {
>         return@forEach
>       }
>       else -> view.visibleOrGone(clazz in receivedTypesOfInputFields)
>   }
> }
> ```
> 
> But I think the showing or hiding input fields should definitely be moved to UI renderer, or else these kinds of race conditions will occur. wdyt?

This looks acceptable for now. I think this screen was also converted from the v1 architecture to Mobius, so it definitely doesn't conform to a traditional Mobius loop based structure. Maybe we can take a look at this after we migrate all our screens.

msasikanth:
> This looks acceptable for now. I think this screen was also converted from the v1 architecture to Mobius, so it definitely doesn't conform to a traditional Mobius loop based structure. Maybe we can take a look at this after we migrate all our screens.

Just to confirm, you're talking about the issue and not the `when` condition right?

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -519,8 +428,6 @@ class EditPatientScreen : BaseScreen<
     identifiers.forEach { identifier -> inflateBpPassportView(identifier) }
 
     bpPassportsLabel.visibleOrGone(identifiers.isNotEmpty())
-
-    post { requestLayout() }
```
vinaysshenoy, line 447: Would it be possible to check why this change was added and how it's safe to remove now?

vinaysshenoy:
> > This looks acceptable for now. I think this screen was also converted from the v1 architecture to Mobius, so it definitely doesn't conform to a traditional Mobius loop based structure. Maybe we can take a look at this after we migrate all our screens.
> 
> Just to confirm, you're talking about the issue and not the `when` condition right?

I meant the workaround for the issue, not the issue itself. The method definitely doesn't look too good, so we should look at how to resolve this correctly later on.

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
msasikanth: Not entirely sure why it was added before, but when we are inflating and adding child views it will automatically request layout internally. So there is no need to manually call it anymore.

I have tested the BP passports layout (single & multiple) and it is working fine.

vinaysshenoy (dismissed):

vinaysshenoy (approved):

PR #2960, opened 2021-08-10
Migrate `IntroVideoScreen` to a Fragment

Honey14:
https://app.clubhouse.io/simpledotorg/story/4459/migrate-introvideoscreen-to-a-fragment


app/src/main/java/org/simple/clinic/introvideoscreen/IntroVideoScreen.kt
```diff
@@ -131,4 +135,18 @@ class IntroVideoScreen(
...
     }
   }
+
+  @Parcelize
+  data class Key(
+      val registrationEntry: OngoingRegistrationEntry
+  ) : ScreenKey() {
+
+    @IgnoredOnParcel
+    override val analyticsName = "Onboarding Intro Video"
```
msasikanth, line 144: Can we move this inside the `data class` constructor, please? We are starting to do that to avoid the lint issues when using `IgnoredOnParcel` annotation.

app/src/main/java/org/simple/clinic/introvideoscreen/IntroVideoScreen.kt
```diff
@@ -131,4 +135,18 @@ class IntroVideoScreen(
...
+
+    @IgnoredOnParcel
+    override val type = ScreenType.FullScreen
```
msasikanth, line 147: The default type is already `FullScreen` for `ScreenKey`, so we don't need to override this.

app/src/main/java/org/simple/clinic/introvideoscreen/IntroVideoScreen.kt
```diff
@@ -106,7 +106,7 @@ class IntroVideoScreen(
   }
 
   override fun openHome() {
-    val screenKey = screenKeyProvider.keyFor<IntroVideoScreenKey>(this)
+    val screenKey = screenKeyProvider.keyFor<Key>(this)
```
msasikanth: Just an FYI, we don't need to replace it here. When we extend the `BaseScreen` we can get the `screenKey` variable from the screen instead of the provider. So we would end up removing this piece of code.

app/src/main/java/org/simple/clinic/introvideoscreen/IntroVideoScreen.kt
```diff
@@ -79,7 +82,7 @@ class IntroVideoScreen : BaseScreen<
   }
 
   override fun openHome() {
-    val screenKey = screenKeyProvider.keyFor<Key>(this)
+    val screenKey = screenKeyProvider.keyFor<Key>(rootView)
```
msasikanth, line 85: We should not be using the screen key provider anymore after migrating to a fragment. Instead, we can directly get the `screenKey` variable from the screen. So we can remove this implementation.

Also, the commit message doesn't reflect what's happening here, we are using root view for neither inflation nor transitions here. So, please update that as well.

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreen.kt
```diff
@@ -11,6 +11,7 @@ import io.reactivex.rxkotlin.ofType
 import org.simple.clinic.ReportAnalyticsEvents
```
msasikanth, line 11: We should remove the unused imports here.

app/src/main/java/org/simple/clinic/introvideoscreen/IntroVideoScreen.kt
```diff
@@ -18,7 +18,6 @@ import org.simple.clinic.R
 import org.simple.clinic.ReportAnalyticsEvents
```
msasikanth, line 18: Can we also remove the `IntroVideoScreenKey` class as well, please?

msasikanth (changes requested):
LGTM, left few comments PTAL

app/src/main/java/org/simple/clinic/introvideoscreen/IntroVideoScreen.kt
Honey14: This has been removed in the later commit, please check

app/src/main/java/org/simple/clinic/registration/facility/RegistrationFacilitySelectionScreen.kt
Honey14: This is also removed in the last commit where we remove the interface. Please check

app/src/main/java/org/simple/clinic/introvideoscreen/IntroVideoScreen.kt
msasikanth: The `IntroVideoScreenKey` still seems to be present in this branch. Please, check that.

app/src/main/java/org/simple/clinic/introvideoscreen/IntroVideoScreen.kt
Honey14: Can you link the reference to the issue that you're talking about?

msasikanth (approved):

PR #2966, opened 2021-08-12
Nudge users to add BP or Blood Sugar after creating patient - final

msasikanth:
https://app.clubhouse.io/simpledotorg/story/4387/nudge-users-to-add-bp-or-blood-sugar-after-creating-patient


Honey14 (changes requested):
Looks good overall. If we update the diagnostics from `NO` to `YES` in the patient summary screen with no records, we don't show the dialog. Looks like an edge case. PTAL?

msasikanth:
> Looks good overall. If we update the diagnostics from `NO` to `YES` in the patient summary screen with no records, we don't show the dialog. Looks like an edge case. PTAL?

It's working fine 🤔 . It should show the warning dialogs once per session, so it's possible you might have checked after the warning dialogs are already shown once.

Honey14 (approved):

PR #2974, opened 2021-08-16
Fix app syncing prescription drugs with empty name

msasikanth:
https://app.clubhouse.io/simpledotorg/story/4527/disable-save-button-when-creating-new-custom-drug-and-drug-name-is-not-present


Honey14 (approved):
LGTM

PR #2977, opened 2021-08-16
Migrate `NewMedicalHistoryScreen` to a Fragment

Honey14:
https://app.clubhouse.io/simpledotorg/story/4532/migrate-newmedicalhistoryscreen-to-a-fragment


app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryScreen.kt
```diff
@@ -172,6 +174,24 @@ class NewMedicalHistoryScreen : BaseScreen<
     }
   }
 
+  override fun onAttach(context: Context) {
+    super.onAttach(context)
+    requireContext().injector<Injector>().inject(this)
```
msasikanth, line 160: Just an FYI, we can use the `context` from the `onAttach` directly. There is no need to use `requireContext` here. It doesn't make any difference, just a visual consistency with other screens in the project.

app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryScreen.kt
```diff
@@ -164,8 +168,8 @@ class NewMedicalHistoryScreen : BaseScreen<
       router.pop()
     }
 
-    post {
-      hideKeyboard()
+    rootView.post {
+      rootView.hideKeyboard()
```
msasikanth, line 172: Previously we were doing this to hide the keyboard when coming from the patient entry screen. Can we verify if it's still needed? or else we can remove this code as well.

app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryScreen.kt
Honey14: Oh right, my bad.

janhavisinghh (approved):
LGTM

PR #2979, opened 2021-08-17
Change `ScanSimpleIdScreen` to use Mobius view effects

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/4559/change-scansimpleidscreen-to-use-view-effects

### DIFF
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdViewEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdViewEffectHandler.kt
@@ -1,0 +1,20 @@
+package org.simple.clinic.scanid
+
+import dagger.assisted.Assisted
+import dagger.assisted.AssistedFactory
+import dagger.assisted.AssistedInject
+import org.simple.clinic.mobius.ViewEffectsHandler
+
+class ScanSimpleIdViewEffectHandler @AssistedInject constructor(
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdViewEffectHandler.kt
```diff
@@ -0,0 +1,20 @@
+package org.simple.clinic.scanid
+
+import dagger.assisted.Assisted
+import dagger.assisted.AssistedFactory
+import dagger.assisted.AssistedInject
+import org.simple.clinic.mobius.ViewEffectsHandler
+
+class ScanSimpleIdViewEffectHandler @AssistedInject constructor(
```
msasikanth, line 8: I don't particularly see a point in making it part of the Dagger graph. Something that can be removed in the next PR?

msasikanth (approved):

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdViewEffectHandler.kt
vinaysshenoy: Oh yeah, you are right there. I initially made it this way so I could provide the schedulers, then realised later that it's not necessary and removed that. But I forgot to remove the Dagger graph related change.

Will make this change in another PR.

PR #2980, opened 2021-08-17
Setup isolated server instances on Heroku for CI instrumented tests

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/4518/set-up-isolated-server-instances-at-a-pr-level-for-the-android-integration-tests

Note: There's a commit titled "Update documentation for instrumented test setup" which contains some documentation about the Heroku server configuration which might be useful to read first.

### DIFF
--- a/.github/workflows/ci_checks.yml
+++ b/.github/workflows/ci_checks.yml
@@ -76,6 +76,26 @@ jobs:
           fetch-depth: 0 # Full clone needed for Heroku deploys (https://devcenter.heroku.com/changelog-items/775)
           path: server-app
 
+      - name: Create node dependency hash
+        id: create_node_modules_hash
+        run: android-app/.github/scripts/compute_node_dependency_hash.sh
+
+      - name: Cache node modules
+        id: node-dependency-cache
+        uses: actions/cache@v2
+        with:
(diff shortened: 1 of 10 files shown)


.github/workflows/ci_checks.yml
```diff
@@ -142,6 +157,13 @@ jobs:
           disable-animations: true
           script: android-app/gradlew -p android-app --build-cache --no-daemon connectedQaDebugAndroidTest
 
+      - name: Teardown Heroku instance
+        if: always() && steps.start-simple-server.outcome != 'success'
```
msasikanth: What does `always` do? 🤔

msasikanth (approved):
LGTM

.github/workflows/ci_checks.yml
vinaysshenoy: `always()` tells GitHub Actions to always run a step regardless of the outcome of the previous step. However, if you want to conditionally execute a step (without taking the outcome of the previous step into consideration), unless you `AND` the condition with `always()`, GitHub Actions will still ignore checking the condition if the previous step failed. So we need this here.

PR #2987, opened 2021-08-18
Wire up `CustomDrugEntrySheet` mobius loop with the UI

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/2905/implement-customdrugentrysheet

In this PR - 
- Added `FetchDrug` effect in the mobius loop as we can only pass parcelables through keys. So changed the `OpenAs.New.FromDrugList` parameter from `Drug` to `drugUuid`
- Made some corrections in the mobius loop
- Removed the implementation of passing `Unknown("None")` when converting `DrugFrequency` to `MedicineFrequency` as  in the list of frequencies we will not have an option of Unknown. Unknown is only for cases when there is a new addition to the latest version in `DrugFrequency` and if we have to load it in the older version then we use it. Although in case of `Unknown("None")` this only means that user has not selected any frequency in which case we will just set it to null.
- Added `sheet_custom_drug_entry`
- Implemented the UI layer, mobius delegate and intents.

What's not implemented yet 
- Drug Frequency dialog is still not implemented. Will be raising another PR for that.
- If the user edits the dosage and removes the entire dosage text. The placeholder value stays there and if the user saves after it the placeholder value is saved instead of null in dosage. I will resolve this in the next PR.

### DIFF
--- a/app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
@@ -88,7 +88,8 @@ class DrugsSearchScreen : BaseScreen<
   override fun events() = Observable
       .mergeArray(
           searchQueryChanges(),
-          drugListItemClicks())
+          drugListItemClicks(),
+          newCustomDrugItemClicks())
       .compose(ReportAnalyticsEvents())
       .cast<DrugSearchEvent>()
 
@@ -159,6 +160,13 @@ class DrugsSearchScreen : BaseScreen<
         .itemEvents
         .ofType<DrugSearchListItem.Event.DrugClicked>()
         .map { DrugListItemClicked(it.drug.id, screenKey.patientId) }
+  }
+
+  private fun newCustomDrugItemClicks(): Observable<UiEvent> {
+    return adapter
+        .itemEvents
+        .ofType<DrugSearchListItem.Event.NewCustomDrugClicked>()
+        .map { NewCustomDrugClicked(it.name, screenKey.patientId) }
   }
 
   interface Injector {
--- a/app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
@@ -129,6 +129,10 @@ class DrugsSearchScreen : BaseScreen<
     router.push(CustomDrugEntrySheet.Key(OpenAs.New.FromDrugList(drugUuid), patientUuid))
   }
 
+  override fun openCustomDrugEntrySheetFromDrugName(drugName: String, patientUuid: UUID) {
+    router.push(CustomDrugEntrySheet.Key(OpenAs.New.FromDrugName(drugName), patientUuid))
+  }
+
   private fun drugSearchLoadStateListener(combinedLoadStates: CombinedLoadStates) {
     val isLoading = combinedLoadStates.refresh is LoadState.Loading
     progressIndicator.visibleOrGone(isLoading)
--- a/app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
@@ -1,6 +1,7 @@
 package org.simple.clinic.drugs.search
 
 import android.content.Context
+import android.content.Intent
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
@@ -12,6 +13,7 @@ import com.jakewharton.rxbinding3.widget.textChanges
 import com.spotify.mobius.functions.Consumer
 import io.reactivex.Observable
 import io.reactivex.rxkotlin.cast
+import io.reactivex.rxkotlin.ofType
 import kotlinx.parcelize.Parcelize
 import org.simple.clinic.R
 import org.simple.clinic.ReportAnalyticsEvents
@@ -83,7 +85,10 @@ class DrugsSearchScreen : BaseScreen<
 
   override fun uiRenderer() = DrugSearchUiRenderer(this)
 
-  override fun events() = searchQueryChanges()
+  override fun events() = Observable
+      .mergeArray(
+          searchQueryChanges(),
+          drugListItemClicks())
       .compose(ReportAnalyticsEvents())
       .cast<DrugSearchEvent>()
 
@@ -147,6 +152,13 @@ class DrugsSearchScreen : BaseScreen<
         .map { searchQuery ->
           SearchQueryChanged(searchQuery.toString())
         }
+  }
+
+  private fun drugListItemClicks(): Observable<UiEvent> {
+    return adapter
+        .itemEvents
+        .ofType<DrugSearchListItem.Event.DrugClicked>()
+        .map { DrugListItemClicked(it.drug.id, screenKey.patientId) }
   }
 
   interface Injector {
--- a/app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
@@ -20,6 +20,8 @@ import org.simple.clinic.databinding.ListItemDrugSearchCornerCapBinding
 import org.simple.clinic.databinding.ListItemDrugSearchDividerBinding
 import org.simple.clinic.databinding.ScreenDrugsSearchBinding
 import org.simple.clinic.di.injector
+import org.simple.clinic.drugs.selection.custom.CustomDrugEntrySheet
+import org.simple.clinic.drugs.selection.custom.OpenAs
 import org.simple.clinic.navigation.v2.Router
 import org.simple.clinic.navigation.v2.ScreenKey
 import org.simple.clinic.navigation.v2.fragments.BaseScreen
@@ -121,6 +123,10 @@ class DrugsSearchScreen : BaseScreen<
     val searchQuery = searchQueryEditText.text?.toString().orEmpty()
     drugSearchResultsList.scrollToPosition(0)
     adapter.submitData(lifecycle, DrugSearchListItem.from(searchResults, searchQuery))
+  }
+
+  override fun openCustomDrugEntrySheetFromDrugList(drugUuid: UUID, patientUuid: UUID) {
+    router.push(CustomDrugEntrySheet.Key(OpenAs.New.FromDrugList(drugUuid), patientUuid))
   }
 
   private fun drugSearchLoadStateListener(combinedLoadStates: CombinedLoadStates) {
--- a/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryEffectHandler.kt
@@ -9,6 +9,7 @@ import io.reactivex.Observable
 import io.reactivex.ObservableTransformer
 import org.simple.clinic.drugs.PrescriptionRepository
 import org.simple.clinic.drugs.search.DrugFrequency
+import org.simple.clinic.drugs.search.DrugRepository
 import org.simple.clinic.facility.Facility
 import org.simple.clinic.teleconsultlog.medicinefrequency.MedicineFrequency
 import org.simple.clinic.util.nullIfBlank
@@ -18,6 +19,7 @@ import org.simple.clinic.uuid.UuidGenerator
 class CustomDrugEntryEffectHandler @AssistedInject constructor(
     private val schedulersProvider: SchedulersProvider,
     private val prescriptionRepository: PrescriptionRepository,
+    private val drugRepository: DrugRepository,
     private val currentFacility: Lazy<Facility>,
     private val uuidGenerator: UuidGenerator,
     @Assisted private val uiActions: CustomDrugEntrySheetUiActions
@@ -40,8 +42,18 @@ class CustomDrugEntryEffectHandler @AssistedInject constructor(
         .addTransformer(UpdatePrescription::class.java, updatePrescription())
         .addAction(CloseBottomSheet::class.java, uiActions::close, schedulersProvider.ui())
         .addTransformer(FetchPrescription::class.java, fetchPrescription())
+        .addTransformer(FetchDrug::class.java, fetchDrug())
         .addTransformer(RemoveDrugFromPrescription::class.java, removeDrugFromPrescription())
         .build()
+  }
+
+  private fun fetchDrug(): ObservableTransformer<FetchDrug, CustomDrugEntryEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.io())
+          .map { drugRepository.drugImmediate(it.drugUuid) }
+          .map(::DrugFetched)
+    }
   }
 
   private fun setSheetTitle(setSheetTitle: SetSheetTitle) {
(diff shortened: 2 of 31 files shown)


app/src/main/res/layout/sheet_custom_drug_entry.xml
```diff
@@ -0,0 +1,116 @@
...
+    android:layout_marginTop="24dp"
+    android:layout_marginEnd="24dp"
+    android:minWidth="@dimen/spacing_0"
+    android:minHeight="@dimen/spacing_0"
+    android:text="@string/custom_drug_entry_sheet_remove"
+    app:layout_constraintEnd_toEndOf="parent"
+    app:layout_constraintTop_toTopOf="parent"
+    tools:visibility="visible" />
+
+  <LinearLayout
```
msasikanth, line 46: We can probably avoid this nesting and create a chain of these two text fields.

app/src/main/res/layout/sheet_custom_drug_entry.xml
```diff
@@ -0,0 +1,116 @@
...
+
+    </com.google.android.material.textfield.TextInputLayout>
+
+  </LinearLayout>
+
+  <FrameLayout
+    style="@style/Widget.Simple.Button.Frame"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:layout_marginTop="@dimen/spacing_32"
```
msasikanth, line 104: Spacing between text fields and button frame is 24dp, please check that.

app/src/main/res/layout/sheet_custom_drug_entry.xml
```diff
@@ -0,0 +1,116 @@
...
+    app:layout_constraintTop_toBottomOf="@+id/textFieldsLinearLayout">
+
+    <com.google.android.material.button.MaterialButton
+      android:id="@+id/saveButton"
+      android:layout_width="match_parent"
+      android:layout_height="wrap_content"
+      android:text="@string/customprescription_save_drug"
+      app:cornerRadius="@dimen/spacing_0"
```
msasikanth, line 112: I don't think we need to remove the corner radius, based on the designs it has a 2dp corner radius and default button style will take care of it. Can we remove this, please?

app/src/main/res/layout/sheet_custom_drug_entry.xml
```diff
@@ -0,0 +1,116 @@
...
+      tools:ignore="UnusedAttribute" />
```
msasikanth, line 113: Why are we doing this? I don't see any attribute that should show this lint warning, can we remove this if not needed, please?

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryEvent.kt
```diff
@@ -15,7 +16,7 @@ data class EditFrequencyClicked(val frequency: DrugFrequency) : CustomDrugEntryE
 
 data class FrequencyEdited(val frequency: DrugFrequency) : CustomDrugEntryEvent()
 
-object AddMedicineButtonClicked : CustomDrugEntryEvent()
+data class AddMedicineButtonClicked(val patientUuid: UUID) : CustomDrugEntryEvent()
```
msasikanth, line 19: Can't we just have the `patientUuid` in the model instead of passing for every event that requires it? 🤔

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntrySheet.kt
```diff
@@ -0,0 +1,186 @@
...
+  private fun removeClicks(): Observable<UiEvent> =
+      removeButton
+          .clicks()
+          .map { RemoveDrugButtonClicked }
+
+  override fun showEditFrequencyDialog(frequency: DrugFrequency) {
+    // to do
+  }
+
+  @Suppress("IMPLICIT_CAST_TO_ANY")
```
msasikanth, line 128: Why are we suppressing this? 🤔

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntrySheet.kt
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  override fun setButtonTextAsAdd() {
+    saveButton.text = getString(R.string.custom_drug_entry_sheet_add_button_text)
+  }
+
+  @Parcelize
+  data class Key(val openAs: OpenAs, val patientUuid: UUID) : ScreenKey() {
+    @IgnoredOnParcel
+    override val analyticsName = "Custom Drug Entry Sheet"
```
msasikanth, line 181: Can we move this inside the constructor of the `Key`, please?

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryEffectHandler.kt
```diff
@@ -57,7 +57,7 @@ class CustomDrugEntryEffectHandler @AssistedInject constructor(
   }
 
   private fun setSheetTitle(setSheetTitle: SetSheetTitle) {
-    val sheetTitle = constructSheetTitle(setSheetTitle.name, setSheetTitle.dosage, setSheetTitle.frequency)
+    val sheetTitle = constructSheetTitle(setSheetTitle.name, setSheetTitle.dosage.nullIfBlank(), setSheetTitle.frequency)
```
msasikanth, line 61: Is there a scenario where this can be blank instead of `null` or `Unknown`? 🤔

app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
```diff
@@ -40,6 +43,9 @@ class DrugsSearchScreen : BaseScreen<
     DrugSearchEffect,
     Unit>(), DrugSearchUi, UiActions {
 
+  @Inject
+  lateinit var activity: AppCompatActivity
```
msasikanth, line 49: Why are we inject activity here?

app/src/test/java/org/simple/clinic/drugs/search/DrugSearchUpdateTest.kt
```diff
@@ -1,15 +1,12 @@
 package org.simple.clinic.drugs.search
 
 import androidx.paging.PagingData
-import com.spotify.mobius.test.NextMatchers.hasEffects
-import com.spotify.mobius.test.NextMatchers.hasModel
-import com.spotify.mobius.test.NextMatchers.hasNoEffects
-import com.spotify.mobius.test.NextMatchers.hasNoModel
+import com.spotify.mobius.test.NextMatchers.*
```
msasikanth, line 4: Can we avoid wild card imports, please?

app/src/test/java/org/simple/clinic/drugs/search/DrugSearchUpdateTest.kt
```diff
@@ -1,15 +1,12 @@
...
 import com.spotify.mobius.test.UpdateSpec
 import com.spotify.mobius.test.UpdateSpec.assertThatNext
 import org.junit.Test
 import org.simple.clinic.TestData
-import java.util.UUID
+import java.util.*
```
msasikanth, line 9: Same comment as above.

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntrySheet.kt
```diff
@@ -142,6 +143,10 @@ class CustomDrugEntrySheet : BaseBottomSheet<
     router.pop()
   }
 
+  override fun goToEditMedicineScreen() {
+    router.clearHistoryAndPush(PrescribedDrugsScreenKey(screenKey.patientUuid))
```
msasikanth, line 147: Clearing history means it will drop all the screens before edit medicines screen as well. You can try one of `popUntil`, `popUntilInclusive` or `replaceKeyOfSameType`, see which one fits this use case.

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
```diff
@@ -21,7 +21,7 @@ class CustomDrugEntryUpdate : Update<CustomDrugEntryModel, CustomDrugEntryEvent,
       is EditFrequencyClicked -> dispatch(ShowEditFrequencyDialog(event.frequency))
       is FrequencyEdited -> next(model.frequencyEdited(event.frequency), SetDrugFrequency(event.frequency), SetSheetTitle(model.drugName, model.dosage, event.frequency))
       is AddMedicineButtonClicked -> createOrUpdatePrescriptionEntry(model, event.patientUuid)
-      is CustomDrugSaved, ExistingDrugRemoved -> dispatch(CloseBottomSheet)
+      is CustomDrugSaved, ExistingDrugRemoved -> dispatch(CloseBottomSheet, GoToEditMedicineScreen)
```
msasikanth, line 24: I don't think this would work, here we are closing and trying to launch the edit medicines screen. Here you're trying to do two navigation actions, I don't think this would work the same way you expect it to.

msasikanth (changes requested):
Left few comments PTAL

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryModel.kt
```diff
@@ -37,4 +37,8 @@ data class CustomDrugEntryModel(
   fun drugNameLoaded(drugName: String): CustomDrugEntryModel {
     return copy(drugName = drugName)
   }
+
+  fun rxNormCodeEdited(rxNormCOde: String?): CustomDrugEntryModel {
```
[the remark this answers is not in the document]
janhavisinghh: Yes, because we are not tracking the drug object in the model and whenever we are updating or saving anything in the db we are creating a new entry. So, whenever we are fetching a prescription or a drug from their ids we are also saving the rxNormCode value so that when we save the medicine we save that value as well.

app/src/main/java/org/simple/clinic/drugs/selection/custom/OpenAs.kt
```diff
@@ -1,14 +1,18 @@
...
-import org.simple.clinic.drugs.search.Drug
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
 import java.util.UUID
 
-sealed class OpenAs {
-  sealed class New : OpenAs() {
+sealed class OpenAs : Parcelable {
+  sealed class New : Parcelable, OpenAs() {
+    @Parcelize
```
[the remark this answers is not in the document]
janhavisinghh: I needed this implementation for the ui changes which is why kept it here. 😅

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryInit.kt
```diff
@@ -20,7 +20,7 @@ class CustomDrugEntryInit : Init<CustomDrugEntryModel, CustomDrugEntryEffect> {
   ): First<CustomDrugEntryModel, CustomDrugEntryEffect> {
     val updatedModel = model.drugNameLoaded(drugName)
 
-    val effects = setOf(SetSheetTitle(drugName, null, null), SetDrugFrequency(null))
+    val effects = setOf(SetSheetTitle(drugName, null, null), SetDrugFrequency(null), SetDrugDosage(null))
```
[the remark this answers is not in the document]
janhavisinghh: Because, setting frequency as null in this case will mean that setting it to `None` And as we are not rendering this from the ui. I am calling it from here.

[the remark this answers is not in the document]
janhavisinghh: I don't need to `SetDrugDosage` here tho

app/src/main/res/layout/sheet_custom_drug_entry.xml
janhavisinghh: 🤦🏻 I was checking from the button

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryEvent.kt
janhavisinghh: Umm.. We only needed it when the save button is clicked and the `patientUuid` parameter is present in the screen key which is why I didn't see a reason to add it.

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntrySheet.kt
janhavisinghh: It was giving me a waning earlier. It's not anymore 😅

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryModel.kt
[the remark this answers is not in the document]
msasikanth: Hmm, okay. I guess we can look if there is a better approach after the entire thing is implemented.

app/src/main/java/org/simple/clinic/drugs/selection/custom/OpenAs.kt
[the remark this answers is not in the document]
msasikanth: Ah, usually I push it as a separate commit and then push the UI changes commit after that. But something to keep in mind.

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryEvent.kt
msasikanth: Ah, makes sense if this is the only place 👍🏾

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryInit.kt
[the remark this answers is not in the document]
msasikanth: hmm, can't we just have the default text as `None`? That way it only changes we are passing the actual frequency value.

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryEffectHandler.kt
janhavisinghh: This is for dosage text. How it is currently implemented is if dosage doesn't have focus and there is only `mg` in the edit text then it sets it to an empty string. You can go to `CustomDrugEntryUiRenderer#showDefaultDosagePlaceholder` method to see the implementation. And  I'm using `listOfNotNull` for constructing the sheet title which led to the title having `DrugName` and a `,`. Hence I added this.

app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
janhavisinghh: I had initially implemented the screen as an activity. #Mistake 😅 Altho, it led to me understanding the navigation framework more clearly. Which was great. :P

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
janhavisinghh: So can I just do a `popUntil` here instead?

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryInit.kt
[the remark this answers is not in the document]
janhavisinghh: Umm fair enough

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
msasikanth: Yep

app/src/main/res/layout/sheet_custom_drug_entry.xml
janhavisinghh: Will resolve this in the next PR

msasikanth:
The spacing of the sheet doesn't match the one from the designs
image

The button click radius is really small as well. It makes it hard for the user to click on the button. Please look at the existing prescription entry sheet.
image

You can resolve these UI issues in the next PR. 

There is also an issue with navigation, once you have added a prescription it takes you back to the edit medicines screen. But if you try to go back from there, the app exists. 

For now, I am approving it, since it's behind a feature flag. But please make sure to resolve these before releasing this feature.

msasikanth (approved):

PR #2989, opened 2021-08-19
UI improvements for medical history screen

msasikanth:
https://app.clubhouse.io/simpledotorg/story/4467/ui-improvements-for-medical-history-screen


app/src/main/java/org/simple/clinic/medicalhistory/MedicalHistoryDiagnosisWithTreatment.kt
```diff
@@ -0,0 +1,115 @@
...
+import org.simple.clinic.databinding.ListMedicalhistoryDiagnosisWithTreatmentBinding
+
+class MedicalHistoryDiagnosisWithTreatment(
+    context: Context,
+    attrs: AttributeSet?
+) : MaterialCardView(context, attrs) {
+  private var binding: ListMedicalhistoryDiagnosisWithTreatmentBinding? = null
+
+  private val diagnosisTextView
+    get() = binding!!.diagnosisTextView
```
janhavisinghh, line 16: No necessary to change. We can reassign `binding!!` to a variable here as well to avoid using nullable variables, no?

app/src/main/res/values/strings.xml
```diff
@@ -816,7 +816,7 @@
   <string name="nhid_error_cannot_assign_to_patient_with_existing_ok">OK</string>
 
   <!-- Select ongoing hypertension treatment dialog -->
-  <string name="select_ongoing_hypertension_treatment_title">Did patient take BP medicines any time in the last month?</string>
+  <string name="select_ongoing_hypertension_treatment_title">Confirm if patient took BP medicines anytime in the last month</string>
```
janhavisinghh, line 819: This error dialog is not showing the entire message after we made this change. The expected behaviour is to show the entire message ([figma design ref](https://www.figma.com/file/vmZlmlFSCOrEYzTMkqsNB8/Simple-Production?node-id=17674%3A285)) Also the UI of this dialog is off. Can you PTAL?

janhavisinghh (changes requested):
LGTM left a few comments PTAL?

app/src/main/res/values/strings.xml
msasikanth: Yep, that's resolved in a later PR, it requires us to change a style.

msasikanth: >Also the UI of this dialog is off. Can you PTAL?

WDYM? The UI seems the same as the designs 🤔

msasikanth: So, we can ignore this change for now. I have talked to Mahima about a small issue with the current dialog, we will be changing them, will open a different PR for it.

app/src/main/res/layout/list_medicalhistory_hypertension_treatment.xml
```diff
@@ -8,15 +8,14 @@
 
   <androidx.constraintlayout.widget.ConstraintLayout
     android:layout_width="match_parent"
-    android:layout_height="wrap_content">
+    android:layout_height="wrap_content"
+    android:padding="@dimen/spacing_16">
 
     <TextView
       android:id="@+id/labelTextView"
       android:layout_width="0dp"
```
janhavisinghh, line 16: Not important to change here but we are using hardcoded dimen value here

msasikanth: That change is not from this PR, we will removing this file anyway.

janhavisinghh (approved):

PR #2993, opened 2021-08-20
Show appointment results in contact patient sheet for patient with no phone number

msasikanth:
Slack message: https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1629447233229600

Since this UI doesn't have to be reactive, we don't have to worry about hiding the results section. Basically we are not adjusting UI based on some reactive stream, we just render the UI when creating the sheet.

This results UI is hidden by default, and only get's shown if the patient doesn't have no phone number and has appointment.

### DIFF
--- a/app/src/main/res/layout/contactpatient_callpatient.xml
+++ b/app/src/main/res/layout/contactpatient_callpatient.xml
@@ -346,4 +346,11 @@
     android:layout_height="wrap_content"
     android:visibility="gone"
     app:constraint_referenced_ids="resultOfCallLabel, agreedToVisitTextView, agreedToVisitSeparator, remindToCallLaterTextView, remindToCallLaterSeparator, removeFromOverdueListTextView" />
+
+  <androidx.constraintlayout.widget.Group
+    android:id="@+id/patientWithNoPhoneNumberResultGroup"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:visibility="gone"
+    app:constraint_referenced_ids="resultOfCallLabel, agreedToVisitTextView, agreedToVisitSeparator, removeFromOverdueListTextView" />
 </merge>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Update `MaterialAlertDialog` theme and style
 - Run integration tests on discrete Heroku servers instead of QA
 - Bump lottie to v4.1.0
+- Show appointment results in contact patient sheet for patient with no phone number
 
 ### Changes
 - [In Progress: 19 Aug 2021] UI improvements for medical history screen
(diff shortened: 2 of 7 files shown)


app/src/main/res/layout/contactpatient_callpatient.xml
```diff
@@ -346,4 +346,11 @@
     android:layout_height="wrap_content"
```
janhavisinghh, line 346: @Honey14 I wanted to know why did we separate `patientWithCallResultGroup`  out from `patientWithPhoneNumberGroup`? 

@msasikanth I feel like adding `patientWithNoPhoneNumberResultGroup` is redundant as we can just not handle hiding and showing of views that are common between `patientWithNoPhoneNumberResultGroup` and `patientWithCallResultGroup`. So, instead of adding one more group we can just remove `resultOfCallLabel, agreedToVisitTextView, agreedToVisitSeparator, removeFromOverdueListTextView` from `patientWithCallResultGroup`. As they are going to be present irrespective of phone number being present or not. I feel like the groups should only have unique attributes that won't be present in the other scenario. Wdyt?

janhavisinghh (changes requested):
Left a comment PTAL

app/src/main/res/layout/contactpatient_callpatient.xml
msasikanth: We cannot do that right now, this sheet is being used in both the overdue list and patient summary screen. So in the patient summary screen if there is no appointment we cannot show call results. Without this group, I don't think it was possible for honey to fix the crash that has been happening with the previous implementation.

For now, let's go with this. We will be refactoring this code later to simplify the approach.

janhavisinghh: Umm alright.

msasikanth: @janhavisinghh So basically, this sheet has 4 different states (counting secure calling state).

1. Patient with phone number and appointment (Need to show the patient info, diagnosis, call results, and call buttons)
2. Patient with phone number and no appointment (Need to show the patient info, diagnosis, call buttons, and no call results)
3. Patient with no phone number and appointment (Need to show the patient info, diagnosis, call results and no phone number view)
4. Finally we have to toggle between showing a secure calling button or not depending on if the feature is enabled or not, and if the country supports it.

msasikanth: So, call results are not tied to patients with or without phone numbers. But rather tied to if the patient has an appointment or not.

janhavisinghh: What about patient with no phone number and no appointment?

msasikanth: How would they even trigger that? We won't show a call option in the patient summary if the patient doesn't have a phone number. So they can only open it from the overdue screen.

janhavisinghh (approved):

PR #3000, opened 2021-08-23
Migrate `EnterOtpScreen` to a Fragment

Honey14:
https://app.clubhouse.io/simpledotorg/story/4671/migrate-enterotpscreen-to-a-fragment

### DIFF
--- a/app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
+++ b/app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
@@ -124,6 +124,8 @@ class EnterOtpScreen : BaseScreen<
     super.onAttachedToWindow()
     delegate.start()
   }
+  private val rootLayout
+  get() = binding.rootLayout
 
   override fun onDetachedFromWindow() {
     delegate.stop()
@@ -201,7 +203,7 @@ class EnterOtpScreen : BaseScreen<
   }
 
   override fun goBack() {
-    hideKeyboard()
+    rootLayout.hideKeyboard()
     router.pop()
   }
 
@@ -234,13 +236,13 @@ class EnterOtpScreen : BaseScreen<
   }
 
   override fun showProgress() {
-    TransitionManager.beginDelayedTransition(this)
+    TransitionManager.beginDelayedTransition(rootLayout)
     validateOtpProgressBar.visibility = View.VISIBLE
     otpEntryContainer.visibility = View.INVISIBLE
   }
 
   override fun hideProgress() {
-    TransitionManager.beginDelayedTransition(this)
+    TransitionManager.beginDelayedTransition(rootLayout)
     validateOtpProgressBar.visibility = View.INVISIBLE
     otpEntryContainer.visibility = View.VISIBLE
   }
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
```diff
@@ -124,6 +130,12 @@ class EnterOtpScreen(
   override fun onRestoreInstanceState(state: Parcelable?) {
     super.onRestoreInstanceState(delegate.onRestoreInstanceState(state))
   }
+
```
janhavisinghh: This is a breaking commit. Are we purposefully doing this? 
Although, I think we can add all the overridden mobius methods in this commit.

janhavisinghh: We should probably squash this with the next commit as we are already adding `defaultModel` here

app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
```diff
@@ -137,6 +144,24 @@ class EnterOtpScreen : BaseScreen<
       layoutInflater: LayoutInflater,
       container: ViewGroup?) = ScreenEnterotpBinding.inflate(layoutInflater, container, false)
 
+  override fun events() =  Observable
+      .mergeArray(
+          otpSubmits(),
+          resendSmsClicks()
```
janhavisinghh, line 84: Also can we reformat all the files whenever we commit something please?

app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
```diff
@@ -19,6 +22,8 @@ import org.simple.clinic.databinding.ScreenEnterotpBinding
 import org.simple.clinic.di.injector
 import org.simple.clinic.mobius.MobiusDelegate
 import org.simple.clinic.navigation.v2.Router
+import org.simple.clinic.navigation.v2.ScreenKey
+import org.simple.clinic.router.screen.FullScreenKey
 import org.simple.clinic.util.unsafeLazy
 import org.simple.clinic.widgets.UiEvent
 import org.simple.clinic.widgets.hideKeyboard
```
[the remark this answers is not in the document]
janhavisinghh: I see you have removed this in the later commit but I think we should follow reformatting code in every commit. Wdyt?

janhavisinghh (changes requested):
LGTM left some small comments PTAL?

app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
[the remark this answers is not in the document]
Honey14: So with we migrations we usually remove all the unused code all at once and with other feature related work we format in every commit.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
Honey14: So with fragment migration, mostly all commits would be breaking if we see it individually. So we're anyway squashing and merging so it doesn't matter, 
I've made these as two different commits because the bindview and model overridden methods were mandatory implemented from BaseScreen and that particular commit extends baseScreen and the rest of the Mobius methods into different one.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
Honey14: This is done in the later commit when we remove all the unused code. PTAL?

janhavisinghh (approved):

PR #3001, opened 2021-08-23
Add Clubhouse overview in the Pivotal documentation

Honey14:
https://app.clubhouse.io/simpledotorg/story/4633/update-the-pivotal-adr-to-include-clubhouse

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Fix TDS drug frequency string value
 - Remove unused `DrugFrequencyChoiceItem` class
 - Change `ScheduleAppointmentSheet` to use Mobius view effects
+- Add Clubhouse overview description in the Pivotal documentation
 - [In Progress: 30 Aug 2021] Record call results instead of updating the same appointment record
 - [In Progress: 30 Aug 2021] Migrate v1 `Country` usages to v2 `Country`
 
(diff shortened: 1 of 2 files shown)


vinaysshenoy (changes requested):
We should _never_ be editing the content of ADRS, only the status of them (Accepted, Rejected, Superceded, etc).

What should be done in this scenario is to create a new ADR describing our current process and then update the status of the older one as "Superceded by" and link to the new one.

doc/tracking-stories-on-clubhouse.md
```diff
@@ -0,0 +1,7 @@
+# Tracking stories on Clubhouse
```
vinaysshenoy, line 1: Ok, I am _really_ sorry about this, but I think I got confused when the PR title referred to a change in the ADR. An ADR is a very specific type of documentation separate from our typical documentation (as can be seen in the `doc/arch` directory) and it needs to follow the conventions I mentioned earlier.

I just noticed that this documentation is not actually an ADR and is just a regular piece of documentation in which case your original set of changes was actually correct, except for the labelling of these changes as an ADR when it should just be documentation.

Could you please revert these changes and update the commit and PR messages to remove the references to "ADR"? I could make these changes as well if you're occupied.

vinaysshenoy (changes requested):

doc/tracking-stories-on-clubhouse.md
Honey14: Oh okay. Understood. I'll update this

vinaysshenoy (approved):

PR #3002, opened 2021-08-23
Implement drug frequency dialog in `CustomDrugEntrySheet`

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/2905/implement-customdrugentrysheet

In this PR : 
✅ Implemented Drug Frequency dialog 
✅ Resolved the below issue
> If the user edits the dosage and removes the entire dosage text. The placeholder value stays there and if the user saves after it the placeholder value is saved instead of null in dosage. 

✅ What I've done here instead is I'm saving an empty string instead of `dosagePlaceHolder` in the repository and also not showing the dosage value in the sheet title if it's just `dosagePlaceHolder` without any numeric value.

👍🏻 Also addressed the small issues that were raised in the last [PR](https://github.com/simpledotorg/simple-android/pull/2987#issuecomment-901803068) 

> There is also an issue with navigation, once you have added a prescription it takes you back to the edit medicines screen. But if you try to go back from there, the app exists.

✅ I tried this out and the app crashed with an error message "Dosage cannot be null and blank at the same time". So, I added a `nullIfBlank()` while saving dosage values in the repository. Altho, lmk if this is something that is happening again. 

> The spacing of the sheet doesn't match the one from the designs.

✅ I rechecked all the margins. Fixed the ones I could find. 

> The button click radius is really small as well. It makes it hard for the user to click on the button. Please look at the existing prescription entry sheet.

✅ I added padding instead of margin in `removeButton` layout. 

> We can probably avoid this nesting and create a chain of these two text fields.

✅ done (from [this](https://github.com/simpledotorg/simple-android/pull/2987#discussion_r691720733) PR comment)


app/src/main/java/org/simple/clinic/drugs/selection/custom/drugfrequency/SelectDrugFrequencyDialog.kt
```diff
@@ -0,0 +1,91 @@
...
+  private val screenKey: Key by unsafeLazy { ScreenKey.key(this) }
+
+  override fun onAttach(context: Context) {
+    super.onAttach(context)
+
+    context.injector<Injector>().inject(this)
+  }
+
+  override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+    var selectedValueIndex = frequenciesArrayIndexFromDrugFrequency(screenKey.drugFrequency)
```
msasikanth, line 38: We are tracking a state as a mutable property here. Also, if there is a state we should switch to using Mobius loop. We go with a standard dialog fragment if there is no state and it's just handling a normal dialog.

app/src/main/java/org/simple/clinic/drugs/selection/custom/drugfrequency/SelectDrugFrequencyDialog.kt
```diff
@@ -0,0 +1,91 @@
...
+    requireActivity().onBackPressed()
+  }
+
+  private fun drugFrequencyFromFrequenciesArrayIndex(index: Int): DrugFrequency {
+    return when (index) {
+      1 -> OD
+      2 -> BD
+      3 -> QDS
+      4 -> TDS
+      else -> Unknown("None")
```
msasikanth, line 63: `Unknown` is different from selecting `None`. `Unknown` is for API backwards compatibility, so we should not be doing this. Instead, we can send back a `null` if the drug frequency is selected as `None`.

app/src/main/java/org/simple/clinic/drugs/selection/custom/drugfrequency/SelectDrugFrequencyDialog.kt
```diff
@@ -0,0 +1,91 @@
...
+  override fun onCancel(dialog: DialogInterface) {
+    backPressed()
+    super.onCancel(dialog)
+  }
+
+  private fun backPressed() {
+    requireActivity().onBackPressed()
+  }
+
+  private fun drugFrequencyFromFrequenciesArrayIndex(index: Int): DrugFrequency {
```
msasikanth, line 57: Instead of doing this, we can create a list of drug frequencies and use the initial screen key passed to determine the selected index.

app/src/main/java/org/simple/clinic/drugs/selection/custom/drugfrequency/SelectDrugFrequencyDialog.kt
```diff
@@ -0,0 +1,91 @@
...
+    return when (index) {
+      1 -> OD
+      2 -> BD
+      3 -> QDS
+      4 -> TDS
+      else -> Unknown("None")
+    }
+  }
+
+  private fun frequenciesArrayIndexFromDrugFrequency(drugFrequency: DrugFrequency?): Int {
```
msasikanth, line 67: Same comment as using a list of drug frequencies, instead of doing this.

app/src/main/java/org/simple/clinic/drugs/selection/custom/drugfrequency/SelectDrugFrequencyDialog.kt
msasikanth: Also, is there a reason for us to do this in the first place 🤔 ? The dialog internally takes care of maintaining selection states. So we just need to get the selected frequency when the "Done" button is clicked. So, we can probably remove this tracking altogether.

app/src/main/java/org/simple/clinic/drugs/selection/custom/drugfrequency/SelectDrugFrequencyDialog.kt
```diff
@@ -0,0 +1,91 @@
...
+    context.injector<Injector>().inject(this)
+  }
+
+  override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+    var selectedValueIndex = frequenciesArrayIndexFromDrugFrequency(screenKey.drugFrequency)
+    return MaterialAlertDialogBuilder(requireContext(), R.style.ThemeOverlay_Simple_Dialog)
+        .setTitle(getString(R.string.custom_drug_entry_sheet_frequency))
+        .setSingleChoiceItems(resources.getStringArray(R.array.custom_drug_entry_sheet_frequencies), selectedValueIndex) { _, indexSelected -> selectedValueIndex = indexSelected }
+        .setPositiveButton(getString(R.string.custom_drug_entry_sheet_frequency_dialog_done)) { _, _ ->
+          router.popWithResult(Succeeded(drugFrequencyFromFrequenciesArrayIndex(selectedValueIndex)))
```
msasikanth, line 43: We can create a list of drug frequencies and use that for rendering the list items and getting frequency based on index.

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryModel.kt
```diff
@@ -25,6 +25,9 @@ data class CustomDrugEntryModel(
         rxNormCode = null)
   }
 
+  val isDrugFrequencyUnknown
```
msasikanth, line 31: We should not be using `Unknown` as a `None` option.

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
```diff
@@ -31,6 +31,12 @@ class CustomDrugEntryUpdate : Update<CustomDrugEntryModel, CustomDrugEntryEvent,
     }
   }
 
+  private fun frequencyEdited(nextModel: CustomDrugEntryModel, event: FrequencyEdited): Next<CustomDrugEntryModel, CustomDrugEntryEffect> {
+    val frequency = if (nextModel.isDrugFrequencyUnknown) null else event.frequency
+
+    return next(nextModel, SetDrugFrequency(frequency), SetSheetTitle(nextModel.drugName, nextModel.dosage, frequency))
```
msasikanth, line 50: Correct me if I am wrong, instead of triggering UI actions for this, can't we just render it in the UI renderer? Is there any reason we are not doing that?

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryModel.kt
```diff
@@ -11,23 +11,29 @@ data class CustomDrugEntryModel(
...
         frequency = null,
-        rxNormCode = null)
+        rxNormCode = null,
+        dosagePlaceholder = dosagePlaceholder)
   }
 
   val isDrugFrequencyUnknown
     get() = frequency is DrugFrequency.Unknown
 
+  val isDosageEqualToPlaceHolderOrEmpty
```
msasikanth, line 34: We should not be doing this to track the placeholder content. We can switch to using suffix feature of the text field ([refer](https://obvious-hq.slack.com/archives/CE0F1TLTC/p1629780370039200)).

But in general, we need to track the text field focus and if a text field has text or not to set the place holder. Please refer how `CustomPrescriptionEntryUiRenderer` does this.

msasikanth: It seems like we are doing this to set empty string if there is only `mg` unit? It is not how it should be handled.

app/src/main/res/layout/sheet_custom_drug_entry.xml
```diff
@@ -34,8 +34,10 @@
     android:layout_height="wrap_content"
     android:layout_alignParentEnd="true"
     android:layout_gravity="center_vertical"
-    android:layout_marginTop="@dimen/spacing_24"
-    android:layout_marginEnd="@dimen/spacing_24"
+    android:layout_marginTop="@dimen/spacing_18"
```
msasikanth, line 36: These margins seem wrong. Usually, spacing dimens in Material are multiples of 4. 18dp & 10dp seems odd. Can you take a look at design spec again, please?

app/src/main/res/layout/sheet_custom_drug_entry.xml
```diff
@@ -35,80 +35,76 @@
...
-        android:id="@+id/drugDosageEditText"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
-        android:hint="@string/custom_drug_entry_sheet_dosage"
-        android:imeOptions="actionNext"
-        android:inputType="textCapWords"
-        android:textColor="@color/color_on_surface_67" />
-
-    </com.google.android.material.textfield.TextInputLayout>
+  <com.google.android.material.textfield.TextInputLayout
```
msasikanth, line 71: The frequency label color always seems to be the primary color. I have confirmed with design, and only text fields that are focused/active need to have label color as primary.

Also, the text color in designs has been updated PTAL.

msasikanth (changes requested):
Left few comments PTAL

app/src/main/java/org/simple/clinic/drugs/selection/custom/drugfrequency/SelectDrugFrequencyDialog.kt
msasikanth: Based on whether or not we need to track the state, please change the approach.

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryModel.kt
janhavisinghh: ✅ But in general, we need to track the text field focus and if a text field has text or not to set the place holder. Please refer how CustomPrescriptionEntryUiRenderer does this.

As discussed offline going with the previous approach of showing `mg` once the user has focused on the dosage edit text and pressed saved even if there is no numeric value added by the user for dosage.

msasikanth (approved):

PR #3007, opened 2021-08-25
Restricting login OTP entries to 5 attempts #1

Honey14:
https://app.clubhouse.io/simpledotorg/story/4514/limit-incorrect-otp-entries

### DIFF
--- a/app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtection.kt
+++ b/app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtection.kt
@@ -1,0 +1,29 @@
+package org.simple.clinic.enterotp
+
+import com.f2prateek.rx.preferences2.Preference
+import org.simple.clinic.util.UtcClock
+import javax.inject.Inject
+
+class BruteForceOtpEntryProtection @Inject constructor(
+    private val utcClock: UtcClock,
+    private val config: BruteForceOtpEntryProtectionConfig,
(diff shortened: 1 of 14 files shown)


msasikanth (changes requested):
There seems to be a circular dagger dependency with the release build, please check.

app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtection.kt
```diff
@@ -0,0 +1,29 @@
...
+      maxAllowedFailedOtpAttempts: Int
+  ): BruteForceOtpProtectionState {
+    val otpState = otpStatePreference.get()
+    val isOtpAttemptLimitReached = otpState.failedLoginOtpAttempt >= maxAllowedFailedOtpAttempts
+    return if (isOtpAttemptLimitReached && !otpState.limitReachedAt.isPresent) {
+      otpState.failedAttemptLimitReached(utcClock)
+    } else {
+      otpState.loginAttemptFailed()
+    }
+  }
```
msasikanth, line 28: We can probably combine these 2 into 1 function? No need to do this now, you can do this in a different PR.

msasikanth (dismissed):

app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtection.kt
Honey14: Okay, adding this in a different PR

msasikanth (approved):

PR #3015, opened 2021-08-30
Migrate v1 `Country` usages to v2 `Country` - 1

msasikanth:
https://app.clubhouse.io/simpledotorg/story/4753/migrate-v1-country-usages-to-v2-country

In later PRs we will migrate all the `Country` usages with `CountryV2`. In this PR, we migrated the `Country#endPoint` usage with `Deployment` usage.

**Testing steps**:
- Before running this PR, sign in into the `master` branch app
- Run the app from this branch
- You should see updated preferences in Flipper
image
- Verify sync is working as expected. You can look at the network panel of Flipper to make sure all calls are successful.

### DIFF
--- a/app/src/main/java/org/simple/clinic/appconfig/AppConfigRepository.kt
+++ b/app/src/main/java/org/simple/clinic/appconfig/AppConfigRepository.kt
@@ -6,6 +6,7 @@ import io.reactivex.Observable
 import io.reactivex.Single
 import org.simple.clinic.appconfig.api.ManifestFetchApi
 import org.simple.clinic.util.ErrorResolver
+import org.simple.clinic.util.toNullable
 import java.util.Optional
 import javax.inject.Inject
 
@@ -30,6 +31,10 @@ class AppConfigRepository @Inject constructor(
 
   fun currentCountryObservable(): Observable<Optional<Country>> {
     return selectedCountryPreference.asObservable()
+  }
+
+  fun currentDeployment(): Deployment? {
+    return selectedDeployment.get().toNullable()
   }
 
   fun fetchAppManifest(): Single<ManifestFetchResult> {
--- a/app/src/main/java/org/simple/clinic/appconfig/AppConfigRepository.kt
+++ b/app/src/main/java/org/simple/clinic/appconfig/AppConfigRepository.kt
@@ -19,7 +19,8 @@ import javax.inject.Inject
  **/
 class AppConfigRepository @Inject constructor(
     private val manifestFetchApi: ManifestFetchApi,
-    private val selectedCountryPreference: Preference<Optional<Country>>
+    private val selectedCountryPreference: Preference<Optional<Country>>,
+    private val selectedCountryV2Preference: Preference<Optional<CountryV2>>
 ) {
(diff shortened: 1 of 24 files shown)


msasikanth:
Moving to draft, I want to check with designers to see if we can have a UI flow to handle this issue instead of this kind of migration.

app/src/main/java/org/simple/clinic/appconfig/AppConfigRepository.kt
```diff
@@ -32,6 +33,10 @@ class AppConfigRepository @Inject constructor(
     return selectedCountryPreference.asObservable()
   }
 
+  fun currentDeployment(): Deployment? {
+    return selectedDeployment.get().toNullable()
```
Honey14, line 37: In which would deployment be nullable? Oh is it because the old versions may not have deployments?

app/src/main/java/org/simple/clinic/di/network/RetrofitModule.kt
```diff
@@ -54,4 +55,22 @@ class RetrofitModule {
...
+  @AppScope
+  @Named("for_deployment")
+  fun retrofitForDeployment(
+      deployment: Deployment,
+      commonRetrofitBuilder: Retrofit.Builder
+  ): Retrofit {
+    // Since the endpoint is not under our control, and is defined at the server level,
+    // this is a safety check that will generate the right endpoint regardless of whether the
+    // endpoint defined in the manifest has a trailing slash or not.
+    val baseUrl = deployment.endPoint.toString().removeSuffix("/")
```
Honey14, line 50: I didn't understand why we're checking this? So there is no possibility of getting the slash in the url that we receive from server but we're checking as a safety precaution?

Honey14 (changes requested):
Looks good overall. Left 2 comments. PTAL
Q: In this PR, we've migrated `country#endpoint` usage to `deployment` usage. Just what the description says. Can you update the PR title to say the same?

msasikanth:
> Looks good overall. Left 2 comments. PTAL
> Q: In this PR, we've migrated `country#endpoint` usage to `deployment` usage. Just what the description says. Can you update the PR title to say the same?

This is part of the migrating v1 country to v2 country. That's why the changelog is still marked as `in-progress` and the it's added as `Migrate v1 Country usages to v2 Country - 1`

app/src/main/java/org/simple/clinic/appconfig/AppConfigRepository.kt
msasikanth: The existing preferences don't contain deployments. So it will be nullable when we try to get it.

app/src/main/java/org/simple/clinic/di/network/RetrofitModule.kt
msasikanth: Retrofit expects the base url to end with one trailing slash. Since we are getting this from server and don't know how it's defined in manifest, we are removing all ending slashes and adding just one.

Honey14 (approved):
LGTM

PR #3017, opened 2021-08-30
Migrate v1 `Country` usages to v2 `Country`

msasikanth:
https://app.clubhouse.io/simpledotorg/story/4753/migrate-v1-country-usages-to-v2-country

**Testing Steps**:
**Migration**:
- Checkout this commit 4ad4d41e57e8862e8b3fec3bb5a23ea8298640fe (The last commit with full old `Country` support)
- Then run the app and sign-in
- Use Flipper to check the shared preferences to make sure the data is in old format
- Checkout this PR
- Run the app
- Verify the data is migrated to the new format
- Check if sync is working

**Fresh install**:
- Clear app data
- Checkout this PR
- Run the app and sign-in
- Check Flipper to make sure the data is in new format.

**Important**: This PR doesn't support state selection yet, so any country with multiple deployments will potentially break the app and use the fallback endpoint. This support will be added in later PRs.

### DIFF
--- a/app/src/main/java/org/simple/clinic/appconfig/AppConfigModule.kt
+++ b/app/src/main/java/org/simple/clinic/appconfig/AppConfigModule.kt
@@ -34,8 +34,8 @@ class AppConfigModule {
   fun provideSelectedCountryV2Preference(
       rxSharedPreferences: RxSharedPreferences,
       moshi: Moshi
-  ): Preference<Optional<CountryV2>> {
-    val countryPreferenceConverter = MoshiObjectPreferenceConverter(moshi, CountryV2::class.java)
+  ): Preference<Optional<Country>> {
+    val countryPreferenceConverter = MoshiObjectPreferenceConverter(moshi, Country::class.java)
     return rxSharedPreferences.getOptional("preference_selected_country_v2", countryPreferenceConverter)
   }
 
--- a/app/src/main/java/org/simple/clinic/appconfig/AppConfigModule.kt
+++ b/app/src/main/java/org/simple/clinic/appconfig/AppConfigModule.kt
@@ -25,8 +25,8 @@ class AppConfigModule {
   fun provideSelectedCountryPreference(
       rxSharedPreferences: RxSharedPreferences,
       moshi: Moshi
-  ): Preference<Optional<Country>> {
-    val countryPreferenceConverter = MoshiObjectPreferenceConverter(moshi, Country::class.java)
+  ): Preference<Optional<Country_Old>> {
+    val countryPreferenceConverter = MoshiObjectPreferenceConverter(moshi, Country_Old::class.java)
     return rxSharedPreferences.getOptional("preference_selected_country_v1", countryPreferenceConverter)
   }
 
--- a/app/src/main/java/org/simple/clinic/appconfig/CountryModule.kt
+++ b/app/src/main/java/org/simple/clinic/appconfig/CountryModule.kt
@@ -30,7 +30,7 @@ object CountryModule {
   }
 
   private fun createDeploymentIfOldCountryIsPresent(appConfigRepository: AppConfigRepository) {
-    val selectedCountry = appConfigRepository.currentCountry().toNullable()
+    val selectedCountry = appConfigRepository.currentCountry_Old().toNullable()
     val selectedDeployment = appConfigRepository.currentDeployment()
 
     if (selectedCountry != null && selectedDeployment == null) {
@@ -44,7 +44,7 @@ object CountryModule {
   }
 
   private fun createV2CountryIfOldCountryIsPresent(appConfigRepository: AppConfigRepository) {
-    val selectedOldCountry = appConfigRepository.currentCountry().toNullable()
+    val selectedOldCountry = appConfigRepository.currentCountry_Old().toNullable()
     val selectedNewCountry = appConfigRepository.currentCountryV2()
 
     if (selectedOldCountry != null && selectedNewCountry == null) {
--- a/app/src/main/java/org/simple/clinic/appconfig/CountryModule.kt
+++ b/app/src/main/java/org/simple/clinic/appconfig/CountryModule.kt
@@ -11,7 +11,7 @@ object CountryModule {
   fun providesCountry(appConfigRepository: AppConfigRepository): Country {
     createV2CountryIfOldCountryIsPresent(appConfigRepository)
 
-    val selectedCountry = appConfigRepository.currentCountryV2()
+    val selectedCountry = appConfigRepository.currentCountry()
 
     requireNotNull(selectedCountry) { "There is no stored country available!" }
 
@@ -45,7 +45,7 @@ object CountryModule {
 
   private fun createV2CountryIfOldCountryIsPresent(appConfigRepository: AppConfigRepository) {
     val selectedOldCountry = appConfigRepository.currentCountry_Old().toNullable()
-    val selectedNewCountry = appConfigRepository.currentCountryV2()
+    val selectedNewCountry = appConfigRepository.currentCountry()
 
     if (selectedOldCountry != null && selectedNewCountry == null) {
       // Since V1 country doesn't have deployment names, we are going with country name
--- a/app/src/main/java/org/simple/clinic/appconfig/CountryModule.kt
+++ b/app/src/main/java/org/simple/clinic/appconfig/CountryModule.kt
@@ -8,7 +8,7 @@ import org.simple.clinic.util.toNullable
 object CountryModule {
 
   @Provides
-  fun providesCountry(appConfigRepository: AppConfigRepository): CountryV2 {
+  fun providesCountry(appConfigRepository: AppConfigRepository): Country {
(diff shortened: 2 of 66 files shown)


msasikanth:
I would have preferred to open a different PR migrating v1 country usages in `SelectCountryScreen`. But since I migrated the provides method and app config repository, it became impossible to do this iteratively. So I had to do this in this PR.

app/src/main/java/org/simple/clinic/appconfig/CountryModule.kt
```diff
@@ -40,4 +42,26 @@ object CountryModule {
       appConfigRepository.saveDeployment(deployment)
     }
   }
+
+  private fun createV2CountryIfOldCountryIsPresent(appConfigRepository: AppConfigRepository) {
+    val selectedOldCountry = appConfigRepository.currentCountry().toNullable()
```
vinaysshenoy, line 47: Given that the old `Country` structure will be needed _exactly once_ at the time of migration and never again, I think we can plan for removing getting the old country via the repository and instead manually parse it using Moshi in this one migration path so that we don't end up using the method to read the old country anywhere else. What do you think?

app/src/main/java/org/simple/clinic/appconfig/CountryModule.kt
```diff
@@ -40,4 +42,26 @@ object CountryModule {
...
+      // Since V1 country doesn't have deployment names, we are going with country name
+      val deployment = Deployment(
+          displayName = selectedOldCountry.displayName,
+          endPoint = selectedOldCountry.endpoint
+      )
+      val country = CountryV2(
+          isoCountryCode = selectedOldCountry.isoCountryCode,
+          displayName = selectedOldCountry.displayName,
+          isdCode = selectedOldCountry.isdCode,
+          deployments = listOf(deployment)
```
vinaysshenoy, line 60: Hmm, for a given installation of the app, we will never need to store all the deployments, just the one the app needs to communicate to the server.

I think we can simplify this by having the `CountryPayload` store a list of deployments and only have a single `Deployment` for each `Country` saved in the local repository when we complete the state and deployment selection. This would mean that we don't need to inject two different things at usage sites and can continue just injecting a single `Country` instance that has a reference to the `Deployment`.

What do you think?

app/src/main/java/org/simple/clinic/appconfig/AppConfigModule.kt
```diff
@@ -25,8 +25,8 @@ class AppConfigModule {
   fun provideSelectedCountryPreference(
       rxSharedPreferences: RxSharedPreferences,
       moshi: Moshi
-  ): Preference<Optional<Country>> {
-    val countryPreferenceConverter = MoshiObjectPreferenceConverter(moshi, Country::class.java)
+  ): Preference<Optional<Country_Old>> {
+    val countryPreferenceConverter = MoshiObjectPreferenceConverter(moshi, Country_Old::class.java)
```
vinaysshenoy, line 29: The eventual plan is to remove providers for `Country_Old` once we finish implementing state selection, right?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/selectcountry/SelectCountryUpdate.kt
```diff
@@ -24,11 +23,13 @@ class SelectCountryUpdate : Update<SelectCountryModel, SelectCountryEvent, Selec
...
 
   private fun countrySaved(model: SelectCountryModel): Next<SelectCountryModel, SelectCountryEffect> {
-    return if (model.hasSelectedACountry() && !model.hasMoreThanOneDeployment) {
+    val effect = if (model.hasSelectedACountry() && !model.hasMoreThanOneDeployment) {
       val selectedCountryDeployment = model.selectedCountry!!.deployments.first()
-      dispatch(SaveDeployment(selectedCountryDeployment))
+      SaveDeployment(selectedCountryDeployment)
     } else {
-      noChange()
+      GoToNextScreen
```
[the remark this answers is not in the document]
msasikanth: Yep, that's the plan. I was planning to do that on the state selection screen. For now, I was just reusing the existing effect.

app/src/main/java/org/simple/clinic/selectcountry/SelectCountryUpdate.kt
```diff
@@ -24,11 +23,13 @@ class SelectCountryUpdate : Update<SelectCountryModel, SelectCountryEvent, Selec
   }
 
   private fun countrySaved(model: SelectCountryModel): Next<SelectCountryModel, SelectCountryEffect> {
-    return if (model.hasSelectedACountry() && !model.hasMoreThanOneDeployment) {
+    val effect = if (model.hasSelectedACountry() && !model.hasMoreThanOneDeployment) {
```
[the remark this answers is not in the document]
msasikanth: No particular reason tbh, to I was just being extra careful I guess. It's expected to have a selected state at this point, so there is no need to have this, and since the model will be restored during process deaths properly. There is no need to have it. I can remove this.

app/src/main/java/org/simple/clinic/selectcountry/SelectCountryUpdate.kt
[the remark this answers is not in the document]
msasikanth: If there is more than one deployment, we let the state selection screen decide which deployment to save. But if there is only one deployment, we are saving it.

app/src/main/java/org/simple/clinic/appconfig/AppConfigModule.kt
msasikanth: I am not sure if that would be possible since we would need a type to parse the JSON from preferences. So for now, it will stay for a while? Unless there is a different approach I am missing.

app/src/main/java/org/simple/clinic/appconfig/CountryModule.kt
msasikanth: Yeah, I am thinking about it. In my mind, I think I liked having a different `Deployment` being injected in only sync APIs, and use `Country` for the rest of the areas. But obviously like you suggested we can just place one deployment after state selection. We can revisit this once state selection is done. I think this should be a straightforward change.

app/src/main/java/org/simple/clinic/appconfig/CountryModule.kt
msasikanth: Yeah, we can do that, we can also just mark the function as deprecated to avoid usage. I personally prefer `deprecated` since I don't have to make many changes 😂 . But let me know what works for you.

vinaysshenoy: Deprecated indicates we plan to remove it. Removal is not what we are intending here since we need to older code to migrate old installations. So I would prefer to do the local parsing and conversion.

app/src/main/java/org/simple/clinic/appconfig/CountryModule.kt
vinaysshenoy: Yeah, I think it will also simplify the migration of the `Country` and enforce safety since we won't be able to save a `Country` separately from a deployment.

app/src/main/java/org/simple/clinic/appconfig/AppConfigModule.kt
vinaysshenoy: Moshi allows us to parse JSON dynamically without creating a type. We could directly read the string from preferences and parse this eventually and remove all traces of the old `Country` model eventually.

vinaysshenoy: I would personally take this approach since we don't need the older type for anything except a one-time migration.

msasikanth: Ah, we would have to directly use `Types.newParameterizedType` right? Let's do that once the entire feature is done then?

app/src/main/java/org/simple/clinic/appconfig/CountryModule.kt
msasikanth: Cool, will make this change after state selection is done. For now creating a story for it.

app/src/main/java/org/simple/clinic/appconfig/CountryModule.kt
msasikanth: If it's fine, I am gonna open a different PR that combines this and the removal of `Country_Old` data class. wdyt?

vinaysshenoy: Yeah, no issues with that.

app/src/main/java/org/simple/clinic/selectcountry/SelectCountryUpdate.kt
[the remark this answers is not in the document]
vinaysshenoy: Ah, got it.

vinaysshenoy (approved):

PR #3019, opened 2021-08-31
Implement the `CallResultSync` layer 

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/4667/add-support-for-the-callresult-sync-resource-in-the-android-app

### Note
The tests are currently disabled because the server side code is not merged in yet. Will raise another PR enabling them once it's available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Change `NewMedicalHistoryScreen` to use Mobius view effects
 - Parse `Country_Old` manually when migrating to v2 `Country`
 - Remove `Country_Old`
-- [In Progress: 30 Aug 2021] Record call results instead of updating the same appointment record
+- [In Progress: 31 Aug 2021] Record call results instead of updating the same appointment record
 - [In Progress: 31 Aug 2021] Refactor logic around providing drug frequencies label depending on the country
 
 ### Changes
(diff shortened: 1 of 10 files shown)


janhavisinghh (changes requested):
The changes look good to me overall. I have a few questions tho. Why are we not testing the `CallResultRepository` changes? Ideally we should have integration tests for all the changes in `CallResultRepositoryAnroidTest`, right? Also there are some code climate issues can you take a look at them as well, please?

> The tests are currently disabled because the server side code is not merged in yet. Will raise another PR enabling them once it's available.

How do we disable tests? 😅

vinaysshenoy:
> The changes look good to me overall. I have a few questions tho. Why are we not testing the `CallResultRepository` changes? Ideally we should have integration tests for all the changes in `CallResultRepositoryAnroidTest`, right? Also there are some code climate issues can you take a look at them as well, please?
> 
> > The tests are currently disabled because the server side code is not merged in yet. Will raise another PR enabling them once it's available.
> 
> How do we disable tests? sweat_smile

> How do we disable tests? sweat_smile
The `CallResultSyncIntegrationTest` class has an `@Ignore` annotation which tells the test suite to not run these tests. In a later PR, I will remove the annotation which will start running these tests as part of the test suite.

> Why are we not testing the `CallResultRepository` changes?
It is not necessary that each class we write also should have a corresponding test class. What matters is that the functionality that the class is supposed to provide is verified in some way. Here, the `CallResultSyncIntegrationTest` is an integration test that will exercise and verify the entire sync resource and related repository methods, so we don't need to test these methods separately.

However, if we add new methods to the repository that are not exercised as part of this suite, we should add tests for those.

>  there are some code climate issues can you take a look at them as well, please
Yup, will correct those.

vinaysshenoy:
I've decided not to correct the CodeClimate issues for now because none of them are related to complexity, but purely method parameter counts for factory methods. These are just stricter constructors and doing a full constructor call would not have triggered these CodeClimate issues anyway, so I'm manually approving these for now. I'll see if there's a better way to handle these kind of issues in CodeClimate at some point.

janhavisinghh (approved):

PR #3023, opened 2021-09-01
Fetch states for selected country deployments

msasikanth:
https://app.clubhouse.io/simpledotorg/story/4035/allow-state-selection-during-app-onboarding?vc_group_by=day&ct_workflow=all&cf_workflow=500000005

### DIFF
--- a/app/src/main/java/org/simple/clinic/appconfig/StatesFetcher.kt
+++ b/app/src/main/java/org/simple/clinic/appconfig/StatesFetcher.kt
@@ -1,0 +1,35 @@
+package org.simple.clinic.appconfig
+
+import okhttp3.HttpUrl.Companion.toHttpUrl
+import org.simple.clinic.appconfig.api.StatesApi
+import retrofit2.Retrofit
+import java.net.ConnectException
+import javax.inject.Inject
+
+class StatesFetcher @Inject constructor(
+    private val retrofitBuilder: Retrofit.Builder
+) {
+
+  @Throws(ConnectException::class)
+  fun fetchStates(deployment: Deployment): List<State> {
+    val statesApi = createStatesApi(deployment)
+    return statesApi
+        .fetchStates()
+        .map { state ->
+          State(
+              displayName = state,
+              deployment = deployment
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/simple/clinic/appconfig/StatesFetcher.kt
```diff
@@ -0,0 +1,35 @@
+package org.simple.clinic.appconfig
+
+import okhttp3.HttpUrl.Companion.toHttpUrl
+import org.simple.clinic.appconfig.api.StatesApi
+import retrofit2.Retrofit
+import java.net.ConnectException
+import javax.inject.Inject
+
+class StatesFetcher @Inject constructor(
+    private val retrofitBuilder: Retrofit.Builder
```
janhavisinghh, line 10: > Since it's possible for the underlying `StatesApi` call to throw connection exception we are marking it for the `fetchStates` call, so that mockito can throw the exception properly.

@msasikanth can you please elaborate on this more? In what cases we have a connection exception? Are we handling this in this case because we are going to be choosing a different server end point which can be rejected? Why are we not handling this when we are initially connected to different servers based on the country?

janhavisinghh (approved):

janhavisinghh:
LGTM. Left a question. PTAL?

app/src/main/java/org/simple/clinic/appconfig/StatesFetcher.kt
msasikanth: >Are we handling this in this case because we are going to be choosing a different server endpoint that can be rejected?

No, basically the API call can throw a `ConnectionException`, it might be because of some connection issues to the server. But since this method is not marked as a can throw an error, Mockito was not able to resolve the connection exception test. So, I have added that to the public method we are exposing.

There might be a better way of doing this, but at this moment this was the best I can think of. @vinaysshenoy can you take a look at it and see if there was a better way of handling it? 🤔  (There won't be any changes to the way it functions in prod, this is just for testing)

PR #3024, opened 2021-09-02
Implement Mobius loop for state selection screen

msasikanth:

### DIFF
--- a/app/src/test/java/org/simple/clinic/appconfig/AppConfigRepositoryTest.kt
+++ b/app/src/test/java/org/simple/clinic/appconfig/AppConfigRepositoryTest.kt
@@ -27,13 +27,15 @@ class AppConfigRepositoryTest {
   private val manifestFetchApi = mock<ManifestFetchApi>()
   private val selectedCountryV2Preference = mock<Preference<Optional<Country>>>()
   private val selectedDeployment = mock<Preference<Optional<Deployment>>>()
+  private val selectedStatePreference = mock<Preference<Optional<State>>>()
   private val statesFetcher = mock<StatesFetcher>()
 
   private val repository = AppConfigRepository(
       manifestFetchApi,
       selectedCountryV2Preference,
       selectedDeployment,
-      statesFetcher
+      selectedStatePreference,
+      statesFetcher,
   )
 
   @Test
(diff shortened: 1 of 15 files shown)


msasikanth:
Blocked by #3023

msasikanth:
While we are saving `State` with deployment to preferences now, eventually we can just combine all of these things in `Country` in preferences. At this moment, `Country` with deployments is needed to load states.

We can drop the `Deployment` from preferences now, since selected state will have the `Deployment`

app/src/test/java/org/simple/clinic/selectstate/SelectStateUpdateTest.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.spotify.mobius.test.UpdateSpec
+import com.spotify.mobius.test.UpdateSpec.assertThatNext
+import okhttp3.MediaType.Companion.toMediaType
+import okhttp3.ResponseBody.Companion.toResponseBody
+import org.junit.Test
+import org.simple.clinic.TestData
+import org.simple.clinic.appconfig.StatesResult.FetchError
+import org.simple.clinic.appconfig.StatesResult.StatesFetched
+import org.simple.clinic.util.ResolvedError
+import retrofit2.HttpException
```
vinaysshenoy, line 14: Having retrofit imports in a test for the business logic seems like a code smell to me. I feel we should move the conversion of http errors to the screen level errors in the effect handler so that the update function does not know about anything outside its domain. What do you think?

app/src/test/java/org/simple/clinic/appconfig/AppConfigRepositoryTest.kt
```diff
@@ -290,4 +292,24 @@ class AppConfigRepositoryTest {
     val fetchedStatesResult = repository.fetchStatesInSelectedCountry()
     assertThat(fetchedStatesResult).isEqualTo(StatesResult.FetchError(ResolvedError.Unexpected(cause)))
   }
+
+  @Test
+  fun `saving the state, should save it to local persistence`() {
```
vinaysshenoy, line 297: Hmm, I feel this verification is more suited to an instrumented test rather than a unit test since it actually verifies the saving instead of a method call.

vinaysshenoy (changes requested):

app/src/test/java/org/simple/clinic/selectstate/SelectStateUpdateTest.kt
msasikanth: Ah, that explains why we are doing it like that in the manifest/country selection screen. I will move them.

app/src/test/java/org/simple/clinic/appconfig/AppConfigRepositoryTest.kt
msasikanth: Yeah, I tried to keep it consistent with existing `AppConfigRepository` tests. Like we discussed will create an integration test, would that be fine?

msasikanth: Along with the API integration test, I think we can move this entire test to be an integration test as well, wdyt?

vinaysshenoy: That works.

vinaysshenoy (approved):

PR #3026, opened 2021-09-02
Remove login failed errors from UiRenderer, and add via effect handler when logging in

Honey14:
https://app.clubhouse.io/simpledotorg/story/4844/add-effects-for-network-error-unexpected-error-to-have-consistent-behaviour-for-login-otp-entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Fetch states for selected country deployments
 - [In Progress: 31 Aug 2021] Record call results instead of updating the same appointment record
 - [In Progress: 31 Aug 2021] Refactor logic around providing drug frequencies label depending on the country
+- Remove login failed errors from UiRenderer, and add via effect handler when logging in
 
 ### Changes
 - Implement providing drug frequencies label depending on the country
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/simple/clinic/enterotp/EnterOtpEffectHandler.kt
```diff
@@ -42,6 +42,7 @@ class EnterOtpEffectHandler @AssistedInject constructor(
         .addTransformer(RequestLoginOtp::class.java, activateUser())
         .addAction(ShowSmsSentMessage::class.java, uiActions::showSmsSentMessage, schedulers.ui())
         .addConsumer(FailedLoginOtpAttempt::class.java, { bruteForceProtection.incrementFailedOtpAttempt() }, schedulers.io())
+        .addAction(ShowNetworkError::class.java, { uiActions.showNetworkError() }, schedulers.ui())
```
vinaysshenoy, line 45: This is a personal preference, I like using method references instead of full lambdas for this sort of thing.

```kotlin
```

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 - [In Progress: 06 Aug 2021] Add support for Sri Lanka
 - [In Progress: 25 Aug 2021] Implement providing drug frequencies label depending on the country
 - [In-progress: 27 Aug 2021] Restrict OTP entries to 5 attempts
+  - Remove login failed errors from UiRenderer, and add via effect handler when logging in
```
vinaysshenoy, line 32: The indentation seems to be incorrect. Can you correct this please?

vinaysshenoy (changes requested):
Overall PR looks good. One error that needs to be corrected in the changelog and we're good to merge this in.

vinaysshenoy (approved):

msasikanth:
This change would be mean that the errors will no longer get restored after a model is restored. Are you planning on going with a different approach for that @Honey14 ? 🤔

PR #3028, opened 2021-09-02
Mobius loop implementations to add a progress state in `CustomDrugEntrySheet` - 1

janhavisinghh:
https://app.clubhouse.io/simpledotorg/story/4783/add-a-progress-state-in-customdrugentrysheet-as-it-s-loading-slowly

### DIFF
--- a/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryInit.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryInit.kt
@@ -7,7 +7,7 @@ import org.simple.clinic.mobius.first
 class CustomDrugEntryInit : Init<CustomDrugEntryModel, CustomDrugEntryEffect> {
   override fun init(model: CustomDrugEntryModel): First<CustomDrugEntryModel, CustomDrugEntryEffect> {
     return when (model.openAs) {
-      is OpenAs.New.FromDrugList -> first(model, FetchDrug(model.openAs.drugUuid), LoadDrugFrequencyChoiceItems)
+      is OpenAs.New.FromDrugList -> first(model.customDrugEntryProgressStateLoading(), FetchDrug(model.openAs.drugUuid), LoadDrugFrequencyChoiceItems)
       is OpenAs.New.FromDrugName -> first(model.drugNameLoaded(model.openAs.drugName), LoadDrugFrequencyChoiceItems)
       is OpenAs.Update -> first(model, FetchPrescription(model.openAs.prescribedDrugUuid), LoadDrugFrequencyChoiceItems)
     }
--- a/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryInit.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryInit.kt
@@ -7,9 +7,9 @@ import org.simple.clinic.mobius.first
 class CustomDrugEntryInit : Init<CustomDrugEntryModel, CustomDrugEntryEffect> {
   override fun init(model: CustomDrugEntryModel): First<CustomDrugEntryModel, CustomDrugEntryEffect> {
     return when (model.openAs) {
-      is OpenAs.New.FromDrugList -> first(model.customDrugEntryProgressStateLoading(), FetchDrug(model.openAs.drugUuid), LoadDrugFrequencyChoiceItems)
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryProgressState.kt
```diff
@@ -0,0 +1,6 @@
+package org.simple.clinic.drugs.selection.custom
+
+enum class CustomDrugEntryProgressState {
```
msasikanth, line 3: Can we rename this to `DrugInfoProgressState`, please? Because we would also want to add a progress state when we are saving a prescribed drug/updating it right? This name doesn't exactly specify which progress state it is. So I think `DrugInfoProgressState` indicates the progress states are for loading drug info. WDYT?

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryInit.kt
```diff
@@ -7,7 +7,7 @@ import org.simple.clinic.mobius.first
 class CustomDrugEntryInit : Init<CustomDrugEntryModel, CustomDrugEntryEffect> {
   override fun init(model: CustomDrugEntryModel): First<CustomDrugEntryModel, CustomDrugEntryEffect> {
     return when (model.openAs) {
-      is OpenAs.New.FromDrugList -> first(model, FetchDrug(model.openAs.drugUuid), LoadDrugFrequencyChoiceItems)
+      is OpenAs.New.FromDrugList -> first(model.customDrugEntryProgressStateLoading(), FetchDrug(model.openAs.drugUuid), LoadDrugFrequencyChoiceItems)
       is OpenAs.New.FromDrugName -> first(model.drugNameLoaded(model.openAs.drugName), LoadDrugFrequencyChoiceItems)
       is OpenAs.Update -> first(model, FetchPrescription(model.openAs.prescribedDrugUuid), LoadDrugFrequencyChoiceItems)
```
msasikanth: I think in general we can probably extract out the model and effect from this when statement, to make it a little bit easier to read?

For example, something like
```kt
val model = when (model.openAs) {
  // handle states
}
```

WDYT?

msasikanth (changes requested):

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryInit.kt
janhavisinghh: I had thought of this before but this doesn't seem like a better approach in this case.. as we have different effects that we are sending in different cases. So, I'll have to make a when case for effects as well. Which ends up increasing the code than reducing it.

msasikanth: Having more lines of code doesn't necessarily mean increased complexity every time, in this scenario we have a couple of repetitions of the same model updates or effects. 

Something like this can be easier to read, or even just making the params in separate lines is also better.

```kt
    val effect = when (model.openAs) {
      is FromDrugList -> FetchDrug(model.openAs.drugUuid)
      is Update -> FetchPrescription(model.openAs.prescribedDrugUuid)
      is FromDrugName -> throw IllegalArgumentException("Should not be called")
    }
    val effects = mutableSetOf(LoadDrugFrequencyChoiceItems, effect)

    val updatedModel = when (model.openAs) {
      is FromDrugList, is Update -> model.customDrugEntryProgressStateLoading()
      else -> model.customDrugEntryProgressStateLoaded()
    }

    return first(updatedModel, effects)
```

msasikanth (commented):
Left a comment PTAL

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryInit.kt
msasikanth: We can address this in a different PR as well

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryModel.kt
```diff
@@ -69,10 +69,10 @@ data class CustomDrugEntryModel(
   }
 
   fun customDrugEntryProgressStateLoaded(): CustomDrugEntryModel {
```
msasikanth, line 71: Can we update these function names as well, please?

msasikanth (dismissed):

msasikanth (approved):

PR #3029, opened 2021-09-03
Move opening of facility selection screen to patient summary screen

vinaysshenoy:
https://app.clubhouse.io/simpledotorg/story/4822/changing-the-assigned-facility-for-a-patient-goes-to-home-screen-instead-of-staying-on-summary-screen

### Note: 
In order to fix this, we need to migrate the facility selection screen to extend the `BaseScreen` and be a part of the navigation graph. However, there is an issue in the current setup where `AssignedFacilityView` is one of the components which starts this screen. This makes migration hard because we cannot handle results in a child view.

This PR moves opening of the facility selection screen to `PatientSummaryScreen`, which will then make it easier for us to migrate `FacilitySelectionScreen`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,10 @@
 - Parse `Country_Old` manually when migrating to v2 `Country`
 - Remove `Country_Old`
 - Fetch states for selected country deployments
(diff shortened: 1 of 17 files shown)


msasikanth (approved):
LGTM

PR #3074, opened 2021-09-06
Implement Mobius loop for state selection screen - 2

msasikanth:

I forgot to add this when creating the Mobius loop 😅 , next PR will wire up the Mobius loop with UI

### DIFF
--- a/app/src/main/java/org/simple/clinic/selectstate/SelectStateUiRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/selectstate/SelectStateUiRenderer.kt
@@ -3,6 +3,7 @@ package org.simple.clinic.selectstate
 import org.simple.clinic.mobius.ViewRenderer
 import org.simple.clinic.selectstate.StatesFetchError.NetworkError
 import org.simple.clinic.selectstate.StatesFetchError.ServerError
+import org.simple.clinic.selectstate.StatesFetchError.UnexpectedError
 
 class SelectStateUiRenderer(
     private val ui: SelectStateUi
@@ -19,6 +20,7 @@ class SelectStateUiRenderer(
     when (model.statesFetchError) {
       NetworkError -> ui.showNetworkErrorMessage()
       ServerError -> ui.showServerErrorMessage()
+      UnexpectedError -> ui.showGenericErrorMessage()
     }
   }
 
--- a/app/src/main/java/org/simple/clinic/selectstate/SelectStateUiRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/selectstate/SelectStateUiRenderer.kt
@@ -2,6 +2,7 @@ package org.simple.clinic.selectstate
 
 import org.simple.clinic.mobius.ViewRenderer
 import org.simple.clinic.selectstate.StatesFetchError.NetworkError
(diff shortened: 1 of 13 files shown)


msasikanth:
Adding missing parts of Mobius loop

app/src/test/java/org/simple/clinic/selectstate/SelectStateUiRendererTest.kt
```diff
@@ -43,16 +44,7 @@ class SelectStateUiRendererTest {
     // then
     verify(ui).showStates(states = states, selectedState = andhraPradesh)
     verify(ui).showNextButton()
-    verifyNoMoreInteractions(ui)
-  }
-
-  @Test
-  fun `when states are not fetched, then hide the states list`() {
```
vinaysshenoy: Why was this test removed? 🤔

app/src/main/java/org/simple/clinic/selectstate/SelectStateUiRenderer.kt
```diff
@@ -10,6 +10,11 @@ class SelectStateUiRenderer(
 ) : ViewRenderer<SelectStateModel> {
 
   override fun render(model: SelectStateModel) {
+    if (model.isFetching) {
+      ui.showProgress()
+      return
```
vinaysshenoy, line 15: I prefer to use `if-else` blocks instead of early returns. Seems more readable to me. What do you think?

app/src/test/java/org/simple/clinic/selectstate/SelectStateUpdateTest.kt
```diff
@@ -47,6 +49,24 @@ class SelectStateUpdateTest {
...
   }
 
+  @Test
+  fun `when state is saved, then go to registration screen`() {
+    val states = listOf(
+        TestData.state(displayName = "Andhra Pradesh"),
+        TestData.state(displayName = "Kerala")
+    )
+    val statesLoadedModel = defaultModel
+        .statesLoaded(states)
```
vinaysshenoy, line 59: We cannot save a state without having selected it, so this model should also have a selected state, right?

vinaysshenoy (changes requested):

app/src/test/java/org/simple/clinic/selectstate/SelectStateUiRendererTest.kt
msasikanth: Basically, this is the same as states fetch error. So removed this.

app/src/test/java/org/simple/clinic/selectstate/SelectStateUpdateTest.kt
msasikanth: Ah, I thought I update it. Adding it 👍🏾

app/src/main/java/org/simple/clinic/selectstate/SelectStateUiRenderer.kt
msasikanth: Yeah, we can do that. Personally I prefer this, but `if-else` also makes sense

vinaysshenoy (approved):

PR #3132, opened 2021-09-07
Allow state selection during app onboarding

msasikanth:

**Testing steps**

- Enable clear text traffic in debug `AndroidManifest` using `android:usesCleartextTraffic="true"` (This is temporary since some deployments have `http` instead of `https`, so we don't need it in prod)
- Replace manifest and fallback endpoints in `gradle.properties` with this: `https://simple-review-pr-2870.herokuapp.com/api/`
- Clear app data and run the app

### DIFF
--- a/app/src/main/java/org/simple/clinic/appconfig/StatesFetcher.kt
+++ b/app/src/main/java/org/simple/clinic/appconfig/StatesFetcher.kt
@@ -13,11 +13,16 @@ class StatesFetcher @Inject constructor(
   @Throws(ConnectException::class)
   fun fetchStates(deployment: Deployment): List<State> {
     val statesApi = createStatesApi(deployment)
-    return statesApi
+    val statesPayload = statesApi
         .fetchStates()
+        .execute()
+        .body()
+    val states = statesPayload?.states.orEmpty()
+
+    return states
         .map { state ->
           State(
-              displayName = state,
+              displayName = state.displayName,
               deployment = deployment
           )
         }
--- a/app/src/main/java/org/simple/clinic/selectstate/SelectStateEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/selectstate/SelectStateEffectHandler.kt
@@ -32,7 +32,8 @@ class SelectStateEffectHandler @AssistedInject constructor(
     return ObservableTransformer { effects ->
       effects
           .observeOn(schedulers.io())
-          .map { appConfigRepository.saveState(it.state) }
+          .doOnNext { appConfigRepository.saveState(it.state) }
+          .doOnNext { appConfigRepository.saveDeployment(it.state.deployment) }
           .map { StateSaved }
     }
   }
--- a/app/src/main/java/org/simple/clinic/selectstate/SelectStateScreen.kt
+++ b/app/src/main/java/org/simple/clinic/selectstate/SelectStateScreen.kt
@@ -1,0 +1,186 @@
+package org.simple.clinic.selectstate
+
+import android.content.Context
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import com.jakewharton.rxbinding3.view.clicks
+import com.spotify.mobius.functions.Consumer
+import io.reactivex.Observable
+import io.reactivex.rxkotlin.cast
+import kotlinx.parcelize.Parcelize
+import org.simple.clinic.R
+import org.simple.clinic.ReportAnalyticsEvents
+import org.simple.clinic.appconfig.State
+import org.simple.clinic.databinding.ListSelectstateStateViewBinding
+import org.simple.clinic.databinding.ScreenSelectStateBinding
+import org.simple.clinic.di.injector
+import org.simple.clinic.navigation.v2.Router
+import org.simple.clinic.navigation.v2.ScreenKey
+import org.simple.clinic.navigation.v2.fragments.BaseScreen
+import org.simple.clinic.registration.phone.RegistrationPhoneScreenKey
+import org.simple.clinic.widgets.ItemAdapter
+import org.simple.clinic.widgets.UiEvent
+import javax.inject.Inject
+
+class SelectStateScreen : BaseScreen<
+    SelectStateScreen.Key,
+    ScreenSelectStateBinding,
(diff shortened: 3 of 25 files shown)


msasikanth:
Let me know if there are too many additions, I can break it down further. @Honey14

msasikanth:
Converting it to draft, the states API response has been updated, so this would break. I Will open PR again with that support.

app/src/main/java/org/simple/clinic/appconfig/StatesFetcher.kt
```diff
@@ -13,11 +13,16 @@ class StatesFetcher @Inject constructor(
...
         .fetchStates()
+        .execute()
+        .body()
+    val states = statesPayload?.states.orEmpty()
+
+    return states
         .map { state ->
           State(
-              displayName = state,
+              displayName = state.displayName,
```
Honey14, line 25: Q: What happens here if we receive an empty statePayload?

app/src/main/java/org/simple/clinic/selectstate/SelectStateEffectHandler.kt
```diff
@@ -32,7 +32,8 @@ class SelectStateEffectHandler @AssistedInject constructor(
     return ObservableTransformer { effects ->
```
Honey14, line 32: Commit message has `persistencem`?

app/src/main/java/org/simple/clinic/selectstate/SelectStateEffectHandler.kt
```diff
@@ -32,7 +32,8 @@ class SelectStateEffectHandler @AssistedInject constructor(
...
       effects
           .observeOn(schedulers.io())
-          .map { appConfigRepository.saveState(it.state) }
+          .doOnNext { appConfigRepository.saveState(it.state) }
+          .doOnNext { appConfigRepository.saveDeployment(it.state.deployment) }
```
Honey14, line 36: Why do we save deployments? Is that for our reference or something else that I'm missing?

Honey14 (commented):
Looks great. Left few comments. PTAL?

app/src/test/java/org/simple/clinic/selectstate/SelectableStateItemTest.kt
```diff
@@ -0,0 +1,54 @@
...
+    assertThat(listItems)
+        .containsExactly(
+            SelectableStateItem(state = andhraPradesh, isStateSelectedByUser = false, showDivider = true),
+            SelectableStateItem(state = kerala, isStateSelectedByUser = true, showDivider = true),
+            SelectableStateItem(state = punjab, isStateSelectedByUser = false, showDivider = false)
+        ).inOrder()
+  }
+
+  @Test
+  fun `if there is only one item in the list, then divider must not be shown`() {
```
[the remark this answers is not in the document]
msasikanth: Deployment is different from the state, so basically we skip state selection if we have only one deployment. But multiple deployments can have the same state? (I don't think that's possible right now). But this is mostly just a divider handling nothing more than that.

app/src/test/java/org/simple/clinic/selectstate/SelectableStateItemTest.kt
```diff
@@ -0,0 +1,54 @@
+package org.simple.clinic.selectstate
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+import org.simple.clinic.TestData
+
+class SelectableStateItemTest {
```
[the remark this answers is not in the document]
msasikanth: If we are doing some transformation from original data, then yeah.

app/src/main/java/org/simple/clinic/selectstate/SelectStateScreen.kt
```diff
@@ -0,0 +1,186 @@
...
+    progressBar.visibility = View.GONE
+  }
+
+  override fun goToRegistrationScreen() {
+    router.push(RegistrationPhoneScreenKey())
+  }
+
+  private fun stateSelectedEvents(): Observable<UiEvent> {
+    return statesAdapter
+        .itemEvents
```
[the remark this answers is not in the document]
msasikanth: Yeah. Once the user clicks on a state, we dispatch a UI event.

app/src/main/java/org/simple/clinic/appconfig/StatesFetcher.kt
msasikanth: I don't think that's possible. But in that scenario, this `map` won't be called and we send an empty list and we show an empty recycler view. Since this list is generated by server folks and not user dependent, I don't think we have to handle those cases.

app/src/main/java/org/simple/clinic/selectstate/SelectStateEffectHandler.kt
msasikanth: Deployments are different from the state, for example, each deployment can have multiple states. So instead of tying a state and deployment together, we are saving them separately.

app/src/main/java/org/simple/clinic/selectstate/SelectStateEffectHandler.kt
msasikanth: Rewording it 👍🏾

app/src/test/java/org/simple/clinic/selectstate/SelectableStateItemTest.kt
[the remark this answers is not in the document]
Honey14: So we're still going to have one radio button with the state even though that state has the same deployment as the country? I know we don't have this scenario right now but something that we should look out for.

Honey14 (approved):
LGTM

app/src/test/java/org/simple/clinic/selectstate/SelectableStateItemTest.kt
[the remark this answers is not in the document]
msasikanth: Not really, this test was mainly to test the divider visibility and nothing else. We don't show a divider for the last item, so just used one item in the list to verify that. If the country just has one deployment, we won't come to this screen in the first place, we will just save that deployment and go to the registration screen from the select country screen.

PR #3133, opened 2021-09-07
Restricting OTP entries to 5 attempts #2

Honey14:
https://app.clubhouse.io/simpledotorg/story/4514/limit-incorrect-otp-entries

### DIFF
--- a/app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtection.kt
+++ b/app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtection.kt
@@ -1,7 +1,12 @@
 package org.simple.clinic.enterotp
 
 import com.f2prateek.rx.preferences2.Preference
+import io.reactivex.Observable
+import io.reactivex.rxkotlin.Observables
 import org.simple.clinic.util.UtcClock
+import org.simple.clinic.util.timer
+import java.time.Duration
+import java.time.Instant
 import javax.inject.Inject
 
 class BruteForceOtpEntryProtection @Inject constructor(
@@ -10,20 +15,83 @@ class BruteForceOtpEntryProtection @Inject constructor(
     private val otpStatePreference: Preference<BruteForceOtpProtectionState>
 ) {
 
+  sealed class ProtectedState {
+    data class Allowed(val attemptsMade: Int, val attemptsRemaining: Int) : ProtectedState()
+
+    data class Blocked(val attemptsMade: Int, val blockedTill: Instant) : ProtectedState()
+  }
+
   fun incrementFailedOtpAttempt() {
-    val updatedBruteForceOtpProtectionState = updateFailedOtpAttemptLimitReached(config.limitOfFailedAttempts)
+    val updatedBruteForceOtpProtectionState = updateFailedOtpAttemptLimitReached()
     otpStatePreference.set(updatedBruteForceOtpProtectionState)
   }
 
-  private fun updateFailedOtpAttemptLimitReached(
-      maxAllowedFailedOtpAttempts: Int
-  ): BruteForceOtpProtectionState {
+  private fun updateFailedOtpAttemptLimitReached(): BruteForceOtpProtectionState {
     val otpState = otpStatePreference.get()
-    val isOtpAttemptLimitReached = otpState.failedLoginOtpAttempt >= maxAllowedFailedOtpAttempts
+    val updatedOtpState = otpState.loginAttemptFailed()
+    val isOtpAttemptLimitReached = config.limitOfFailedAttempts == updatedOtpState.failedLoginOtpAttempt
     return if (isOtpAttemptLimitReached && !otpState.limitReachedAt.isPresent) {
-      otpState.failedAttemptLimitReached(utcClock)
+      updatedOtpState.failedAttemptLimitReached(utcClock)
     } else {
-      otpState.loginAttemptFailed()
+      updatedOtpState
     }
+  }
+
+  fun protectedStateChanges(): Observable<ProtectedState> {
+    val bruteForceProtectionReset = otpStatePreference
+        .asObservable()
+        .switchMap(::signalBruteForceTimerReset)
+        .doOnNext { resetFailedAttempts() }
+        .startWith(Unit)
+
+    return Observables
+        .combineLatest(
+            otpStatePreference.asObservable(),
+            bruteForceProtectionReset
+        )
+        .map { (state, _) -> generateProtectedState(state) }
+        .distinctUntilChanged()
+  }
+
+  private fun generateProtectedState(state: BruteForceOtpProtectionState): ProtectedState {
+    val blockedAt = state.limitReachedAt
+    return blockedAt
+        .map<ProtectedState> {
+          val blockedTill = it + config.blockDuration
+          ProtectedState.Blocked(attemptsMade = state.failedLoginOtpAttempt, blockedTill = blockedTill)
+        }
+        .orElseGet {
+          val attemptsRemaining = config.limitOfFailedAttempts - state.failedLoginOtpAttempt
+          ProtectedState.Allowed(attemptsMade = state.failedLoginOtpAttempt, attemptsRemaining = attemptsRemaining)
+        }
+  }
+
+  private fun signalBruteForceTimerReset(
+      state: BruteForceOtpProtectionState
+  ): Observable<Unit> {
+    return state
+        .limitReachedAt
+        .map {
+          val resetDuration = resetBruteForceTimerIn(it, config.blockDuration)
+          Observables
+              .timer(resetDuration)
+              .map { Unit }
+        }
+        .orElse(Observable.empty())
+  }
+
+  private fun resetBruteForceTimerIn(
+      blockedAt: Instant,
+      blockDuration: Duration
+  ): Duration {
+    val blockExpiresAt = blockedAt + blockDuration
+    val now = Instant.now(utcClock)
+
+    val millisTillExpiry = (blockExpiresAt.toEpochMilli() - now.toEpochMilli()).coerceAtLeast(0)
+    return Duration.ofMillis(millisTillExpiry + 1)
+  }
+
+  private fun resetFailedAttempts() {
+    otpStatePreference.delete()
   }
 }
--- a/app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtection.kt
+++ b/app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtection.kt
@@ -1,8 +1,10 @@
 package org.simple.clinic.enterotp
 
+import android.os.Parcelable
 import com.f2prateek.rx.preferences2.Preference
 import io.reactivex.Observable
 import io.reactivex.rxkotlin.Observables
+import kotlinx.parcelize.Parcelize
 import org.simple.clinic.util.UtcClock
 import org.simple.clinic.util.timer
 import java.time.Duration
@@ -15,9 +17,11 @@ class BruteForceOtpEntryProtection @Inject constructor(
     private val otpStatePreference: Preference<BruteForceOtpProtectionState>
 ) {
 
-  sealed class ProtectedState {
+  sealed class ProtectedState : Parcelable {
+    @Parcelize
     data class Allowed(val attemptsMade: Int, val attemptsRemaining: Int) : ProtectedState()
 
+    @Parcelize
     data class Blocked(val attemptsMade: Int, val blockedTill: Instant) : ProtectedState()
   }
 
--- a/app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtectionConfigModule.kt
+++ b/app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtectionConfigModule.kt
@@ -8,6 +8,6 @@ import java.time.Duration
 class BruteForceOtpEntryProtectionConfigModule {
   @Provides
   fun config(): BruteForceOtpEntryProtectionConfig {
-    return BruteForceOtpEntryProtectionConfig(limitOfFailedAttempts = 4, blockDuration = Duration.ofMinutes(20))
+    return BruteForceOtpEntryProtectionConfig(limitOfFailedAttempts = 5, blockDuration = Duration.ofMinutes(20))
   }
 }
--- a/app/src/main/java/org/simple/clinic/enterotp/EnterOtpEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/enterotp/EnterOtpEffectHandler.kt
@@ -44,7 +44,17 @@ class EnterOtpEffectHandler @AssistedInject constructor(
         .addConsumer(FailedLoginOtpAttempt::class.java, { bruteForceProtection.incrementFailedOtpAttempt() }, schedulers.io())
         .addAction(ShowNetworkError::class.java, uiActions::showNetworkError, schedulers.ui())
         .addAction(ShowUnexpectedError::class.java, uiActions::showUnexpectedError, schedulers.ui())
+        .addTransformer(LoadOtpEntryProtectedStates::class.java, loadOtpEntryStates())
         .build()
+  }
+
+  private fun loadOtpEntryStates(): ObservableTransformer<LoadOtpEntryProtectedStates, EnterOtpEvent> {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulers.io())
+          .switchMap { bruteForceProtection.protectedStateChanges() }
+          .map(::OtpEntryProtectedStateChanged)
+    }
   }
 
   private fun loadUser(): ObservableTransformer<LoadUser, EnterOtpEvent> {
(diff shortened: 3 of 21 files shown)


app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtection.kt
```diff
@@ -26,4 +40,69 @@ class BruteForceOtpEntryProtection @Inject constructor(
...
+  fun protectedStateChanges(): Observable<ProtectedState> {
+    val bruteForceProtectionResets = Observables
+        .combineLatest(config, otpStatePreference.asObservable()) { config, state -> config.blockDuration to state.limitReachedAt }
+        .switchMap { (blockDuration, blockedAt) -> signalBruteForceTimerReset(blockedAt, blockDuration) }
+        .flatMapCompletable { resetFailedAttempts() }
+        .toObservable<Any>()
+
+    return Observables
+        .combineLatest(
+            config,
```
msasikanth, line 63: Same comment as above, `config` is not `Observable`. So we can directly call the params for it in `map`

app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtection.kt
```diff
@@ -26,4 +40,69 @@ class BruteForceOtpEntryProtection @Inject constructor(
...
+      blockDuration: Duration
+  ): Duration {
+    val blockExpiresAt = blockedAt + blockDuration
+    val now = Instant.now(utcClock)
+
+    val millisTillExpiry = (blockExpiresAt.toEpochMilli() - now.toEpochMilli()).coerceAtLeast(0)
+    return Duration.ofMillis(millisTillExpiry + 1)
+  }
+
+  private fun resetFailedAttempts(): Completable {
```
msasikanth, line 115: This can probably be a synchronous call? 🤔

app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtection.kt
```diff
@@ -26,4 +40,69 @@ class BruteForceOtpEntryProtection @Inject constructor(
...
+        )
+        .map { (config, state) ->
+          generateProtectedState(
+              blockedAt = state.limitReachedAt,
+              attemptsMade = state.failedLoginOtpAttempt,
+              maxAllowedFailedAttempts = config.limitOfFailedAttempts,
+              blockAttemptsFor = config.blockDuration
+          )
+        }
+        .distinctUntilChanged()
```
msasikanth, line 75:
```suggestion
val bruteForceProtectionResets = otpStatePreference
    .asObservable()
    .switchMap { state -> signalBruteForceTimerReset(state.limitReachedAt, config.blockDuration) }
    .doOnNext { resetFailedAttempts() }
return Observables
    .combineLatest(
        otpStatePreference.asObservable(),
        bruteForceProtectionResets
    )
    .map { (state) ->
      generateProtectedState(
          blockedAt = state.limitReachedAt,
          attemptsMade = state.failedLoginOtpAttempt,
          maxAllowedFailedAttempts = config.limitOfFailedAttempts,
          blockAttemptsFor = config.blockDuration
      )
    }
    .distinctUntilChanged()
```

Overall, something like this perhaps? WDYT?

app/src/main/res/layout/screen_enterotp.xml
```diff
@@ -74,6 +74,7 @@
       android:indeterminate="true"
       android:visibility="invisible" />
 
+
```
msasikanth, line 77: Extra line?

app/src/main/res/layout/screen_enterotp.xml
```diff
@@ -94,17 +95,53 @@
...
+        android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:layout_gravity="center_horizontal"
-        android:gravity="center_horizontal"
-        android:imeOptions="actionDone"
-        android:inputType="number"
-        android:maxLength="6"
-        tools:text="123456" />
+        android:layout_marginTop="@dimen/spacing_24"
+        android:measureAllChildren="false"
```
msasikanth, line 103: It's by default false, so no need to specify it.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpEffectHandler.kt
```diff
@@ -44,6 +45,7 @@ class EnterOtpEffectHandler @AssistedInject constructor(
         .addConsumer(FailedLoginOtpAttempt::class.java, { bruteForceProtection.incrementFailedOtpAttempt() }, schedulers.io())
         .addAction(ShowNetworkError::class.java, uiActions::showNetworkError, schedulers.ui())
         .addAction(ShowUnexpectedError::class.java, uiActions::showUnexpectedError, schedulers.ui())
+        .addAction(AllowOtpEntry::class.java, { uiActions.showOtpEntryMode(OtpEntry) }, schedulers.ui())
```
msasikanth, line 50: Is there a reason why this is not part of the model and UI renderer? One of the reasons why that is important is, UI actions don't get restored after state restoration or process death.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
```diff
@@ -22,6 +24,14 @@ class EnterOtpUpdate(
...
       is RequestLoginOtpCompleted -> requestOtpCompleted(model, event)
       is EnterOtpResendSmsClicked -> next(model.requestLoginOtpStarted(), RequestLoginOtp as EnterOtpEffect)
+      is OtpEntryProtectedStateChanged -> effectsForStateChanged(event.stateChanged)
+    }
+  }
+
+  private fun effectsForStateChanged(stateChanged: ProtectedState): Next<EnterOtpModel, EnterOtpEffect> {
+    return when (stateChanged) {
+      is Allowed -> dispatch(AllowOtpEntry)
+      else -> // do something
```
msasikanth: Correct me if I am wrong, depending on the protected state we dispatch either allow or block effects right? Instead of having 2 effects for changing the OTP entry mode. Why not pass the `OtpEntryMode` to a single effect from here? 🤔 

Like this
```kt
val otpEntryMode = when (stateChanged) {
  // do something
}
return dispatch(ChangeOtpEntryMode(otpEntryMode))
```

While this is working with UI effects, I think ideally this should be done in UI renderer.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
```diff
@@ -30,11 +30,18 @@ class EnterOtpUpdate(
...
     return when (stateChanged) {
-      is Allowed -> dispatch(AllowOtpEntry)
+      is Allowed -> dispatch(generateEffectForAllowingOtpEntry(stateChanged), AllowOtpEntry)
       else -> // do something
     }
   }
 
+  private fun generateEffectForAllowingOtpEntry(protectedState: Allowed): EnterOtpEffect {
+    return if (protectedState.attemptsMade == 0) {
+      HideErrors
```
msasikanth, line 41: Usually, we just return `HideErrors` directly and in later commits we add the `if-else` condition.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
```diff
@@ -31,7 +32,7 @@ class EnterOtpUpdate(
   private fun effectsForStateChanged(stateChanged: ProtectedState): Next<EnterOtpModel, EnterOtpEffect> {
     return when (stateChanged) {
       is Allowed -> dispatch(generateEffectForAllowingOtpEntry(stateChanged), AllowOtpEntry)
-      else -> // do something
+      is Blocked -> dispatch(BlockOtpEntryUntil(stateChanged.blockedTill))
```
msasikanth: Same comment as before, where we can have just one UI effect for changing the OTP entry mode.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpEffectHandler.kt
```diff
@@ -49,6 +49,7 @@ class EnterOtpEffectHandler @AssistedInject constructor(
         .addAction(AllowOtpEntry::class.java, { uiActions.showOtpEntryMode(OtpEntry) }, schedulers.ui())
         .addConsumer(BlockOtpEntryUntil::class.java, { uiActions.showOtpEntryMode(BruteForceOtpEntryLocked(it.blockTill)) }, schedulers.ui())
         .addAction(HideErrors::class.java, { uiActions.hideError() }, schedulers.ui())
+        .addConsumer(ShowIncorrectOtpLimitReachedError::class.java, { uiActions.showLimitReachedError(it.attemptsMade) }, schedulers.ui())
```
msasikanth, line 53: I feel like we are handling all the UI rendering in UI actions, we should not be doing that. UI actions are mostly for navigation or fire and forget UI operations.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpEffectHandler.kt
```diff
@@ -50,6 +50,7 @@ class EnterOtpEffectHandler @AssistedInject constructor(
         .addConsumer(BlockOtpEntryUntil::class.java, { uiActions.showOtpEntryMode(BruteForceOtpEntryLocked(it.blockTill)) }, schedulers.ui())
         .addAction(HideErrors::class.java, { uiActions.hideError() }, schedulers.ui())
         .addConsumer(ShowIncorrectOtpLimitReachedError::class.java, { uiActions.showLimitReachedError(it.attemptsMade) }, schedulers.ui())
+        .addConsumer(ShowIncorrectOtpError::class.java, { uiActions.showIncorrectOtpErrorAttempt(it.attemptsMade, it.attemptsRemaining) }, schedulers.ui())
```
msasikanth, line 54: Same comment as before, is there a reason why we are using UI actions for everything? 😅

app/src/test/java/org/simple/clinic/enterotp/EnterOtpEffectHandlerTest.kt
```diff
@@ -126,4 +126,14 @@ class EnterOtpEffectHandlerTest {
...
     verifyNoMoreInteractions(uiActions)
   }
+
+  @Test
+  fun `when load the otp entry protected state effect is received, then load the protected states`() {
+    // when
+    testCase.dispatch(LoadOtpEntryProtectedStates)
+
+    // then
+    verify(bruteForceOtpEntryProtection).protectedStateChanges()
```
msasikanth, line 136: We also should be verifying the outgoing event here.

app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtectionConfigModule.kt
```diff
@@ -2,12 +2,13 @@ package org.simple.clinic.enterotp
...
 import dagger.Provides
+import io.reactivex.Observable
 import java.time.Duration
 
 @Module
 class BruteForceOtpEntryProtectionConfigModule {
   @Provides
-  fun config(): BruteForceOtpEntryProtectionConfig {
-    return BruteForceOtpEntryProtectionConfig(limitOfFailedAttempts = 4, blockDuration = Duration.ofMinutes(20))
+  fun config(): Observable<BruteForceOtpEntryProtectionConfig> {
```
msasikanth, line 11: I don't personally see a reason for this to be `Observable`, since the values are constant for the config. Let me know if I got something wrong here.

app/src/main/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtection.kt
```diff
@@ -18,26 +18,36 @@ class BruteForceOtpEntryProtection @Inject constructor(
...
-
   sealed class ProtectedState {
     data class Allowed(val attemptsMade: Int, val attemptsRemaining: Int) : ProtectedState()
     data class Blocked(val attemptsMade: Int, val blockedTill: Instant) : ProtectedState()
   }
 
-  fun incrementFailedOtpAttempt() {
-    val updatedBruteForceOtpProtectionState = updateFailedOtpAttemptLimitReached(config.limitOfFailedAttempts)
-    otpStatePreference.set(updatedBruteForceOtpProtectionState)
+  fun incrementFailedOtpAttempt(): Completable {
```
msasikanth, line 26: This can be a synchronous call. Is there a reason we changed it back to return `Completable`? 🤔

msasikanth (changes requested):
Overall approach LGTM, left few comments. There are also seem to be quite a few unit tests that are breaking. PTAL.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
Honey14: Ah I see, noted boss

app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
Honey14: Ah makes sense, updating this.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
Honey14: Ah makes sense, updating this to be included in UiRenderer

app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
```diff
@@ -31,7 +32,7 @@ class EnterOtpUpdate(
   private fun effectsForStateChanged(model: EnterOtpModel, stateChanged: ProtectedState): Next<EnterOtpModel, EnterOtpEffect> {
     return when (stateChanged) {
       is Allowed -> next(model.setOtpEntryMode(stateChanged))
-      else -> // do something
+      is Blocked -> next(model.setOtpEntryMode(stateChanged))
```
msasikanth: We can just extract out the state and just call `model.setOtpEntryMode` one time.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpScreen.kt
```diff
@@ -186,6 +186,15 @@ class EnterOtpScreen : BaseScreen<
     showError(resources.getString(R.string.otpentry_error_incorrect_otp_attempts_limit_reached, attemptsMade.toString()))
   }
 
+  override fun showFailedAttemptOtpError(attemptsMade: Int, attemptsRemaining: Int) {
+    if (attemptsMade in 1..2) {
+      showError(resources.getString(R.string.otpentry_error_incorrect_otp_attempt))
+    } else if (attemptsMade in 3..4) {
+      showError(resources.getString(R.string.otpentry_error_incorrect_otp_attempts_remaining, attemptsRemaining.toString()))
+    }
```
msasikanth, line 248: Instead of this, could we have just used this? 

```suggestion
if (attemptsMode >= 3) {
  // show error with attempts remaning
} else {
  // show error
}
```
I think this is much more understandable, wdyt?

app/src/main/java/org/simple/clinic/enterotp/EnterOtpUiRenderer.kt
```diff
@@ -41,12 +41,20 @@ class EnterOtpUiRenderer(
...
       }
       is Blocked -> {
         ui.showOtpEntryMode(BruteForceOtpEntryLocked(model.protectedState.blockedTill))
         ui.showLimitReachedError(model.protectedState.attemptsMade)
       }
     }
   }
+
+  private fun generateUiForAllowingOtpEntry(attemptsMade: Int, attemptsRemaining: Int) {
+    if (attemptsMade == 0) {
```
msasikanth, line 54: Hmm, I think we are using magic numbers in this commit. For example, deciding which error format to show or whether to hide/show an error. Instead, can we move them to a model, so that they don't lose context?

Like, `model.hasNoIncorrectPinEntries` or `mode.hasReachedMinPinRetries` (for showing error with attempts remaining, etc.,)

app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
```diff
@@ -53,11 +53,12 @@ class EnterOtpUpdate(
...
     val updatedModel = model.loginFinished()
+    val loginFailedModel = model.loginFailed()
     return when (val result = event.result) {
       LoginResult.Success -> next(updatedModel, ClearLoginEntry, TriggerSync)
-      is LoginResult.ServerError -> dispatch(FailedLoginOtpAttempt(result), ClearPin)
-      LoginResult.NetworkError -> dispatch(ShowNetworkError, ClearPin)
-      LoginResult.UnexpectedError -> dispatch(ShowUnexpectedError, ClearPin)
+      is LoginResult.ServerError -> next(loginFailedModel, FailedLoginOtpAttempt(result), ClearPin)
+      LoginResult.NetworkError -> next(loginFailedModel, ShowNetworkError, ClearPin)
+      LoginResult.UnexpectedError -> next(loginFailedModel, ShowUnexpectedError, ClearPin)
```
msasikanth, line 61: This works, but I think this could've been handled a little bit cleaner. We are creating 2 different models here at the same time.

msasikanth (changes requested):
Left few comments PTAL. Primarily I am looking for magic number fixes. We can just move those to `model` and give a name to make it more meaningful.

msasikanth (approved):

PR #3144, opened 2021-09-13
Remove magic numbers from `EnterOtpScreen`

Honey14:
https://app.shortcut.com/simpledotorg/story/4975/remove-magic-numbers-from-enterotpscreen

### DIFF
--- a/app/src/main/java/org/simple/clinic/enterotp/EnterOtpModel.kt
+++ b/app/src/main/java/org/simple/clinic/enterotp/EnterOtpModel.kt
@@ -15,17 +15,21 @@ data class EnterOtpModel(
     val isAsyncOperationOngoing: Boolean,
     val protectedState: ProtectedState,
 
+    val minOtpRetries: Int
     ) : Parcelable {
 
   companion object {
 
-    fun create(): EnterOtpModel {
+    fun create(
+        minOtpRetries: Int
+    ): EnterOtpModel {
       return EnterOtpModel(
           user = null,
           otpValidationResult = ValidationResult.NotValidated,
           asyncOpError = null,
           isAsyncOperationOngoing = false,
-          protectedState = ProtectedState.Allowed(0,5)
+          protectedState = ProtectedState.Allowed(0,5),
+          minOtpRetries = minOtpRetries
       )
     }
   }
@@ -34,7 +38,7 @@ data class EnterOtpModel(
     get() = (protectedState as ProtectedState.Allowed).attemptsMade == 0
 
   val hasReachedMinPinRetries: Boolean
-    get() = (protectedState as ProtectedState.Allowed).attemptsMade >= 3
+    get() = (protectedState as ProtectedState.Allowed).attemptsMade >= minOtpRetries
 
   val hasLoadedUser: Boolean
     get() = user != null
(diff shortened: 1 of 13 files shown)


app/src/test/java/org/simple/clinic/enterotp/EnterOtpUiRendererTest.kt
```diff
@@ -64,4 +64,21 @@ class EnterOtpUiRendererTest {
     verify(ui).showOtpEntryMode(BruteForceOtpEntryLocked(blockedUntil))
     verify(ui).showLimitReachedError(attemptsMade)
   }
+
+  @Test
+  fun `when the otp failed attempt is more than 3, then allow otp entries with failed attempts error`() {
```
vinaysshenoy, line 69: The problem with this test is that it is dependent on the implementation detail that the maximum number of PIN attempts is 3 and that it is hardcoded in the model. This makes the test covariant with the production code and should be avoided.

The right way to fix this is to accept the entry limit as a dependency in the model and then provide it when initialising the model. In production code, we can make this part of the DI graph with a configuration object, while in tests they can be directly instantiated.

vinaysshenoy (changes requested):

app/src/test/java/org/simple/clinic/enterotp/EnterOtpUiRendererTest.kt
Honey14: So the maximum OTP entries is configured in the DI graph in `BruteForceOtpEntryProtectionConfigModule`, this is minimum entries for rendering error messages. So allowing and blocking is handled via the config. Do we need to add the minimumOTPEntries number in the DI graph as well? 
Refer `EnterOtpUiRenderer#generateUiForAllowingOtpEntry` for context.

Honey14: So the maximum OTP entries is configured in the DI graph in BruteForceOtpEntryProtectionConfigModule, this is minimum entries for rendering error messages. So allowing and blocking is handled via the config. Do we need to add the minimumOTPEntries number in the DI graph as well?
Refer EnterOtpUiRenderer#generateUiForAllowingOtpEntry for context.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpModel.kt
```diff
@@ -15,24 +15,31 @@ data class EnterOtpModel(
...
+    fun create(
+        minOtpRetries: Int
+    ): EnterOtpModel {
       return EnterOtpModel(
           user = null,
           otpValidationResult = ValidationResult.NotValidated,
           asyncOpError = null,
           isAsyncOperationOngoing = false,
-          protectedState = ProtectedState.Allowed(0,5)
+          protectedState = ProtectedState.Allowed(0,5),
```
vinaysshenoy, line 31: Hmm, I also see we are hardcoding 5 here for the `ProtectedState.Allowed` constructor. If that is meant to be the same as a limit of failed attempts configuration property, then this should be accepted via a dependency instead of being hardcoded as well.

vinaysshenoy (changes requested):
Overall PR looks good, I noticed one more magic number being hardcoded which needs clarification and then this PR will be good to go.

vinaysshenoy (approved):

PR #3145, opened 2021-09-13
Reset OTP attempts in preferences on success and resend of OTP

Honey14:
https://app.clubhouse.io/simpledotorg/story/4840/on-resend-and-success-delete-the-otp-entries-from-preference

### DIFF
--- a/app/src/main/java/org/simple/clinic/enterotp/EnterOtpEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/enterotp/EnterOtpEffectHandler.kt
@@ -45,6 +45,7 @@ class EnterOtpEffectHandler @AssistedInject constructor(
         .addAction(ShowNetworkError::class.java, uiActions::showNetworkError, schedulers.ui())
         .addAction(ShowUnexpectedError::class.java, uiActions::showUnexpectedError, schedulers.ui())
         .addTransformer(LoadOtpEntryProtectedStates::class.java, loadOtpEntryStates())
+        .addAction(ResetOtpAttemptLimit::class.java, { bruteForceProtection.resetFailedAttempts() }, schedulers.io())
         .build()
   }
 
--- a/app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
@@ -36,7 +36,7 @@ class EnterOtpUpdate(
     val updatedModel = model.requestLoginOtpFinished()
 
     return when (val result = event.result) {
-      is ActivateUser.Result.Success -> next(updatedModel, ClearPin, ShowSmsSentMessage)
+      is ActivateUser.Result.Success -> next(updatedModel, ClearPin, ShowSmsSentMessage, ResetOtpAttemptLimit)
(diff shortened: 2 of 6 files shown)


app/src/main/java/org/simple/clinic/enterotp/EnterOtpEffectHandler.kt
```diff
@@ -45,6 +45,7 @@ class EnterOtpEffectHandler @AssistedInject constructor(
         .addAction(ShowNetworkError::class.java, uiActions::showNetworkError, schedulers.ui())
         .addAction(ShowUnexpectedError::class.java, uiActions::showUnexpectedError, schedulers.ui())
         .addTransformer(LoadOtpEntryProtectedStates::class.java, loadOtpEntryStates())
+        .addConsumer(ResetOtpAttemptLimit::class.java, { bruteForceProtection.resetFailedAttempts() }, schedulers.io())
```
msasikanth, line 48: Since we are not expecting any data from the effect here, we can use `addAction` instead.

```suggestion
addAction(ResetOtpAttemptLimit::class.java, bruteForceProtection::resetFailedAttempts, schedulers.io())
```

app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
```diff
@@ -48,7 +48,7 @@ class EnterOtpUpdate(
     val updatedModel = model.loginFinished()
     val loginFailedModel = updatedModel.loginFailed()
```
msasikanth, line 49: Can we please update this approach in a future PR. We are creating a login failed model and login finished model irrespective of the results.

msasikanth (changes requested):
Left a comment PTAL

app/src/main/java/org/simple/clinic/enterotp/EnterOtpEffectHandler.kt
Honey14: Ah right, my bad.

app/src/main/java/org/simple/clinic/enterotp/EnterOtpUpdate.kt
Honey14: Yes, will do

app/src/main/java/org/simple/clinic/enterotp/EnterOtpEffectHandler.kt
```diff
@@ -45,6 +45,7 @@ class EnterOtpEffectHandler @AssistedInject constructor(
         .addAction(ShowNetworkError::class.java, uiActions::showNetworkError, schedulers.ui())
         .addAction(ShowUnexpectedError::class.java, uiActions::showUnexpectedError, schedulers.ui())
         .addTransformer(LoadOtpEntryProtectedStates::class.java, loadOtpEntryStates())
+        .addAction(ResetOtpAttemptLimit::class.java, { bruteForceProtection.resetFailedAttempts() }, schedulers.io())
```
msasikanth, line 48: Just a personal preference, we can use method reference here.

```suggestion
        .addAction(ResetOtpAttemptLimit::class.java, bruteForceProtection::resetFailedAttempts, schedulers.io())
```

msasikanth (approved):

PR #3149, opened 2021-09-15
Record call results when updating appointments in `ContactPatientBottomSheet`

vinaysshenoy:
https://app.shortcut.com/simpledotorg/story/4669/record-a-callresult-whenever-a-call-is-made-in-contactpatientbottomsheet

### DIFF
--- a/app/src/main/java/org/simple/clinic/contactpatient/RecordPatientAgreedToVisit.kt
+++ b/app/src/main/java/org/simple/clinic/contactpatient/RecordPatientAgreedToVisit.kt
@@ -1,16 +1,39 @@
 package org.simple.clinic.contactpatient
 
+import dagger.Lazy
 import org.simple.clinic.overdue.Appointment
 import org.simple.clinic.overdue.AppointmentRepository
+import org.simple.clinic.overdue.callresult.CallResult
+import org.simple.clinic.overdue.callresult.CallResultRepository
+import org.simple.clinic.patient.SyncStatus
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/simple/clinic/contactpatient/RecordPatientAgreedToVisit.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.simple.clinic.util.UserClock
+import javax.inject.Inject
+
+class RecordPatientAgreedToVisit @Inject constructor(
+    private val appointmentRepository: AppointmentRepository,
+    private val userClock: UserClock
+) {
+
+  fun execute(appointment: Appointment) {
+    appointmentRepository.markAsAgreedToVisit(appointment.uuid, userClock)
```
Honey14, line 25: This removes a few dependencies from the effect handler. Is our goal here to increase readability with extraction or something else?

Honey14 (approved):
LGTM. Left one comment,

app/src/main/java/org/simple/clinic/contactpatient/RecordPatientAgreedToVisit.kt
vinaysshenoy: Sort of. The goal is to extract related logic into an atomic flow so that it can be reused if needed. Here, this was done because the updating of the appointment and creating the call result should be a single operation.

PR #3151, opened 2021-09-15
Change `SynceableRepository#save` to be a synchronous call

vinaysshenoy:


app/src/main/java/org/simple/clinic/drugs/PrescriptionRepository.kt
```diff
@@ -82,7 +82,7 @@ class PrescriptionRepository @Inject constructor(
     return Completable.fromAction { dao.save(records) }
   }
 
-  fun saveImmediate(records: List<PrescribedDrug>) {
+  override fun saveImmediate(records: List<PrescribedDrug>) {
     return dao.save(records)
```
msasikanth, line 83: We can probably drop the `return` here?

msasikanth (changes requested):
LGTM. More synchronous code 🙌🏾 

Left one comment, just a small change.

app/src/main/java/org/simple/clinic/drugs/PrescriptionRepository.kt
vinaysshenoy: Good catch, done.

msasikanth (approved):

PR #3155, opened 2021-09-16
Remove maximum character limit for phone numbers

msasikanth:
https://app.clubhouse.io/simpledotorg/story/5018/remove-maximum-character-limit-for-phone-numbers


janhavisinghh (approved):
LGTM overall. Might be a nitpick but I was wondering why don't we have any validation check for special characters in phone number fields? Currently I can type `######` and it will pass as a valid phone number.

msasikanth:
> LGTM overall. Might be a nitpick but I was wondering why don't we have any validation check for special characters in phone number fields? Currently I can type `######` and it will pass as a valid phone number.

This PR is just for removing the other length based number validations. Also, that we are just trying to keep it simple in terms of validation, that's the reason why we are removing the different length based validation for different countries.

janhavisinghh (approved):

msasikanth:
reopening to see if the CI tests will be resolved

PR #3157, opened 2021-09-16
Record call results when removing an appointment from the overdue list

vinaysshenoy:
https://app.shortcut.com/simpledotorg/story/4669/record-a-callresult-whenever-a-call-is-made-in-contactpatientbottomsheet

### DIFF
--- a/app/src/main/java/org/simple/clinic/removeoverdueappointment/RemoveOverdueEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/removeoverdueappointment/RemoveOverdueEffectHandler.kt
@@ -56,7 +56,7 @@ class RemoveOverdueEffectHandler @AssistedInject constructor(
     return ObservableTransformer { effects ->
       effects
           .observeOn(schedulersProvider.io())
-          .doOnNext { appointmentRepository.cancelWithReason(it.appointmentUuid, it.reason) }
+          .doOnNext { appointmentRepository.cancelWithReason(it.appointment.uuid, it.reason) }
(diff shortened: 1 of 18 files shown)


app/src/main/java/org/simple/clinic/removeoverdueappointment/RemoveOverdueEffectHandler.kt
```diff
@@ -56,7 +57,7 @@ class RemoveOverdueEffectHandler @AssistedInject constructor(
     return ObservableTransformer { effects ->
       effects
           .observeOn(schedulersProvider.io())
-          .doOnNext { appointmentRepository.cancelWithReason(it.appointment.uuid, it.reason) }
+          .doOnNext { cancelAppointmentWithReason.execute(it.appointment, it.reason) }
```
janhavisinghh, line 60: Seems like this was a part of the next commit?

janhavisinghh (changes requested):

app/src/main/java/org/simple/clinic/removeoverdueappointment/RemoveOverdueEffectHandler.kt
vinaysshenoy: It wasn't. This commit represents the extraction of current behaviour (cancelling an appointment) to a discrete use case. The next commit is the one in which we add an additional side effect to that use case to record a call result.

janhavisinghh (approved):
LGTM

janhavisinghh (approved):

PR #3158, opened 2021-09-16
Hide keyboard when OTP entries are blocked

Honey14:
https://app.shortcut.com/simpledotorg/story/5038/hide-keyboard-when-otp-entries-are-blocked

### DIFF
--- a/app/src/test/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtectionTest.kt
+++ b/app/src/test/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtectionTest.kt
@@ -53,7 +53,7 @@ class BruteForceOtpEntryProtectionTest {
   }
 
   @Test
-  fun `when incrementing the otp failed attempts and the otp limit was already reached then the blocked-at time should not be updated`() {
+  fun `when otp limit was already reached then the blocked-at time and the number of attempts should not be updated`() {
     val timeOfLastAttempt = Instant.now(clock)
     val bruteForceOtpProtectionState = BruteForceOtpProtectionState(
         failedLoginOtpAttempt = config.limitOfFailedAttempts,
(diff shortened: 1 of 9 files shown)


app/src/test/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtectionTest.kt
```diff
@@ -64,7 +64,7 @@ class BruteForceOtpEntryProtectionTest {
     bruteForceOtpEntryProtection.incrementFailedOtpAttempt()
 
     verify(preferenceState).set(BruteForceOtpProtectionState(
-        failedLoginOtpAttempt = config.limitOfFailedAttempts + 1,
+        failedLoginOtpAttempt = config.limitOfFailedAttempts,
```
vinaysshenoy, line 67: Hmm, how is reducing the number of limits reached sufficient to test the new behaviour? We added a new branch, but no new tests have been added, so I assume we are missing at least one more test to verify the new branch, right?

vinaysshenoy (changes requested):

app/src/test/java/org/simple/clinic/enterotp/BruteForceOtpEntryProtectionTest.kt
Honey14: So +1 was the problem. After the user has reached the max attempts, the blocked at time wasn't getting updated but the number of attempts was getting updated. I think the only thing we need to update in this test is the test name.
  fun `when otp limit was already reached then the blocked-at time and the number of attempts should not be updated`

Honey14: If you look at `BruteForceOtpEntryProtectionTest`, all three scenarios are covered.

vinaysshenoy (approved):

PR #3159, opened 2021-09-16
Implement providing drug frequencies label depending on the country in `DrugSearchScreen`

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/5043/implement-providing-drug-frequencies-label-depending-on-the-country-in-drugsearchscreen

### DIFF
--- a/app/src/main/java/org/simple/clinic/drugs/search/DrugSearchListItem.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/search/DrugSearchListItem.kt
@@ -22,7 +22,11 @@ sealed class DrugSearchListItem : PagingItemAdapter.Item<Event> {
 
   companion object {
 
-    fun from(searchResults: PagingData<Drug>, searchQuery: String, drugFrequencyToFrequencyChoiceItemMap: Map<DrugFrequency?, DrugFrequencyChoiceItem>): PagingData<DrugSearchListItem> {
+    fun from(
+        searchResults: PagingData<Drug>,
+        searchQuery: String,
+        drugFrequencyToFrequencyChoiceItemMap: Map<DrugFrequency?, DrugFrequencyChoiceItem>
+    ): PagingData<DrugSearchListItem> {
       return searchResults
           .map{DrugSearchResult(it, drugFrequencyToFrequencyChoiceItemMap)}
           .insertSeparators(SOURCE_COMPLETE) { oldItem, newItem ->
--- a/app/src/main/java/org/simple/clinic/drugs/search/DrugSearchListItem.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/search/DrugSearchListItem.kt
@@ -13,6 +13,7 @@ import org.simple.clinic.R
 import org.simple.clinic.databinding.ListItemDrugSearchBinding
 import org.simple.clinic.databinding.ListItemDrugSearchCornerCapBinding
 import org.simple.clinic.drugs.search.DrugSearchListItem.Event
+import org.simple.clinic.drugs.selection.custom.drugfrequency.country.DrugFrequencyChoiceItem
 import org.simple.clinic.widgets.PagingItemAdapter
 import org.simple.clinic.widgets.dp
 import org.simple.clinic.widgets.recyclerview.BindingViewHolder
@@ -21,9 +22,9 @@ sealed class DrugSearchListItem : PagingItemAdapter.Item<Event> {
 
   companion object {
 
-    fun from(searchResults: PagingData<Drug>, searchQuery: String): PagingData<DrugSearchListItem> {
+    fun from(searchResults: PagingData<Drug>, searchQuery: String, drugFrequencyToFrequencyChoiceItemMap: Map<DrugFrequency?, DrugFrequencyChoiceItem>): PagingData<DrugSearchListItem> {
       return searchResults
-          .map(::DrugSearchResult)
+          .map{DrugSearchResult(it, drugFrequencyToFrequencyChoiceItemMap)}
           .insertSeparators(SOURCE_COMPLETE) { oldItem, newItem ->
             insertSeparators(oldItem, newItem, searchQuery)
           }
@@ -49,7 +50,7 @@ sealed class DrugSearchListItem : PagingItemAdapter.Item<Event> {
     }
   }
 
-  data class DrugSearchResult(val drug: Drug) : DrugSearchListItem() {
+  data class DrugSearchResult(val drug: Drug, val drugFrequencyToFrequencyChoiceItemMap: Map<DrugFrequency?, DrugFrequencyChoiceItem>) : DrugSearchListItem() {
 
     override fun layoutResId() = R.layout.list_item_drug_search
 
@@ -72,7 +73,8 @@ sealed class DrugSearchListItem : PagingItemAdapter.Item<Event> {
         }
 
         if (drug.frequency != null) {
-          append("$drugNameSeparator${drug.frequency}")
+          val frequencyLabel = drugFrequencyToFrequencyChoiceItemMap[drug.frequency]!!.label
+          append("$drugNameSeparator${frequencyLabel}")
         }
       }
     }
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
```diff
@@ -128,7 +131,7 @@ class DrugsSearchScreen : BaseScreen<
   override fun setDrugSearchResults(searchResults: PagingData<Drug>, drugFrequencyToFrequencyChoiceItemMap: Map<DrugFrequency?, DrugFrequencyChoiceItem>) {
     val searchQuery = searchQueryEditText.text?.toString().orEmpty()
     drugSearchResultsList.scrollToPosition(0)
-    adapter.submitData(lifecycle, DrugSearchListItem.from(searchResults, searchQuery))
+    adapter.submitData(lifecycle, DrugSearchListItem.from(searchResults, searchQuery, drugFrequencyToFrequencyChoiceItemMap))
```
msasikanth, line 134: Just an FYI, this should be done in the next commit.

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchModel.kt
```diff
@@ -2,16 +2,19 @@ package org.simple.clinic.drugs.search
 
 import android.os.Parcelable
 import kotlinx.parcelize.Parcelize
+import org.simple.clinic.drugs.selection.custom.drugfrequency.country.DrugFrequencyChoiceItem
 
 @Parcelize
 data class DrugSearchModel(
-    val searchQuery: String
+    val searchQuery: String,
+    val drugFrequencyToFrequencyChoiceItemMap: Map<DrugFrequency?, DrugFrequencyChoiceItem>?
```
msasikanth, line 10: If we are using this in the UI layer only, can't we just inject it directly to the adapter or screen? WDYT? 🤔

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchUpdate.kt
```diff
@@ -17,10 +16,19 @@ class DrugSearchUpdate : Update<DrugSearchModel, DrugSearchEvent, DrugSearchEffe
...
-      is DrugFrequencyChoiceItemsLoaded -> noChange()
+      is DrugFrequencyChoiceItemsLoaded -> drugFrequencyChoiceItemsLoaded(model, event)
     }
   }
 
+  private fun drugFrequencyChoiceItemsLoaded(
+      model: DrugSearchModel,
+      event: DrugFrequencyChoiceItemsLoaded
+  ): Next<DrugSearchModel, DrugSearchEffect> {
+    val drugFrequencyToFrequencyChoiceItemMap = event.drugFrequencyChoiceItems.items.associateBy({ it.drugFrequency }, { it })
```
msasikanth, line 27: I see we are converting the list to a map for the drug choice items, can't we just provide it as a map directly?

msasikanth (changes requested):
Looks good overall, I have left one question related to the approach. Also, I feel `DrugChoiceItems` might be a good name for this, we might have to consider changing it in future. For example, the name indicates we are using it for choice items, but we are using it for the conversion of drug frequency labels.

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchUpdate.kt
janhavisinghh: Umm.. We are mapping it as DrugFrequency or MedicineFrequency to DrugFrequencyChoiceItem. Although you are right that approach sounds better and will remove a lot of code. We can merge this in for now. I will make that change in all the places and remove the unnecessary code.

msasikanth: Yeah, sounds good. I am not saying we need to do this change here. I was just checking to see if there was any particular reason for going with this approach. We can also combine `DrugFrequency` & `MedicineFrequency` into one thing, right? Technically, I don't think there is any difference between them 🤔

janhavisinghh: Yeah, we can take the whole DrugFrequency implementation out in a separate package and use DrugFrequency everywhere. Maintaining two different data classes is not necessary at all here.

app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
```diff
@@ -49,6 +52,11 @@ class DrugsSearchScreen : BaseScreen<
   @Inject
   lateinit var router: Router
 
+  @Inject
+  lateinit var drugFrequencyFactory: DrugFrequencyFactory
+
+  private var drugFrequencyToFrequencyChoiceItemMap: Map<DrugFrequency?, DrugFrequencyChoiceItem>? = null
```
msasikanth, line 58: Is there a reason why it has to be a mutable property? and also can't we just have this inside the `onViewCreated` as well?

app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
```diff
@@ -105,6 +113,9 @@ class DrugsSearchScreen : BaseScreen<
   override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
     super.onViewCreated(view, savedInstanceState)
 
+    val drugFrequencyChoiceItems = DrugFrequencyChoiceItems(drugFrequencyFactory.provideFields())
+    drugFrequencyToFrequencyChoiceItemMap = drugFrequencyChoiceItems.items.associateBy({ it.drugFrequency }, { it })
```
msasikanth, line 117: Can't we just use this instead? 🤔 . We can even do this inside the adapter as well and just pass in the factory/items in the screen.

```suggestion
val drugFrequencyToFrequencyChoiceItemMap = drugFrequencyFactory
    .provideFields()
    .associateBy({ it.drugFrequency }, { it })
```

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchListItem.kt
```diff
@@ -21,9 +22,9 @@ sealed class DrugSearchListItem : PagingItemAdapter.Item<Event> {
 
   companion object {
 
-    fun from(searchResults: PagingData<Drug>, searchQuery: String): PagingData<DrugSearchListItem> {
+    fun from(searchResults: PagingData<Drug>, searchQuery: String, drugFrequencyToFrequencyChoiceItemMap: Map<DrugFrequency?, DrugFrequencyChoiceItem>): PagingData<DrugSearchListItem> {
```
msasikanth, line 25: Can we put the arguments in separate lines, please? It crosses our character limit in a single line.

app/src/main/java/org/simple/clinic/drugs/search/DrugSearchListItem.kt
```diff
@@ -49,7 +50,7 @@ sealed class DrugSearchListItem : PagingItemAdapter.Item<Event> {
     }
   }
 
-  data class DrugSearchResult(val drug: Drug) : DrugSearchListItem() {
+  data class DrugSearchResult(val drug: Drug, val drugFrequencyToFrequencyChoiceItemMap: Map<DrugFrequency?, DrugFrequencyChoiceItem>) : DrugSearchListItem() {
```
msasikanth, line 53: Same comment as above

app/src/main/java/org/simple/clinic/drugs/search/DrugsSearchScreen.kt
```diff
@@ -138,7 +138,7 @@ class DrugsSearchScreen : BaseScreen<
   override fun setDrugSearchResults(searchResults: PagingData<Drug>) {
     val searchQuery = searchQueryEditText.text?.toString().orEmpty()
     drugSearchResultsList.scrollToPosition(0)
-    adapter.submitData(lifecycle, DrugSearchListItem.from(searchResults, searchQuery))
+    adapter.submitData(lifecycle, DrugSearchListItem.from(searchResults, searchQuery, drugFrequencyToFrequencyChoiceItemMap!!))
```
msasikanth, line 141: So yeah, if we follow the above suggestion where we create this inside the adapter. We can avoid creating a nullable property in the screen.

msasikanth (changes requested):

msasikanth (approved):

PR #3161, opened 2021-09-17
Fix patient summary going back to the home screen when opening teleconsult links

msasikanth:
https://app.shortcut.com/simpledotorg/story/5073/fix-patient-summary-going-back-to-the-home-screen-when-opening-teleconsult-links


Honey14 (approved):
LGTM

PR #3166, opened 2021-09-19
Use Kotlin script for PR comment check GH Action

msasikanth:
https://app.shortcut.com/simpledotorg/story/5085/use-kotlin-script-for-pr-comment-check-gh-action

### DIFF
--- a/.github/scripts/PrCommentCheck.main.kts
+++ b/.github/scripts/PrCommentCheck.main.kts
@@ -1,0 +1,16 @@
+#!/usr/bin/env kscript
+
+import kotlin.system.exitProcess
+
+val prDescription = args.firstOrNull().orEmpty()
+val storyRegex = Regex("https:\\/\\/app.shortcut.com\\/simpledotorg\\/story\\/[0-9]+\\/[A-Za-z0-9_-]+")
+
+println("Checking PR description!")
+
+if (prDescription.contains(storyRegex)) {
+  println("PR description contains a valid story link.")
+  exitProcess(0)
+} else {
+  println("PR description doesn't contain any valid story links.")
+  exitProcess(1)
+}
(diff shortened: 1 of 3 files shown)


.github/scripts/PrCommentCheck.main.kts
```diff
@@ -0,0 +1,16 @@
...
+
+import kotlin.system.exitProcess
+
+val prDescription = args.firstOrNull().orEmpty()
+val storyRegex = Regex("https:\\/\\/app.shortcut.com\\/simpledotorg\\/story\\/[0-9]+\\/[A-Za-z0-9_-]+")
+
+println("Checking PR description!")
+
+if (prDescription.contains(storyRegex)) {
+  println("PR description contains a valid story link.")
```
Honey14, line 11: Why do we need printing statements?

Honey14 (changes requested):
LGTM. Left one comment

.github/scripts/PrCommentCheck.main.kts
msasikanth: Just for action logs, that's all.

msasikanth: and we are not exposing any sensitive content, so I think this should be good for now.

Honey14: Got it. Let's delete this later then

Honey14 (approved):
LGTM

.github/scripts/PrCommentCheck.main.kts
msasikanth: I don't think there is any reason to delete them. The print statements are basically just a message we show when an action is successful or a failure. Since the exit process cannot have a message, we are doing it with this.

PR #3167, opened 2021-09-20
Hide resend sms button when OTP attempts are blocked

Honey14:
https://app.shortcut.com/simpledotorg/story/5079/when-the-timer-has-blocked-otp-attempts-hide-the-resend-sms-button


msasikanth (approved):
LGTM

PR #3172, opened 2021-09-20
Add progress state when the drug is being added/updated in `CustomDrugEntrySheet`

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/5045/add-progress-state-when-the-drug-is-being-added-updated

### DIFF
--- a/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryModel.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryModel.kt
@@ -86,4 +86,8 @@ data class CustomDrugEntryModel(
   fun buttonProgressStateIsSaved(): CustomDrugEntryModel {
     return copy(saveButtonState = SAVED)
   }
+
+  fun saveButtonStateChanged(saveButtonState: ButtonState): CustomDrugEntryModel {
+    return copy(saveButtonState = saveButtonState)
+  }
 }
--- a/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryModel.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryModel.kt
@@ -3,6 +3,8 @@ package org.simple.clinic.drugs.selection.custom
 import android.os.Parcelable
 import kotlinx.parcelize.Parcelize
 import org.simple.clinic.drugs.search.DrugFrequency
+import org.simple.clinic.drugs.selection.custom.ButtonState.SAVED
+import org.simple.clinic.drugs.selection.custom.ButtonState.SAVING
 import org.simple.clinic.drugs.selection.custom.DrugInfoProgressState.DONE
 import org.simple.clinic.drugs.selection.custom.DrugInfoProgressState.IN_PROGRESS
 import org.simple.clinic.drugs.selection.custom.drugfrequency.country.DrugFrequencyChoiceItem
@@ -18,7 +20,8 @@ data class CustomDrugEntryModel(
     val dosagePlaceholder: String,
     val drugFrequencyChoiceItems: List<DrugFrequencyChoiceItem>?,
     val drugFrequencyToFrequencyChoiceItemMap: Map<DrugFrequency?, DrugFrequencyChoiceItem>?,
-    val drugInfoProgressState: DrugInfoProgressState?
+    val drugInfoProgressState: DrugInfoProgressState?,
+    val saveButtonState: ButtonState?
 ) : Parcelable {
   companion object {
     fun default(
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
```diff
@@ -6,7 +6,6 @@ import com.spotify.mobius.Update
 import org.simple.clinic.drugs.PrescribedDrug
```
msasikanth, line 6: The commit message seems to be wrong, can you check it, please?

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryModel.kt
```diff
@@ -71,4 +78,12 @@ data class CustomDrugEntryModel(
...
     return copy(drugInfoProgressState = IN_PROGRESS)
   }
+
+  fun buttonProgressStateIsSaving(): CustomDrugEntryModel {
+    return copy(saveButtonState = SAVING)
+  }
+
+  fun buttonProgressStateIsSaved(): CustomDrugEntryModel {
+    return copy(saveButtonState = SAVED)
+  }
```
msasikanth, line 88: We can just have one function that says, `saveButtonStateChanged` and pass in the params right? 

```kt
fun saveButtonStateChanged(buttonState: ButtonState): CustomDrugEntryModel {
  return copy(saveButtonState = buttonState)
}
```

app/src/test/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdateTest.kt
```diff
@@ -107,51 +107,51 @@ class CustomDrugEntryUpdateTest {
...
   }
 
 
   @Test
-  fun `when add button is clicked and the sheet is opened in create mode from drug name with edited dosage and frequency values, then add the drug to the custom drug list`() {
+  fun `when add button is clicked and the sheet is opened in create mode from drug name with edited dosage and frequency values, then add the drug to the custom drug list and set button progress state to saving`() {
     val dosage = "200 mg"
     val frequency = OD
-    val defaultModel = CustomDrugEntryModel.default(openAs = OpenAs.New.FromDrugName(drugName), dosagePlaceholder).drugNameLoaded(drugName)
+    val model = CustomDrugEntryModel.default(openAs = OpenAs.New.FromDrugName(drugName), dosagePlaceholder).drugNameLoaded(drugName).dosageEdited(dosage).frequencyEdited(frequency)
```
msasikanth, line 131: Can we break these into multiple lines, please? There is too much happening in a single line.

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUiRenderer.kt
```diff
@@ -22,6 +22,9 @@ class CustomDrugEntryUiRenderer(
         setSheetTitle(model.drugName, model.dosage, model.drugFrequencyToFrequencyChoiceItemMap[model.frequency]!!.label)
 
       showDefaultDosagePlaceholder(model.dosage, model.dosageHasFocus)
+
+      if (model.isSaveButtonInProgressState)
+        ui.showSaveButtonProgressState()
```
msasikanth, line 26: Just an FYI, if there is no else condition we can just have it in a single line.

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryEffect.kt
```diff
@@ -43,3 +43,5 @@ data class RemoveDrugFromPrescription(val drugUuid: UUID) : CustomDrugEntryEffec
 object LoadDrugFrequencyChoiceItems : CustomDrugEntryEffect()
 
 object HideKeyboard : CustomDrugEntryEffect()
+
```
msasikanth, line 46: Shouldn't the keyboard changes be in a different PR, considering this PR is for adding progress state in the buttons?

msasikanth (changes requested):
LGTM. Left few comments PTAL

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryModel.kt
janhavisinghh: umm yeah. We can do that in `drugInfoProgressStateLoading` as well right? Can you tell me how do we decide which one is the right approach for doing this in which scenario? I took this approach to maintain consistency in the model for progress states.

msasikanth: I guess this is more of a personal preference than the correct approach. For example, if we decided to add another state to the button state, we would need to add another function here. Instead of that, we have been using a single function and just passing the state from the update function. 

You can refer few other places where we used `ButtonState`s.

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryEffect.kt
janhavisinghh: Raising a different pr for it 👍🏻

msasikanth (approved):

PR #3174, opened 2021-09-21
Don't show appointment sheet when patient is dead

msasikanth:
https://app.shortcut.com/simpledotorg/story/5098/go-back-to-the-previous-screen-when-done-back-is-clicked-in-the-patient-summary-screen-when-the-patient-is-marked-as-dead

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
@@ -24,7 +24,7 @@ class PatientSummaryUpdate : Update<PatientSummaryModel, PatientSummaryEvent, Pa
     return when (event) {
       is PatientSummaryProfileLoaded -> patientSummaryProfileLoaded(model, event)
       is PatientSummaryBackClicked -> backClicked(model, event)
-      is PatientSummaryDoneClicked -> dispatch(LoadDataForDoneClick(model.patientUuid))
+      is PatientSummaryDoneClicked -> doneClicked()
       is CurrentUserAndFacilityLoaded -> currentUserAndFacilityLoaded(model, event)
       PatientSummaryEditClicked -> dispatch(HandleEditClick(model.patientSummaryProfile!!, model.currentFacility!!))
       is ScheduledAppointment -> dispatch(TriggerSync(event.sheetOpenedFrom))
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
```diff
@@ -54,6 +54,10 @@ class PatientSummaryUpdate : Update<PatientSummaryModel, PatientSummaryEvent, Pa
     }
   }
 
+  private fun backClicked(): Next<PatientSummaryModel, PatientSummaryEffect> {
+    return dispatch(GoBackToPreviousScreen)
```
vinaysshenoy: The commit message includes the qualifier of the patient being dead, but it doesn't seem like we are checking the patient status in this commit. Should we reword the message?

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
```diff
@@ -54,6 +54,10 @@ class PatientSummaryUpdate : Update<PatientSummaryModel, PatientSummaryEvent, Pa
     }
   }
 
+  private fun doneClicked(): Next<PatientSummaryModel, PatientSummaryEffect> {
+    return dispatch(GoToHomeScreen)
```
vinaysshenoy: Same comment as earlier about the commit message not matching the actual change in this commit.

app/src/test/java/org/simple/clinic/summary/PatientSummaryUpdateTest.kt
```diff
@@ -995,6 +995,21 @@ class PatientSummaryUpdateTest {
...
+
+    updateSpec
+        .given(model)
+        .whenEvent(PatientSummaryDoneClicked(patientUuid))
+        .then(assertThatNext(
+            hasNoModel(),
+            hasEffects(LoadDataForDoneClick(patientUuid))
+        ))
+  }
+
```
vinaysshenoy, line 1012: Given that we have added some Mobius style tests for existing user flows in `PatientSummaryScreen`, should we add the remaining effect handler tests and then remove the tests pertaining to this flow from the older regression tests?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
msasikanth: Ah, I forgot I can just return `noChange`.  Making that change.

app/src/test/java/org/simple/clinic/summary/PatientSummaryUpdateTest.kt
msasikanth: Yeah, I can do that in a different PR?

vinaysshenoy: Yup, that works.

vinaysshenoy (approved):

PR #3175, opened 2021-09-21
Refactor `ContactPatientUiRenderer`

msasikanth:


app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
```diff
@@ -316,11 +316,11 @@ class ContactPatientBottomSheet : BaseBottomSheet<
     )
```
Honey14, line 316: Multiple things are happening in one commit, can we rename and remove in two different commits

Honey14 (changes requested):
LGTM, left feedback. PTAL

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientBottomSheet.kt
msasikanth: It's mostly just reshuffling of functions, extracting them out. I can separate the renaming functions to a new commit, though.

msasikanth: done ✅

Honey14 (approved):
LGTM

PR #3176, opened 2021-09-21
Show patient died status in contact patient bottom sheet

msasikanth:
https://app.shortcut.com/simpledotorg/story/5090/show-patient-died-status-in-contact-patient-bottom-sheet

### DIFF
--- a/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
@@ -47,6 +47,12 @@ class ContactPatientUiRenderer(
       renderPatientFacilityLabel(model.appointmentIsInRegisteredFacility)
     }
 
+    renderPhoneNumberAndCallResults(model)
+
+    ui.switchToCallPatientView()
+  }
+
+  private fun renderPhoneNumberAndCallResults(model: ContactPatientModel) {
     when {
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
```diff
@@ -57,6 +58,14 @@ class ContactPatientUiRenderer(
     ui.switchToCallPatientView()
   }
 
+  private fun renderDeadPatientWithPhoneNumber(isSecureCallingEnabled: Boolean) {
+    ui.showPatientWithPhoneNumberUi()
```
janhavisinghh, line 66: are we not rendering patient dead status in other ui states like patient with no phone number for example because we just want to prevent the user from calling a dead patient?

janhavisinghh: ohh nvm I got it. This is because in the other cases the sheet will not be opened. Right?

janhavisinghh (approved):
LGTM

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
msasikanth: A patient who is dead, and doesn't have a phone number cannot reach this state. We don't show dead patients in the overdue list. So they can only trigger this sheet from the patient summary screen. If the patient doesn't have a phone number, we don't show the call button in the summary screen. So I think we can just handle that one case.

PR #3179, opened 2021-09-21
Show change diagnosis dialog when patient is not diagnosed with HTN or diabetes when registering

msasikanth:
https://app.shortcut.com/simpledotorg/story/5087/show-change-diagnosis-dialog-when-patient-is-not-diagnosed-with-htn-or-diabetes-when-registering

### DIFF
--- a/app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryModel.kt
+++ b/app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryModel.kt
@@ -64,6 +64,9 @@ data class NewMedicalHistoryModel(
   val hasNoDiabetes: Boolean
     get() = ongoingMedicalHistoryEntry.hasDiabetes == No
 
+  val showChangeDiagnosisError: Boolean
+    get() = !hasShownChangeDiagnosisError && hasNoHypertension && hasNoDiabetes
+
   companion object {
     fun default(country: Country): NewMedicalHistoryModel = NewMedicalHistoryModel(
         country = country,
--- a/app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryModel.kt
+++ b/app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryModel.kt
@@ -94,4 +94,8 @@ data class NewMedicalHistoryModel(
   fun patientRegistered(): NewMedicalHistoryModel {
     return copy(nextButtonState = ButtonState.SAVED)
   }
+
+  fun changeDiagnosisErrorShown(): NewMedicalHistoryModel {
+    return copy(hasShownChangeDiagnosisError = true)
+  }
 }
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryScreen.kt
```diff
@@ -94,7 +94,7 @@ class NewMedicalHistoryScreen : BaseScreen<
   private val diabetesDiagnosis
     get() = binding.diabetesDiagnosis
 
-  private val questionViewEvents: Subject<NewMedicalHistoryEvent> = PublishSubject.create()
+  private val hotEvents: Subject<NewMedicalHistoryEvent> = PublishSubject.create()
```
Honey14, line 97: What does hotEvents mean?

app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryModel.kt
```diff
@@ -56,6 +57,12 @@ data class NewMedicalHistoryModel(
   val showOngoingDiabetesTreatment: Boolean
     get() = diagnosedWithDiabetes && country.isoCountryCode == Country.INDIA
 
+  val hasNoHypertension: Boolean
+    get() = ongoingMedicalHistoryEntry.diagnosedWithHypertension == No
+
+  val hasNoDiabetes: Boolean
```
Honey14, line 64: We're only using `hasNoHypertension` and `hasNoDiabetes` in the model, can we make this private

Honey14 (changes requested):
LGTM, left feedback. PTAL?

app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryModel.kt
msasikanth: Ah, good catch. I moved them to resolve to code climate issue and forgot to change visibility.

app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryScreen.kt
msasikanth: It's basically a stream that receives any `NewMedicalHistoryEvent`. For example for certain scenarios we cannot create a stream similar to button clicks or text changes, etc., So we just pass the event to this. This will be then added to `events` function.

Honey14: I understand the meaning, what I meant was the name is weirdly off.

msasikanth: Yeah, it's something we are using in rest of the project 😅

Honey14 (approved):
LGTM

PR #3180, opened 2021-09-22
Convert commit and push bash script to Kotlin script

msasikanth:


msasikanth:
I was just trying out how to convert the existing bash script to Kotlin script. Feel free to close it if you don't want to migrate existing bash scripts.

.github/scripts/Extensions.kts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+fun File.runCommand(command: String): Process {
+  val parts = command.split("\\s".toRegex())
+
+  return ProcessBuilder(*parts.toTypedArray())
+      .directory(this)
+      .start()
+      .apply { waitFor() }
+}
```
vinaysshenoy, line 17: I don't think these should be extension functions on `java.io.File`, because they are not applicable to all `File` instances, only specific ones. I feel these should just be top level functions instead which accept a `File` as a parameter. What do you think?

vinaysshenoy (changes requested):

.github/scripts/Extensions.kts
msasikanth: I initially went with that but had to pass the working directory in every function. So I just went with extensions, we can move these extensions in the `CommitAndPush.main.kts` file so that they are scoped to that script. WDYT?

vinaysshenoy: Yeah, that works

vinaysshenoy (approved):

PR #3181, opened 2021-09-22
Provide drug frequencies to label as a map

msasikanth:
https://app.shortcut.com/simpledotorg/story/5142/provide-drug-frequencies-to-label-as-a-map


Honey14 (approved):
LGTM

PR #3182, opened 2021-09-22
Custom Drug Entry Sheet UI Improvements

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/5046/custom-drug-entry-sheet-ui-improvements

This PR:
- Removes the focus from dosage when user clicks on frequency
- Saves `null` in the repository if dosage edit text value is equal to "mg"
- Changes from numeric keyboard to the default keyboard for drug dosage edit text
- Uses `joinToString` to correctly display dosage and frequency values in `EditMedicineScreen`

Next PR:
- Make the cursor appear right before 'mg' whenever the sheet is opened

### DIFF
--- a/app/src/main/java/org/simple/clinic/drugs/selection/DrugListItem.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/selection/DrugListItem.kt
@@ -45,9 +45,8 @@ data class ProtocolDrugListItem(
     viewBinding.protocoldrugItemName.isChecked = prescribedDrug != null
 
     viewBinding.protocoldrugItemDosageAndFrequency.visibleOrGone(prescribedDrug != null)
-    val medicineFrequency = if (prescribedDrug?.frequency == null) "" else medicineFrequencyToLabelMap[prescribedDrug.frequency]!!.label
-    val dosage = if(prescribedDrug?.dosage == null) "" else prescribedDrug.dosage
-    val dosageAndFrequency = "$dosage, $medicineFrequency"
+    val medicineFrequency = if (prescribedDrug?.frequency == null) null else medicineFrequencyToLabelMap[prescribedDrug.frequency]!!.label
+    val dosageAndFrequency = listOfNotNull(prescribedDrug?.dosage, medicineFrequency).joinToString()
     viewBinding.protocoldrugItemDosageAndFrequency.text = dosageAndFrequency
 
     viewBinding.prescribeddrugItemProtocoldrugRootlayout.shapeAppearanceModel = shapeAppearanceModel(hasTopCorners)
@@ -76,9 +75,8 @@ data class CustomPrescribedDrugListItem(
     viewBinding.prescribeddrugItemCustomdrugName.text = prescribedDrug.name
     viewBinding.prescribeddrugItemCustomdrugName.isChecked = true
 
-    val medicineFrequency = if (prescribedDrug.frequency == null) "" else medicineFrequencyToLabelMap[prescribedDrug.frequency]!!.label
-    val dosage = prescribedDrug.dosage
-    val dosageAndFrequency = "$dosage, $medicineFrequency"
+    val medicineFrequency = if (prescribedDrug.frequency == null) null else medicineFrequencyToLabelMap[prescribedDrug.frequency]!!.label
+    val dosageAndFrequency = listOfNotNull(prescribedDrug.dosage, medicineFrequency).joinToString()
     viewBinding.protocoldrugItemDosageAndFrequency.text = dosageAndFrequency
 
     viewBinding.root.setOnClickListener {
--- a/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
@@ -19,7 +19,7 @@ class CustomDrugEntryUpdate : Update<CustomDrugEntryModel, CustomDrugEntryEvent,
     return when (event) {
       is DosageEdited -> next(model.dosageEdited(event.dosage))
       is DosageFocusChanged -> next(model.dosageFocusChanged(event.hasFocus))
-      is EditFrequencyClicked -> dispatch(ShowEditFrequencyDialog(model.frequency))
+      is EditFrequencyClicked -> dispatch(ShowEditFrequencyDialog(model.frequency), ClearFocusFromDosageEditText)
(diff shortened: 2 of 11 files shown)


app/src/main/java/org/simple/clinic/drugs/selection/DrugListItem.kt
```diff
@@ -45,9 +45,8 @@ data class ProtocolDrugListItem(
     viewBinding.protocoldrugItemName.isChecked = prescribedDrug != null
 
     viewBinding.protocoldrugItemDosageAndFrequency.visibleOrGone(prescribedDrug != null)
-    val medicineFrequency = if (prescribedDrug?.frequency == null) "" else medicineFrequencyToLabelMap[prescribedDrug.frequency]!!.label
-    val dosage = if(prescribedDrug?.dosage == null) "" else prescribedDrug.dosage
-    val dosageAndFrequency = "$dosage, $medicineFrequency"
+    val medicineFrequency = if (prescribedDrug?.frequency == null) null else medicineFrequencyToLabelMap[prescribedDrug.frequency]!!.label
+    val dosageAndFrequency = listOfNotNull(prescribedDrug?.dosage, medicineFrequency).joinToString()
```
msasikanth, line 49: Honestly, for a couple of items, it's just fine to hardcode. Was IDE showing some lint warning?

app/src/main/java/org/simple/clinic/drugs/selection/DrugListItem.kt
```diff
@@ -76,9 +75,8 @@ data class CustomPrescribedDrugListItem(
     viewBinding.prescribeddrugItemCustomdrugName.text = prescribedDrug.name
     viewBinding.prescribeddrugItemCustomdrugName.isChecked = true
 
-    val medicineFrequency = if (prescribedDrug.frequency == null) "" else medicineFrequencyToLabelMap[prescribedDrug.frequency]!!.label
-    val dosage = prescribedDrug.dosage
-    val dosageAndFrequency = "$dosage, $medicineFrequency"
+    val medicineFrequency = if (prescribedDrug.frequency == null) null else medicineFrequencyToLabelMap[prescribedDrug.frequency]!!.label
+    val dosageAndFrequency = listOfNotNull(prescribedDrug.dosage, medicineFrequency).joinToString()
```
msasikanth, line 79: Same comment as above.

app/src/main/java/org/simple/clinic/drugs/selection/DrugListItem.kt
msasikanth: Here we are ending up creating a new list and then getting string from that. Instead, we can directly just create a string. wdyt?

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
```diff
@@ -76,10 +76,12 @@ class CustomDrugEntryUpdate : Update<CustomDrugEntryModel, CustomDrugEntryEvent,
       patientUuid: UUID,
       dosagePlaceHolder: String
   ): Next<CustomDrugEntryModel, CustomDrugEntryEffect> {
+    val dosage = if (model.dosage == dosagePlaceHolder) null else model.dosage
```
msasikanth, line 79: Instead of passing the placeholder, can't we just check if the dosage contains any numeric value or not? WDYT?

msasikanth (changes requested):

app/src/main/java/org/simple/clinic/drugs/selection/DrugListItem.kt
janhavisinghh: So the benefit of using `joinToString` with `listOfNull` is that it will not show nulls and remove the commas if there is a null string. If we hardcode it we had to handle all the null cases and add empty strings if there was null and also add/remove commas depending on if it had a value or not. So `joinToString` seemed like a better and a more cleaner approach to handle this.

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
janhavisinghh: uumm yeah 👍🏻

app/src/main/java/org/simple/clinic/drugs/selection/DrugListItem.kt
msasikanth: Hmm, that makes sense. But readability wise I don't think this is good, for example, we can create a function that creates the dosage and frequency string and do the string manipulations there.

I feel like this would be more understandable than the current implementation, WDYT? Also, this way we don't have to create a new list of objects and then covert them to a string.

```kt
private fun dosageAndFrequency(dosage: String?, frequency: MedicineFrequency): String {
  return if (dosage != null) {
    "$dosage, $frequency"
  } else {
    "$frequency"
  }
}
```

janhavisinghh: We will have to add a check for frequency as well `val medicineFrequency = if (prescribedDrug?.frequency == null) null else medicineFrequencyToLabelMap[prescribedDrug.frequency]!!.label` because we don't want to display `None` frequencies. 
It just seems like we are doing a lot of things here to do something that those methods handle already. 😅

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
janhavisinghh: @msasikanth 
Does using regex to handle this sound like a good approach?

  ```
  val atleastOneDigitRegex = "\\d".toRegex()
    val dosage = if (atleastOneDigitRegex.containsMatchIn(model.dosage!!)) model.dosage else null

```

File: app/src/main/java/org/simple/clinic/drugs/selection/DrugListItem.kt (hunk 1 above)
Comment (msasikanth) [reply app/src/main/java/org/simple/clinic/drugs/selection/DrugListItem.kt:49]: Cool, sounds good 👍🏾

File: app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt (hunk 3 above)
Comment (msasikanth) [reply app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt:79]: Yeah, that works. But we can probably go with this approach instead?

```kt
val isAValidDosage = dosage.filter { it.isDigit() }.isNotBlank()
```

janhavisinghh:
@msasikanth I removed the commits that were adding `dosagePlaceHolder` to `AddMedicineClicked` event and implemented the suggested changes as new commits.

msasikanth (approved):
LGTM

PR #3183, opened 2021-09-23
Fix `Country` v2 migration not running

msasikanth:
https://app.shortcut.com/simpledotorg/story/5158/app-not-displaying-national-id-and-proper-illustration-in-sri-lanka

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Provide drug frequencies to label as a map
 - Show ongoing diabetes treatment required dialog in diagnosis screen
 - Record call results instead of updating the same appointment record
+- Fix `Country` v2 migration not running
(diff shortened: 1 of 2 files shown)


msasikanth:
**Testing steps**

- Checkout `release/2021-09-06` branch
- Run the app and sign-in
- Using Flipper verify the saved country is in v1 format
- Check out this PR and run the app (don't clear existing app data, we are trying to update the existing app)
- Using Flipper verify that the preferences now contains v2 country and deployment. With the same values as in v1 country.

msasikanth:
Ideally, we would want to run this kind of shared preferences migrations during setup activity creation. Will look into this approach later.

vinaysshenoy (approved):

PR #3188, opened 2021-09-24
Set the cursor in the correct position whenever `CustomDrugEntrySheet` is opened 

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/5046/custom-drug-entry-sheet-ui-improvements

### DIFF
--- a/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
@@ -45,7 +45,7 @@ class CustomDrugEntryUpdate : Update<CustomDrugEntryModel, CustomDrugEntryEvent,
       model: CustomDrugEntryModel,
       drug: Drug
   ): Next<CustomDrugEntryModel, CustomDrugEntryEffect> {
-    val cursorPosition = if (drug.dosage != null) cursorPositionFromDosage(drug.dosage) else 0
+    val cursorPosition = cursorPositionFromDosage(drug.dosage)
 
     val updatedModel = model
         .drugNameLoaded(drug.name)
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
```diff
@@ -45,14 +45,21 @@ class CustomDrugEntryUpdate : Update<CustomDrugEntryModel, CustomDrugEntryEvent,
       model: CustomDrugEntryModel,
       drug: Drug
   ): Next<CustomDrugEntryModel, CustomDrugEntryEffect> {
+    val cursorPosition = if (drug.dosage != null) cursorPositionFromDosage(drug.dosage) else 0
```
Honey14, line 48: Do we need the to return 0 twice here? `cursorPositionFromDosage` is going to return us 0, can we instead check the nullability inside the function and have one condition (either x position or 0 position. WDYT?

app/src/test/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdateTest.kt
```diff
@@ -242,7 +244,32 @@ class CustomDrugEntryUpdateTest {
...
+        .given(drugFrequencyChoiceItemsLoaded)
+        .whenEvent(DrugFetched(drug))
+        .then(
+            assertThatNext(
+                hasModel(drugFrequencyChoiceItemsLoaded
+                    .drugNameLoaded(drug.name)
+                    .dosageEdited(drug.dosage)
+                    .frequencyEdited(drug.frequency)
+                    .rxNormCodeEdited(drug.rxNormCode)
+                    .drugInfoProgressStateLoaded()),
```
Honey14, line 272: Can we extract the model out for readability?

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntryUpdate.kt
```diff
@@ -67,6 +67,7 @@ class CustomDrugEntryUpdate : Update<CustomDrugEntryModel, CustomDrugEntryEvent,
       prescription: PrescribedDrug
   ): Next<CustomDrugEntryModel, CustomDrugEntryEffect> {
     val frequency = DrugFrequency.fromMedicineFrequency(prescription.frequency)
+    val cursorPosition = if (prescription.dosage != null) cursorPositionFromDosage(prescription.dosage) else 0
```
Honey14, line 70: I see why we have 2 conditions now. Apart from having `cursorPositionFromDosage` usage twice? Is there some other reason that I'm missing? 
If not, can we have one condition for these into their individual functions.

Honey14 (changes requested):
LGTM, left feedback, PTAL

janhavisinghh:
> Steps to reproduce: New medicine -> type custom med -> click on the dialog -> as soon as start typing we see the mg.

Yes this is the expected behaviour @Honey

Honey14 (approved):
LGTM

PR #3192, opened 2021-09-28
Fix state selection screen content overlapping next button

msasikanth:
https://app.shortcut.com/simpledotorg/story/5226/state-selection-screen-content-overlapping-next-button

There is already a changelog entry for fixing the content overlap on the next button. So I haven't added a new one. 

**Testing steps**
- Change the manifest endpoint to prod API
- Select India in the country selection screen
- State selection screen should be fully visible and scrollable
- Once the state is selected the content should not overlap the next button

**This should be hotfixed in the current demo**


msasikanth:
Closing, going with a different approach. Discussed this with Mahima and decided to remove the next button and use the state selection directly.

PR #3198, opened 2021-09-29
Navigate back to `SelectCountryScreen` from `RegistrationPhoneScreen` when there's only one state present

Honey14:
https://app.shortcut.com/simpledotorg/story/5236/navigate-back-to-selectcountryscreen-from-registrationphonescreen-when-there-s-only-one-state-present-in-the-country


msasikanth (approved):
LGTM

PR #3203, opened 2021-10-01
Use overdue list count to display the count in the tab bar

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/5339/use-overdue-list-count-to-display-the-count-in-the-tab-bar


msasikanth (approved):
LGTM

PR #3204, opened 2021-10-01
Replace the entire screen history in `TheActivityUpdate` instead of loading a single screen

vinaysshenoy:
https://app.shortcut.com/simpledotorg/story/5181/restore-correct-navigation-state-after-app-is-resumed-from-background


Honey14 (approved):
LGTM

PR #3207, opened 2021-10-04
Scroll to top when drugs list content is changed in edit medicines screen

msasikanth:
- Scroll to top when drugs list content is changed in edit medicines screen
- Update CHANGELOG


janhavisinghh (approved):
LGTM

PR #3209, opened 2021-10-05
Fix state selection list content hidden behind Android navigation bar

msasikanth:
https://app.shortcut.com/simpledotorg/story/5357/fix-state-selection-list-content-hidden-behind-android-navigation-bar

**Testing steps**
- Change the manifest endpoint to `https://api.simple.org/api/` in `gradle.properties`
- Run the app and verify entire state list is visible when scrolling


janhavisinghh (approved):
LGTM

PR #3226, opened 2021-10-06
Remove unused params from `OverdueAppointment`

msasikanth:


msasikanth:
Moving it to draft, I forgot to do the DB version bump and migration 😅

PR #3245, opened 2021-10-11
Scope shared preference instantiation to the application

vinaysshenoy:


msasikanth:
I have checked it as a sample app and you're right, it's a singleton. So even without this change, it should work as expected. But for now, we can make this change just to be safe.

PR #3267, opened 2021-10-25
Remove flaky test monitoring tool

vinaysshenoy:
https://app.shortcut.com/simpledotorg/story/5563/remove-flaky-test-tool


vinaysshenoy:
Adding @msasikanth as a reviewer since @Honey14 is on leave today.

app/build.gradle.kts
```diff
@@ -418,6 +418,8 @@ dependencies {
 
   testImplementation(libs.truth)
 
+  testImplementation(libs.kotlin.reflect)
```
msasikanth, line 421: Why do we need this dependency? 🤔

vinaysshenoy: The Quarantine library earlier was pulling this in transitively. Now that the library is removed, some tests started failing so I needed to add the dependency again.

msasikanth (approved):

PR #3285, opened 2021-11-05
Cancel in progress GH actions when new action is started with same group name

msasikanth:
**docs**: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency


Honey14 (approved):
LGTM

PR #3288, opened 2021-11-08
Migrate `PatientEntryScreen` to use view effects

msasikanth:
https://app.shortcut.com/simpledotorg/story/5691/migrate-patiententryscreen-to-use-view-effects


app/src/main/java/org/simple/clinic/newentry/PatientEntryViewEffectHandler.kt
```diff
@@ -53,4 +55,36 @@ class PatientEntryViewEffectHandler(
       showDOBExceedsMinLimitError(false)
     }
   }
+
+  private fun showValidationErrors(errors: List<PatientEntryValidationError>) {
+    errors
+        .onEach { Analytics.reportInputValidationError(it.analyticsName) }
+        .forEach {
+          when (it) {
+            PatientEntryValidationError.FullNameEmpty -> validationActions.showEmptyFullNameError(true)
```
Honey14, line 64: Why do we have a different interface for `PatientEntryValidationActions` (UiActions)? It should ideally be included in the `PatientEntryUiActions`. WDYT?

Honey14 (changes requested):
Looks good. left one comment

app/src/main/java/org/simple/clinic/newentry/PatientEntryViewEffectHandler.kt
msasikanth: Yep, good point. Will make that change.

Honey14 (approved):
LGTM

PR #3291, opened 2021-11-09
Fix diabetes treatment input required dialog showing when diabetes management is disabled

msasikanth:
https://app.shortcut.com/simpledotorg/story/5822/fix-diabetes-treatment-input-required-dialog-showing-when-diabetes-management-is-disabled

### DIFF
--- a/app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryModel.kt
+++ b/app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryModel.kt
@@ -59,7 +59,7 @@ data class NewMedicalHistoryModel(
     get() = diagnosedWithHypertension && country.isoCountryCode == Country.INDIA
 
   val showOngoingDiabetesTreatment: Boolean
-    get() = diagnosedWithDiabetes && country.isoCountryCode == Country.INDIA
+    get() = facilityDiabetesManagementEnabled && diagnosedWithDiabetes && country.isoCountryCode == Country.INDIA
 
   private val hasNoHypertension: Boolean
     get() = ongoingMedicalHistoryEntry.diagnosedWithHypertension == No
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@
 
 - [In Progress: 29 Oct 2021] Add option to download & share overdue list
 
+### Fixes
+
+- Fix diabetes treatment input required dialog showing when diabetes management is disabled
+
 ## 2021-11-02-8005
 
 ### Internal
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryModel.kt
```diff
@@ -59,7 +59,7 @@ data class NewMedicalHistoryModel(
     get() = diagnosedWithHypertension && country.isoCountryCode == Country.INDIA
 
   val showOngoingDiabetesTreatment: Boolean
-    get() = diagnosedWithDiabetes && country.isoCountryCode == Country.INDIA
+    get() = facilityDiabetesManagementEnabled && diagnosedWithDiabetes && country.isoCountryCode == Country.INDIA
```
vinaysshenoy, line 62: Hmm, does this not require a test? I feel adding an extra condition to an if-clause that did not break an existing test means there was some functionality we skipped adding a test for.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/medicalhistory/newentry/NewMedicalHistoryModel.kt
msasikanth: We already have a test for it (kind of), in update tests we are checking when save is clicked if we are showing registering patient or showing the errors.

msasikanth: ...and for the test that verifies we are showing the error dialog, I already added a diabetes management facility. I just forgot to have this condition here as well 😅

vinaysshenoy: No, I mean we are adding a new behaviour here:

- Clicking on "Next" should show a pop-up if diabetes management is enabled but a diagnosis has not been selected.
- Clicking on "Next" should not show a pop-up if diabetes management is disabled.

So adding this behaviour into the app but not adding a test is an issue, right?

msasikanth: We already have a test for the first case, for the second scenario we usually continue registering patients. We have a test for that as well. We can add one more test with a facility without a diabetes management if you want.

vinaysshenoy: We should do that, it's a behaviour we want to support.

msasikanth: Done

vinaysshenoy (approved):

PR #3299, opened 2021-11-17
Remove post delayed callbacks when view is detached from window

msasikanth:


msasikanth:
Showing keyboard in `doOnAttach` action is not working perfectly in pin entry card view. Switching to the new `WindowInsetsControllerCompat` to show/hide the keyboard fixes this, but breaks showing keyboard in the instant search screen. So, for now, I just stuck with an approach that works and just removed the callback when the view is detached.

app/src/main/java/org/simple/clinic/widgets/Views.kt
```diff
@@ -31,11 +31,13 @@ import timber.log.Timber
 import java.time.Duration
 
 fun EditText.showKeyboard() {
-  postDelayed({
-    this.requestFocus()
+  fun showKeyboard() {
```
vinaysshenoy, line 35: Can we rename this to something else? Seems weird to have a local function named the same as the enclosing function.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/widgets/Views.kt
msasikanth: Yep, renaming it to `openKeyboard`

vinaysshenoy (approved):

PR #3300, opened 2021-11-17
Track SQL query performance with begin and end hooks

vinaysshenoy:
https://app.shortcut.com/simpledotorg/story/5505/enable-end-to-end-application-monitoring

This change will be necessary in order for us to start tracking SQL performance in Datadog.


app/src/main/java/org/simple/clinic/storage/monitoring/SqlPerformanceReporter.kt
```diff
@@ -15,23 +15,32 @@ class SqlPerformanceReporter {
     }
 
     @JvmStatic
-    fun report(
+    fun begin(
```
msasikanth, line 18: So, is there a reason why we didn't just pass the start and end time in a single function and instead broke it down to two? 🤔

vinaysshenoy: This will be necessary for integrating Datadog. The way the telemetry API works is that we need to have explicit `start` and `stop` for their spans instead of just recording just how long an operation took.

msasikanth: Got it, I was just thinking instead of two methods if we can just have a single method `report` with `start. end, methodName`.

vinaysshenoy: That won't work, we need callbacks to start and stop the spans. The actual SQL query execution has to happen between the start and stop calls. You can take a look at https://github.com/simpledotorg/room-metadata-generator/blob/main/src/main/resources/MeasureMethod.java to see what the Room tracking method needs to look like.

vinaysshenoy: This other branch where we did the DD spike should also give you an idea why we need to make the change in this manner: https://github.com/simpledotorg/simple-android/commit/fc8c292c67d02a924286558e9cefe646d1309872

msasikanth (approved):

msasikanth (approved):

PR #3307, opened 2021-11-22
Send application performance metrics to Datadog

vinaysshenoy:
https://app.shortcut.com/simpledotorg/story/5505/enable-end-to-end-application-monitoring

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -15,6 +15,7 @@ plugins {
   id("io.sentry.android.gradle")
   id("com.google.firebase.firebase-perf")
   id("plugins.git.install-hooks")
+  id("dd-sdk-android-gradle-plugin")
 }
 
 sentry {
@@ -455,6 +456,8 @@ dependencies {
   lintChecks(projects.lint)
 
   runtimeOnly(libs.jackson.core)
+
+  implementation(libs.datadog.sdk)
 }
 
 // This must always be present at the bottom of this file, as per:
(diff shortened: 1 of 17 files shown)


app/build.gradle.kts
```diff
@@ -455,6 +456,8 @@ dependencies {
   lintChecks(projects.lint)
 
   runtimeOnly(libs.jackson.core)
+
+  implementation(libs.datadog.sdk)
```
msasikanth, line 464: Just a personal preference, we can probably add dependencies in alphabetical order. 😅

app/src/main/java/org/simple/clinic/ClinicApp.kt
```diff
@@ -86,6 +91,11 @@ abstract class ClinicApp : Application(), CameraXConfig.Provider {
   }
 
   private fun setupApplicationPerformanceMonitoring() {
+    val samplingRate = remoteConfig
+        .double("datadog_sample_rate", 0.0)
```
msasikanth, line 95: Currently we have this turned off?

app/build.gradle.kts
vinaysshenoy: Hmm, I prefer to add new dependencies at the bottom of the file, so it matches the version control history 😅

app/src/main/java/org/simple/clinic/ClinicApp.kt
vinaysshenoy: It's turned off by default, I have enabled it on Remote Config consoles.

msasikanth (approved):

app/build.gradle.kts
msasikanth: Yeah, I figured to keep this alphabetical to make it easier to find dependencies. But sure, this works. We can refactor it once in a while I guess.

msasikanth (approved):

PR #3308, opened 2021-11-22
Fix overdue list file name in `OverdueListDownloader`

msasikanth:
Forgot to update the file name extension when refactoring this class

### DIFF
--- a/app/src/main/java/org/simple/clinic/overdue/download/OverdueListDownloader.kt
+++ b/app/src/main/java/org/simple/clinic/overdue/download/OverdueListDownloader.kt
@@ -32,7 +32,11 @@ class OverdueListDownloader @Inject constructor(
         .download()
         .map { responseBody ->
           val localDateNow = LocalDate.now(userClock)
-          val fileName = "$DOWNLOAD_FILE_NAME_PREFIX$localDateNow.csv"
+          val fileExtension = when (downloadFormat) {
+            OverdueListDownloadFormat.CSV -> "csv"
+            OverdueListDownloadFormat.PDF -> "pdf"
+          }
+          val fileName = "$DOWNLOAD_FILE_NAME_PREFIX$localDateNow.$fileExtension"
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/overdue/download/OverdueListDownloader.kt
```diff
@@ -32,7 +32,11 @@ class OverdueListDownloader @Inject constructor(
         .download()
         .map { responseBody ->
           val localDateNow = LocalDate.now(userClock)
-          val fileName = "$DOWNLOAD_FILE_NAME_PREFIX$localDateNow.csv"
+          val fileExtension = when (downloadFormat) {
```
vinaysshenoy, line 35: Is there a test for this? Seems weird that we changed this without changing a test somewhere.

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/overdue/download/OverdueListDownloader.kt
msasikanth: We have an integration test that verifies if a file is saved to the disk or not, not particularly for the file name. Since Android provides us with a `Uri`, I think we can read the media query to get the file name to verify it, I guess.

msasikanth: I am not particularly sure we have to verify the downloaded file name, since it's a bit of effort to get this in Android. We can instead have a file extension provider function in the enum class to do this and have a test for a simple `when` statement. wdyt?

vinaysshenoy: Ah, I see. I don't think that test is necessary, just wanted to confirm what was happening here.

vinaysshenoy (approved):

PR #3325, opened 2021-11-29
Use hardcoded user in server integration tests

vinaysshenoy:
https://app.shortcut.com/simpledotorg/story/6149/hardcode-a-user-for-use-in-server-integration-tests

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/rules/ServerAuthenticationRule.kt
+++ b/app/src/androidTest/java/org/simple/clinic/rules/ServerAuthenticationRule.kt
@@ -107,13 +107,7 @@ class ServerAuthenticationRule : TestRule {
 
   private fun ensureLoggedInUser() {
     ensureFacilities()
-    val cachedUserInformation = readCachedUserInformation()
-    if (cachedUserInformation != null) {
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/org/simple/clinic/rules/ServerAuthenticationRule.kt
```diff
@@ -24,6 +24,8 @@ import org.simple.clinic.security.PasswordHasher
 import org.simple.clinic.user.User
 import org.simple.clinic.user.UserSession
 import org.simple.clinic.user.UserStatus
+import org.simple.clinic.user.finduser.FindUserResult.*
```
msasikanth, line 23: Not that important, can do this later. We are not using wildcard imports according to our code style, so can we break this down to individual imports, please?

msasikanth (approved):

app/src/androidTest/java/org/simple/clinic/rules/ServerAuthenticationRule.kt
vinaysshenoy: Oh yes, I usually set this to happen automatically when I format code, but it seems like it didn't run. Let me see what's going on.

PR #3331, opened 2021-12-01
Add support to set custom animations for `Router` transactions

msasikanth:


msasikanth:
Even though Android provides some of these animations resources, the timing of it is a little bit longer. So, created these resources to go with the shortest animation time.

app/src/main/java/org/simple/clinic/navigation/v2/Router.kt
```diff
@@ -453,6 +440,24 @@ class Router(
...
+        transactionOptions?.exitAnim ?: R.anim.router_slide_out_left
+    )
+    Direction.Backward -> Pair(
+        transactionOptions?.enterAnim ?: R.anim.router_slide_in_left,
+        transactionOptions?.exitAnim ?: R.anim.router_slide_out_right
+    )
+    Direction.Replace -> Pair(
+        transactionOptions?.enterAnim ?: R.anim.router_fade_in,
+        transactionOptions?.exitAnim ?: R.anim.router_fade_out
+    )
```
vinaysshenoy, line 459: Another option we have is to add the default animation IDs as properties of the `Direction` enum and then we can do something like:

```kotlin
val defaultEnterAnim = direction.defaultEnterAnim
val defaultExitAnim = direction.defaultExitAnim

return when(direction) {
}
```

What do you think?

vinaysshenoy (changes requested):

app/src/main/java/org/simple/clinic/navigation/v2/Router.kt
msasikanth: We can do that, but we would be exposing `Direction` enum to public APIs. That's why went with a separate class since this can be used for other use cases in future to customise the transaction/navigation behaviour.

msasikanth: We can probably move the `Direction` enum inside the `TransactionOptions` as well

vinaysshenoy: Ah cool. It was just a suggestion, whatever's here now works as well.

vinaysshenoy (approved):

app/src/main/java/org/simple/clinic/navigation/v2/Router.kt
msasikanth: Yeah 😅

vinaysshenoy:
@msasikanth It looks like CodeClimate is still throwing issues.

msasikanth:
> @msasikanth It looks like CodeClimate is still throwing issues.

Ah, we cannot simplify that function further. Gonna approve it for now.

PR #3333, opened 2021-12-02
Migrate `DeletePatientScreen` to a fragment

msasikanth:
https://app.shortcut.com/simpledotorg/story/6199/migrate-deletepatientscreen-to-a-fragment


app/src/main/java/org/simple/clinic/editpatient/deletepatient/DeletePatientScreen.kt
```diff
@@ -182,4 +182,9 @@ class DeletePatientScreen(
 
     override fun instantiateFragment() = DeletePatientScreen()
   }
+
+  interface Injector {
+
```
Honey14, line 160: Extra space here

app/src/main/java/org/simple/clinic/editpatient/deletepatient/DeletePatientScreen.kt
```diff
@@ -1,6 +1,7 @@
 package org.simple.clinic.editpatient.deletepatient
```
Honey14, line 1: Commit message does not make sense, can you reword it better?

Honey14 (changes requested):
LTGM, left 2 comments. PTAL

app/src/main/java/org/simple/clinic/editpatient/deletepatient/DeletePatientScreen.kt
msasikanth: That's intentional.

app/src/main/java/org/simple/clinic/editpatient/deletepatient/DeletePatientScreen.kt
msasikanth: I think the commit message specifies exactly what we are doing here. We are injecting into `DeletePatientScreen` in `onAttach`. 🤔

PR #3353, opened 2021-12-20
Update command for setting up server side Heroku review apps

vinaysshenoy:


.github/scripts/setup_heroku_instance.sh
```diff
@@ -41,7 +41,7 @@ resultOfServerPush=$?
 resultOfSeedDataSetup=0
 if [ $serverAppAlreadyExists = false ]; then
   echo "Setting up initial seed data"
-  (cd $serverAppDirectory && heroku run rails db:schema:load db:seed)
+  (cd $serverAppDirectory && heroku run rails db:structure:load:with_data db:seed)
```
Honey14, line 44: Why did the command change? Is it a path to setting up server side heroku instance?

Honey14 (changes requested):
LGTM

.github/scripts/setup_heroku_instance.sh
vinaysshenoy: Yes, this sets up the server side database schema with seed data. This was something that was changed on the server side setup recently, so we needed to update our local setup script to make things work again.

Honey14 (approved):

PR #3356, opened 2021-12-24
Add support for writing performance regression instrumented tests

vinaysshenoy:
https://app.shortcut.com/simpledotorg/story/6246/set-up-a-performance-regression-test-suite-that-alerts-performance-regressions

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/benchmark/BackupBenchmarkDatabase.kt
+++ b/app/src/androidTest/java/org/simple/clinic/benchmark/BackupBenchmarkDatabase.kt
@@ -1,0 +1,37 @@
+package org.simple.clinic.benchmark
+
+import android.app.Application
+import java.io.File
+import javax.inject.Inject
+
+class BackupBenchmarkDatabase @Inject constructor(
+    private val application: Application,
+    private val database: org.simple.clinic.AppDatabase
+) {
+
+  fun backup() {
+    val databaseDirectory = File(database.openHelper.readableDatabase.path).parentFile!!
+    val databaseBackupDirectory = resolveDatabaseBackupDirectory()
+    databaseDirectory.copyRecursively(databaseBackupDirectory, overwrite = true)
+  }
+
+  fun restore() {
+    val databaseDirectory = File(database.openHelper.readableDatabase.path).parentFile!!
+    val databaseBackupDirectory = resolveDatabaseBackupDirectory()
+
+    database.close()
+    databaseDirectory.deleteRecursively()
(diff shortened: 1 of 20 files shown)


app/src/androidTest/java/org/simple/clinic/benchmark/CanaryBenchmarkTest.kt
```diff
@@ -1,11 +1,33 @@
...
 import org.junit.Test
+import org.simple.clinic.TestClinicApp
+import org.simple.clinic.user.User
+import javax.inject.Inject
+import javax.inject.Provider
 
 class CanaryBenchmarkTest : BaseBenchmarkTest() {
 
+  @Inject
+  lateinit var userDaoProvider: Provider<User.RoomDao>
```
msasikanth, line 13: CMIIW, the reason why we are doing this, is to make sure the user data is inserted in the DB?

app/src/androidTest/java/org/simple/clinic/benchmark/BackupBenchmarkDatabase.kt
```diff
@@ -0,0 +1,37 @@
...
+import javax.inject.Inject
+
+class BackupBenchmarkDatabase @Inject constructor(
+    private val application: Application,
+    private val database: org.simple.clinic.AppDatabase
+) {
+
+  fun backup() {
+    val databaseDirectory = File(database.openHelper.readableDatabase.path).parentFile!!
+    val databaseBackupDirectory = resolveDatabaseBackupDirectory()
```
msasikanth, line 14: If these don't change, we can move them to class `init`?

msasikanth (approved):
LGTM. Left one comment that might require a change, I am approving it and we can make the change later if needed.

app/src/androidTest/java/org/simple/clinic/benchmark/CanaryBenchmarkTest.kt
vinaysshenoy: The reason we are doing this is to add a regression for verifying that the databases are saved and restored across multiple runs of a test.

app/src/androidTest/java/org/simple/clinic/benchmark/BackupBenchmarkDatabase.kt
vinaysshenoy: I vaguely remember trying to do that, but there was an issue because the open helper changes across test runs so the original file handle is not valid and throws an exception. Something that can be looked at later, I feel.

PR #3357, opened 2021-12-27
Return early if the required `Timeout` conditions are not satisfied in `TimeoutCallAdapterFactory`

msasikanth:
If the required conditions are not satisfied, we are returning early to avoid checking annotations or creating a new call adapter.


msasikanth:
@Honey14 Just a reminder, this hasn't been reviewed yet.

Honey14 (approved):
LGTM

PR #3373, opened 2022-01-04
Adding support for scanning Identifier for `EditPatientScreen`

Honey14:
https://app.shortcut.com/simpledotorg/story/6437/adding-support-for-scanning-identifier-for-editpatientscreen

### DIFF
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
@@ -112,6 +112,9 @@ class ScanSimpleIdScreen : BaseScreen<
   private val scanErrorTextView
     get() = binding.scanErrorTextView
 
+  private val enteredCodeContainer
+    get() = binding.enteredCodeContainer
+
   private val keyboardVisibilityDetector = KeyboardVisibilityDetector()
   private val cameraExecutor = Executors.newSingleThreadExecutor()
   private val cameraProviderFuture by unsafeLazy {
@@ -328,6 +331,10 @@ class ScanSimpleIdScreen : BaseScreen<
 
   override fun showScanError() {
     scanErrorTextView.visibility = View.VISIBLE
+  }
+
+  override fun hideEnteredCodeContainerView() {
+    enteredCodeContainer.visibility = View.GONE
   }
 
   interface Injector {
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
@@ -300,7 +300,10 @@ class ScanSimpleIdScreen : BaseScreen<
   }
 
   override fun showPatientWithIdentifierExistsError() {
-    // Will handle in a later commit
+    Snackbar
+        .make(binding.root, R.string.scansimpleid_identfier_already_exists, Snackbar.LENGTH_SHORT)
+        .setAction(R.string.scansimpleid_error_state_snackbar_ok) {}
+        .show()
   }
 
   override fun showInvalidQrCodeError() {
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
@@ -21,6 +21,7 @@ import androidx.camera.lifecycle.ProcessCameraProvider
 import androidx.core.content.ContextCompat
 import com.google.android.gms.common.ConnectionResult
 import com.google.android.gms.common.GoogleApiAvailability
+import com.google.android.material.snackbar.Snackbar
 import com.jakewharton.rxbinding3.widget.editorActionEvents
 import com.jakewharton.rxbinding3.widget.textChangeEvents
 import com.spotify.mobius.functions.Consumer
@@ -303,7 +304,10 @@ class ScanSimpleIdScreen : BaseScreen<
   }
 
   override fun showInvalidQrCodeError() {
-    // Will handle in a later commit
+    Snackbar
+        .make(binding.root, R.string.scansimpleid_invalid_qr_code, Snackbar.LENGTH_SHORT)
+        .setAction(R.string.scansimpleid_error_state_snackbar_ok) {}
+        .show()
   }
 
   override fun showEnteredCodeValidationError(failure: EnteredCodeValidationResult) {
--- a/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
+++ b/app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
@@ -36,6 +36,7 @@ import org.simple.clinic.feature.Feature.IndiaNationalHealthID
 import org.simple.clinic.feature.Features
 import org.simple.clinic.instantsearch.InstantSearchScreenKey
 import org.simple.clinic.navigation.v2.Router
+import org.simple.clinic.navigation.v2.Succeeded
 import org.simple.clinic.navigation.v2.fragments.BaseScreen
 import org.simple.clinic.patient.PatientPrefillInfo
 import org.simple.clinic.patient.businessid.Identifier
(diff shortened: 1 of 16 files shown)


app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -144,6 +144,11 @@ class ScanSimpleIdUpdate @Inject constructor(
     if (model.isSearching) return noChange()
 
     val clearInvalidQrCodeModel = model.clearInvalidQrCodeError()
+
+    if (model.openedFrom == OpenedFrom.EditPatientScreen.ToAddNHID) {
+      return searchPatientWhenNHIDEnabled(clearInvalidQrCodeModel, event)
+    }
```
msasikanth, line 188: This will be automatically done by the try/catch block. I don't think we don't need to handle it specifically. Is there a reason we are doing this? 🤔

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -155,16 +155,24 @@ class ScanSimpleIdUpdate @Inject constructor(
...
+      }
+      else -> {
+        try {
+          val bpPassportCode = UUID.fromString(event.text)
+          val identifier = Identifier(bpPassportCode.toString(), BpPassport)
+          next(model = clearInvalidQrCodeModel.searching(), SearchPatientByIdentifier(identifier))
+        } catch (e: Exception) {
+          searchPatientWhenNHIDEnabled(clearInvalidQrCodeModel, event)
+        }
+      }
```
msasikanth: We can just use the try/catch block we already have here.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -122,7 +122,7 @@ class ScanSimpleIdUpdate @Inject constructor(
       model: ScanSimpleIdModel,
       identifier: Identifier
   ): ScanSimpleIdViewEffect {
-    return if (model.openedFrom == OpenedFrom.EditPatientScreen.ToAddNHID)
+    return if (model.openedFrom == OpenedFrom.EditPatientScreen.ToAddNHID || model.openedFrom == OpenedFrom.EditPatientScreen.ToAddBpPassport)
```
msasikanth: We can go with this `model.openedFrom is OpenedFrom.EditPatientScreen` instead, WDYT?

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -102,7 +103,13 @@ class ScanSimpleIdUpdate @Inject constructor(
     return if (event.patients.isEmpty()) {
       searchPatientOnlineWhenOnlinePatientLookupEnabled(event, model)
     } else {
-      next(model = model.notSearching(), patientFoundByIdentifierSearch(patients = event.patients, identifier = event.identifier))
+      when (model.openedFrom) {
+        OpenedFrom.EditPatientScreen.ToAddBpPassport, OpenedFrom.EditPatientScreen.ToAddNHID -> dispatch(ShowScannedQrCodeError(IdentifierAlreadyExists))
```
msasikanth: Same as above comment, since we are not displaying specific messages based on if it's NHID or BP Passport, we can just go with checking if it's opened from `EditPatientScreen`

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -51,7 +51,14 @@ class ScanSimpleIdUpdate @Inject constructor(
   ): Next<ScanSimpleIdModel, ScanSimpleIdEffect> {
     return when (event.result) {
       is NotFound, is OtherError -> next(model.notSearching(), checkIfOpenedFromEditPatient(model, event.identifier))
-      is Found -> next(model.notSearching(), SaveCompleteMedicalRecords(event.result.medicalRecords))
+      is Found -> {
+        when (model.openedFrom) {
+          OpenedFrom.EditPatientScreen.ToAddBpPassport, OpenedFrom.EditPatientScreen.ToAddNHID -> next(model.notSearching(), ShowScannedQrCodeError(IdentifierAlreadyExists))
```
msasikanth: Same comment as above about using `OpenedFrom.EditPatientScreen`

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
```diff
@@ -39,12 +44,19 @@ class ScanSimpleIdUpdate @Inject constructor(
...
-      InvalidQrCode -> next(model.notSearching().invalidQrCode())
+      InvalidQrCode -> invalidQrCode(model)
       is OnlinePatientLookupWithIdentifierCompleted -> onlinePatientLookupWithIdentifierCompleted(model, event)
       is CompleteMedicalRecordsSaved -> patientsFoundByOnlineLookup(event.completeMedicalRecords)
     }
   }
 
+  private fun invalidQrCode(model: ScanSimpleIdModel): Next<ScanSimpleIdModel, ScanSimpleIdEffect> {
+    return when (model.openedFrom) {
+      is EditPatientScreen -> dispatch(ShowScannedQrCodeError(ScanErrorState.InvalidQrCode))
```
msasikanth, line 55: Yep, just like this. Let's do this for other places as well.

I think we don't specifically need to have subclasses of this to like `ToAddNHID` or `ToAddBpPassport`. Since we will be returning the identified directly, and we already have code to identifiy if the scanned QR is BP Passport or NHID (the try/catch block)

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
```diff
@@ -303,7 +304,9 @@ class ScanSimpleIdScreen : BaseScreen<
   }
 
   override fun showInvalidQrCodeError() {
-    // Will handle in a later commit
+    val invalidQrCodeSnackbar = Snackbar.make(binding.root, R.string.scansimpleid_invalid_qr_code, Snackbar.LENGTH_SHORT)
+
+    invalidQrCodeSnackbar.setAction(R.string.scansimpleid_error_state_snackbar_ok) { invalidQrCodeSnackbar.dismiss() }.show()
```
msasikanth, line 311: When the action button is clicked, it will automatically get dismissed, so we don't have to do that manually. You can just leave it as an empty lambda.

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
```diff
@@ -300,7 +300,9 @@ class ScanSimpleIdScreen : BaseScreen<
   }
 
   override fun showPatientWithIdentifierExistsError() {
-    // Will handle in a later commit
+    val identifierAlreadyExistsSnackbar = Snackbar.make(binding.root, R.string.scansimpleid_identfier_already_exists, Snackbar.LENGTH_SHORT)
+
+    identifierAlreadyExistsSnackbar.setAction(R.string.scansimpleid_error_state_snackbar_ok) { identifierAlreadyExistsSnackbar.dismiss() }.show()
```
msasikanth, line 305: Same comment as above for action button

msasikanth (changes requested):

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdUpdate.kt
janhavisinghh: So the reason why we are doing it like this is because when we are opening the scanner to add bp passport, but actually scanning an NHID it should give an error. In the previous behaviour it would correctly scan the NHID as it would go inside the catch block which is what we wanted to avoid. Which is why if you even look at the test case we have written, we are testing if we scan BPPassport when we are supposed to scan NHID then it should output `InvalidQrCode` and not scan bp passport.

msasikanth: Ah, got it. That makes sense, but we could've just handled the error states instead of breaking down the search functionality into different functions, right? For example, if the user scans an NHID when adding BP Passport, we can catch the error and check if it's opened to add the BP Passport and return early, no?
image

Same, vice versa
image

That way we just return early if the ID that is scanned is not what we want. Anyway, what we have works as well. Maybe we can make this change later on.

janhavisinghh (approved):

msasikanth (approved):

PR #3374, opened 2022-01-05
Refresh next appointment when appointment schedule or assigned facility is changed

msasikanth:
https://app.shortcut.com/simpledotorg/story/6476/refresh-appointment-in-nextappointmentcard

### DIFF
--- a/app/src/test/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentUpdateTest.kt
+++ b/app/src/test/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentUpdateTest.kt
@@ -3,12 +3,14 @@ package org.simple.clinic.scheduleappointment
 import com.spotify.mobius.test.NextMatchers.hasEffects
 import com.spotify.mobius.test.NextMatchers.hasModel
 import com.spotify.mobius.test.NextMatchers.hasNoEffects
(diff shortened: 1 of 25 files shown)


app/src/test/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentUpdateTest.kt
```diff
@@ -148,4 +150,28 @@ class ScheduleAppointmentUpdateTest {
             )
         )
   }
+
+  @Test
+  fun `when patient defaulter status is loaded and patient is defaulter, then schedule an automatic appointment`() {
```
janhavisinghh, line 156: This test is testing when the patient is a defaulter then do something but the commit message is for when not now is clicked. Can we change the commit message to represent what's happening here?

janhavisinghh (changes requested):

app/src/test/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentUpdateTest.kt
msasikanth: That's intentional. When we say defaulter in commit it doesn't describe the behaviour. But in tests, the naming indicates what's happening.

janhavisinghh (approved):

PR #3375, opened 2022-01-07
Change appointment bottom sheet display logic

msasikanth:
https://app.shortcut.com/simpledotorg/story/6431/change-appointment-bottom-sheet-display-logic

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
@@ -197,7 +197,7 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
             )
 
             DataForBackClickLoaded(
-                hasPatientDataChangedSinceScreenCreated = patientRepository.hasPatientDataChangedSince(patientUuid, timestamp),
+                hasPatientMeasurementDataChangedSinceScreenCreated = patientRepository.hasPatientDataChangedSince(patientUuid, timestamp),
                 countOfRecordedBloodPressures = countOfRecordedBloodPressures,
                 countOfRecordedBloodSugars = countOfRecordedBloodSugars,
                 medicalHistory = medicalHistory
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
@@ -212,16 +212,21 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
     return ObservableTransformer { effects ->
       effects
           .observeOn(scheduler)
-          .map { loadDataForBackClick ->
-            val patientUuid = loadDataForBackClick.patientUuid
+          .map { loadDataForDoneClick ->
+            val patientUuid = loadDataForDoneClick.patientUuid
             val countOfRecordedBloodPressures = bloodPressureRepository.bloodPressureCountImmediate(patientUuid)
             val countOfRecordedBloodSugars = bloodSugarRepository.bloodSugarCountImmediate(patientUuid)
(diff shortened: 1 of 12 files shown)


app/src/test/java/org/simple/clinic/summary/PatientSummaryUpdateTest.kt
```diff
@@ -377,7 +377,7 @@ class PatientSummaryUpdateTest {
     updateSpec
         .given(model)
         .whenEvent(DataForDoneClickLoaded(
-            hasPatientMeasurementDataChangedSinceScreenCreated = true,
+            hasPatientMeasurementDataChangedSinceScreenCreated = false,
```
janhavisinghh, line 380: Shouldn't this be `true`? according to the test name? `when patient measurement data is changed`

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
```diff
@@ -197,7 +197,7 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
             )
```
janhavisinghh, line 199: Can we correct the commit message, please?
We have mistakenly spelled `hasPatientMeasurementDataChangedSinceScreenCreated` to `sPatientMeasurementDataChangedSinceScreenCreated`

app/src/test/java/org/simple/clinic/summary/PatientSummaryUpdateTest.kt
```diff
@@ -230,14 +230,14 @@ class PatientSummaryUpdateTest {
   }
 
   @Test
-  fun `when there are patient summary changes and no measurements are recorded and warning is shown, clicking on back link id with patient screen must go back to home screen`() {
+  fun `when there are no patient measurement changes and warnings, clicking on back link id with patient screen must go back to home screen`() {
```
janhavisinghh, line 233: In all the test cases we are going back to the previous screen or the home screen. Don't you think we should also write a test case for when the appointment sheet is shown?

janhavisinghh (changes requested):

app/src/test/java/org/simple/clinic/summary/PatientSummaryUpdateTest.kt
msasikanth: If there are no changes we don't show the appointment sheet, but we show if there are changes. We have a test for that case though.

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
msasikanth: Ooh, good catch.

app/src/test/java/org/simple/clinic/summary/PatientSummaryUpdateTest.kt
msasikanth: Huh, it should be. I am not sure how this got reset. Good catch, taking a look at it.

msasikanth: OMG, this is from an entirely different test, because that test name was not changed. It seemed like it's from the same test.

msasikanth:
> android tests are failing can you ptal? @msasikanth

Yeah, it seems to be an issue with APIs. I have tagged server devs on Slack.

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
janhavisinghh: @msasikanth I can't see the change here. Can you ptal?

janhavisinghh (changes requested):

janhavisinghh (approved):
LGTM

PR #3377, opened 2022-01-10
Fix appointment not refreshing after assigned facility is changed

msasikanth:
It also has a small UI fix for `NextAppointmentCardView`


janhavisinghh (approved):
LGTM

PR #3397, opened 2022-01-13
Implement adding support for BpPassport in `EditPatientScreen`

Honey14:
https://app.shortcut.com/simpledotorg/story/6391/implement-adding-bppassport-in-editpatientscreen

### DIFF
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientEffectHandler.kt
@@ -8,6 +8,7 @@ import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
 import io.reactivex.Maybe
 import io.reactivex.Observable
+import io.reactivex.ObservableSource
 import io.reactivex.ObservableTransformer
 import io.reactivex.Scheduler
 import io.reactivex.Single
@@ -101,9 +102,31 @@ class EditPatientEffectHandler @AssistedInject constructor(
       Observable.merge(
           createOrUpdatePhoneNumber(sharedSavePatientEffects),
           savePatient(sharedSavePatientEffects),
-          handleAlternativeId(sharedSavePatientEffects)
+          handleAlternativeId(sharedSavePatientEffects),
+          linkBpPassportToThePatient(sharedSavePatientEffects)
       )
     }
+  }
+
+  private fun linkBpPassportToThePatient(
+      savePatientEffects: Observable<SavePatientEffect>
+  ): Observable<EditPatientEvent> {
+    return savePatientEffects
+        .filter(::isBpPassportAdded)
+        .flatMapCompletable { savePatientEffect ->
+          patientRepository
+              .addIdentifierToPatient(
+                  uuid = uuidGenerator.v4(),
+                  assigningUser = currentUser.get(),
+                  patientUuid = savePatientEffect.ongoingEntry.patientUuid,
+                  identifier = Identifier(
+                      value = savePatientEffect.ongoingEntry.bpPassports?.map { it.displayValue() }.toString(),
+                      type = BpPassport
+                  )
+              )
+              .ignoreElement()
+        }
+        .toObservable()
   }
 
   private fun savePatient(savePatientEffects: Observable<SavePatientEffect>): Observable<EditPatientEvent> {
@@ -305,6 +328,8 @@ class EditPatientEffectHandler @AssistedInject constructor(
   private fun isAlternativeIdAdded(it: SavePatientEffect) = it.saveAlternativeId == null && it.ongoingEntry.alternativeId.isNotBlank()
 
   private fun isAlternativeIdCleared(it: SavePatientEffect) = it.saveAlternativeId != null && it.ongoingEntry.alternativeId.isBlank()
+
+  private fun isBpPassportAdded(it: SavePatientEffect) = it.ongoingEntry.bpPassports!!.isNotEmpty()
 
   private fun loadInputFields(): ObservableTransformer<LoadInputFields, EditPatientEvent> {
     return ObservableTransformer { effects ->
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
@@ -2,6 +2,7 @@ package org.simple.clinic.editpatient
 
 import android.content.Context
 import android.os.Bundle
+import android.os.Parcelable
 import android.view.LayoutInflater
 import android.view.View
 import android.view.View.GONE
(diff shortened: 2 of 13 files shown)


janhavisinghh:
@Honey14 unit tests are failing in this branch. Can you ptal?

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
```diff
@@ -360,6 +362,9 @@ class ScanSimpleIdScreen : BaseScreen<
   interface Injector {
     fun inject(target: ScanSimpleIdScreen)
   }
+
+  @Parcelize
+  data class SendScannedIdentifier(val identifier: Identifier) : Parcelable
```
janhavisinghh, line 367: I think we can just rename this to `ScannedIdentifier`

app/src/main/java/org/simple/clinic/editpatient/EditPatientEffectHandler.kt
```diff
@@ -101,11 +102,33 @@ class EditPatientEffectHandler @AssistedInject constructor(
...
           createOrUpdatePhoneNumber(sharedSavePatientEffects),
           savePatient(sharedSavePatientEffects),
-          handleAlternativeId(sharedSavePatientEffects)
+          handleAlternativeId(sharedSavePatientEffects),
+          linkBpPassportToThePatient(sharedSavePatientEffects)
       )
     }
   }
 
+  private fun linkBpPassportToThePatient(
```
janhavisinghh, line 111: Hey, seems like we haven't added a test for this flow. Can we do that please?

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -351,6 +355,14 @@ class EditPatientScreen : BaseScreen<
     }
 
     deletePatient.setOnClickListener { router.push(DeletePatientScreen.Key(screenKey.patient.uuid)) }
+
+    setFragmentResultListener(RequestToScanIdentifier) { requestKey, result ->
+      if (result !is Succeeded) return@setFragmentResultListener
+
+      if (requestKey is RequestToScanIdentifier) {
+        val scannedIdentifier = (result.result as ScanSimpleIdScreen.SendScannedIdentifier).identifier
```
janhavisinghh, line 373: I think we should change the name of `RequestToScanIdentifier` to something more specific like `ScannedBPPassport` or `ScannedNHID`. As currently, we have created an implicit object which would then not be helpful in the case of NHID. So, let's rename it to `ScannedBPPassport` and while passing it in `openScanSimpleIdScreen` we should check if opened from is `ToAddBPPassport` and then set the request type as `ScannedBPPassport`. WDYT?

janhavisinghh (changes requested):
Left a few comments PTAL

janhavisinghh:
There is one code climate issue as well. Can you take a look at that as well? @Honey14

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
Honey14: ScannedIdentifier does not define the variable really well, hence the name

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
Honey14: Oh I'll rename it to `RequestToScanBpPassport`, makes more sense

janhavisinghh: We can maybe rename it to `ScannedBPPassportResult` or something like that as this key for the result and not request. You can also look at `ContactPatientSheet` for reference

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
janhavisinghh: No, I meant, we don't need to add the `Send` prefix. This type is only referred to as the `Succeeded` result. And it doesn't actually refer to us sending anything.

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
Honey14: It is requesting to scan bp passport, not the result

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
Honey14: It is sending result, maybe you can look at `SendBlankScannedQrCodeResult` to get the better idea of what's happening?

janhavisinghh: That's an effect name. The request type it's passing is also named as `BlankScannedQRCodeResult`

app/src/main/java/org/simple/clinic/editpatient/EditPatientUpdate.kt
```diff
@@ -2,6 +2,7 @@ package org.simple.clinic.editpatient
 
 import com.spotify.mobius.Next
 import com.spotify.mobius.Next.next
+import com.spotify.mobius.Next.noChange
```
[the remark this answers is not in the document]
Honey14: Removed

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
janhavisinghh: No.. So we are actually using `pushExpectingResult` in which the result is of type `RequestToScanIdentifier` 

Even in the example, you mentioned in another comment. They have named it `BlankScannedQrCode`

app/src/main/java/org/simple/clinic/scanid/ScanSimpleIdScreen.kt
janhavisinghh: > The idea is to make the code readable and to understand what's happening better

That is correct but in this case, this data class doesn't refer to an act of sending something but it just refers to the `ScannedQrCodeResult` or `ScannedQRCode` the action `router.popWithResult` is actually returning this back.

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
Honey14: So we're pushing expecting result where the key says what's going to happen. The key does not have to describe the result, the key is to basically to identify what's happening. So I'll remove the request and result both and rename it to something like `ScanBpPassport`

msasikanth: Just an FYI, we can also pass the identifier directly as a result. Since it's a parcelable it should be fine. We can then check the request type and then read the result directly as well.

PR #3422, opened 2022-01-19
Run renovate bot daily after 12 am

msasikanth:
With the previous configuration we would have received 5 PR's every 2 months and have to manually trigger the PR generation if we want from the dashboard. But we are gonna run this job daily to generate max of 5 concurrent PR's to avoid too much noise.


Honey14 (approved):
LGTM

PR #3424, opened 2022-01-19
Display National Health ID before BP Passports in `EditPatientScreen`

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/6781/display-national-health-id-before-bp-passports-in-editpatientscreen

### DIFF
--- a/app/src/main/res/layout/screen_edit_patient.xml
+++ b/app/src/main/res/layout/screen_edit_patient.xml
@@ -137,38 +137,6 @@
       </LinearLayout>
 
       <TextView
-        android:id="@+id/bpPassportsLabel"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:layout_marginTop="@dimen/spacing_24"
-        android:layout_marginEnd="@dimen/spacing_16"
-        android:text="@string/patientedit_bp_passports_label"
-        android:visibility="gone"
-        tools:visibility="visible" />
-
-      <LinearLayout
-        android:id="@+id/bpPassportsContainer"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
-        android:layout_marginTop="@dimen/spacing_8"
-        android:orientation="vertical" />
-
-      <com.google.android.material.button.MaterialButton
-        android:id="@+id/addBpPassportButton"
-        style="?attr/borderlessButtonDenseStyle"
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/screen_edit_patient.xml
```diff
@@ -196,7 +197,6 @@
         android:text="@string/patientedit_add_button"
         android:textAlignment="viewStart"
         android:textAppearance="?attr/textAppearanceButton"
-        android:visibility="gone"
```
Honey14, line 199: We need bp passport button to be gone by default, I have a story for this [here](https://app.shortcut.com/simpledotorg/story/6763/show-bp-passport-text-when-bp-passport-is-null-as-well) which will update this when feature flag is true for this feature. Can you undo this, please?

Honey14 (approved):
LGTM, left one comment, ptal

Honey14 (changes requested):
LGTM, Left one comment, PTAL

app/src/main/res/layout/screen_edit_patient.xml
janhavisinghh: oh good catch. I've already made this change tho in the feature flag PR. So we don't need to handle [this](https://app.shortcut.com/simpledotorg/story/6393/move-adding-health-ids-from-edit-patient-behind-a-feature-flag) anymore.

Honey14: We'll still need to remove this from this PR, if you've added in that and that PR is merged in, here you're removing it again

Honey14 (changes requested):

app/src/main/res/layout/screen_edit_patient.xml
janhavisinghh: yeah, I've changed that and pushed

app/src/main/res/layout/screen_edit_patient.xml
```diff
@@ -197,9 +197,9 @@
...
         app:icon="@drawable/ic_add_circle_outline_24px"
         app:iconGravity="start"
+        android:visibility="gone"
```
janhavisinghh, line 202: @Honey14 ^

Honey14 (approved):
LGTM

PR #3425, opened 2022-01-19
Set toolbar title in `ScanSimpleIdScreen` based on where it's opened from

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/6780/change-the-title-of-scansimpleidscreen-to-either-bp-passport-or-national-health-id-depending-on-where-it-s-opened-from

### DIFF
--- a/app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUiRendererTest.kt
+++ b/app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUiRendererTest.kt
@@ -22,6 +22,7 @@ class ScanSimpleIdUiRendererTest {
     // then
     verify(ui).showSearchingForPatient()
     verify(ui).hideScanError()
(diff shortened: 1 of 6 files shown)


app/src/test/java/org/simple/clinic/scanid/ScanSimpleIdUiRendererTest.kt
```diff
@@ -22,6 +22,7 @@ class ScanSimpleIdUiRendererTest {
     // then
     verify(ui).showSearchingForPatient()
     verify(ui).hideScanError()
+    verify(ui).setToolBarTitle(searchingModel.openedFrom)
```
msasikanth, line 25: Can we pass the actual `OpenedFrom` value that we expect to have when asserting, please?

msasikanth (changes requested):

msasikanth (approved):

PR #3432, opened 2022-01-20
Move adding health ids from edit patient behind a feature flag

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/6393/move-adding-health-ids-from-edit-patient-behind-a-feature-flag

### DIFF
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientModel.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientModel.kt
@@ -71,8 +71,8 @@ data class EditPatientModel(
   val hasInputFields
     get() = inputFields != null
 
-  val isUserCountryIndiaAndAddingNHIDsFeatureEnabled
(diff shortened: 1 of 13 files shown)


app/src/main/java/org/simple/clinic/editpatient/EditPatientModel.kt
```diff
@@ -71,6 +71,9 @@ data class EditPatientModel(
   val hasInputFields
     get() = inputFields != null
 
+  val isUserCountryIndiaAndAddingNHIDsFeatureEnabled
```
msasikanth, line 74: We can simplify this variable name to say `canAddNHID`, WDYT?

msasikanth: Also, do you think we can just move the `alternateId == null` check to this variable as well? 🤔

msasikanth (commented):
LGTM. Left a comment, PTAL

msasikanth (approved):

PR #3433, opened 2022-01-20
Support scanning list of unique Bp passports in `EditPatientScreen`

Honey14:
https://app.shortcut.com/simpledotorg/story/6760/support-scanning-list-of-bp-passports-in-editpatientscreen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,11 +42,12 @@
 - Bump Mixpanel to v5.9.6
 - Bump androidx-annotation
 - Add benchmark test for loading overdue appointments in a facility
-- [In Progress: 17 Jan 2022] Implement adding scan support for BpPassport in `EditPatientScreen`
+- Support scanning list of unique Bp passports in `EditPatientScreen`
 
 ### Changes
 
 - Implement adding scan support for India NHID in `EditPatientScreen`
+- Implement adding scan support for BpPassport in `EditPatientScreen`
 
 ## On Demo
 
--- a/app/src/main/java/org/simple/clinic/editpatient/EditablePatientEntry.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditablePatientEntry.kt
@@ -141,8 +141,9 @@ data class EditablePatientEntry @Deprecated("Use the `from` factory function ins
   fun updateAlternativeId(alternativeId: String): EditablePatientEntry =
       copy(alternativeId = alternativeId)
 
-  fun addBpPassports(bpPassports: List<Identifier>): EditablePatientEntry =
-      copy(bpPassports = bpPassports)
+  fun addBpPassports(bpPassports: List<Identifier>): EditablePatientEntry {
+    return copy(bpPassports = this.bpPassports.orEmpty() + bpPassports)
+  }
 
   fun validate(
       alreadySavedNumber: PatientPhoneNumber?,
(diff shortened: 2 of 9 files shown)


janhavisinghh:
@Honey14 

Currently when we scan multiple bp passports; in that, 
- Only allowing unique bp passports to be added is working great. 
- When we click on save button after adding multiple bp passports and go back to the edit screen again - Only the last bp passport is saved. Is this the expected behaviour or are we planning to handle this in a different pr?

CHANGELOG.md
```diff
@@ -33,8 +33,11 @@
 - Bump firebase config to v21.0.1
 - Add facility protocol and group UUID in demo facility
 - [In Progress: 11 Jan 2022] Implement adding NHID in `PatientEditScreen`
-- [In Progress: 17 Jan 2022] Implement adding scan support for BpPassport in `EditPatientScreen`
 
+### Changes
+- Implement adding scan support for BpPassport in `EditPatientScreen`
+  - Support scanning list of unique Bp passports in `EditPatientScreen`
```
janhavisinghh, line 39: There is an extra space here, can we refactor this please?

janhavisinghh (changes requested):
Looks good to me overall. Have added one question in comments, ptal. :)

Honey14:
> @Honey14
> 
> Currently when we scan multiple bp passports; in that,
> 
> * Only allowing unique bp passports to be added is working great.
> * When we click on save button after adding multiple bp passports and go back to the edit screen again - Only the last bp passport is shown. Is this the expected behaviour or are we planning to handle this in a different pr?

Can you rebase and test again because the second scenario is also working great for me.

CHANGELOG.md
Honey14: The space is intended because it comes under the change

app/src/main/java/org/simple/clinic/editpatient/EditablePatientEntry.kt
```diff
@@ -142,7 +142,10 @@ data class EditablePatientEntry @Deprecated("Use the `from` factory function ins
       copy(alternativeId = alternativeId)
 
   fun addBpPassports(bpPassports: List<Identifier>): EditablePatientEntry {
-    return copy(bpPassports = this.bpPassports.orEmpty() + bpPassports)
+    return if (this.bpPassports.orEmpty().contains(bpPassports.last()))
+      copy(bpPassports = this.bpPassports.orEmpty())
+    else
+      copy(bpPassports = this.bpPassports.orEmpty() + bpPassports)
```
msasikanth, line 148: @Honey14 Since this is business logic, can we move this to the `Update` function, please?

CHANGELOG.md
msasikanth: We don't need to list it here, since this feature is not out yet for users yet. We can move the second line to internal change as it's a bug fix.

janhavisinghh:
> > @Honey14
> > Currently when we scan multiple bp passports; in that,
> > 
> > * Only allowing unique bp passports to be added is working great.
> > * When we click on save button after adding multiple bp passports and go back to the edit screen again - Only the last bp passport is shown. Is this the expected behaviour or are we planning to handle this in a different pr?
> 
> Can you rebase and test again because the second scenario is also working great for me.

@Honey14 can you please recheck if you have pushed all the changes? (correct me if I'm wrong here) As I don't see any commit which would solve the second scenario. We have not changed anything in the `EditPatientEffectHandler#linkBpPassportToThePatient` method so it is saving the last bp passport only.

janhavisinghh (approved):
LGTM

janhavisinghh (approved):

PR #3440, opened 2022-01-24
Handle visibility of Bp Passport label and button when scan health id from edit patient is enabled

Honey14:
https://app.shortcut.com/simpledotorg/story/6763/show-bp-passport-text-when-bp-passport-is-null-as-well

### DIFF
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
@@ -455,6 +455,10 @@ class EditPatientScreen : BaseScreen<
     bpPassportsLabel.visibility = GONE
   }
 
+  override fun hideBpPassportButton() {
+    addBpPassportButton.visibility = GONE
+  }
+
   private fun showOrHideInputFields(inputFields: InputFields) {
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -457,6 +457,10 @@ class EditPatientScreen : BaseScreen<
     bpPassportsLabel.visibility = VISIBLE
   }
 
+  override fun showBpPassportLabelIfIdentifierNotEmpty(identifiers: List<String>) {
```
msasikanth, line 460: Instead of passing all the identifies list to toggle the visibility. Can we just have functions to hide and show the label, please? 

We already have the show label function from the last commit, we can have the hide label function as well and then have the logic to toggle between these two in the UI renderer.

msasikanth (changes requested):
Left one comment, PTAL

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
Honey14: Ah this makes more sense, updating this

msasikanth (approved):

PR #3448, opened 2022-01-27
Migrate `SettingsScreen` to use view effects

msasikanth:
https://app.shortcut.com/simpledotorg/story/6893/migrate-settingsscreen-to-use-view-effects

### DIFF
--- a/app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
@@ -1,5 +1,6 @@
 package org.simple.clinic.settings
 
+import com.spotify.mobius.functions.Consumer
 import com.spotify.mobius.rx2.RxMobius
 import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/simple/clinic/settings/SettingsEffectHandler.kt
```diff
@@ -17,12 +18,16 @@ class SettingsEffectHandler @AssistedInject constructor(
     private val schedulersProvider: SchedulersProvider,
     private val appVersionFetcher: AppVersionFetcher,
     private val appUpdateAvailability: CheckAppUpdateAvailability,
-    @Assisted private val uiActions: UiActions
+    @Assisted private val uiActions: UiActions,
+    @Assisted val viewEffectsConsumer: Consumer<SettingsViewEffect>
```
janhavisinghh, line 21: Is there a reason why `viewEffectsConsumer` is not private?

msasikanth: Because I missed it 😂. Good catch.

janhavisinghh (approved):

PR #3451, opened 2022-01-27
Add benchmark test for creating a new bp entry

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/6852/add-benchmark-for-bp-entry-queries

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/benchmark/bp/BloodPressureEntryBenchmark.kt
+++ b/app/src/androidTest/java/org/simple/clinic/benchmark/bp/BloodPressureEntryBenchmark.kt
@@ -39,7 +39,7 @@ class BloodPressureEntryBenchmark : BaseBenchmarkTest() {
   }
 
   @Test
-  fun `creating_new_blood_pressure_entry_for_a_patient`() {
+  fun creating_new_blood_pressure_entry_for_a_patient() {
     val patientUuid = UUID.fromString("9a59d117-34ad-44e5-b0e2-93213ebfea01")
     val recordedAt = Instant.now()
 
--- a/app/src/androidTest/java/org/simple/clinic/benchmark/bp/BloodPressureEntryBenchmark.kt
+++ b/app/src/androidTest/java/org/simple/clinic/benchmark/bp/BloodPressureEntryBenchmark.kt
@@ -1,0 +1,51 @@
+package org.simple.clinic.benchmark.bp
+
+import org.junit.Test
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/org/simple/clinic/benchmark/bp/BloodPressureEntryBenchmark.kt
```diff
@@ -0,0 +1,51 @@
...
+  @Inject
+  lateinit var uuidGenerator: UuidGenerator
+
+  init {
+    TestClinicApp.appComponent().inject(this)
+  }
+
+  @Test
+  fun `creating_new_blood_pressure_entry_for_a_patient`() {
+    val patientUuid = uuidGenerator.v4()
```
msasikanth, line 35: Instead of creating a new patient id, can we use an existing patient ID information?

app/src/androidTest/java/org/simple/clinic/benchmark/bp/BloodPressureEntryBenchmark.kt
```diff
@@ -0,0 +1,51 @@
...
+    val recordedAt = Instant.now()
+
+    bloodPressureRepository.saveMeasurement(
+        patientUuid = patientUuid,
+        reading = BloodPressureReading(systolic = 120, diastolic = 80),
+        loggedInUser = TestData.loggedInUser(uuidGenerator.v4(), name = "Anand Rai"),
+        currentFacility = TestData.facility(uuidGenerator.v4()),
```
msasikanth, line 42: Same for user and current facility, we should be able to inject these values, since we are logging in with a test account.

msasikanth (changes requested):
Overall looks good, left couple of comments PTAL

app/src/androidTest/java/org/simple/clinic/benchmark/bp/BloodPressureEntryBenchmark.kt
```diff
@@ -26,20 +28,29 @@ class BloodPressureEntryBenchmark : BaseBenchmarkTest() {
   @Inject
   lateinit var uuidGenerator: UuidGenerator
 
+  @Inject
+  lateinit var patient: Patient
```
msasikanth, line 32: We cannot inject a patient here 😅. We can find a patient ID from the facility, either from the dashboard or signing in into our test account and getting it manually.

msasikanth (approved):

PR #3455, opened 2022-01-28
Migrate `CustomDrugEntrySheet` to user view effects

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/6924/migrate-customdrugentrysheet-to-use-view-effects

### DIFF
--- a/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntrySheet.kt
+++ b/app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntrySheet.kt
@@ -102,7 +102,13 @@ class CustomDrugEntrySheet : BaseBottomSheet<
 
   override fun createInit() = CustomDrugEntryInit()
 
-  override fun createEffectHandler(viewEffectsConsumer: Consumer<CustomDrugEntryViewEffect>) = effectHandlerFactory.create(this).build()
+  override fun createEffectHandler(viewEffectsConsumer: Consumer<CustomDrugEntryViewEffect>) =
+      effectHandlerFactory
+          .create(
+              uiActions = this,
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntrySheet.kt
```diff
@@ -104,6 +104,8 @@ class CustomDrugEntrySheet : BaseBottomSheet<
 
   override fun createEffectHandler(viewEffectsConsumer: Consumer<CustomDrugEntryViewEffect>) = effectHandlerFactory.create(this).build()
 
+  override fun viewEffectsHandler() = CustomDrugEntryViewEffectHandler(this)
```
msasikanth, line 110: I think there is a typo here, the method name is just `viewEffectHandler`

msasikanth (changes requested):
Looks good overall, left one comment PTAL

app/src/main/java/org/simple/clinic/drugs/selection/custom/CustomDrugEntrySheet.kt
janhavisinghh: 🤔 Umm seems right to me. @msasikanth It's an overridden method from `BaseBottomSheet`. Or, do you mean we want to change the name in `BaseBottomSheet` to `viewEffectHandler`?

msasikanth: Ah, interesting. The `BaseScreen` and `BaseBottomSheet` has different function name. Yep, this looks fine.

msasikanth (approved):

PR #3458, opened 2022-01-28
Highlight newly scanned bp passports in `EditPatientScreen`

Honey14:
https://app.shortcut.com/simpledotorg/story/6834/add-highlight-when-bp-passports-are-added


app/src/main/java/org/simple/clinic/editpatient/EditPatientViewRenderer.kt
```diff
@@ -82,7 +82,7 @@ class EditPatientViewRenderer(private val ui: EditPatientUi) : ViewRenderer<Edit
     val newlyAddedBpPassports = model.ongoingEntry.bpPassports?.map { it.displayValue() }.orEmpty()
     val identifiers = model.bpPassports?.map { it.identifier.displayValue() }.orEmpty()
 
-    ui.displayBpPassports(identifiers + newlyAddedBpPassports)
+    ui.displayBpPassports(identifiers, newlyAddedBpPassports)
```
msasikanth, line 85: This approach works fine, but we end up passing two lists and then checking the highlight status at the view level. 

Instead, maybe we can map the identifiers to a specific type and then render those? Something like this.

```kt
val newlyAddedBpPassports = model.ongoingEntry.bpPassports
    ?.map { BpPassportListItem(identifier = it.displayValue(), isHighlighted = true) }
    .orEmpty()

val identifiers = model.bpPassports
    ?.map {
      BpPassportListItem(identifier = it.identifier.displayValue(), isHighlighted = false)
    }
    .orEmpty()
```

So, now we don't have to use nested loops to identify the highlight status at view level and can simply use the boolean value. This way we can actually have the tests to verify the highlight status as well. WDYT?

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -563,6 +563,15 @@ class EditPatientScreen : BaseScreen<
     }
   }
 
+  private fun highlightNewlyScannedBpPassports(bpPassportView: PatientEditBpPassportViewBinding) {
+    bpPassportView.bpPassportIdentifier.setBackgroundColor(resources.getColor(R.color.simple_yellow_100))
```
msasikanth, line 567: Can we use `context.resolveColor` instead, please? `resources.getColor` is deprecated.

msasikanth (changes requested):
Left couple of comments, PTAL

janhavisinghh:
@msasikanth added the requested changes as new commits

app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -557,9 +557,8 @@ class EditPatientScreen : BaseScreen<
 
   private fun dateOfBirthFocusChanges(): Observable<EditPatientEvent> = dateOfBirthEditText.focusChanges.map(::DateOfBirthFocusChanged)
 
-  override fun displayBpPassports(identifiers: List<BPPassportListItem>, newlyScannedIdentifiers: List<BPPassportListItem>) {
+  override fun displayBpPassports(listOfBpPassports: List<BPPassportListItem>) {
```
msasikanth, line 560: Personally I would have preferred to just name it `bpPassports` 😅

janhavisinghh: haha. oki. Changing it.

msasikanth (approved):
LGTM

PR #3488, opened 2022-02-03
Migrate `OnboardingScreen` to use view effects

msasikanth:
https://app.shortcut.com/simpledotorg/story/7081/migrate-onboardingscreen-to-use-view-effects


app/src/test/java/org/simple/clinic/onboarding/OnboardingEffectHandlerTest.kt
```diff
@@ -0,0 +1,49 @@
...
+  ).build()
+  private val testCase = EffectHandlerTestCase(effectHandler)
+
+  @After
+  fun tearDown() {
+    testCase.dispose()
+  }
+
+  @Test
+  fun `when complete onboarding effect is received, then mark user has completed onboarding`() {
```
janhavisinghh, line 29: 👍🏻  This is great. Thought about this: we should follow TDD while adding these tests. Like, write the test then see if it's working with the effect handler previously and then check if it's working with the view effect handler. Saying this because once I had mistakenly added the wrong UI action both in the test and in the view effect. 😅

msasikanth: Yeah, it's a good idea. Currently this is moving existing controller/logic tests into the new tests, so all of the effect handler and update functions code is already present. So, we can move these tests directly.

PR #3493, opened 2022-02-04
Fix the broken `backButton` in the “Edit” screen

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/7103/bug-the-x-button-in-the-edit-screen-is-broken

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@
 - Add highlight when newly scanned bp passports are added
 - Index `patientUuid` in tables that are joined in `OverdueAppointment` database view
 - Migrate `RegistrationPinScreen` to use view effects
+- Fix broken back button in `EditPatientScreen`
 
 ### Changes
 
--- a/app/src/main/res/layout/screen_edit_patient.xml
+++ b/app/src/main/res/layout/screen_edit_patient.xml
@@ -7,17 +7,6 @@
   android:layout_height="match_parent"
   android:orientation="horizontal">
 
-  <ImageButton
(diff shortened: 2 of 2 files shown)


janhavisinghh:
@msasikanth this is a fix we need to hotfix in `DEMO` so can we please review this on priority?

CHANGELOG.md
```diff
@@ -78,6 +78,7 @@
 - Add highlight when newly scanned bp passports are added
 - Index `patientUuid` in tables that are joined in `OverdueAppointment` database view
 - Migrate `RegistrationPinScreen` to use view effects
+- Fix broken back button in `EditPatientScreen`
```
msasikanth, line 81: Ideally this should be in the fixes section, since it's in demo and we are hotfixing it. But I am merging this in for now.

msasikanth (approved):

CHANGELOG.md
janhavisinghh: umm but this is not out there to the users yet so them seeing this as a fix would not make sense. @msasikanth

janhavisinghh: We caught this bug internally in a pre-production environment which is why I thought we can have this in the internal section

msasikanth (approved):

CHANGELOG.md
msasikanth: Yeah, that's fair. So far, if it goes into demo and we have to fix an bug fix, we place it in `Fixes` section of the changelog.

PR #3503, opened 2022-02-07
Fix uneven spacing between health ID labels and `ADD` button

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/7110/ui-improvement-the-spacing-between-health-id-labels-and-add-button-is-uneven

### DIFF
--- a/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
+++ b/app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
@@ -560,6 +560,9 @@ class EditPatientScreen : BaseScreen<
 
   override fun displayBpPassports(bpPassports: List<BPPassportListItem>) {
     bpPassportsContainer.removeAllViews()
+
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/simple/clinic/editpatient/EditPatientScreen.kt
```diff
@@ -560,6 +560,8 @@ class EditPatientScreen : BaseScreen<
 
   override fun displayBpPassports(bpPassports: List<BPPassportListItem>) {
     bpPassportsContainer.removeAllViews()
+    bpPassportsContainer.visibility = VISIBLE
```
msasikanth, line 563: I think just this might not be enough to fix the spacing issue. For example, we call `displayBpPassports` even if there are no BP passports. So, that would end up just toggling the visibility. 

So, we need to toggle this view based on whether we have any BP passports are not.

msasikanth (changes requested):

msasikanth (approved):

PR #3509, opened 2022-02-08
Ignore material update dependency

Honey14:


renovate.json5
```diff
@@ -8,7 +8,11 @@
   "timezone": "Asia/Kolkata",
   "ignoreDeps": [
     // Ignoring zxing update. Since update requires Java 8+ runtime.
-    "com.google.zxing:core"
+    "com.google.zxing:core",
+
+    /* Ignoring this version because it introduces Material 3 specs and it requires changes on theming structure and components.
+    Depending on how much we're adapting we will re-evaluate this later */
```
msasikanth, line 14: It doesn't look like we will pick Material 3 spec for Simple. So, let's just have a line saying "Ignoring Material update. Since it requires us to use Material 3 spec"

msasikanth (changes requested):

renovate.json5
```diff
@@ -8,7 +8,10 @@
   "timezone": "Asia/Kolkata",
   "ignoreDeps": [
     // Ignoring zxing update. Since update requires Java 8+ runtime.
-    "com.google.zxing:core"
+    "com.google.zxing:core",
+
+    // Ignoring this version because it introduces Material 3 specs
```
msasikanth, line 13: Just a nitpick. Here we are saying "Ignoring this version" but we are not specifying any version. We should probably start considering having the version in the comment or see if there is an equivalent renovate config for version locking.

msasikanth (dismissed):

msasikanth (approved):

PR #3510, opened 2022-02-08
Migrate `IntroVideoScreen` to use view effects

janhavisinghh:
[Migrate `IntroVideoScreen` to use view effects](https://app.shortcut.com/simpledotorg/story/7179/migrate-introvideoscreen-to-use-view-effects)

### DIFF
--- a/app/src/main/java/org/simple/clinic/introvideoscreen/IntroVideoViewEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/introvideoscreen/IntroVideoViewEffectHandler.kt
@@ -1,12 +1,15 @@
 package org.simple.clinic.introvideoscreen
 
 import org.simple.clinic.mobius.ViewEffectsHandler
+import org.simple.clinic.util.exhaustive
 
 class IntroVideoViewEffectHandler(
     private val uiActions: UiActions
 ) : ViewEffectsHandler<IntroVideoViewEffect> {
   
   override fun handle(viewEffect: IntroVideoViewEffect) {
-    // nothing to see here, yet. -\(".")/-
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/simple/clinic/introvideoscreen/IntroVideoViewEffectHandler.kt
```diff
@@ -0,0 +1,12 @@
+package org.simple.clinic.introvideoscreen
+
+import org.simple.clinic.mobius.ViewEffectsHandler
+
+class IntroVideoViewEffectHandler(
+    private val uiActions: UiActions
```
Honey14, line 7: We don't need the uiactions yet. So ideally let's not add it in this commit. We should only add it as a parameter when we need it.

Honey14 (changes requested):
LGTM, left one comment PTAL

app/src/main/java/org/simple/clinic/introvideoscreen/IntroVideoViewEffectHandler.kt
janhavisinghh: We are migrating to use view effects so I think it's okay to have a few commits with implementations that are not being used yet. For example, here in this commit, we are not using the effect handler as well. This would also endure that in the next few commits while overriding effect handler we are passing the ui action.

Honey14: Should be fine

Honey14 (approved):
LGTM

PR #3515, opened 2022-02-09
Remove Google Guava dependency

msasikanth:
Currently we are using Google Guava dependency for getting a [executor service](https://guava.dev/releases/19.0/api/docs/com/google/common/util/concurrent/MoreExecutors.html#newDirectExecutorService()) in `MobiusTestFixture`, for running tasks in the same thread they are invoked from. The usage can be replaced by [`ImmediateWorker`](https://github.com/spotify/mobius/blob/master/mobius-core/src/main/java/com/spotify/mobius/runners/ImmediateWorkRunner.java) from `mobius` package, which does the same.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - Migrate `AppLockScreen` to use view effects
 - Bump Gradle to v7.4
 - Bump RootBeer to 0.1.0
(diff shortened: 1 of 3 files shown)


Honey14 (approved):
LGTM

PR #3550, opened 2022-02-14
Migrate `HelpScreen` to use view effects

Honey14:
https://app.shortcut.com/simpledotorg/story/7199/migrate-helpscreen-to-view-effects


app/src/main/java/org/simple/clinic/home/help/HelpScreenEffectHandler.kt
```diff
@@ -13,12 +13,16 @@ class HelpScreenEffectHandler @AssistedInject constructor(
     private val helpRepository: HelpRepository,
     private val helpSync: HelpSync,
     private val schedulersProvider: SchedulersProvider,
-    @Assisted private val uiActions: HelpScreenUiActions
+    @Assisted private val uiActions: HelpScreenUiActions,
+    @Assisted private val viewEffectHandler: HelpScreenViewEffectHandler
```
msasikanth: Just an FYI, we don't have to add the view effect handler here as a dependency. We can directly add the consumer.

msasikanth (approved):

PR #3603, opened 2022-02-21
Fix R8 build error when generating release artifacts

Honey14:
https://app.shortcut.com/simpledotorg/story/7391/fix-sandbox-build-failing-in-bitrise

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Remove `isFlexibleUpdateType` as parameter from `UpdateInfo`
 - Provide `PackageInfo` as dependency for `AppVersionFetcher`
 - Bump Lottie to v5.0.1
+- Update proguard rule to keep enum classes only for iText library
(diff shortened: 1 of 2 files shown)


msasikanth:
Hmm, I agree with Sentry PR comment about this rule being too broad. We should probably look into only keep specific failing classes. WDYT?

Honey14:
> Hmm, I agree with Sentry PR comment about this rule being too broad. Essentially that rule keeps all the enum classes in the project (our code and libraries) away from obfuscation.
> 
> We should probably look into only keep specific failing classes. WDYT?

That's probably a better idea. I'm not sure if this is the need of the moment though, WDYT? @msasikanth

msasikanth:
You can use this Gradle command to test it out @janhavisinghh 
`./gradlew -PrunProguard=true assembleSandboxRelease bundleSandboxRelease`

janhavisinghh (approved):
LGTM

PR #3613, opened 2022-03-01
Use full text search when searching for patients

msasikanth:
https://app.shortcut.com/simpledotorg/story/7452/use-full-text-search-when-searching-for-patients


app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -93,7 +92,11 @@ class PatientSearchResultItemView(
     renderPatientAddress(model.address)
     renderPatientDateOfBirth(model.ageDetails)
     renderPatientPhoneNumber(searchQuery, model)
-    renderLastSeen(model.lastSeen, currentFacilityId)
+    renderLastSeen(
```
Honey14, line 94: We've forgotten to change the name of the function here.

Honey14 (changes requested):
Looks great, left one comment PTAL

Honey14 (approved):

PR #3616, opened 2022-03-02
Show patient died status for dead patients in search results

msasikanth:
https://app.shortcut.com/simpledotorg/story/7495/show-patient-died-status-for-dead-patients-in-search-results


app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
```diff
@@ -98,6 +102,11 @@ class PatientSearchResultItemView(
         assignedFacilityName = model.assignedFacilityName
     )
     renderIdentifier(model.identifier, searchQuery)
+    renderPatientStatus(model.status)
+  }
+
+  private fun renderPatientStatus(status: PatientStatus) {
+    patientDiedStatusView.visibleOrGone(isVisible = status == Dead)
```
janhavisinghh, line 109: Looks like we didn't add a test for this. Is there a reason for that?

janhavisinghh (changes requested):

app/src/main/java/org/simple/clinic/widgets/PatientSearchResultItemView.kt
msasikanth: This is a recycler view item layout. So, there aren't any tests for these 😅

msasikanth: This is a recycler view item layout. There aren't tests for this particular item view. Also, unlike view renderer or UI actions we cannot verify that a particular UI call is made unless we provide another abstraction layer. Probably, something to think about in case tests is important in cases like this. 😅

janhavisinghh (approved):

PR #3620, opened 2022-03-04
Move wildcard formatting to search query

msasikanth:
I initially used the wildcard formatting in the repository layer while testing it out. Moved it to query itself to avoid any confusion/issues when using the dao query somewhere else outside of repository layer.


Honey14 (approved):
LGTM

PR #3629, opened 2022-03-09
Add `DrugStockReminder` API service definition

Honey14:
https://app.shortcut.com/simpledotorg/story/7406/add-drugstockreminder-api-support


app/src/main/java/org/simple/clinic/drugstockreminders/CompleteDrugStockReportPayload.kt
```diff
@@ -0,0 +1,17 @@
+package org.simple.clinic.drugstockreminders
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+import java.util.UUID
+
+@JsonClass(generateAdapter = true)
+data class CompleteDrugStockReportPayload(
```
msasikanth, line 8: Any reason why it's "Complete" drug stock report?. Should we rename this to just `DrugStockReportPayload` instead? 🤔

app/src/main/java/org/simple/clinic/drugstockreminders/DrugStockReminderApi.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import retrofit2.Call
+import retrofit2.http.Body
+import retrofit2.http.GET
+
+interface DrugStockReminderApi {
+
+  @GET("v4/drug_stock")
+  fun drugStockReminder(
+      @Body body: DrugStockReminderRequest
```
msasikanth, line 11: We use `@Body` annotation to pass objects usually, in this case, we are just passing date in the object. Should we instead use the `@Query` parameter to pass the date in the API function, WDYT?

```kt
@GET("v4/drug_stock")
fun drugStockReminder(
  // Depending on whether retrofit takes in a `LocalDate` as query parameter, we might have to use a `String`
  @Query("date") date: LocalDate/String
): Call<DrugStockReminderResponsePayload>
```

app/src/main/java/org/simple/clinic/drugstockreminders/DrugStockReminderApi.kt
```diff
@@ -0,0 +1,13 @@
+package org.simple.clinic.drugstockreminders
+
+import retrofit2.Call
+import retrofit2.http.Body
+import retrofit2.http.GET
+
+interface DrugStockReminderApi {
+
+  @GET("v4/drug_stock")
+  fun drugStockReminder(
```
msasikanth, line 10: I am guessing we are going to add API integration test later? Fwiw we can write the test and use `Ignore` annotation to skip that test until server implements it.

app/src/main/java/org/simple/clinic/drugstockreminders/DrugStockReminderRequest.kt
```diff
@@ -0,0 +1,11 @@
+package org.simple.clinic.drugstockreminders
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+data class DrugStockReminderRequest(
```
msasikanth, line 7: Since we are just passing a date, we can avoid creating a new request type and instead use the `@Query` param in the API function. WDYT?

msasikanth (changes requested):
Left couple of comments PTAL

app/src/main/java/org/simple/clinic/drugstockreminders/DrugStockReminderApi.kt
msasikanth: Also, it looks like the API is expecting the date as a query parameter. So, passing it as a request body might not work iirc.

app/src/main/java/org/simple/clinic/drugstockreminders/DrugStockReminderApi.kt
Honey14: Yes we're going to add it in the next PR.

app/src/main/java/org/simple/clinic/drugstockreminders/DrugStockReminderApi.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import retrofit2.Call
+import retrofit2.http.GET
+import retrofit2.http.Query
+
+interface DrugStockReminderApi {
+
+  @GET("v4/drug_stock")
+  fun drugStockReminder(
+      @Query("date") previousMonth: String
```
msasikanth, line 11: We cannot pass the `LocalDate` as query param directly?

msasikanth: We are passing the entire date, so let's keep the name of the param as `date`

Honey14: We can, but the requested param is a string.

msasikanth: I know, just checking if we have verified whether Retrofit will automatically convert the `LocalDate` to string. Since the date string, we need to pass is basically `LocalDate#toString`. It's not that big of a deal, we can merge this in. I just wanted to check if there was an issue with `LocalDate`.

Let's update the param name as the API definition just expects a date, it doesn't care if it's last month or current month. So, let's rename this to `date`, WDYT?

Honey14: Right so I was thinking for the name to be something that we're sending, more readable. 
I'll check the issue with the Localdate/string in the implementation PR.

msasikanth: I am merging this in for now, we can make this change later.

msasikanth (approved):

app/src/main/java/org/simple/clinic/drugstockreminders/DrugStockReminderApi.kt
Honey14: So the whole point of adding @json strings explicitly is because we want to rename it in the variable and fit in our use-case. Hence this makes more sense to me

PR #3637, opened 2022-03-15
Implement critical app update dialog ui

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/7222/implement-criticalappupdatedialog


app/src/main/res/values/strings.xml
```diff
@@ -928,4 +928,14 @@
...
   <string name="critical_app_update_notification_subtext">Important updates available. Update your Simple app from the Google Playstore.</string>
+  
+  <!-- Critical app update dialog -->
+  <string name="critical_update_required_dialog_title">Critical update required</string>
+  <string name="critical_update_required_dialog_content">Update now for critical security changes and latest features</string>
+  <string name="critical_update_required_dialog_error_waning">App security is compromised</string>
+  <string name="critical_update_required_dialog_update_button_text">UPDATE NOW</string>
+  <string name="critical_update_required_dialog_help_section_title">Need help?</string>
+  <string name="critical_update_required_dialog_help_section_content">Tap on the button below to get help from our friendly support team</string>
+  <string name="critical_update_required_dialog_help_section_phone_number">+91 8397847298</string>
```
Honey14, line 939: Do we have same number throughout all countries? Has an Indian code to it as well. Why did we decide to hardcode this?

app/src/main/res/values/strings.xml
```diff
@@ -928,4 +928,14 @@
   <!-- App update notification -->
   <string name="app_update_notification_subtext">Please update your Simple app in the Google Play Store as soon as possible.</string>
   <string name="critical_app_update_notification_subtext">Important updates available. Update your Simple app from the Google Playstore.</string>
+  
+  <!-- Critical app update dialog -->
+  <string name="critical_update_required_dialog_title">Critical update required</string>
+  <string name="critical_update_required_dialog_content">Update now for critical security changes and latest features</string>
```
Honey14, line 934: The text "changes and latest features" should come on the next line as per Figma design. Can you update this?

app/src/main/res/values/strings.xml
```diff
@@ -928,4 +928,14 @@
...
+  <string name="critical_update_required_dialog_error_waning">App security is compromised</string>
+  <string name="critical_update_required_dialog_update_button_text">UPDATE NOW</string>
+  <string name="critical_update_required_dialog_help_section_title">Need help?</string>
+  <string name="critical_update_required_dialog_help_section_content">Tap on the button below to get help from our friendly support team</string>
```
Honey14, line 938: The text "from our friendly support team" should come on the next line as per Figma design. Can you update this?

Honey14 (changes requested):
LGTM, left few comments, PTAL

app/src/main/res/values/strings.xml
janhavisinghh: Afaik, the lines that come in the next line can vary based on the screen size. I don't think it would be a good idea to hardcode it to be on the next line. Wdyt? @Honey14

janhavisinghh: For example in the figma design, they have set a particular text size and have set boundaries and not specifically put the text in the next line. Correct me if I'm wrong.

app/src/main/res/values/strings.xml
janhavisinghh: This is just for the layout file. I will change it to `tools`

janhavisinghh: We will change this dynamically. I have just set this here for the layout file.

Honey14: Instead of having a random number, can we keep the field empty in the layout file and update programmatically when we receive something based on country then?

msasikanth: +1 on that, we should not be using a random number, since we are not sure if this number is operational or not. Either way, if the number is not loaded we simply hide the bottom section.  So we don't have to provide a text value here.

app/src/main/res/values/strings.xml
msasikanth: I don't think it's that big of a requirement in this case. The design device frame is a 360 dpi based one, so if it's showing 2 lines there, that should be enough. You can confirm with design once @janhavisinghh

app/src/main/res/layout/dialog_critical_app_update.xml
```diff
@@ -0,0 +1,127 @@
...
+    android:textAppearance="?attr/textAppearanceButtonBig"
+    android:textColor="?attr/colorPrimary"
+    android:visibility="gone"
+    app:backgroundTint="?attr/colorPrimaryVariant"
+    app:elevation="@dimen/spacing_4"
+    app:icon="@drawable/ic_telegram_logo"
+    app:iconTint="@null"
+    app:layout_constraintEnd_toEndOf="parent"
+    app:layout_constraintStart_toStartOf="parent"
+    app:layout_constraintTop_toBottomOf="@id/helpSectionDescription" />
```
msasikanth, line 126: Instead of having 2 different buttons here, do you think we can instead have 2 styles and update them programmatically @janhavisinghh ? 🤔

app/src/main/res/values/strings.xml
msasikanth: For example, if we are opening the dialog on a high dpi device. There is going to be a lot of white space on the right side. Showing a single line in that scenario might be a good idea. WDYT?

app/src/main/res/layout/dialog_critical_app_update.xml
msasikanth: For example, how would the UI react when we update the json to have a different share type, how do we decide which one to show in this scenario? I was thinking of updating them programmatically, that way we can set a "default" button type that just renders the text. WDYT?

janhavisinghh:
@Honey14 I have added all the requested changes as new commits

app/src/main/res/values/strings.xml
janhavisinghh: I agree @msasikanth

Honey14 (changes requested):
LGTM, android tests are failing, can you PTAL

app/src/main/res/layout/dialog_critical_app_update.xml
```diff
@@ -0,0 +1,107 @@
...
+    android:layout_height="wrap_content"
+    android:layout_marginTop="@dimen/spacing_16"
+    android:elevation="@dimen/spacing_2"
+    android:textAppearance="?attr/textAppearanceButtonBig"
+    app:elevation="@dimen/spacing_4"
+    app:icon="@drawable/ic_whatsapp_logo"
+    app:iconTint="?attr/colorSecondary"
+    app:layout_constraintEnd_toEndOf="parent"
+    app:layout_constraintStart_toStartOf="parent"
+    app:layout_constraintTop_toBottomOf="@id/helpSectionDescription" />
```
msasikanth, line 106: This is good. But do you think we can move the icon, text, and icon tint to the style itself? 

For example, we can have a style named

```xml
  <style name="Widget.Simple.Button.WhatsApp">
```

and similarly `Telegram`, and so on. That way we can just update the style programmatically WDYT? 

_We can do this in the next PR_

msasikanth (approved):

Honey14 (dismissed):
LGTM

msasikanth (approved):

Honey14 (approved):

app/src/main/res/layout/dialog_critical_app_update.xml
janhavisinghh: Afaiu we can not update the style of a button dynamically. There is no `.setStyle` method available. Correct me if I'm wrong? Seems like having two buttons is a better option after all. Thoughts? @msasikanth

msasikanth: We would have to go with 3 different buttons, WhatsApp, Telegram, and a default fallback button for backward compatibility. We cannot update the style directly, but we can take a look at updating the style using a `[ContextThemeWrapper](https://developer.android.com/reference/android/view/ContextThemeWrapper)` for example or updating the individual properties like background, icon, and text at runtime.

PR #3638, opened 2022-03-16
Render app update nudge reason on critical app update dialog

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/7222/implement-criticalappupdatedialog

### DIFF
--- a/app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUi.kt
+++ b/app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUi.kt
@@ -4,4 +4,5 @@ interface CriticalAppUpdateUi {
   fun showHelp()
   fun hideHelp()
   fun renderCriticalAppUpdateReason(appStalenessInMonths: Int)
+  fun renderCriticalSecurityAppUpdateReason()
 }
--- a/app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUi.kt
+++ b/app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUi.kt
@@ -3,4 +3,5 @@ package org.simple.clinic.appupdate.criticalupdatedialog
 interface CriticalAppUpdateUi {
   fun showHelp()
   fun hideHelp()
+  fun renderCriticalAppUpdateReason(appStalenessInMonths: Int)
 }
--- a/app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUiRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUiRenderer.kt
@@ -17,6 +17,9 @@ class CriticalAppUpdateUiRenderer(
 
     if (model.isCriticalUpdateNudgePriority && model.hasAppStaleness)
       ui.renderCriticalAppUpdateReason(appStalenessInMonths(model.appStaleness!!))
+
+    if (model.isCriticalSecurityUpdateNudgePriority)
+      ui.renderCriticalSecurityAppUpdateReason()
   }
 
   private fun appStalenessInMonths(appStaleness: Int): Int {
--- a/app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUiRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUiRenderer.kt
@@ -15,11 +15,15 @@ class CriticalAppUpdateUiRenderer(
       ui.hideHelp()
     }
 
-    if (model.isCriticalUpdateNudgePriority && model.hasAppStaleness)
-      ui.renderCriticalAppUpdateReason(appStalenessInMonths(model.appStaleness!!))
+    renderCriticalAppUpdateReason(model)
+  }
 
-    if (model.isCriticalSecurityUpdateNudgePriority)
+  private fun renderCriticalAppUpdateReason(model: CriticalAppUpdateModel) {
+    if (model.isCriticalUpdateNudgePriority && model.hasAppStaleness) {
+      ui.renderCriticalAppUpdateReason(appStalenessInMonths(model.appStaleness!!))
+    } else if (model.isCriticalSecurityUpdateNudgePriority) {
       ui.renderCriticalSecurityAppUpdateReason()
+    }
   }
 
   private fun appStalenessInMonths(appStaleness: Int): Int {
--- a/app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUiRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUiRenderer.kt
@@ -5,7 +5,8 @@ import java.time.LocalDate
 import java.time.Period
 
 class CriticalAppUpdateUiRenderer(
-    private val ui: CriticalAppUpdateUi
+    private val ui: CriticalAppUpdateUi,
+    private val currentDate: LocalDate
 ) : ViewRenderer<CriticalAppUpdateModel> {
 
   override fun render(model: CriticalAppUpdateModel) {
@@ -27,7 +28,6 @@ class CriticalAppUpdateUiRenderer(
   }
 
   private fun appStalenessInMonths(appStaleness: Int): Int {
-    val currentDate = LocalDate.now()
     val lastUpdatedDate = currentDate.minusDays(appStaleness.toLong())
     return Period.between(lastUpdatedDate, currentDate).months
   }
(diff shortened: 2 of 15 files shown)


app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUiRenderer.kt
```diff
@@ -12,5 +12,8 @@ class CriticalAppUpdateUiRenderer(
     } else {
       ui.hideHelp()
     }
+
+    if (model.isCriticalUpdateNudgePriority && model.hasAppStaleness)
+      ui.renderCriticalAppUpdateReason(model.appStaleness!!)
```
janhavisinghh, line 17: @msasikanth what are your thoughts about having two different methods for rendering the reason in the dialog? I thought of doing it this way because in this method we will also have to manipulate the string a bit according to the app staleness value and in the other its going to be one particular string. Although, we can have a single method for this. Would also like some opinion on the name. Bit unsure about it.

janhavisinghh: I am inclining towards having a single ui renderer method which handles the rest of the logic.

msasikanth: We can do the check in the UI renderer? That way we can have a UI renderer test for this case as well. WDYT?

janhavisinghh: That's how I have implemented it, currently. The string manipulation that we are going to based on `appStaleness` for example writing 3 months or 23 days is a very ui based logic which I think we don't need to add in the renderer.

msasikanth: Instead of passing the `appStaleness`, could we construct the `LocalDate` or `Duration`, since that is what we needed for rendering? If it's a critical security priority update, we can just render the other message. That way we don't have to construct the date or duration in the UI. Thoughts?

janhavisinghh: 👍🏻

app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUi.kt
```diff
@@ -3,4 +3,5 @@ package org.simple.clinic.appupdate.criticalupdatedialog
 interface CriticalAppUpdateUi {
   fun showHelp()
   fun hideHelp()
+  fun renderCriticalAppUpdateReason(appStalenessInMonths: Int)
```
janhavisinghh, line 6: Not sure about the method name. Lmk what are your thoughts. @msasikanth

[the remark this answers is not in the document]
msasikanth: Yep, this looks good 👌🏾

app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUiRenderer.kt
```diff
@@ -12,5 +14,14 @@ class CriticalAppUpdateUiRenderer(
     } else {
       ui.hideHelp()
     }
+
+    if (model.isCriticalUpdateNudgePriority && model.hasAppStaleness)
+      ui.renderCriticalAppUpdateReason(appStalenessInMonths(model.appStaleness!!))
+  }
+
+  private fun appStalenessInMonths(appStaleness: Int): Int {
+    val currentDate = LocalDate.now()
```
msasikanth, line 26: Instead of creating the current date object here, should we instead pass it from the constructor? Because this method is now impure and we cannot set a specific value in our tests as well. WDYT? 

(_side note: We should also use `UserClock` when getting the current date_)

msasikanth: We can also explore `Days` API from threeten extra library. We can use that to get the period and then convert it to the number of months. Like so

```kt
Days.of(appStaleness).toPeriod().months
```

With this, we don't have to construct a `LocalDate` object. You can take a look at whether this works for us.

app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUiRenderer.kt
```diff
@@ -17,6 +17,9 @@ class CriticalAppUpdateUiRenderer(
 
     if (model.isCriticalUpdateNudgePriority && model.hasAppStaleness)
       ui.renderCriticalAppUpdateReason(appStalenessInMonths(model.appStaleness!!))
+
+    if (model.isCriticalSecurityUpdateNudgePriority)
+      ui.renderCriticalSecurityAppUpdateReason()
```
msasikanth, line 22: We are gonna have either a critical update or critical security update priority, but not both right? Should we use an if-else-if / when block here? Because right now both if conditions will run.

msasikanth (changes requested):
Overall looks good. Left couple of comments PTAL

app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUiRenderer.kt
janhavisinghh: yeah I thought about this as well. Although, app Update priority value can only be either `CRITICAL_SECURITY` or `CRITICAL` at one point. It can never be both.

msasikanth: Yep, so let's just go with a `when` block to simplify it?

janhavisinghh: The if conditionals make more sense here because if this is true we show this or else we do nothing. In a when block we will have to add an empty else case as it wont be exhaustive.

msasikanth: Yeah, we can use an `if` block as well. But instead of two different `if` blocks, we can use `if-else-if` to handle the second if condition. Since it's not possible to have 2 different update priorities, we can just have a single conditional block to handle the update priority state. WDYT?

app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUiRenderer.kt
janhavisinghh: We cannot use this `Days.of(appStaleness).toPeriod().months` as the `period.months` method basically returns the months in the periods object and does not calculate the months from days. I have added `currentDate` as part of the constructor.

msasikanth: Cool, sounds good 👍🏾

msasikanth: I think we can also use `Months.from(Days.of(appStaleness)).amount`. But this works. Merging this in 👍🏾

app/src/main/java/org/simple/clinic/appupdate/criticalupdatedialog/CriticalAppUpdateUiRenderer.kt
```diff
@@ -15,11 +15,15 @@ class CriticalAppUpdateUiRenderer(
...
+    renderCriticalAppUpdateReason(model)
+  }
 
-    if (model.isCriticalSecurityUpdateNudgePriority)
+  private fun renderCriticalAppUpdateReason(model: CriticalAppUpdateModel) {
+    if (model.isCriticalUpdateNudgePriority && model.hasAppStaleness) {
+      ui.renderCriticalAppUpdateReason(appStalenessInMonths(model.appStaleness!!))
+    } else if (model.isCriticalSecurityUpdateNudgePriority) {
       ui.renderCriticalSecurityAppUpdateReason()
+    }
```
msasikanth, line 27: Should we throw an error in `else` condition here? If there is an else condition we currently don't render any text and show empty text space.

janhavisinghh: Honestly, that would be over optimising for a situation that's never gonna happen.

janhavisinghh: This dialog will only be shown when there is an app update nudge priority and only when it is critical or critical security

msasikanth: Hmm, I don't think it's over-optimizing in this scenario, since it fails silently. For example, whenever we are using a type in a conditional we either have a "default" state in the else condition or crash it. But yeah, this might not be updated as often. I guess we can leave it as it is.

msasikanth (approved):

PR #3640, opened 2022-03-17
Adding API support for `DrugStockReminder` - 2

Honey14:
https://app.shortcut.com/simpledotorg/story/7406/add-drugstockreminder-api-support

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/drugstockreminders/DrugStockReminderApiIntegrationTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/drugstockreminders/DrugStockReminderApiIntegrationTest.kt
@@ -2,6 +2,7 @@ package org.simple.clinic.drugstockreminders
 
 import com.google.common.truth.Truth.assertThat
 import org.junit.Before
+import org.junit.Ignore
 import org.junit.Test
 import org.simple.clinic.TestClinicApp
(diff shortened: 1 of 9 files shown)


app/src/androidTest/java/org/simple/clinic/drugstockreminders/DrugStockReminderApiIntegrationTest.kt
```diff
@@ -0,0 +1,47 @@
...
+    clock.setDate(LocalDate.parse("2022-03-01"))
+  }
+
+  @Test
+  fun drug_stock_reports_are_filled_or_not_in_the_previous_month_should_be_fetched_to_show_reminders() {
+    // given
+    val dateFormat = DateTimeFormatter.ofPattern("YYYY-MM-DD", Locale.ENGLISH)
+    val previousMonthsDate = LocalDateTime.now(clock).minusMonths(1)
+
+    val formattedDate = dateFormat.format(previousMonthsDate)
```
msasikanth, line 37: We can directly use `LocalDate#toString` which returns a string representation of a date in `yyyy-MM-dd` format (ISO-8601 format). So, we can use `LocalDate.now(clock).toString()`

It's not important to change in this PR, we can do it later as well.

msasikanth (dismissed):
LGTM

msasikanth (approved):

PR #3652, opened 2022-03-28
Fix recycler view crash when refilling drugs

msasikanth:
https://app.shortcut.com/simpledotorg/story/7740/fix-recycler-view-crash-when-refilling-drugs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Fix `trigger_demo_release` workflow running on every newly created branch
 - Implement critical app update card in `PatientsTabScreen`
 - Bump flipper to v0.140.0
+- Use `RecyclerView.AdapterDataObserver` to scroll to top when drugs are added
 - [In Progress: 29 Mar 2022] Implement showing app update nudges based on the priority in `PatientTabScreen`
 
 ### Features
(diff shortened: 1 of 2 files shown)


msasikanth:
While I was not able to reproduce the issue on my device. The issue itself occurs from the `postDelyed` usage on the `RecyclerView` to scroll to the top whenever we are populating the drug's list. The best way to avoid this is to use fragment lifecycle methods to remove any listeners. We have also switched to using `RecyclerView.AdapterDataObserver` to scroll to the top whenever a new item is added to the list.

msasikanth:
> > While I was not able to reproduce the issue on my device. The issue itself occurs from the `postDelyed` usage on the `RecyclerView` to scroll to the top whenever we are populating the drug's list. The best way to avoid this is to use fragment lifecycle methods to remove any listeners. We have also switched to using `RecyclerView.AdapterDataObserver` to scroll to the top whenever a new item is added to the list.
> 
> How can we test the crash fixed?

At this point, the crash itself was not reproducible on my device. But we can try clicking on the refill medicines button to see if the app crashes. Even without reproducing this change is mostly to avoid using the `postDelayed` API and instead use a safer API.

Honey14 (approved):
LGTM

PR #3656, opened 2022-03-29
Update AVD disk size to 8GB in GH Actions

msasikanth:
I have noticed couple of Android tests failed with no space error. So, set the disk size to 8 GB for AVD


Honey14 (approved):
LGTM

PR #3660, opened 2022-03-31
Implement showing app update nudges based on the priority in `PatientTabScreen`

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/7236/implement-showing-app-update-nudges-based-on-the-priority-in-patienttabscreen

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientsTabUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientsTabUpdate.kt
@@ -121,9 +121,12 @@ class PatientsTabUpdate(private val isNotifyAppUpdateAvailableV2Enabled: Boolean
       model: PatientsTabModel,
       event: RequiredInfoForShowingAppUpdateLoaded
   ): Next<PatientsTabModel, PatientsTabEffect> {
-    return when (event.appUpdateNudgePriority!!) {
-      LIGHT, MEDIUM -> showAppUpdateAvailableDialog(model, event)
-      CRITICAL, CRITICAL_SECURITY -> next(model.appUpdateNudgePriorityUpdated(event.appUpdateNudgePriority), ShowCriticalAppUpdateDialog(event.appUpdateNudgePriority))
+    val updatedModel = model.appUpdateNudgePriorityUpdated(event.appUpdateNudgePriority)
+
+    return when (event.appUpdateNudgePriority) {
+      LIGHT, MEDIUM -> showAppUpdateAvailableDialog(updatedModel, event)
+      CRITICAL, CRITICAL_SECURITY -> next(updatedModel, ShowCriticalAppUpdateDialog(event.appUpdateNudgePriority))
+      else -> noChange()
     }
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/simple/clinic/home/patients/PatientsTabUpdate.kt
```diff
@@ -100,4 +105,42 @@ class PatientsTabUpdate(private val isNotifyAppUpdateAvailableV2Enabled: Boolean
...
+  ): Next<PatientsTabModel, PatientsTabEffect> {
+    val appUpdateLastShownOn = event.appUpdateLastShownOn
+    val currentDate = event.currentDate
+
+    val hasADayPassedSinceUpdateLastShown = appUpdateLastShownOn.isBefore(currentDate)
+
+    val shouldShowAppUpdate = event.isAppUpdateAvailable && hasADayPassedSinceUpdateLastShown && model.hasAppUpdateNudgePriority
+
+    return if (shouldShowAppUpdate)
+      next(model.appUpdateNudgePriorityUpdated(event.appUpdateNudgePriority!!), ShowAppUpdateAvailable, TouchAppUpdateShownAtTime)
```
msasikanth, line 142: This is still the older app update dialog but we are updating the nudge priority right?

app/src/main/java/org/simple/clinic/home/patients/PatientsTabUpdate.kt
```diff
@@ -123,7 +123,7 @@ class PatientsTabUpdate(private val isNotifyAppUpdateAvailableV2Enabled: Boolean
   ): Next<PatientsTabModel, PatientsTabEffect> {
     return when (event.appUpdateNudgePriority!!) {
       LIGHT, MEDIUM -> showAppUpdateAvailableDialog(model, event)
-      CRITICAL, CRITICAL_SECURITY -> noChange()
+      CRITICAL, CRITICAL_SECURITY -> next(model.appUpdateNudgePriorityUpdated(event.appUpdateNudgePriority), ShowCriticalAppUpdateDialog(event.appUpdateNudgePriority))
```
msasikanth, line 126: Since we are updating the model in both cases with update priority. Do you think we can extract it out of the when block and pass in something like `updatedModel` to the functions? This can be done in a later PR as well. Please make a note of it if so.

msasikanth (changes requested):
Looks good overall. A test seems to be failing please take a look. The change I suggested can be done in later PR if required.

app/src/main/java/org/simple/clinic/home/patients/PatientsTabUpdate.kt
janhavisinghh: yep!

janhavisinghh: As we were going to remove the old one, I didn't think of refactoring this much.

msasikanth (approved):

PR #3664, opened 2022-04-01
Remove marking older appointments for a patient as a side effect when scheduling an appointment

msasikanth:
https://app.shortcut.com/simpledotorg/story/7763/remove-marking-older-appointments-for-a-patient-as-a-side-effect-when-scheduling-an-appointment

### DIFF
--- a/app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
+++ b/app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
@@ -37,7 +37,6 @@ class AppointmentRepository @Inject constructor(
       appointmentFacilityUuid: UUID,
       creationFacilityUuid: UUID
   ): Appointment {
-
     val appointment = Appointment(
         uuid = appointmentUuid,
         patientUuid = patientUuid,
@@ -55,8 +54,6 @@ class AppointmentRepository @Inject constructor(
         creationFacilityUuid = creationFacilityUuid
     )
 
-    // TODO (vs) 20/05/20: Remove this side effect from this method
-    markOlderAppointmentsAsVisited(patientUuid)
     appointmentDao.save(listOf(appointment))
     return appointment
   }
--- a/app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
+++ b/app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
@@ -1,5 +1,6 @@
 package org.simple.clinic.overdue
 
+import androidx.annotation.VisibleForTesting
 import androidx.paging.PagingSource
 import io.reactivex.Observable
 import org.simple.clinic.home.overdue.OverdueAppointment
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
```diff
@@ -1,5 +1,6 @@
 package org.simple.clinic.overdue
 
+import androidx.annotation.VisibleForTesting
```
janhavisinghh, line 3: Seems like this import was part of the previous commit.

app/src/test/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentEffectHandlerTest.kt
```diff
@@ -142,20 +142,32 @@ class ScheduleAppointmentEffectHandlerTest {
   }
 
   @Test
-  fun `when schedule sppointment for patient from next is received then schedule appointment`() {
+  fun `when schedule appointment for patient from next effect is received, then schedule appointment`() {
```
janhavisinghh, line 146: Can we add `and mark older appointment as visited` in the test name as well?

app/src/test/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentEffectHandlerTest.kt
```diff
@@ -171,6 +172,36 @@ class ScheduleAppointmentEffectHandlerTest {
     verifyZeroInteractions(uiActions)
   }
 
+  @Test
+  fun `when schedule appointment for patient effect is received, then schedule appointment`() {
```
janhavisinghh, line 176: Can we add `and mark older appointment as visited` in the test name as well?

janhavisinghh (changes requested):

app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
msasikanth: Good catch

app/src/test/java/org/simple/clinic/scheduleappointment/ScheduleAppointmentEffectHandlerTest.kt
msasikanth: Ah right, previously it was implicit. Now that we have made it explicit in the effect handler that makes sense. Although I am wondering if adding those side effects increase the names of the tests. For example, the flow itself is scheduling an appointment, and a side effect of it is marking older appointments, that can still be read from the tests themselves. So were essentially verifying the flow rather than everything that might happen in there. WDYT? 

I can rename the test if you think that makes more sense and is easier to understand. I am fine with either.

janhavisinghh: Umm, I personally prefer making the test names explicit. As, I look at tests as a representation of the business logic as well.

msasikanth: Tests themselves will still represent that. Anyway, I have updated the names and pushed the changes.

app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
msasikanth: We don't need to rebase this as we are going to squash the commits anyway. So, it's fine. We can go ahead and merge this in.

janhavisinghh (approved):

PR #3672, opened 2022-04-05
Hide clinical decision support alert banner when edit medicines are clicked

msasikanth:
https://app.shortcut.com/simpledotorg/story/7770/hide-clinical-decision-support-alert-banner-when-edit-medicines-are-clicked

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
@@ -658,11 +658,19 @@ class PatientSummaryScreen :
     clinicalDecisionSupportAlertView.translationY = clinicalDecisionSupportAlertView.height.unaryMinus().toFloat()
 
     val spring = clinicalDecisionSupportAlertView.spring(DynamicAnimation.TRANSLATION_Y)
+
+    val transition = AutoTransition().apply {
+      excludeChildren(clinicalDecisionSupportAlertView, true)
+      excludeTarget(R.id.newBPItemContainer, true)
+      excludeTarget(R.id.bloodSugarItemContainer, true)
+      excludeTarget(R.id.drugsSummaryContainer, true)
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -664,6 +664,8 @@ class PatientSummaryScreen :
       excludeTarget(R.id.newBPItemContainer, true)
       excludeTarget(R.id.bloodSugarItemContainer, true)
       excludeTarget(R.id.drugsSummaryContainer, true)
+      // We are doing this to wait for the router transitions to be done before we start this.
+      startDelay = 500
```
Honey14, line 668: Not sure if its a good idea to set hardcoded delay? WDYT?

Honey14 (changes requested):
LGTM, left one comment PTAL?

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
msasikanth: This is an animation-delay, it's normal mostly? Also, this is not the final implementation of this feature. It's just for field study, in the final version when we pick up, we have to revisit this animation and how it should work. 

So, I am guessing this should be fine for now. WDYT?

msasikanth: or do you mean to have this in the resources? That I can do.

Honey14 (approved):

app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
Honey14: No No, not the resources. If we're going to anyway revisit then its fine.

PR #3682, opened 2022-04-11
Use personal access token when checking out repo in `cut_release_branch` workflow

msasikanth:
Since we want the release branch creation to run a different workflow (triggering demo release). We need to use PAT (Personal Access Token) instead of GITHUB_TOKEN in GH Actions to do that.

### DIFF
--- a/.github/workflows/cut_release_branch.yml
+++ b/.github/workflows/cut_release_branch.yml
@@ -10,6 +10,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: '0'
+          token: ${{ secrets.SERVICES_ACCESS_TOKEN }}
 
       - name: Cut release branch
         run: kotlinc -script ".github/scripts/CutReleaseBranch.main.kts"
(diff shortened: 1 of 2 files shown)


.github/workflows/cut_release_branch.yml
```diff
@@ -10,6 +10,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: '0'
+          token: ${{ secrets.SERVICES_ACCESS_TOKEN }}
```
Honey14, line 13: Is the default token `GITHUB_TOKEN`, if we don't mention PAT?

Honey14 (approved):
LGTM, have to test this once merged

.github/workflows/cut_release_branch.yml
msasikanth: Yep. You can look at the checkout default over here
https://github.com/actions/checkout

```
    # Personal access token (PAT) used to fetch the repository. The PAT is configured
    # with the local git config, which enables your scripts to run authenticated git
    # commands. The post-job step removes the PAT.
    #
    # We recommend using a service account with the least permissions necessary. Also
    # when generating a new PAT, select the least scopes necessary.
    #
    # [Learn more about creating and using encrypted secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets)
    #
    # Default: ${{ github.token }}
    token: ''
```

PR #3811, opened 2022-04-28
Add `updatePriorities` for security app update reminders as a Firebase remote config

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/8037/add-updatepriority-for-security-app-update-reminders-as-a-firebase-remote-config

### DIFF
--- a/app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
+++ b/app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
@@ -31,6 +31,10 @@ class CheckAppUpdateAvailability @Inject constructor(
     private val appVersionFetcher: AppVersionFetcher
 ) {
 
+  companion object {
+    const val UPDATE_PRIORITIES_DEFAULT_VALUE = 0
+  }
+
   fun listen(): Observable<AppUpdateState> {
     return updateManager
         .updateInfo()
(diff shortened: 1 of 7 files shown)


msasikanth:
Since Honey is on leave, I am picking this up.

app/src/main/java/org/simple/clinic/main/TypedMap.kt
```diff
@@ -0,0 +1,11 @@
+package org.simple.clinic.main
+
+import javax.inject.Qualifier
+
+@Qualifier
+annotation class TypedMap(val type: Type) {
```
msasikanth, line 6: Nice 👌🏾

app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
```diff
@@ -31,6 +31,10 @@ class CheckAppUpdateAvailability @Inject constructor(
     private val appVersionFetcher: AppVersionFetcher
 ) {
 
+  companion object {
+    const val UPDATE_PRIORITIES_DEFAULT_VALUE = 0
```
msasikanth, line 35: We can shorten the name and call it `DEFAULT_UPDATE_PRIORITY`, wdyt?

app/src/main/java/org/simple/clinic/home/patients/PatientsInit.kt
```diff
@@ -14,7 +14,8 @@ class PatientsInit(
         LoadUser,
         RefreshUserDetails,
         LoadNumberOfPatientsRegistered,
-        LoadInfoForShowingAppUpdateMessage
+        LoadInfoForShowingAppUpdateMessage,
+        LoadAppStaleness
```
msasikanth, line 18: I think trying to load app staleness here is causing the tests to fail? 

I am just curious, we are already passing priority in `ShowAppUpdate`, any reason why we didn't do this for staleness as well? Is there any issue with that? 🤔

msasikanth (changes requested):
It looks good overall. Unit tests are failing, can you take a look at that, please?

app/src/main/java/org/simple/clinic/home/patients/PatientsInit.kt
janhavisinghh: Not a bad idea. Will do this in the next pr

msasikanth (approved):

PR #3818, opened 2022-05-03
Fix `availableVersionCode` being used to get `updatePriorities` instead of `currentVersionCode`

janhavisinghh:


app/src/main/java/org/simple/clinic/appupdate/CheckAppUpdateAvailability.kt
```diff
@@ -73,14 +73,15 @@ class CheckAppUpdateAvailability @Inject constructor(
...
+    val currentVersionCode = appVersionFetcher.appVersionCode().toString()
     val appStaleness = appVersionStaleness(availableVersionCode)
 
     val appUpdatePriority = config
         .map {
           appUpdateNudgePriority(
               appStaleness = appStaleness,
               config = it,
-              appUpdatePriority = updatePriorities.getOrDefault(availableVersionCode.toString(), DEFAULT_UPDATE_PRIORITY)
+              appUpdatePriority = updatePriorities.getOrDefault(currentVersionCode, DEFAULT_UPDATE_PRIORITY)
```
msasikanth, line 84: We need to use the `availableVersionCode` and not the current version code. We don't need to show an update dialog for the current version right? We only show it for the version that is available for update. 

So, I think we can close this?

janhavisinghh: This change is for getting the update priorities value. Update priorities is dependent on the current version code. Available version code is the value that is available on google play.

msasikanth:
Resolved on a call. Closing this.

PR #3819, opened 2022-05-04
Enable CDSS alerts only in pilot facilities

msasikanth:
https://app.shortcut.com/simpledotorg/story/8158/enable-cdss-alerts-only-in-pilot-facilities


Honey14 (approved):
LGTM

PR #3820, opened 2022-05-05
Fix critical app update dialog UI

janhavisinghh:
From slack [message](https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1651754513474169)


app/src/main/res/layout/dialog_critical_app_update.xml
```diff
@@ -92,13 +92,12 @@
     app:layout_constraintTop_toBottomOf="@+id/helpSectionTitle" />
 
   <com.google.android.material.button.MaterialButton
+    style="@style/Widget.Simple.Button"
```
msasikanth, line 95: The elevation is still not working. I don't think it's the style that's causing the issue, this is applied by default to all buttons. The issue seems to be because of the parent padding, the shadow seems to be getting clipped. Can you take a look at that once, please?

msasikanth: We can get rid of the style and elevation from this button if that's the case.

msasikanth (changes requested):

msasikanth:
@janhavisinghh can we also remove elevation attrs from `supportTeamContactButton`? It will automatically be taken care of by the default style. So, we don't need to set the elevation at view level.

msasikanth (approved):

PR #3825, opened 2022-05-09
Redact measurements PHI from logs

msasikanth:
https://app.shortcut.com/simpledotorg/story/8195/redact-measurements-phi-from-logs


janhavisinghh (approved):
LGTM.

PR #3838, opened 2022-05-16
Show call result status for the appointment in `ContactPatientBottomSheet` - 2

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/8168/show-call-result-status-for-the-appointment-in-contactpatientbottomsheet

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Bump Sentry to v5.7.4  
 - Fix CDSS alert hiding when multiple BPs are added on the same day
 - Bump flipper to v0.145.0
-- [In Progress: 11 May 2022] Show call result status for the appointment in `ContactPatientBottomSheet`
+- [In Progress: 16 May 2022] Show call result status for the appointment in `ContactPatientBottomSheet`
 - [In Progress: 27 Apr 2022] Migrate `BloodPressureHistoryScreenEffectHandler` to use view effect
 
 ### Changes
--- a/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
@@ -36,6 +36,8 @@ class ContactPatientUiRenderer(
     if (model.hasCallResult) {
       val callResult = model.callResult!!.get()
       ui.showCallResult(callResult.outcome, callResult.timestamps.updatedAt)
+    } else {
+      ui.hideCallResult()
     }
   }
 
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 ### Internal
 
 - Bump Sentry to v5.7.4  
-- [In Progress: 11 May 2022] Show call result status for the appointment in `ContactPatientBottomSheet`
+- [In Progress: 16 May 2022] Show call result status for the appointment in `ContactPatientBottomSheet`
```
msasikanth, line 8: nit: there is an extra space at end, can we format the file, please?

app/src/test/java/org/simple/clinic/contactpatient/ContactPatientUpdateTest.kt
```diff
@@ -88,12 +88,12 @@ class ContactPatientUpdateTest {
         .whenEvent(OverdueAppointmentLoaded(appointment))
         .then(assertThatNext(
             hasModel(defaultModel.overdueAppointmentLoaded(appointment)),
-            hasNoEffects()
+            hasEffects(LoadCallResultForAppointment(overdueAppointment.uuid))
```
msasikanth, line 91: I think this is something we need to take a look at later. Whether to load them in a blocking manner one after the other or do them in a single call, both the appointment and call result. For now we can go ahead with this 👍🏾

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
```diff
@@ -31,6 +32,13 @@ class ContactPatientUiRenderer(
     }
   }
 
+  private fun renderCallResultOutcome(model: ContactPatientModel) {
+    if (model.hasCallResult) {
+      val callResult = model.callResult!!.get()
+      ui.showCallResult(callResult.outcome, callResult.timestamps.updatedAt)
+    }
```
msasikanth, line 39: Will we raise another PR for hiding the call result?

msasikanth (changes requested):
See inline comments

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUiRenderer.kt
janhavisinghh: resolved

msasikanth (dismissed):

app/src/main/java/org/simple/clinic/contactpatient/ContactPatientUpdate.kt
```diff
@@ -56,7 +56,7 @@ class ContactPatientUpdate(
           .contactPatientInfoLoaded()
     }
 
-    return next(updatedModel)
+    return next(updatedModel, LoadCallResultForAppointment(event.overdueAppointment.get().uuid))
```
msasikanth, line 59: Overdue appointment here is of type `Optional`, would this cause any issues for us? Are we sure we will have overdue appointment here always?

janhavisinghh: umm good catch. Added a commit for this

msasikanth (approved):

PR #3864, opened 2022-05-24
Show call result section by default in `ContactPatientBottomSheet`

Honey14:
https://app.shortcut.com/simpledotorg/story/8169/show-call-result-section-by-default-in-contactpatientbottomsheet

### DIFF
--- a/app/src/main/java/org/simple/clinic/overdue/Appointment.kt
+++ b/app/src/main/java/org/simple/clinic/overdue/Appointment.kt
@@ -310,8 +310,8 @@ data class Appointment(
       WHERE 
         patientUuid = :patientUUID
         AND scheduledDate < :scheduledDate
-        AND (remindOn < :scheduledDate OR remindOn IS NULL)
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/simple/clinic/overdue/Appointment.kt
```diff
@@ -310,8 +310,8 @@ data class Appointment(
       WHERE 
         patientUuid = :patientUUID
         AND scheduledDate < :scheduledDate
-        AND (remindOn < :scheduledDate OR remindOn IS NULL)
-        AND deletedAt IS NULL AND status = 'scheduled'
+        AND deletedAt IS NULL
```
janhavisinghh, line 313: Is there a reason why we have not added an android test for this use case?

janhavisinghh (changes requested):

janhavisinghh:
Android tests seem to be failing. Can you ptal? @Honey14

PR #3866, opened 2022-05-25
Limit instant search to alphanumeric characters

msasikanth:
https://app.shortcut.com/simpledotorg/story/8260/handle-special-characters-when-searching-for-patient

Another potential option we can take is replacing the special characters with a "safe" character. But that would require 
us to 

1. Either change the characters saved in DB or
2. Replace them while fetching and do transformations

Both of them felt like a complex solutions for the use case. So, we can go ahead with the alphanumeric search for now and 
based on user feedback take the call

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,15 @@
 
 - Bump flipper to v0.146.0
 - Fix test name in `PatientSummaryUpdateTest`
-- Bump mixpanel to v6.2.2  
-- Bump datadog plugin to v1.4.0  
+- Bump mixpanel to v6.2.2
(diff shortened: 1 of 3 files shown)


msasikanth:
* master
  * **PR #3866** <a href="https://app.graphite.dev/github/pr/simpledotorg/simple-android/3866" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/simpledotorg/simple-android/3866?utm_source=stack-comment).

Honey14 (approved):
LGTM

PR #3870, opened 2022-05-27
Add query to load all overdue appointments for a facility

msasikanth:
https://app.shortcut.com/simpledotorg/story/8325/add-query-to-load-all-overdue-appointments

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentNew.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentNew.kt
@@ -1,0 +1,130 @@
+package org.simple.clinic.home.overdue
+
+import android.os.Parcelable
+import androidx.room.Dao
+import androidx.room.Embedded
+import androidx.room.Query
+import kotlinx.parcelize.Parcelize
+import org.simple.clinic.overdue.Appointment
+import org.simple.clinic.overdue.callresult.CallResult
+import org.simple.clinic.patient.Gender
+import org.simple.clinic.patient.PatientAgeDetails
+import org.simple.clinic.patient.PatientPhoneNumber
+import java.time.LocalDate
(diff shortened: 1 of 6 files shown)


msasikanth:
* master
  * **PR #3870** <a href="https://app.graphite.dev/github/pr/simpledotorg/simple-android/3870" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/simpledotorg/simple-android/3870?utm_source=stack-comment).

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointmentNew.kt
```diff
@@ -0,0 +1,130 @@
...
+import org.simple.clinic.overdue.Appointment
+import org.simple.clinic.overdue.callresult.CallResult
+import org.simple.clinic.patient.Gender
+import org.simple.clinic.patient.PatientAgeDetails
+import org.simple.clinic.patient.PatientPhoneNumber
+import java.time.LocalDate
+import java.util.UUID
+
+@Parcelize
+data class OverdueAppointmentNew(
```
janhavisinghh, line 17: Is there a reason why we did we not go with `OverdueAppointment_Old` and `OverdueAppointment` names for this like we usually do when we are creating a new version of something?

janhavisinghh (changes requested):
See inline comments

msasikanth (commented):
> Is there a reason why we did we not go with `OverdueAppointment_Old` and `OverdueAppointment` names for this like we usually do when we are creating a new version of something?

Yeah, renaming it would cause a lot of changes. So went with instead creating a new data type. Once we replace all the older usages, we can delete it and rename it. Similar process, but order changes I guess 😂

janhavisinghh (approved):

PR #3876, opened 2022-05-30
Enable monthly drug stock reminders only in India

Honey14:
https://app.shortcut.com/simpledotorg/story/8361/enable-drug-stock-only-for-india

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/patients/PatientsTabScreen.kt
+++ b/app/src/main/java/org/simple/clinic/home/patients/PatientsTabScreen.kt
@@ -209,7 +209,7 @@ class PatientsTabScreen : BaseScreen<
 
   override fun createInit() = PatientsInit(
       isNotifyAppUpdateAvailableV2Enabled = features.isEnabled(NotifyAppUpdateAvailableV2),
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/simple/clinic/home/patients/PatientsTabScreen.kt
```diff
@@ -230,7 +230,8 @@ class PatientsTabScreen : BaseScreen<
     homeIllustration.setImageResource(illustrationResourceId())
     simpleVideoIllustration.setImageResource(videoIllustrationResourceId())
 
-    if (features.isEnabled(MonthlyDrugStockReportReminder)) {
+    val isMonthlyDrugStockReminderEnabledInIndia = features.isEnabled(MonthlyDrugStockReportReminder) && country.isoCountryCode == Country.INDIA
+    if (isMonthlyDrugStockReminderEnabledInIndia) {
```
msasikanth, line 234: This prevents us from scheduling drug stock check outside of India. But what about the side effect we are loading when patients screen is created? Should we have a check for that as well?

msasikanth (changes requested):

msasikanth (approved):

PR #3891, opened 2022-06-01
Add feature flag for overdue sections

Honey14:
https://app.shortcut.com/simpledotorg/story/8347/add-feature-flag


app/src/main/java/org/simple/clinic/feature/Feature.kt
```diff
@@ -35,5 +35,6 @@ enum class Feature(
   NextAppointment(false, "next_appointment_v1"),
   AddingHealthIDsFromEditPatient(false, "adding_health_ids_from_edit_patient"),
   NotifyAppUpdateAvailableV2(false, "appupdate_enabled_v2"),
-  MonthlyDrugStockReportReminder(false, "monthly_drug_stock_report_reminders_v1")
+  MonthlyDrugStockReportReminder(false, "monthly_drug_stock_report_reminders_v1"),
+  OverdueSections(false, "overdue_section_improvements_v1")
```
msasikanth, line 39: Until this feature is ready, let's change the version of the feature flag to v0? `overdue_section_improvements_v0`. 

Because when this feature is finished, we would have to change this version considering it might take a couple of weeks to finish this. To avoid any backward compatibility issues. So, we have to change this to v2 at that time.

Honey14: Oh yes, my bad. Updating this

msasikanth (approved):

PR #3908, opened 2022-06-06
Fetch the last created appointment for the patient when fetching the latest overdue appointment

msasikanth:
https://app.shortcut.com/simpledotorg/story/8437/fetch-the-last-created-appointment-for-the-patient-when-fetching-the-latest-overdue-appointment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - Bump core-ktx to 1.8.0
 - Fix `RetentionType` getting obfuscated by R8/Proguard
 - Bump appcompat to v1.4.2
+- Fetch the last created appointment for the patient when fetching the latest overdue appointment
 - [In Progress: 27 Apr 2022] Migrate `BloodPressureHistoryScreenEffectHandler` to use view effect
 - [In Progress: 03 Jun 2022] Implement `OverdueAppointmentListItemNew` adapter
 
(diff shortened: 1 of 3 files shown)


msasikanth:
* master
  * **PR #3908** <a href="https://app.graphite.dev/github/pr/simpledotorg/simple-android/3908" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/simpledotorg/simple-android/3908?utm_source=stack-comment).

janhavisinghh (approved):
Honestly would have preferred if we had changes the test in a way first that this particular case would have failed and then changed the query and then changed the tests to pass for this scenario as well. Currently we have just added more details to the already passing tests so it doesn't really seem very useful. LGTG for now tho.

msasikanth:
> Honestly would have preferred if we had changes the test in a way first that this particular case would have failed and then changed the query and then changed the tests to pass for this scenario as well. Currently we have just added more details to the already passing tests so it doesn't really seem very useful. LGTG for now tho.

Yeah. But failing the tests means the CI checks won't pass and not allow us to merge in the changes, so there is no way for us to do this iteratively. You can verify the tests by adding back the remind check in query and leaving the test changes. This makes the tests fail FWIW.

PR #3913, opened 2022-06-06
When there's no overdue appointment, show no patients overdue

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/8340/when-there-s-no-overdue-appointment-show-no-patients-overdue

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/OverdueScreen.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/OverdueScreen.kt
@@ -277,7 +277,6 @@ class OverdueScreen : BaseScreen<
   }
 
   override fun showOverdueAppointments(overdueAppointmentSections: OverdueAppointmentSections, pendingListState: PendingListState) {
-    overdueRecyclerView.visibility = View.VISIBLE
     overdueListAdapter.submitList(OverdueAppointmentListItemNew.from(
         overdueAppointmentSections = overdueAppointmentSections,
         clock = userClock,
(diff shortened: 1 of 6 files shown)


msasikanth (approved):

msasikanth (commented):
@janhavisinghh looks like unit tests are failing, PTAL.

app/src/main/java/org/simple/clinic/home/overdue/OverdueScreen.kt
```diff
@@ -297,6 +297,11 @@ class OverdueScreen : BaseScreen<
     overdueProgressBar.visibility = View.GONE
   }
 
+  override fun handleNoOverduePatientsViewVisibility(shouldShowEmptyView: Boolean) {
```
msasikanth, line 300: I am not sure about this. Usually we handle these visibility check in UI renderer. We end up toggling `overdueRecyclerView` visibility here and `showOverdueAppointments` function as well.

Maybe we can instead move the conditional check in UI renderer and toggle between `showOverdueAppointments` and ` hideOverdueAppointment`? That way we can have separate functions for handling no overdue appointments view as well. Like `showNoOverdueAppointmentsView` and `hideNoOverdueAppointmentsView`. WDYT?

app/src/main/java/org/simple/clinic/home/overdue/OverdueUiRenderer.kt
```diff
@@ -18,8 +18,25 @@ class OverdueUiRenderer(
...
+    ui.hideNoOverduePatientsView()
+  }
+
+  private fun renderOverdueListLoadedViews(model: OverdueModel) {
+    if (model.isOverdueAppointmentSectionsListEmpty) {
+      ui.showNoOverduePatientsView()
+      ui.hideOverdueRecyclerView()
+    } else {
+      ui.hideNoOverduePatientsView()
+      ui.showOverdueRecyclerView()
```
msasikanth, line 39: [goodidea](https://app.graphite.dev/macros/simpledotorg)

msasikanth (approved):

PR #3948, opened 2022-06-15
Define overdue appointment query as a `const` in `OverdueAppointment`

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/8519/implement-search-query-for-overdue-patients

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
@@ -6,6 +6,7 @@ import androidx.room.Embedded
 import androidx.room.Query
 import io.reactivex.Observable
 import kotlinx.parcelize.Parcelize
+import org.intellij.lang.annotations.Language
 import org.simple.clinic.overdue.Appointment
 import org.simple.clinic.overdue.callresult.CallResult
 import org.simple.clinic.patient.Gender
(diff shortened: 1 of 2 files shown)


janhavisinghh:
* master
  * **PR #3948** <a href="https://app.graphite.dev/github/pr/simpledotorg/simple-android/3948" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/simpledotorg/simple-android/3948?utm_source=stack-comment).

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
```diff
@@ -44,7 +45,10 @@ data class OverdueAppointment(
   @Dao
   interface RoomDao {
 
-    @Query("""
+    companion object {
+
+      @Language("RoomSql")
+      private const val OVERDUE_APPOINTMENTS_QUERY = """
```
msasikanth, line 51: Since we moved `WHERE`, `GROUP BY` and `ORDER BY` to const. We cannot add new conditions like search later on. Ideally, the const query should only contain the query without the conditions, so that we can define them per query function basis. We can get on a short call if you want more information about this issue.

msasikanth (changes requested):
See inline comments

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
janhavisinghh: I got it, thanks for pointing this out 👍🏻

msasikanth (approved):

PR #3957, opened 2022-06-16
Implement query to search overdue patients

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/8519/implement-search-query-for-overdue-patients

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
@@ -2686,6 +2686,105 @@ class AppointmentRepositoryAndroidTest {
       }
     }
   }
+
+  @Test
+  fun searching_for_overdue_appointments_should_work_correctly() {
+    fun createOverdueAppointment(
+        patientName: String,
+        colonyOrVillageName: String,
+        patientAddressUuid: UUID,
+        patientUuid: UUID,
+        facilityUuid: UUID
+    ) {
+      val patientAddress = testData.patientAddress(
+          uuid = patientAddressUuid,
+          colonyOrVillage = colonyOrVillageName
+      )
+
+      val patientProfile = testData.patientProfile(
+          patientUuid = patientUuid,
+          generatePhoneNumber = true,
+          patientAddressUuid = patientAddressUuid,
+          patientName = patientName
+      )
+
+      patientRepository.save(listOf(patientProfile))
+      database.addressDao().save(listOf(patientAddress))
+
+      val bp = testData.bloodPressureMeasurement(
+          patientUuid = patientUuid,
+          facilityUuid = facilityUuid
+      )
+      bpRepository.save(listOf(bp))
+
+      val appointment = testData.appointment(
+          patientUuid = patientUuid,
+          facilityUuid = facilityUuid,
+          scheduledDate = LocalDate.now().minusDays(10),
+          status = Scheduled,
+          cancelReason = null
+      )
+      appointmentRepository.save(listOf(appointment))
+    }
+
+    // given
+    val searchQuery = "bar"
+    val facilityUuid = UUID.fromString("722111d2-459c-47a6-859c-ca3564fc374d")
+
+    val patient1WithMatchingName = UUID.fromString("7c7a09ce-d5b7-4c75-9983-01704d8e9cc6")
+    val patient2WithMatchingName = UUID.fromString("385cf640-ba6d-4fac-8098-612e2620643b")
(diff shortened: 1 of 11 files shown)


janhavisinghh:
* master
  * **PR #3957** <a href="https://app.graphite.dev/github/pr/simpledotorg/simple-android/3957" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/simpledotorg/simple-android/3957?utm_source=stack-comment).

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
```diff
@@ -2686,6 +2686,105 @@ class AppointmentRepositoryAndroidTest {
...
+    }
+
+    // given
+    val searchQuery = "bar"
+    val facilityUuid = UUID.fromString("722111d2-459c-47a6-859c-ca3564fc374d")
+
+    val patient1WithNameMatchingQuery = UUID.fromString("7c7a09ce-d5b7-4c75-9983-01704d8e9cc6")
+    val patient2WithNameMatchingQuery = UUID.fromString("385cf640-ba6d-4fac-8098-612e2620643b")
+    val patientWithVillageNameMatchingQuery = UUID.fromString("0d165e45-d12d-4a4a-bbc4-5f01a7955486")
+    val patientWithNothingMatchingQuery = UUID.fromString("d2046ee2-8474-4202-989d-1f2eac2deb2a")
```
msasikanth, line 2737: Let's remove `Query` suffix from the naming here. This is not a query, so we can just call it `patientWithSomeCondition`

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
```diff
@@ -134,5 +135,44 @@ data class OverdueAppointment(
...
+             JOIN PatientFts patientFts ON patientFts.uuid = A.patientUuid
+             WHERE
+               IFNULL(patientAssignedFacilityUuid, appt_facilityUuid) = :facilityUuid AND
+               appt_scheduledDate < :scheduledBefore AND
+               P.deletedAt IS NULL AND
+               P.status != 'dead' AND 
+               PPN.deletedAt IS NULL AND 
+               A.deletedAt IS NULL AND 
+               (A.status = 'scheduled' OR A.status  = 'cancelled') AND 
+               (BP.recordedAt IS NOT NULL OR BloodSugar.recordedAt IS NOT NULL) AND
```
msasikanth, line 151: Any reason we didn't pull out the common conditions to outside query? Because we are duplicating them in both search sub-queries.

msasikanth (changes requested):
See inline comments

app/src/androidTest/java/org/simple/clinic/overdue/AppointmentRepositoryAndroidTest.kt
janhavisinghh: That's the condition right? Name matching the search query? which is why I have named it like this

msasikanth: So, the condition is just `patientWithMatchingName` or `patient2WithMatchingVillageName`, usually how we name these test patients with different conditions. The query suffix kind of  implies that this is a query rather than an ID.

msasikanth (changes requested):
Addressing performance issues with the query

app/src/main/java/org/simple/clinic/home/overdue/OverdueAppointment.kt
```diff
@@ -134,5 +141,26 @@ data class OverdueAppointment(
...
     ): Observable<List<OverdueAppointment>>
+
+    @Query("""
+      $OVERDUE_APPOINTMENTS_QUERY
+      WHERE
+        P.uuid IN (
+          SELECT uuid FROM PatientFts WHERE fullName MATCH "*"||:query||"*"
+          UNION
+          SELECT P.uuid WHERE P.addressUuid = (
+            SELECT uuid FROM PatientAddressFts WHERE colonyOrVillage MATCH "*"||:query||"*"
```
msasikanth, line 152: Instead of creating another sublist with this nested subquery. Could we have used the `OR` condition instead? For example


```sql
WHERE
  (
     P.uuid IN (SELECT uuid FROM PatientFts WHERE fullName MATCH "*"||:query||"*") OR
     P.addressUuid IN (SELECT uuid FROM PatientAddressFts WHERE colonyOrVillage MATCH "*"||:query||"*")
  )
```

WDYT?

msasikanth (changes requested):
See inline comments

msasikanth (approved):

PR #3958, opened 2022-06-17
Save overdue search history as `String` type in preferences

msasikanth:
Unfortunately, the string set preference type doesn't seem to maintain the order in which the items were inserted/added. This becomes a problem when we have to drop older searches as we save newer ones based on the search history limit. To avoid this we will save the search history as a `String` with delimiter and split it. Another potential option is to save it as a JSON array and use Moshi to parse it. Either way, we have to transform the data into a string. 

This is in preparation to extract this into a separate `OverdueSearchHistory` class that takes care of fetching, transforming and updating the overdue search history.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 - Add overdue search query validator
 - Validate search query when overdue search query is changed
 - Bump overdue sections feature flag version to v1
+- Save overdue search history as `String` type in preferences
 - [In Progress: 27 Apr 2022] Migrate `BloodPressureHistoryScreenEffectHandler` to use view effect
 
 ## 2022-06-13-8291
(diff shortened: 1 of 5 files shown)


msasikanth:
* master
  * **PR #3958** <a href="https://app.graphite.dev/github/pr/simpledotorg/simple-android/3958" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/simpledotorg/simple-android/3958?utm_source=stack-comment).

msasikanth:
https://app.shortcut.com/simpledotorg/story/8542/add-overduesearchhistory-class

janhavisinghh (approved):

janhavisinghh (approved):

PR #4018, opened 2022-06-23
Implement view effects to show overdue search results

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/8598/implement-view-effects-to-show-overdue-search-results

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchUiActions.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchUiActions.kt
@@ -1,8 +1,11 @@
 package org.simple.clinic.home.overdue.search
 
+import androidx.paging.PagingData
+import org.simple.clinic.home.overdue.OverdueAppointment
 import java.util.UUID
 
 interface OverdueSearchUiActions {
   fun openPatientSummaryScreen(patientUuid: UUID)
   fun openContactPatientSheet(patientUuid: UUID)
+  fun showOverdueSearchResults(searchResults: PagingData<OverdueAppointment>)
 }
--- a/app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchUiRenderer.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchUiRenderer.kt
@@ -14,15 +14,6 @@ class OverdueSearchUiRenderer(
       ui.hideSearchResults()
     } else {
       ui.hideSearchHistory()
-      showOverdueSearchResults(model)
-    }
-  }
-
-  private fun showOverdueSearchResults(model: OverdueSearchModel) {
-    if (model.hasOverdueSearchResults) {
-      ui.showOverdueSearchResults(model.overdueSearchResults!!)
-    } else {
-      ui.hideSearchResults()
     }
   }
 }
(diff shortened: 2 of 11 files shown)


janhavisinghh:
* master
  * **PR #4018** <a href="https://app.graphite.dev/github/pr/simpledotorg/simple-android/4018" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/simpledotorg/simple-android/4018?utm_source=stack-comment).

app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchUiActions.kt
```diff
@@ -1,8 +1,11 @@
 package org.simple.clinic.home.overdue.search
 
+import androidx.paging.PagingData
+import org.simple.clinic.home.overdue.OverdueAppointment
 import java.util.UUID
 
 interface OverdueSearchUiActions {
   fun openPatientSummaryScreen(patientUuid: UUID)
   fun openContactPatientSheet(patientUuid: UUID)
+  fun showOverdueSearchResults(expectedPagingData: PagingData<OverdueAppointment>)
```
msasikanth, line 10: Let's change the param name to `searchResults`

app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchUiRenderer.kt
```diff
@@ -7,20 +7,11 @@ class OverdueSearchUiRenderer(
 ) : ViewRenderer<OverdueSearchModel> {
 
   override fun render(model: OverdueSearchModel) {
-    if (!model.hasSearchQuery) {
+    if (!model.hasSearchQuery && !model.hasOverdueSearchResults) {
```
msasikanth, line 10: Now that we are no longer having search results in model. We can remove the `hasOverdueSearchResults` check? WDYT?

msasikanth (changes requested):
See inline comments

app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchUiRenderer.kt
msasikanth: We can also drop the code related to search results in the model

janhavisinghh: We are using this to show/hide search history and overdue search result recycler view. If we don't have any search results and we don't have a search query then we show the search history for example.

msasikanth: But since we no longer want to update the search results in the model. This wouldn't work no? For example, we can now just hide the search history when the search query is not empty and hide the search results when the search query is empty.

janhavisinghh: Yeah but then what about the case when the search query is empty and we clicked on a search history item and search results are shown. In that case also we need to hide the search history list

msasikanth: So, we can hide the search results list when search query is empty and show search history list instead, and vice versa.

janhavisinghh: So I understand why the current logic is wrong. But, we still have to handle the case I talked about above. Hiding search history when search query is empty and search results are loaded

msasikanth (changes requested):
Discussed changes on-call

msasikanth (approved):

PR #4041, opened 2022-06-29
Ignore `ReportsModel#reportsContent` on parcel

msasikanth:
Since `ReportsModel#reportsContent` is a HTML string, the size can vary significantly depending on how much data it has. Since the parcel limit is shared across transactions in a process. It's not ideal to save huge data into parcel. Instead of saving and restoring the data in reports screen, we can fetch it from DB on restore to avoid this issue. 

You can use [this](git@github.com:guardian/toolargetool.git) tool to check how much space each transaction operation is taking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Increase overdue search query text field debounce time
 - Add remote key for overdue search feature flag
 - Fix `JSONDataException` when parsing the drug stock response
+- Ignore `ReportsModel#reportsContent` on parcel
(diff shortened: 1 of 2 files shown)


msasikanth:
* master
  * **PR #4041** <a href="https://app.graphite.dev/github/pr/simpledotorg/simple-android/4041" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/simpledotorg/simple-android/4041?utm_source=stack-comment).

janhavisinghh (approved):

PR #4064, opened 2022-07-08
Show a checkbox when select overdue patients feature flag is enable

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/8628/show-a-checkbox-when-select-overdue-patients-feature-flag-is-enable

### DIFF
--- a/app/src/main/res/layout/list_item_overdue_patient.xml
+++ b/app/src/main/res/layout/list_item_overdue_patient.xml
@@ -19,9 +19,24 @@
       android:id="@+id/patientGenderIcon"
       android:layout_width="wrap_content"
       android:layout_height="wrap_content"
+      android:visibility="gone"
+      app:layout_constraintEnd_toStartOf="@id/patientNameTextView"
       app:layout_constraintStart_toStartOf="parent"
       app:layout_constraintTop_toTopOf="parent"
       tools:src="@drawable/ic_patient_female" />
+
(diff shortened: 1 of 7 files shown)


app/src/main/res/layout/list_item_overdue_patient.xml
```diff
@@ -23,6 +23,18 @@
       app:layout_constraintTop_toTopOf="parent"
       tools:src="@drawable/ic_patient_female" />
 
+    <com.google.android.material.checkbox.MaterialCheckBox
+      android:id="@+id/checkbox"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:layout_marginEnd="@dimen/spacing_12"
+      android:minWidth="@dimen/spacing_0"
+      android:minHeight="@dimen/spacing_0"
```
msasikanth, line 32: Why are we setting the `minWidth` and `minHeight` here?  I don't think we need it?

msasikanth (changes requested):
See inline comments

app/src/main/res/layout/list_item_overdue_patient.xml
janhavisinghh: Because there is some extra padding by default in the checkbox for at the end sending a scree shot in dm

msasikanth (approved):

PR #4072, opened 2022-07-12
Ignore deleted and old phone numbers when generating overdue CSV from database

msasikanth:
https://app.shortcut.com/simpledotorg/story/8680/ignore-deleted-and-old-phone-numbers-when-generating-overdue-csv-from-database

### DIFF
--- a/app/src/main/java/org/simple/clinic/overdue/Appointment.kt
+++ b/app/src/main/java/org/simple/clinic/overdue/Appointment.kt
@@ -382,7 +382,11 @@ data class Appointment(
       FROM Appointment A
       INNER JOIN Patient P ON P.uuid = A.patientUuid
       LEFT JOIN PatientAddress PA ON PA.uuid = P.addressUuid
-      LEFT JOIN PatientPhoneNumber PPN ON PPN.patientUuid = P.uuid AND PPN.deletedAt IS NULL
+      LEFT JOIN (
+        SELECT * FROM PatientPhoneNumber
+        WHERE deletedAt IS NULL
+        GROUP BY patientUuid HAVING MAX(createdAt)
+      ) PPN ON PPN.patientUuid = P.uuid AND PPN.deletedAt IS NULL
       LEFT JOIN (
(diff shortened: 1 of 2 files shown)


msasikanth:
* master
  * **PR #4072** <a href="https://app.graphite.dev/github/pr/simpledotorg/simple-android/4072" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/simpledotorg/simple-android/4072?utm_source=stack-comment).

msasikanth:
Technically a phone number cannot be deleted. Whenever we are updating a phone number, it will just be saved as a new one. I am just being a bit careful here with that condition. But the important case here is select the latest phone number for a patient.

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
```diff
@@ -382,7 +382,11 @@ data class Appointment(
       FROM Appointment A
       INNER JOIN Patient P ON P.uuid = A.patientUuid
       LEFT JOIN PatientAddress PA ON PA.uuid = P.addressUuid
-      LEFT JOIN PatientPhoneNumber PPN ON PPN.patientUuid = P.uuid AND PPN.deletedAt IS NULL
+      LEFT JOIN (
+        SELECT * FROM PatientPhoneNumber
+        WHERE deletedAt IS NULL
```
janhavisinghh, line 387: Is there a reason why we haven't added a condition in the tests to verify this behaviour?

janhavisinghh (dismissed):

app/src/main/java/org/simple/clinic/overdue/Appointment.kt
msasikanth: We cannot actually add multiple phone numbers to our test nor do we have a public API to directly add the phone number unless we directly inject the DAO object bypassing repository. That's why.

janhavisinghh (approved):

PR #4078, opened 2022-07-13
Bump Material Design Components to v1.6.1

msasikanth:
Previously we didn't update the Material library as Material 3 spec was just introduced and there were some backward compatability issues for us. With recent version it seems much more stable and those issues were addressed. So we are updating Material library again.

But we are not switching to using Material3 or Material You. We will continue using `Theme.MaterialComponents.*` and as such all the components are themed to Material 2 specs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - Add support for select and download in overdue search screen
 - Add support for select and share in overdue search screen
 - Show no internet connection dialog when download/share button is clicked
+- Bump Material Design Components to v1.6.1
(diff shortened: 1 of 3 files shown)


msasikanth:
* master
  * **PR #4078** <a href="https://app.graphite.dev/github/pr/simpledotorg/simple-android/4078" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/simpledotorg/simple-android/4078?utm_source=stack-comment).

janhavisinghh (approved):

PR #4104, opened 2022-07-18
Add support for selecting all overdue appointment search results

Honey14:
https://app.shortcut.com/simpledotorg/story/8677/add-support-for-selecting-all-overdue-appointment-search-results

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchEffect.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchEffect.kt
@@ -38,3 +38,5 @@ object OpenSelectShareFormatDialog : OverdueSearchViewEffect()
 object OpenShareInProgressDialog : OverdueSearchViewEffect()
 
 object ShowNoInternetConnectionDialog : OverdueSearchViewEffect()
+
+data class AddAllAppointmentIds(val appointmentIds: Set<UUID>): OverdueSearchEffect()
--- a/app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchEffectHandler.kt
@@ -51,7 +51,20 @@ class OverdueSearchEffectHandler @AssistedInject constructor(
         .addConsumer(ClearSelectedOverdueAppointments::class.java, ::clearSelectedOverdueAppointments)
         .addTransformer(ReplaceSelectedAppointmentIds::class.java, replaceSelectedAppointmentIds())
         .addConsumer(ScheduleDownload::class.java, ::scheduleDownload)
+        .addTransformer(AddAllAppointmentIds::class.java, addAllAppointmentIds())
         .build()
+  }
+
+  private fun addAllAppointmentIds(): ObservableTransformer<AddAllAppointmentIds, OverdueSearchEvent>? {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.computation())
+          .map { (appointmentIds) ->
+            overdueAppointmentSelector.addSelectedIds(appointmentIds)
+            return@map appointmentIds
+          }
+          .map(::SelectedOverdueAppointmentsLoaded)
+    }
   }
 
   private fun scheduleDownload(effect: ScheduleDownload) {
--- a/app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchScreen.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchScreen.kt
@@ -125,6 +125,9 @@ class OverdueSearchScreen : BaseScreen<
   private val shareButton
     get() = binding.shareOverdueListButton
 
+  private val selectAllLayout
+    get() = binding.selectAllLayout
+
   private val selectAllCheckbox
     get() = binding.selectAllCheckbox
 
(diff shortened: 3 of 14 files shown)


app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchEffectHandler.kt
```diff
@@ -51,9 +51,22 @@ class OverdueSearchEffectHandler @AssistedInject constructor(
...
 
+  private fun addAllAppointmentIds(): ObservableTransformer<AddAllAppointmentIds, OverdueSearchEvent>? {
+    return ObservableTransformer { effects ->
+      effects
+          .observeOn(schedulersProvider.computation())
+          .map { (appointmentIds) ->
+            overdueAppointmentSelector.addSelectedIds(appointmentIds)
+            return@map appointmentIds
+          }
+          .map(::SelectedOverdueAppointmentsLoaded)
```
msasikanth, line 66: We already have an observable stream from the selector to update model. So, I don't think we need to return those appointments back the `Update` function.

app/src/main/res/layout/screen_overdue_search.xml
```diff
@@ -106,17 +106,48 @@
...
+      android:id="@+id/select_all_textView"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:layout_marginStart="@dimen/spacing_4"
+      android:text="@string/overdue_search_all_appointments_selected"
+      android:textAppearance="?attr/textAppearanceBody2"
+      android:textColor="?attr/colorPrimary"
+      tools:text="SELECT ALL" />
+
+  </LinearLayout>
```
msasikanth, line 137: Have we confirmed with design whether the select all button can be scrolled with the list or just a sticky header? It seems like it's supposed to scroll based on the design. Let's confirm that once.

app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchScreen.kt
```diff
@@ -310,6 +321,14 @@ class OverdueSearchScreen : BaseScreen<
     selectedOverdueCountView.visibility = GONE
   }
 
+  override fun showSelectAllLayout() {
+    selectAllLayout.visibility = VISIBLE
+  }
```
msasikanth, line 326: We need to show the select all button only if the select & download feature is enabled and if country is India. It doesn't look like we are covering that case.

app/src/main/res/layout/screen_overdue_search.xml
```diff
@@ -106,17 +106,48 @@
...
+
+    <TextView
+      android:id="@+id/select_all_textView"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:layout_marginStart="@dimen/spacing_4"
+      android:text="@string/overdue_search_all_appointments_selected"
+      android:textAppearance="?attr/textAppearanceBody2"
+      android:textColor="?attr/colorPrimary"
+      tools:text="SELECT ALL" />
```
msasikanth, line 135: We can add text to the `CheckBox` directly. That ensures if user clicks on text, it's selected as well.

app/src/main/res/layout/screen_overdue_search.xml
```diff
@@ -106,17 +106,48 @@
...
+      android:minHeight="@dimen/spacing_24"
+      android:textColor="?attr/colorPrimary" />
+
+    <TextView
+      android:id="@+id/select_all_textView"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:layout_marginStart="@dimen/spacing_4"
+      android:text="@string/overdue_search_all_appointments_selected"
+      android:textAppearance="?attr/textAppearanceBody2"
```
msasikanth, line 133: The text appearance of text is supposed to be `Tag`, please update this.

app/src/main/res/layout/screen_overdue_search.xml
```diff
@@ -106,17 +106,48 @@
...
+    android:visibility="gone"
+    app:layout_constraintTop_toBottomOf="@+id/overdue_search_app_bar">
+
+    <com.google.android.material.checkbox.MaterialCheckBox
+      android:id="@+id/select_all_checkbox"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:minWidth="@dimen/spacing_24"
+      android:minHeight="@dimen/spacing_24"
+      android:textColor="?attr/colorPrimary" />
```
msasikanth, line 125: Irrespective of the state, the checkbox color is supposed to be primary. Let's update it.

msasikanth (changes requested):
See inline comments

app/src/main/java/org/simple/clinic/home/overdue/search/OverdueSearchEffect.kt
```diff
@@ -38,3 +38,5 @@ object OpenSelectShareFormatDialog : OverdueSearchViewEffect()
 object OpenShareInProgressDialog : OverdueSearchViewEffect()
 
 object ShowNoInternetConnectionDialog : OverdueSearchViewEffect()
+
+data class AddAllAppointmentIds(val appointmentIds: Set<UUID>): OverdueSearchEffect()
```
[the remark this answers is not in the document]
msasikanth: We can even name this as `ToggleOverdueAppointmentsSelection`, as we will be adding or removing selections.

app/src/main/res/layout/screen_overdue_search.xml
Honey14: The color updated here is for the textview, not for the checkbox, PTAL

msasikanth: I am talking about the checkbox color, not the text color. But we might end up removing the checkbox altogether. 

https://simpledotorg.slack.com/archives/CFK2PHJ1L/p1658145294726409

msasikanth: If we end up keeping checkbox. We need to update the text and check box color

PR #4105, opened 2022-07-20
Add support for selecting all overdue appointment search results

Honey14:
https://app.shortcut.com/simpledotorg/story/8677/add-support-for-selecting-all-overdue-appointment-search-results

### DIFF
--- a/app/src/main/res/layout/list_item_search_overdue_select_all_button.xml
+++ b/app/src/main/res/layout/list_item_search_overdue_select_all_button.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<com.google.android.material.button.MaterialButton xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:app="http://schemas.android.com/apk/res-auto"
+  android:id="@+id/search_overdue_select_all_button"
+  style="?attr/borderlessButtonDenseStyle"
+  android:layout_width="wrap_content"
+  android:layout_height="wrap_content"
+  android:layout_marginBottom="@dimen/spacing_8"
+  android:paddingHorizontal="@dimen/spacing_18"
+  android:paddingVertical="@dimen/spacing_18"
+  android:paddingTop="@dimen/spacing_16"
+  android:paddingBottom="@dimen/spacing_16"
+  android:text="@string/selected_overdue_select_all_appointments"
+  android:textAppearance="?attr/textAppearanceButton"
+  app:layout_constraintStart_toStartOf="parent"
+  app:layout_constraintTop_toBottomOf="@+id/overdue_search_app_bar" />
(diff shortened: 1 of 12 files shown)


app/src/test/java/org/simple/clinic/home/overdue/OverdueUpdateTest.kt
```diff
@@ -363,4 +363,18 @@ class OverdueUpdateTest {
             hasEffects(ToggleOverdueAppointmentSelection(appointmentId))
         ))
   }
+
+  @Test
+  fun `when select all button is clicked, then add all appointments Ids`() {
```
janhavisinghh, line 368: Maybe we can rename this test to - `then select all appointment ids`? As in this test name its not very clear on where are we adding these IDs to

app/src/test/java/org/simple/clinic/home/overdue/search/OverdueSearchEffectHandlerTest.kt
```diff
@@ -302,4 +302,18 @@ class OverdueSearchEffectHandlerTest {
     verifyNoMoreInteractions(uiActions)
     effectHandlerTestCase.assertNoOutgoingEvents()
   }
+
+  @Test
+  fun `when add all appointment ids effect is received, then all overdue appointments loaded`() {
```
janhavisinghh, line 307: Can we rename the test to `When select all appointment ids effect is received`?

app/src/test/java/org/simple/clinic/home/overdue/search/OverdueSearchEffectHandlerTest.kt
```diff
@@ -302,4 +302,18 @@ class OverdueSearchEffectHandlerTest {
...
+    // given
+    val allAppointmentIds = setOf(
+        UUID.fromString("9737ad9e-70f6-454e-9226-6eaa01c62645"))
+
+    // when
+    effectHandlerTestCase.dispatch(SelectAllAppointmentsIds(allAppointmentIds))
+
+    // then
+    verifyNoMoreInteractions(uiActions)
+    effectHandlerTestCase.assertNoOutgoingEvents()
```
janhavisinghh, line 317: Seems like we are not testing anything in `//then` here. Maybe we should verify if  `addSelectedIDs` method is called or not in the selector? Also, when we add the then condition we should update the test name as well for the then condition.. we are not loading overdue appointments here, right?

app/src/main/res/layout/list_item_search_overdue_select_all_button.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<com.google.android.material.button.MaterialButton xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:app="http://schemas.android.com/apk/res-auto"
+  android:id="@+id/search_overdue_select_all_button"
+  style="?attr/borderlessButtonDenseStyle"
+  android:layout_width="wrap_content"
+  android:layout_height="wrap_content"
+  android:layout_marginStart="@dimen/spacing_8"
+  android:layout_marginTop="@dimen/spacing_16"
```
janhavisinghh, line 9: The button looks like it's mis aligned. Will send screenshots in DM. The top and bottom margins are higher than in the layout and the button's body is not aligned with the card below. Can you ptal at that?

janhavisinghh (changes requested):

janhavisinghh (approved):

PR #4178, opened 2022-08-17
Migrate `BloodSugarUnitSelectionDialog` to use `BaseDialog`

janhavisinghh:
https://app.shortcut.com/simpledotorg/story/8903/migrate-bloodsugarunitselectiondialog-to-basedialog

### DIFF
--- a/app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionDialog.kt
+++ b/app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionDialog.kt
@@ -7,14 +7,16 @@ import android.content.DialogInterface
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
+import android.view.ViewGroup
 import android.widget.Button
 import androidx.appcompat.app.AlertDialog
-import androidx.appcompat.app.AppCompatDialogFragment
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/simple/clinic/bloodsugar/unitselection/BloodSugarUnitSelectionDialog.kt
```diff
@@ -183,4 +185,16 @@ class BloodSugarUnitSelectionDialog : AppCompatDialogFragment(), BloodSugarUnitS
...
+
+  @Parcelize
+  data class Key(
+      private val bloodSugarUnitPreference: BloodSugarUnitPreference,
+      override val type: ScreenType = ScreenType.Modal,
+      override val analyticsName: String = "Blood Sugar Unit Selection Dialog"
+  ) : ScreenKey() {
+
+    override fun instantiateFragment() = BloodSugarUnitSelectionDialog().apply {
+      isCancelable = false
```
msasikanth, line 150: We use this to avoid dismissing when the user clicks outside the dialog. But it doesn't seem like existing dialog has that implementation. So, we can remove this?

msasikanth (changes requested):
Looks good overall. Left one comment PTAL. Also, it looks like the PR title is different from story?

msasikanth (approved):

PR #4240, opened 2022-09-22
Verify running maestro test on staging environment locally

RishabJaiswal:
https://app.shortcut.com/simpledotorg/story/9184/add-ui-test-for-login-flow

### DIFF
--- a/.github/workflows/maestro_tests.yml
+++ b/.github/workflows/maestro_tests.yml
@@ -1,0 +1,44 @@
+name: Maestro tests
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: maestro-tests
+  cancel-in-progress: true
+
+jobs:
+  maestro-tests:
+    runs-on: [ macos-latest ]
+    steps:
+      - name: Checkout Android source
+        uses: actions/checkout@v3
+
+      - name: set up JDK
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: 11
+
+      - name: Install Maestro
+        run: |
+          brew tap mobile-dev-inc/tap
+          brew install maestro
+
+      - name: Maestro tests
+        id: run-maestro-tests
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: 31
+          arch: x86_64
+          disable-animations: true
+          disk-size: 8G
+          script: |
+            adb root
+            ./gradlew -PmaestroTests=true -PsentryDsn=${{ secrets.SENTRY_DSN }} -PsentryEnvironment=staging :app:installStagingRelease
+            maestro test -e number=1111111111 -e pin=1111 -e otp=000000 uiFlows/login_flow.yaml
+            adb uninstall org.simple.clinic.staging
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -127,7 +127,6 @@ android {
     val datadogApplicationId: String by project
     val datadogClientToken: String by project
     val datadogEnvironment: String by project
-    val maestroTest: String by project
 
     addManifestPlaceholders(mapOf(
         "sentryDsn" to sentryDsn,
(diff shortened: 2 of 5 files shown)


RishabJaiswal:
* master
  * **PR #4240** <a href="https://app.graphite.dev/github/pr/simpledotorg/simple-android/4240" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/simpledotorg/simple-android/4240?utm_source=stack-comment).

app/build.gradle.kts
```diff
@@ -155,6 +156,7 @@ android {
       isDebuggable = false
       isMinifyEnabled = runProguard.toBoolean()
       isShrinkResources = runProguard.toBoolean()
+      signingConfig = if(properties.get("maestroTest") == "true") getByName("debug").signingConfig else null
```
msasikanth, line 159: Instead of setting a `null` value for signing config, can we instead replace it with this?

```kt
if (maestroTest.toBoolean()) {
  signingConfig = getByName("debug").signingConfig
}
```

This allows us to use the proper signing key when making a release build in Bitrise.

msasikanth, line 159: Any reason we aren't using `maestroTest.toBoolean()` directly here?

gradle.properties
```diff
@@ -46,3 +46,4 @@ datadogApplicationId=do_not_change_here
 datadogClientToken=do_not_change_here
 datadogEnvironment=qa
 datadogServiceName=simple-android
+maestroTest=false
```
msasikanth, line 49: Instead of maestroTest, I am thinking we can call it `isCI` or something like that so that it's much more clearer and flexible for us to use. WDYT?

msasikanth (changes requested):
See inline comments

msasikanth (commented):
Looks good overall. Requested a few changes PTAL, Why aren't we adding the GH Actions in this PR?

gradle.properties
RishabJaiswal: Suggestion: `signReleaseWithDebugKey` sounds like a good param here

app/build.gradle.kts
RishabJaiswal: some changes got pushed accidentally, which pushing my first PR via Graphite. Will update it.

gradle.properties
msasikanth: IDK, it's very specific IMO. Since it's a build param, I feel like it would be better to identify it as CI env or not make sense.

msasikanth: Let's just go with `maestroTests` for now. We can change it in future if required.

.github/workflows/maestro_tests.yml
```diff
@@ -0,0 +1,44 @@
...
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: 31
+          arch: x86_64
+          disable-animations: true
+          disk-size: 8G
+          script: |
+            adb root
+            ./gradlew -PmaestroTests=true -PsentryDsn=${{ secrets.SENTRY_DSN }} -PsentryEnvironment=staging :app:installStagingRelease
+            maestro test -e number=1111111111 -e pin=1111 -e otp=000000 uiFlows/login_flow.yaml
```
msasikanth, line 43: Let's add these (number, pin & otp) as GitHub secrets. That way we can control them without having to make changes to the GH Actions file

uiFlows/login_flow.yaml
```diff
@@ -0,0 +1,18 @@
+appId: org.simple.clinic.staging
+---
+- clearState
+- launchApp
+- tapOn: "Next"
+- tapOn: "Get started"
+- tapOn:
+    text: "India"
+- tapOn:
+    text: "Maharashtra"
```
msasikanth, line 10: I don't see Maharashtra in my demo app selection screen. Is there a better way of handling these clicks events instead of relying on data pushed from server? 🤔

msasikanth (changes requested):
See inline comments

.github/workflows/maestro_tests.yml
msasikanth: Also, ideally, we don't want to show this information in our GH Actions logs.

PR #4314, opened 2022-11-04
Add illustrations for Tamil Nadu app

Honey14:
https://app.shortcut.com/simpledotorg/story/9356/add-tamil-nadu-app-illustrations


Honey14:
* master
  * **PR #4314** <a href="https://app.graphite.dev/github/pr/simpledotorg/simple-android/4314" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/simpledotorg/simple-android/4314?utm_source=stack-comment).

app/src/tamilNadu/res/drawable/thanks_logo.xml
```diff
@@ -0,0 +1,84 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
msasikanth, line 1: It's a large vector drawable, it would affect render performance. Is there a way we can simplify this? Can we just use a PNG or WEBP instead for this?

msasikanth (changes requested):
See inline comments

app/src/tamilNadu/res/drawable/thanks.xml
```diff
@@ -0,0 +1,24 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
msasikanth, line 1: Let's combine this and definition illustration into a single drawable. I have combine them on Figma, so you should be able to export them properly now.

We are using `ic_` for icons or `illustration_` for illustrations for all the new drawables we are adding. So, let's rename the drawable

msasikanth (changes requested):
See inline comments (for some reason Graphite decided to not add a comment when I collapsed the file. So, had to re-add it)

app/src/tamilNadu/res/drawable/thanks.xml
msasikanth: That should hopefully help with the long path datas as well for these illustrations as well a bit

app/src/tamilNadu/res/drawable/thanks_logo.xml
msasikanth: https://upday.github.io/blog/vector_drawables_optimisation/

msasikanth (changes requested):
There are couple of issues and changes we need to make

- The illustrations are pixelated. Especially the Tamil Nadu logo. We might want to use a higher res image in `drawable-nodpi` or go with having multiple density images in each drawable folder. There is a Figma plugin to export to multiple densities you can use.
https://www.figma.com/community/plugin/735452896889481850/Android-Resources-Export

- Let's rename the `icon_tamilnadu` to `ic_tamilnadu`, we are not using `icon_` prefix anywhere

- The icon in registration screen, etc., seems small compared to the designs. Can we take a look at what dimension we are supposed to be using there

- The spacing in splash screen seems a bit off, especially how the content is aligned. I think the Tamil Nadu logo is supposed to be centered aligned from bottom of THANKS illustration and top of next button frame.  PTAL and confirm with Tony how it's supposed to be aligned. 

- I don't see the home screen illustration in here. Is it in a different PR?

msasikanth (approved):

PR #4529, opened 2022-12-20
Questionnarie table added to the database

siddh1004:
https://app.shortcut.com/simpledotorg/story/9557/add-sync-resource-for-questionnaire-layout-api


app/src/main/java/org/simple/clinic/monthlyReports/Questionnaire.kt
```diff
@@ -0,0 +1,18 @@
...
+import kotlinx.parcelize.Parcelize
+import java.util.UUID
+
+@Entity(tableName = "Questionnaire")
+@Parcelize
+data class Questionnaire(
+    @PrimaryKey
+    val uuid: UUID,
+
+    val questionnaire_type: QuestionnaireType,
```
msasikanth, line 15: I think we discussed of making `QuestionnaireType` as primary key in PRD. Please check

app/src/main/java/org/simple/clinic/monthlyReports/QuestionnaireType.kt
```diff
@@ -0,0 +1,47 @@
...
+import com.squareup.moshi.ToJson
+import kotlinx.parcelize.Parcelize
+import org.simple.clinic.util.room.SafeEnumTypeAdapter
+
+sealed class QuestionnaireType : Parcelable {
+
+  object TypeAdapter : SafeEnumTypeAdapter<QuestionnaireType>(
+      knownMappings = mapOf(
+          MonthlyScreeningReports to "monthly_screening_reports",
+          MonthlyDrugStock to "monthly_drug_stock"
```
msasikanth, line 15: Is this type already being provided by the server? If not please remove this. We can add it once server has actual implementation of it.

msasikanth (changes requested):
Looks good overall. Left couple of comments, PTAL

app/src/main/java/org/simple/clinic/monthlyReports/Questionnaire.kt
siddh1004: Primary key update to `questionnaire_type`

app/src/main/java/org/simple/clinic/monthlyReports/QuestionnaireType.kt
siddh1004: Data type of `questionnaire_type` changed to string and `QuestionnaireType.kt` deleted

msasikanth: We would still need the questionnaire type class, I am just saying we don't need the drug stock type. Right now it's just screening reports.

siddh1004: I have again added the questionnaire type and removed the monthly drug stock type.

PR #4543, opened 2023-01-06
Questionnaire sync resource removed from sync list

siddh1004:
Questionnaire sync resource removed from the sync list as the feature is still in development mode.


harimohanraj89 (approved):
Looks great.

PR #4548, opened 2023-02-08
Mobius arch setup for questionnaire form

siddh1004:
https://app.shortcut.com/simpledotorg/story/9559/create-epoxy-models-for-screening-view-types


app/src/main/java/org/simple/clinic/monthlyReports/questionnaire/BaseComponentDataPolymorphicJsonAdapter.kt
```diff
@@ -0,0 +1,25 @@
+package org.simple.clinic.monthlyReports.questionnaire
+
+import com.squareup.moshi.adapters.PolymorphicJsonAdapterFactory
+import org.simple.clinic.monthlyReports.questionnaire.component.BaseComponentData
```
olttwa, line 4: The `monthlyReports` namespace will be deleted in a future namespace, right?

app/src/main/java/org/simple/clinic/monthlyReports/questionnaire/entry/QuestionnaireEntryModel.kt
```diff
@@ -0,0 +1,30 @@
+package org.simple.clinic.monthlyReports.questionnaire.entry
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+import org.simple.clinic.facility.Facility
+import org.simple.clinic.monthlyReports.questionnaire.Questionnaire
+
+@Parcelize
+data class QuestionnaireEntryModel(
+    val facility: Facility?,
```
olttwa, line 10: Why does a QuestionnaireEntry have a facility? Facility field will be possessed only by QuestionnaireResponse

Is QuestionnaireEntry synonymous to QuestionnaireResponse? If so, shall we name it as response?

olttwa: Also, facility isn't optional. It is a required field. Thought I'd add a note since I saw a question at the end `Facility?`.

app/src/main/java/org/simple/clinic/monthlyReports/questionnaire/entry/QuestionnaireEntryScreen.kt
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  override fun setFacility(facilityName: String) {
+    questionnaireFormToolbar.title = facilityName
+  }
+
+  override fun displayQuestionnaireFormLayout(layout: BaseComponentData) {
+  }
+
+  override fun goBack() {
```
olttwa, line 84: If user clicks on goBack before Submit(), we want to display a warning saying `Are you sure you want to quit? You'll lose non-submitted data.`

olttwa (changes requested):

olttwa (approved):

PR #4550, opened 2023-02-10
Questionnaire response table added to the room database

siddh1004:

### DIFF
--- a/app/src/main/java/org/simple/clinic/questionnaireResponse/QuestionnaireResponse.kt
+++ b/app/src/main/java/org/simple/clinic/questionnaireResponse/QuestionnaireResponse.kt
@@ -1,10 +1,11 @@
-package org.simple.clinic.monthlyReports.questionnaire
+package org.simple.clinic.questionnaireResponse
 
 import android.os.Parcelable
 import androidx.room.Embedded
 import androidx.room.Entity
 import androidx.room.PrimaryKey
 import kotlinx.parcelize.Parcelize
+import org.simple.clinic.monthlyReports.questionnaire.QuestionnaireType
 import org.simple.clinic.patient.SyncStatus
 import org.simple.clinic.storage.Timestamps
 import java.util.UUID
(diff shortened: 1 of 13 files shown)


app/src/main/java/org/simple/clinic/questionnaireResponse/QuestionnaireResponse.kt
```diff
@@ -0,0 +1,33 @@
...
+    val facilityId: UUID,
+
+    val lastUpdatedByUserId: UUID,
+
+    val content: Map<String, String>,
+
+    @Embedded
+    val timestamps: Timestamps,
+
+    val syncStatus: SyncStatus
```
olttwa, line 32: should there be a deletedAt Timestamp?

olttwa: Questionnaire.kt has one & response also needs one

app/src/main/java/org/simple/clinic/util/room/MapRoomTypeConverter.kt
```diff
@@ -0,0 +1,28 @@
...
+import androidx.room.TypeConverter
+import java.util.TreeMap
+import javax.inject.Inject
+
+class MapRoomTypeConverter @Inject constructor() {
+  @TypeConverter
+  fun fromMap(value: Map<String, String>): String {
+    val sortedMap = TreeMap(value)
+    return sortedMap.keys.joinToString(separator = ",").plus("<divider>")
+        .plus(sortedMap.values.joinToString(separator = ","))
```
olttwa, line 12: just using `,` for a separator is a bad idea. Value might contain a comma, making conversion of string to map impossible.
Can you check Java/Kotlin methods on converting Map to String?

olttwa: GSON also has map->string and string->map conversion logic. Check that out as well.
image

olttwa (changes requested):

olttwa (approved):

PR #4552, opened 2023-02-13
Questionnaire response repository added to the code

siddh1004:


app/src/main/java/org/simple/clinic/questionnaireResponse/QuestionnaireResponse.kt
```diff
@@ -30,4 +36,61 @@ data class QuestionnaireResponse(
...
+      WHERE syncStatus = :syncStatus
+      LIMIT :limit OFFSET :offset
+    """)
+    fun recordsWithSyncStatusBatched(
+        syncStatus: SyncStatus,
+        limit: Int,
+        offset: Int
+    ): List<QuestionnaireResponse>
+
+    @Query("UPDATE QuestionnaireResponse SET syncStatus = :newStatus WHERE syncStatus = :oldStatus")
```
olttwa, line 69: I don't fully understand use-case for all the SQL queries here, like this one. But assuming this is boilerplate, we can keep it

app/src/androidTest/java/org/simple/clinic/di/TestAppComponent.kt
```diff
@@ -26,6 +26,7 @@ import org.simple.clinic.facility.FacilityRepositoryAndroidTest
 import org.simple.clinic.login.LoginUserWithOtpServerIntegrationTest
 import org.simple.clinic.medicalhistory.MedicalHistoryRepositoryAndroidTest
 import org.simple.clinic.monthlyReports.questionnaire.QuestionnaireRepositoryAndroidTest
+import org.simple.clinic.monthlyReports.questionnaire.QuestionnaireResponseRepositoryAndroidTest
```
olttwa, line 29: I know we discussed fixing it towards the end, but this is driving me crazy a bit.
@siddh1004 let's create a separate story for fixing namespaces? I have 3 issues with current heriarchy:
1. `monthlyReports` is a parent for generic things
2. There's redundancy of `questionnaire`. For ex: `questionnaire.QuestionnaireRepositoryAndroidTest`
3. `Questionnaire` & `QuestionnaireResponses` are 2 separate things nested inside 1 same parent `questionnaire`

app/src/main/java/org/simple/clinic/questionnaireResponse/QuestionnaireResponse.kt
siddh1004: It's a boilerplate code. Even i don't understand this fully atm.

app/src/androidTest/java/org/simple/clinic/di/TestAppComponent.kt
siddh1004: @olttwa I will create user stories for this.

PR #4556, opened 2023-02-20
questionnaire content value type changed to Any from String

siddh1004:
https://app.shortcut.com/simpledotorg/story/10087/change-implementation-of-content-map-in-questionnaireresponse


siddh1004:
1. `OrientationType` removed from the display properties of questionnaire form layout response
2. Type of content changed to Map<String, Any> from Map<String, String>
3. QuestionnaireResponse property `last_updated_by_userid` changed to nullable.

app/src/main/java/org/simple/clinic/storage/migrations/RoomMigrationsModule.kt
```diff
@@ -113,7 +113,8 @@ class RoomMigrationsModule {
       migration105: Migration_105,
       migration106: Migration_106,
       migration107: Migration_107,
-      migration108: Migration_108
+      migration108: Migration_108,
+      migration109: Migration_109
```
olttwa, line 117:
```suggestion
      migration109: Migration_109,
```

olttwa: Kotlin accepts comma in the last element of a list. I suggest adding a comma always so it isn't included in a diff next time an entry is added.

app/src/main/java/org/simple/clinic/questionnaireresponse/QuestionnaireResponse.kt
```diff
@@ -29,9 +30,9 @@ data class QuestionnaireResponse(
 
     val facilityId: UUID,
 
-    val lastUpdatedByUserId: UUID,
+    val lastUpdatedByUserId: UUID?,
 
-    val content: Map<String, String>,
+    val content: Map<String, @RawValue Any>,
```
olttwa, line 35: This was a good find. Congrats 🎉

app/src/androidTest/java/org/simple/clinic/sync/QuestionnaireResponseSyncIntegrationTest.kt
```diff
@@ -91,32 +75,18 @@ class QuestionnaireResponseSyncIntegrationTest {
...
   }
 
   private fun resetLocalData() {
-    clearQuestionnaireData()
     clearQuestionnaireResponseData()
     lastPullToken.delete()
-    questionnaireLastPullToken.delete()
-  }
-
-  private fun clearQuestionnaireData() {
```
olttwa, line 118: I see you've cleaned up functions which weren't of any use as of this moment. That's good. We'll add them later if needed.

olttwa (dismissed):

app/src/main/java/org/simple/clinic/storage/migrations/RoomMigrationsModule.kt
olttwa: For example, if `Migration_108,` had a trailing comma, it wouldn't show up in this diff

app/schemas/org.simple.clinic.AppDatabase/109.json
```diff
@@ -0,0 +1,2143 @@
+{
```
prabhanshuguptagit, line 1: Does this file need to be committed?

prabhanshuguptagit (changes requested):
Looks good, one question about committing the AppDatabase json file.

app/schemas/org.simple.clinic.AppDatabase/109.json
siddh1004: yes

siddh1004: @prabhanshuguptagit yes

prabhanshuguptagit (approved):

PR #4560, opened 2023-02-21
Monthly screening reports and Download IHCI patient register strings added

siddh1004:


app/src/main/res/layout/screen_monthly_screening_report_list.xml
```diff
@@ -81,6 +80,16 @@
...
+    <TextView
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:text="@string/monthly_screening_reports_submitted_with_date_and_time" />
+
+    <TextView
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:text="@string/monthly_screening_reports_submitted_with_date" />
+
```
prabhanshuguptagit, line 92: [non-blocking] Which UI elements do these correspond to?

siddh1004: @prabhanshuguptagit these are dummy text, so as the lint error is succeeded and we can then send these strings to transifex before we do the actual implementation.

prabhanshuguptagit: 👍🏼

prabhanshuguptagit (approved):

PR #4700, opened 2023-05-22
Enable clinical decision support for all facilities in Ethiopia

siddh1004:
https://app.shortcut.com/simpledotorg/story/10567/enable-clinical-decision-support-for-all-facilities-in-ethiopia

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
@@ -96,7 +96,9 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
           .observeOn(schedulersProvider.io())
           .map {
             val currentFacilityId = currentFacility.get().uuid
-            CDSSPilotStatusChecked(isPilotEnabledForFacility = cdssPilotFacilities.get().contains(currentFacilityId))
+            CDSSPilotStatusChecked(isPilotEnabledForFacility =
+            country.isoCountryCode == Country.ETHIOPIA || cdssPilotFacilities.get().contains(currentFacilityId)
+            )
           }
     }
   }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
```diff
@@ -96,7 +96,9 @@ class PatientSummaryEffectHandler @AssistedInject constructor(
           .observeOn(schedulersProvider.io())
           .map {
             val currentFacilityId = currentFacility.get().uuid
-            CDSSPilotStatusChecked(isPilotEnabledForFacility = cdssPilotFacilities.get().contains(currentFacilityId))
+            CDSSPilotStatusChecked(isPilotEnabledForFacility =
```
qptr, line 99: shouldn't this parameter just be named `isCDSEnabledForFacility` now? If CDS is enabled for all facilities, there shouldn't be any need for pilots anymore.

qptr: Also, indentation seems off.

qptr (approved):

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffectHandler.kt
siddh1004: @qptr This is not enabled for all facilities yet. Only enabled in ethiopia and selected facilities in other countries. Also, the indention is default for kotlin language and get set automatically.

qptr: Oh, does pilot mean CDS feature here?

siddh1004: yes

PR #4717, opened 2023-06-09
Enabled configuration cache

siddh1004:
https://app.shortcut.com/simpledotorg/story/10662/add-support-for-configuration-cache-in-gradle

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -50,3 +50,5 @@ datadogClientToken=do_not_change_here
 datadogEnvironment=qa
 datadogServiceName=simple-android
 maestroTests=false
+org.gradle.unsafe.configuration-cache=true
+org.gradle.unsafe.configuration-cache-problems=warn
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Add Simple Compose theme in `common-ui` module
 - Bump sentry to v6.22.0
 - Bump Sentry Gradle Plugin to v3.10.0
+- Enable Configuration cache
 
 ### Fixes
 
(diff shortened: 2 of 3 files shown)


gradle.properties
```diff
@@ -50,3 +50,5 @@ datadogClientToken=do_not_change_here
 datadogEnvironment=qa
 datadogServiceName=simple-android
 maestroTests=false
+org.gradle.unsafe.configuration-cache=true
+org.gradle.unsafe.configuration-cache-problems=warn
```
msasikanth, line 54: Let's fallback to default severity which is error. If there is any task that is throwing errors, let's resolve them if we can.

msasikanth (changes requested):
Left one comment PTAL

gradle.properties
siddh1004: Fixed

app/build.gradle.kts
```diff
@@ -504,6 +505,9 @@ apply(plugin = "com.google.gms.google-services")
 
 abstract class TransformGeneratedRoomDaoTask : DefaultTask() {
 
+  @get:Input
+  abstract val projectDirAbsolutePath: Property<String>
```
msasikanth, line 509: Since this property doesn't change, instead of passing it as parameter. We can just have the variable inside this task. WDYT?

msasikanth:
```kt
private val projectDir = project.projectDir.absolutePath
```

siddh1004: Fixed

msasikanth (approved):

PR #4728, opened 2023-06-20
Remove pinned versions from CI workflows

msasikanth:


.github/workflows/maestro_tests.yml
```diff
@@ -74,8 +71,8 @@ jobs:
             
             # When installing Maestro manually, it doesn't get discovered in the path.
             # Hence, providing absolute path to binary.
-            $HOME/.maestro/bin/maestro test --format junit --output maestro-test-report.xml maestroUiFlows
-            
+            maestro test --format junit --output maestro-test-report.xml maestroUiFlows
+
```
siddh1004, line 75: Can we also remove the comment about providing absolute path to binary, as now we have removed the absolute path?

siddh1004 (changes requested):

.github/workflows/maestro_tests.yml
msasikanth: Good catch, fixing it.

siddh1004 (approved):

siddh1004 (approved):

PR #4729, opened 2023-06-26
Add support for database encryption

msasikanth:
https://app.shortcut.com/simpledotorg/story/10755/encrypt-app-database

### DIFF
--- a/app/src/androidTest/java/org/simple/clinic/storage/DatabaseEncryptorTest.kt
+++ b/app/src/androidTest/java/org/simple/clinic/storage/DatabaseEncryptorTest.kt
@@ -5,15 +5,14 @@ import androidx.room.Room
 import com.google.common.truth.Truth.assertThat
 import com.squareup.moshi.Moshi
 import net.sqlcipher.database.SupportFactory
-import org.junit.After
 import org.junit.Before
 import org.junit.Test
 import org.simple.clinic.AppDatabase
 import org.simple.clinic.TestClinicApp
 import org.simple.clinic.questionnaire.component.BaseComponentData
-import org.simple.clinic.storage.DatabaseEncryption.State.DOES_NOT_EXIST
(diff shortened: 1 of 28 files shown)


app/src/androidTest/java/org/simple/clinic/storage/DatabaseEncryptorTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    // when
+    databaseEncryptor.execute(databaseName = DB_NAME)
+
+    // then
+    assertThat(databaseEncryptor.databaseState(DB_NAME)).isEqualTo(ENCRYPTED)
+
+    val encryptedDatabase = Room.databaseBuilder(appContext, AppDatabase::class.java, DB_NAME)
+        .allowMainThreadQueries()
+        .addTypeConverter(BaseComponentData.RoomTypeConverter(moshi))
+        .openHelperFactory(SupportFactory(passphrase, null, false))
```
siddh1004, line 85: Can we remove the default argument values being called to SupportFactory class, as this will make it consistent to what we call in `StorageModule.kt`
```suggestion
        .openHelperFactory(SupportFactory(passphrase))
```

siddh1004 (changes requested):

app/src/androidTest/java/org/simple/clinic/storage/DatabaseEncryptorTest.kt
msasikanth: It's intentional, we don't want to clear the passphrase during test runs.

msasikanth: One more thing is, we are not encrypting by default in tests. In `TestStorageModule` we are still continuing using the requery sqlite for now. So, this support factory is only for the `DatabaseEncryptorTest`

siddh1004 (approved):

PR #4735, opened 2023-06-28
Hide overdue download/share actions when there are no overdue patients

msasikanth:
https://app.shortcut.com/simpledotorg/story/10839/hide-overdue-download-share-actions-when-there-are-no-overdue-patients

### DIFF
--- a/app/src/main/java/org/simple/clinic/home/overdue/OverdueScreen.kt
+++ b/app/src/main/java/org/simple/clinic/home/overdue/OverdueScreen.kt
@@ -214,8 +214,6 @@ class OverdueScreen : BaseScreen<
     super.onViewCreated(view, savedInstanceState)
     overdueRecyclerView.adapter = overdueListAdapter
     overdueRecyclerView.layoutManager = LinearLayoutManager(context)
-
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/simple/clinic/home/overdue/OverdueScreen.kt
```diff
@@ -293,6 +296,7 @@ class OverdueScreen : BaseScreen<
 
   override fun showNoOverduePatientsView() {
     viewForEmptyList.visibility = View.VISIBLE
+    if (isOverdueListDownloadAndShareEnabled) { buttonsFrame.visibility = View.GONE }
```
siddh1004, line 299: Is this condition needed? 
We can just call this -
```suggestion
    buttonsFrame.visibility = View.GONE
```

siddh1004 (changes requested):

app/src/main/java/org/simple/clinic/home/overdue/OverdueScreen.kt
msasikanth: This feature is behind a feature flag, so keeping that for now. We would have to remove it feature flag, which can be done in a different PR.

siddh1004 (approved):

PR #4772, opened 2023-07-20
Add month and year picker dialog for Compose forms

msasikanth:
https://app.shortcut.com/simpledotorg/story/10889/add-date-picker-component-to-questionnaire-forms

### DIFF
--- a/app/src/main/java/org/simple/clinic/widgets/montyeardatepicker/MonthYearPickerDialog.kt
+++ b/app/src/main/java/org/simple/clinic/widgets/montyeardatepicker/MonthYearPickerDialog.kt
@@ -1,0 +1,298 @@
+package org.simple.clinic.widgets.montyeardatepicker
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.requiredHeight
+import androidx.compose.foundation.shape.RoundedCornerShape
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/simple/clinic/widgets/montyeardatepicker/MonthYearPickerDialog.kt
```diff
@@ -0,0 +1,298 @@
...
+@Composable
+fun MonthYearPickerDialog(
+    onDismissRequest: () -> Unit,
+    modifier: Modifier = Modifier,
+    selectedDate: LocalDate? = null,
+    daysRange: IntRange = IntRange(0, 3650), // now to 10 years in future
+    onDateSelected: (LocalDate) -> Unit
+) {
+  val dateFormatter = remember {
+    DateTimeFormatter.ofPattern("MM/yyyy")
```
siddh1004, line 62: Can we pass dateFormatter instance through constructor argument, this would help to initialise the instance through dagger (DateFormatterModule.kt)and make it consistent/similar to other parts of the app.

siddh1004 (changes requested):

app/src/main/java/org/simple/clinic/widgets/montyeardatepicker/MonthYearPickerDialog.kt
msasikanth: That was intentional in this scenario. We don't want to use the Chronology in the date formatters. That's why went with a hardcoded one here. In future if we have a use case that needs support for both, we can revisit this.

msasikanth: Think of it as a separate component similar to material date picker dialog. We just want to render it in this format and don't have anything else.

siddh1004 (approved):

PR #4780, opened 2023-07-31
Replace Mixpanel with Firebase Analytics

msasikanth:


gradle/libs.versions.toml
```diff
@@ -89,6 +89,7 @@ edittext-pinentry = "com.alimuzaffar.lib:pinentryedittext:2.0.6"
 faker = "com.github.blocoio:faker:1.2.9"
 
 firebase-config = "com.google.firebase:firebase-config:21.2.0"
+firebase-analytics = "com.google.firebase:firebase-analytics-ktx:21.3.0"
```
siddh1004, line 93: Shouldn't we also remove the mixpanel packages from the code?

app/src/main/java/org/simple/clinic/plumbing/infrastructure/MixpanelInfrastructure.kt
```diff
@@ -1,31 +0,0 @@
...
-import org.simple.clinic.appconfig.Deployment
-import org.simple.clinic.user.User
-import javax.inject.Inject
-
-class MixpanelInfrastructure @Inject constructor(
-    private val application: Application
-) : Infrastructure {
-
-  private val mixpanel = MixpanelAPI.getInstance(application, BuildConfig.MIXPANEL_TOKEN, true)
-
```
siddh1004, line 16: Can we remove the `BuildConfig.MIXPANEL_TOKEN` from gradle.properties file too?  Also we would need to remove it from CI pipelines as well.

app/src/main/java/org/simple/clinic/analytics/FirebaseAnalyticsReporter.kt
```diff
@@ -0,0 +1,40 @@
...
+    FirebaseAnalytics.getInstance(application)
+  }
+
+  override fun setLoggedInUser(user: AnalyticsUser, isANewRegistration: Boolean) {
+    synchronized(firebaseAnalytics) {
+      val userId = user.id.toString()
+
+      // The current facility, deployment, etc are being tracked in `FirebaseAnalyticsInfrastructure`
+      firebaseAnalytics.setUserId(userId)
+      firebaseAnalytics.setUserProperty("is_new_registration", isANewRegistration.toString())
```
siddh1004, line 24: Why are we setting additional user property `is_new_registration`? It was not present in the mixpanel reporter.

siddh1004 (changes requested):

gradle/libs.versions.toml
msasikanth: Oh, good catch. missed it

app/src/main/java/org/simple/clinic/analytics/FirebaseAnalyticsReporter.kt
msasikanth: Mixpanel had a way to identify new registrations. Since Firebase doesn't have it natively, we are just using a property to identify new registrations.

siddh1004 (approved):

PR #4783, opened 2023-08-02
update PROGUARD file to keep material icon class

siddh1004:
Material Icons were not shown in the dynamic questionnaire form.
**Reason**:

1. We use reflection to generate image vector by name.
2. Because of proguard rules, classes were getting obfuscated.

**Solution**

1. Add proguard rule to not obfuscate class names that extend `androidx.compose.material.icons.filled`

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -118,3 +118,6 @@
   native <methods>;
   public <init>(...);
 }
+
+## Compose icon with reflection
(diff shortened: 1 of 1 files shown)


olttwa (approved):

msasikanth:
This would keep all the classes in that package, which would increase the APK/AAB size as well. Do we have any comparison on how much that would be?

It's not an immediate problem, just something we can visit later.

PR #4825, opened 2023-11-30
Show data protection consent screen during onboarding

msasikanth:
https://app.shortcut.com/simpledotorg/story/11557/show-data-protection-consent-screen-during-onboarding

### DIFF
--- a/app/src/main/java/org/simple/clinic/consent/onboarding/OnboardingConsentScreenFragment.kt
+++ b/app/src/main/java/org/simple/clinic/consent/onboarding/OnboardingConsentScreenFragment.kt
@@ -121,7 +121,7 @@ private fun OnboardingConsentScreen(
             modifier = Modifier
                 .fillMaxWidth()
                 .background(SimpleTheme.colors.material.primaryVariant)
-                .padding(dimensionResource(id = R.dimen.spacing_8))
+                .padding(dimensionResource(id = R.dimen.spacing_12))
         ) {
           FilledButton(
               onClick = onAccept,
--- a/app/src/main/java/org/simple/clinic/consent/onboarding/OnboardingConsentScreenFragment.kt
+++ b/app/src/main/java/org/simple/clinic/consent/onboarding/OnboardingConsentScreenFragment.kt
@@ -1,0 +1,208 @@
+package org.simple.clinic.consent.onboarding
+
+import android.content.Context
+import android.os.Bundle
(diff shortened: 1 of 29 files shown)


app/src/main/java/org/simple/clinic/consent/onboarding/OnboardingConsentScreenFragment.kt
```diff
@@ -0,0 +1,208 @@
...
+        { viewEffect -> viewEffectHandler.handle(viewEffect) },
+        { pausedViewEffects -> pausedViewEffects.forEach { viewEffectHandler.handle(it) } }
+    )
+  }
+
+  override fun moveToRegistrationActivity() {
+    // This navigation should not be done here, we need a way to publish
+    // an event to the parent activity (maybe via the screen router's
+    // event bus?) and handle the navigation there.
+    // TODO(vs): 2019-11-07 Move this to an event that is subscribed in the parent activity
```
siddh1004, line 90: Are we still going to improve `moveToRegistrationActivity`? Can we remove the comment from here and convert it into shortcut story?

app/src/main/java/org/simple/clinic/di/AppModule.kt
```diff
@@ -160,4 +163,9 @@ class AppModule(private val appContext: Application) {
       appContext: Application,
       configReader: ConfigReader
   ): MinimumMemoryChecker = RealMinimumMemoryChecker(appContext)
+
+  @Provides
+  @TypedPreference(DataProtectionConsent)
+  fun hasUserConsentedToDataProtection(rxSharedPreferences: RxSharedPreferences): Preference<Boolean> =
+      rxSharedPreferences.getBoolean("data_protection_consent")
```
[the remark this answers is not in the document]
msasikanth: This needs to be accessed from the home screen as well, for displaying dialog for existing users. That's why I placed it in the app module rather than creating a new module or scoping to a specific feature.

app/src/main/java/org/simple/clinic/consent/onboarding/OnboardingConsentScreenFragment.kt
msasikanth: Yes, eventually when we get back to working on Simple. We can create a new story, but let's leave the TODO in the project. IDE provides a good way to track all the TODOs.

PR #4842, opened 2024-03-07
Fix crash for search patient using special character 

siddh1004:


msasikanth:
@siddh1004 CI checks failed in this PR, PTAL

app/src/main/java/org/simple/clinic/overdue/AppointmentRepository.kt
```diff
@@ -152,8 +152,12 @@ class AppointmentRepository @Inject constructor(
     )
   }
 
-  private fun transformSearchInputsIntoQuery(searchInputs: List<String>) = searchInputs
-      .joinToString(separator = " OR ") { "*$it*" }
+  private fun transformSearchInputsIntoQuery(searchInputs: List<String>): String {
```
msasikanth, line 155: This should handle the overdue search, can we also check if this issue is present in the patient search as well?

siddh1004: @msasikanth For the patient search, only alphaumeric values are allowed. So the bug won't arise over there. Either we remove the alphanumeric restriction and then transform the search input or we leave it as it is. Waiting for your feedback.

msasikanth (approved):

PR #4920, opened 2024-04-23
Update `Patient#isEligibleForReassignment` on BP creation

siddh1004:
https://app.shortcut.com/simpledotorg/story/12440/check-and-update-patient-eligibility-status-when-new-blood-pressure-is-created

### DIFF
--- a/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
@@ -41,6 +41,7 @@ class BloodPressureEntryUpdate(
       is ShowBpClicked -> showBpClicked(model)
       is BloodPressureSaved -> next(model.bloodPressureStateChanged(NOT_SAVING_BLOOD_PRESSURE), SetBpSavedResultAndFinish)
       is DatePrefilled -> next(model.datePrefilled(event.prefilledDate))
+      is PatientReassignmentEligibilityStatusUpdated -> noChange()
     }
   }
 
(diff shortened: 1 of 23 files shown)


app/src/main/java/org/simple/clinic/bp/entry/BloodPressureEntryUpdate.kt
```diff
@@ -39,8 +40,14 @@ class BloodPressureEntryUpdate(
       is ChangeDateClicked -> changeDateClicked(model)
       is SaveClicked -> onSaveClicked(model)
       is ShowBpClicked -> showBpClicked(model)
-      is BloodPressureSaved -> next(model.bloodPressureStateChanged(NOT_SAVING_BLOOD_PRESSURE), SetBpSavedResultAndFinish)
+      is BloodPressureSaved -> next(
+          model.bloodPressureStateChanged(NOT_SAVING_BLOOD_PRESSURE),
+          SetBpSavedResultAndFinish,
+          UpdatePatientReassignmentEligibilityStatus((model.openAs as OpenAs.New).patientUuid)
```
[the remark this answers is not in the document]
siddh1004: Also, We are updating the boolean flag when the BP is created. Should we do the same when the patients status, or any value on which the flag is dependent is updated.

app/src/main/java/org/simple/clinic/patient/PatientRepository.kt
```diff
@@ -783,6 +783,13 @@ class PatientRepository @Inject constructor(
     )
   }
 
+  fun updatePatientReassignmentEligibilityStatus(patientUuid: UUID) {
+    database.patientDao().updatePatientReassignmentEligibilityStatus(
+        patientUuid = patientUuid,
+        isEligibleForReassignment = isPatientEligibleForReassignment(patientUuid)
+    )
+  }
```
siddh1004, line 791: What happens when a sync is performed, does the flag gets reset to false as we are not pushing it to server.

msasikanth: It won't reset, since it's a local field and we use `upsert` function for updating existing patients during sync.

https://github.com/simpledotorg/simple-android/blob/49543a6e8c427ba6b6060ce176ff8ad18d586e4d/app/src/main/java/org/simple/clinic/patient/Patient.kt#L175

msasikanth (changes requested):

msasikanth (dismissed):

msasikanth (approved):

PR #4922, opened 2024-04-25
Add facility reassignment view to list item

siddh1004:
https://app.shortcut.com/simpledotorg/story/12443/show-patient-eligibility-status-in-the-ui

### DIFF
--- a/app/src/main/java/org/simple/clinic/patient/RecentPatient.kt
+++ b/app/src/main/java/org/simple/clinic/patient/RecentPatient.kt
@@ -27,7 +27,9 @@ data class RecentPatient(
 
     val patientRecordedAt: Instant,
 
-    val updatedAt: Instant
+    val updatedAt: Instant,
+
+    val isEligibleForReassignment: Boolean,
 ) : Parcelable {
 
   override fun toString(): String {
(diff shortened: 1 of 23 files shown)


siddh1004:
https://app.shortcut.com/simpledotorg/story/12443/show-patient-eligibility-status-in-the-ui

app/src/main/java/org/simple/clinic/patient/RecentPatient.kt
```diff
@@ -27,7 +27,9 @@ data class RecentPatient(
```
msasikanth, line 27: Looks like the commit message is moved to commit description. PTAL

app/src/main/res/drawable/ic_facility_reassignment.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
msasikanth, line 1: This icon frame doesn't match the one in the design. We need to download the icon and frame, instead of just the icon glyph.

msasikanth (changes requested):

msasikanth (approved):

PR #4942, opened 2024-05-09
Fix incorrect checks in patient reassignment eligibility check

msasikanth:


msasikanth:
Waiting until we receive confirmation of these checks

siddh1004 (approved):

PR #5080, opened 2024-09-13
Fix Statin nudge alert stroke and animation

siddh1004:


app/src/main/java/org/simple/clinic/summary/PatientSummaryScreen.kt
```diff
@@ -831,6 +797,43 @@ class PatientSummaryScreen :
...
+      }
+    }
+    transition.addListener(transitionListener)
+    TransitionManager.beginDelayedTransition(summaryViewsContainer, transition)
+
+    view.visibility = VISIBLE
+  }
+
   override fun hideStatinAlert() {
     statinAlertView.visibility = GONE
```
msasikanth, line 838: We also need to animate hiding as well

msasikanth (changes requested):

msasikanth (approved):

PR #5086, opened 2024-10-21
Migrate `SplashScreen` to Compose

msasikanth:
https://app.shortcut.com/simpledotorg/story/13784/migrate-onboarding-flow-to-compose-ui


siddh1004 (approved):

app/src/main/java/org/simple/clinic/splash/SplashScreen.kt
```diff
@@ -21,33 +44,27 @@ class SplashScreen : Fragment() {
...
     super.onAttach(context)
     context.injector<Injector>().inject(this)
   }
 
-  override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
-    binding = ScreenSplashBinding.inflate(inflater, container, false)
-    return binding?.root
-  }
+  override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
+    return ComposeView(requireContext()).apply {
```
SubhrajyotiSen, line 53: Have you considered using https://developer.android.com/reference/kotlin/androidx/fragment/compose/package-summary#extension-functions ?

msasikanth: We don't have access to that extension since we don't pull the `fragment-compose` dependency. FWIW, it's an easy enough extension to make internally as well. Will consider it for future PRs.

PR #5108, opened 2024-11-06
Update Statin nudge animation and translation

siddh1004:


app/src/main/java/org/simple/clinic/summary/compose/StatinNudgeView.kt
```diff
@@ -35,6 +38,11 @@ fun StatinNudge(
 ) {
   AnimatedVisibility(
       visible = isVisible,
+      enter = expandVertically(
```
msasikanth, line 41: Rather than expand, shouldn't it be sliding in and out? Atleast that's what I remember from the CDSS nudge animation 🤔

common-ui/src/main/java/org/simple/clinic/common/ui/util/TextUtil.kt
```diff
@@ -0,0 +1,36 @@
...
+import android.text.Spanned
+import android.text.style.StyleSpan
+import androidx.compose.ui.text.AnnotatedString
+import androidx.compose.ui.text.SpanStyle
+import androidx.compose.ui.text.buildAnnotatedString
+import androidx.compose.ui.text.font.FontWeight
+
+fun htmlToAnnotatedString(htmlText: String) = getSpannedText(htmlText).toAnnotatedString()
+
+private fun Spanned.toAnnotatedString(): AnnotatedString = buildAnnotatedString {
```
msasikanth, line 15: We can just have a single extension function, no? Something like this

```kt
fun String.toAnnotatedString(): AnnotatedString {
  val spanned = getSpannedText(this)
  ...
}
```

That way we can just call `stringResources(...).toAnnotatedString()`. WDYT?

msasikanth (commented):
LGTM. Left couple of comments, PTAL

msasikanth (approved):

msasikanth (approved):

PR #5110, opened 2024-11-07
Rename `TextUtils` to `Strings`

siddh1004:


common-ui/src/main/java/org/simple/clinic/common/ui/util/Strings.kt
```diff
@@ -13,22 +13,22 @@ import androidx.compose.ui.text.font.FontWeight
 fun String.toAnnotatedString() = getSpannedText(this).toAnnotatedString()
```
msasikanth, line 13: There is already a `Strings.kt` file the app module, since this extension is not needed in the common module or any other modules that depend on the common module. Let's move this extension to app module `Strings.kt` file?

msasikanth (changes requested):

msasikanth (approved):

PR #5119, opened 2024-11-14
Add Simple server Docker Compose file

igbanam:
**Story:** https://app.shortcut.com/simpledotorg/story/13775/fix-integration-tests-on-android-ci-checks

## Description

This would be used in running simple-server as a service during CI. It does not run when introduced alongside the action which needs it.

This file is the same as [simple's docker compose](https://github.com/simpledotorg/simple-server/blob/master/.docker/docker-compose.yml) without…

- `volumes:`, no need for this since we don't want to share files in CI; every run should be contained
- `build:` because we are not building from source, rather we want to use `image:` to point to Simple's latest image
- `env_file:` because this was inlined as `environment:`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Next Release
 
+### Fixes
+
+- Fix app crashing occasionally when viewing all blood sugar history
+
+## 2024-11-11-9265
+
 ### Internal
 
 - Bump AGP to v8.7.2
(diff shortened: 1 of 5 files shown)


.github/docker/simple-server.compose.yml
```diff
@@ -0,0 +1,34 @@
+version: "3"
```
msasikanth, line 1: Can this is maintained on the server repo or does it have to be here? I want to avoid potential drift of setup in future between server docker file changes and this.

During CI we can checkout the server repo and use that file as well.

msasikanth (approved):

PR #5150, opened 2024-12-04
Add `isSmoking` in `MedicalHistory` table

siddh1004:
This includes:

- Updating MedicalHistory table to include isSmoking
- Syncing the new fields with the server

**Shortcut Stories:**
https://app.shortcut.com/simpledotorg/story/14053/update-medical-history-to-add-smoking-and-cholesterol-status


app/src/main/java/org/simple/clinic/medicalhistory/CholesterolReading.kt
```diff
@@ -0,0 +1,7 @@
+package org.simple.clinic.medicalhistory
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+data class CholesterolReading(val value: String) : Parcelable
```
msasikanth, line 7: How about representing this as a float value instead? Since `String` can be anything instead of a numerical value.

app/src/main/java/org/simple/clinic/medicalhistory/sync/MedicalHistoryPayload.kt
```diff
@@ -43,6 +43,12 @@ data class MedicalHistoryPayload(
     @Json(name = "hypertension")
     val hasHypertension: Answer?,
 
+    @Json(name = "smoking")
+    val isSmoker: Answer,
```
msasikanth, line 47: Maybe we can name it as `isSmoking` or `smoking` to match the API naming convention? Same for database column naming as well.

msasikanth (changes requested):

app/src/main/java/org/simple/clinic/medicalhistory/CholesterolReading.kt
siddh1004: I have removed the cholesterol from the code as its not part of the currents scope, maybe we can do it once we start working on phase 3

app/src/main/java/org/simple/clinic/medicalhistory/sync/MedicalHistoryPayload.kt
siddh1004: Renamed to `smoking`

msasikanth (approved):

PR #5154, opened 2024-12-10
Add `CVDRisk` sync resource

siddh1004:
https://app.shortcut.com/simpledotorg/story/14259/add-cv-risk-score-to-android-app-sync-module?team_id=1&iteration_ids=14091

### DIFF
--- a/app/schemas/org.simple.clinic.AppDatabase/118.json
+++ b/app/schemas/org.simple.clinic.AppDatabase/118.json
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 118,
-    "identityHash": "82aff276e9a69eae90d93966437302da",
+    "identityHash": "3b3ca12275a89f31c632031b12f80579",
     "entities": [
       {
         "tableName": "Patient",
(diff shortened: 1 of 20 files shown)


app/src/main/java/org/simple/clinic/cvdrisk/CVDRisk.kt
```diff
@@ -0,0 +1,32 @@
...
+      Index("patientUuid", unique = false)
+    ]
+)
+data class CVDRisk(
+    @PrimaryKey
+    val uuid: UUID,
+
+    val patientUuid: UUID,
+
+    val riskScore: String,
```
msasikanth, line 33: Risk score is a numerical value, no?

app/schemas/org.simple.clinic.AppDatabase/118.json
```diff
@@ -2,7 +2,7 @@
   "formatVersion": 1,
```
msasikanth, line 2: Just a suggestion for future, don't have to change now. "Update json file" seems way to vague, commit message should mention "Update database schema JSON".

msasikanth (changes requested):

app/src/main/java/org/simple/clinic/cvdrisk/CVDRisk.kt
siddh1004: @msasikanth I think it should be string. As risk score will be a range when smoking status is unknown or when bmi is not known.

app/schemas/org.simple.clinic.AppDatabase/118.json
siddh1004: sure. will keep in mind.

msasikanth (approved):

PR #5155, opened 2024-12-13
Add CVD Risk Calculation sheet

siddh1004:
https://app.shortcut.com/simpledotorg/story/14050/save-cv-risk-score-json-model-to-database

### DIFF
--- a/app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculationSheet.kt
+++ b/app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculationSheet.kt
@@ -1,0 +1,36 @@
+package org.simple.clinic.cvdrisk
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+class CVDRiskCalculationSheet(
+    val women: CVDRiskCalculationGenderSheet,
+    val men: CVDRiskCalculationGenderSheet,
+)
+
+@JsonClass(generateAdapter = true)
+data class CVDRiskCalculationGenderSheet(
+    val smoking: SmokingData,
+
+    @Json(name = "nonsmoking")
+    val nonSmoking: SmokingData
+)
+
+@JsonClass(generateAdapter = true)
+data class SmokingData(
+    @Json(name = "40 - 44") val age40to44: List<RiskEntry>?,
+    @Json(name = "45 - 49") val age45to49: List<RiskEntry>?,
+    @Json(name = "50 - 54") val age50to54: List<RiskEntry>?,
+    @Json(name = "55 - 59") val age55to59: List<RiskEntry>?,
+    @Json(name = "60 - 64") val age60to64: List<RiskEntry>?,
+    @Json(name = "65 - 69") val age65to69: List<RiskEntry>?,
+    @Json(name = "70 - 74") val age70to74: List<RiskEntry>?
+)
+
+@JsonClass(generateAdapter = true)
+data class RiskEntry(
+    val sbp: String,
+    val bmi: String,
+    val risk: Int
+)
--- a/app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculationSheet.kt
+++ b/app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculationSheet.kt
@@ -30,7 +30,7 @@ data class SmokingData(
 
 @JsonClass(generateAdapter = true)
 data class RiskEntry(
-    val sbp: String,
+    @Json(name = "sbp") val systolic: String,
     val bmi: String,
     val risk: Int
 )
--- a/app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculationSheet.kt
+++ b/app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculationSheet.kt
@@ -5,12 +5,12 @@ import com.squareup.moshi.JsonClass
 
 @JsonClass(generateAdapter = true)
 class CVDRiskCalculationSheet(
-    val women: CVDRiskCalculationGenderSheet,
-    val men: CVDRiskCalculationGenderSheet,
+    val women: CVDRiskCategory,
+    val men: CVDRiskCategory,
 )
 
 @JsonClass(generateAdapter = true)
-data class CVDRiskCalculationGenderSheet(
+data class CVDRiskCategory(
     val smoking: SmokingData,
 
     @Json(name = "nonsmoking")
--- a/app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculator.kt
+++ b/app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculator.kt
@@ -5,23 +5,16 @@ import org.simple.clinic.patient.Gender
 
 object CVDRiskCalculator {
 
-  fun calculateCvdRisk(
-      cvdRiskInput: CVDRiskInput
-  ): String? {
+  fun calculateCvdRisk(cvdRiskInput: CVDRiskInput): String? {
     with(cvdRiskInput) {
       if (cvdRiskData == null) return null
 
       val genderData = getGenderData(cvdRiskData, gender) ?: return null
       val smokingDataList = getSmokingDataList(genderData, isSmoker)
       val ageRange = getAgeRange(smokingDataList, age) ?: return null
-
       val sbpRange = getSBPRange(sbp)
       val bmiRangeList = getBMIRangeList(bmi)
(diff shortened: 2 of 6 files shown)


app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculator.kt
```diff
@@ -0,0 +1,71 @@
...
+
+object CVDRiskCalculator {
+
+  fun calculateCvdRisk(
+      cvdRiskData: CVDRiskCalculationSheet?,
+      gender: Gender,
+      age: Int,
+      sbp: Int,
+      isSmoker: Answer = Answer.Unanswered,
+      bmi: Double? = null
```
msasikanth: I think a `Float` precision should be enough for BMI

app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculator.kt
```diff
@@ -0,0 +1,71 @@
+package org.simple.clinic.cvdrisk
+
+import org.simple.clinic.patient.Answer
+import org.simple.clinic.patient.Gender
+
+object CVDRiskCalculator {
+
+  fun calculateCvdRisk(
+      cvdRiskData: CVDRiskCalculationSheet?,
```
msasikanth: If I understand correctly, this sheet data won't change from patient to patient right? Why don't we pass this as dependency to this class and use it here instead of passing it to function everywhere we call it? 🤔

app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculator.kt
```diff
@@ -0,0 +1,71 @@
...
+      in 50..54 -> smokingDataList.mapNotNull { it.age50to54 }.flatten()
+      in 55..59 -> smokingDataList.mapNotNull { it.age55to59 }.flatten()
+      in 60..64 -> smokingDataList.mapNotNull { it.age60to64 }.flatten()
+      in 65..69 -> smokingDataList.mapNotNull { it.age65to69 }.flatten()
+      in 70..74 -> smokingDataList.mapNotNull { it.age70to74 }.flatten()
+      else -> return null
+    }
+
+    val sbpRange = when (sbp) {
+      in Int.MIN_VALUE..119 -> "120-"
```
msasikanth: Do we get "120 -" from server? Why not just send it as "<=120", similar to how we have "180+"? Doesn't that make more sense, wdyt?

app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculator.kt
```diff
@@ -0,0 +1,71 @@
...
+      in 120..139 -> "120 - 139"
+      in 140..159 -> "140 - 159"
+      in 160..179 -> "160 - 179"
+      else -> "180+"
+    }
+
+    val bmiRangeList = if (bmi != null) {
+      listOf(
+          when (bmi) {
+            in Double.MIN_VALUE..19.9 -> "20-"
```
msasikanth: Same comment as the sbp min variable

app/src/test/java/org/simple/clinic/cvdrisk/CVDRiskCalculatorTest.kt
```diff
@@ -0,0 +1,89 @@
+package org.simple.clinic.cvdrisk
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.simple.clinic.patient.Answer
+import org.simple.clinic.patient.Gender
+import org.simple.sharedTestCode.TestData
+
+class CVDRiskCalculatorTest {
```
msasikanth, line 9: Nice work on tests 👍🏾

app/src/test/java/org/simple/clinic/cvdrisk/CVDRiskCalculatorTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
+  private val testData = TestData.cvdRiskCalculationSheet()
```
msasikanth, line 11: This variable should be `cvdRiskCalculationSheet` instead of `testData`

app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskModule.kt
```diff
@@ -31,4 +38,19 @@ class CVDRiskModule {
   fun lastPullToken(rxSharedPrefs: RxSharedPreferences): Preference<Optional<String>> {
     return rxSharedPrefs.getOptional("last_cvd_risk_pull_token", StringPreferenceConverter())
   }
+
+  @Provides
+  @AppScope
+  @OptIn(ExperimentalStdlibApi::class)
+  fun cvdRiskCalculationSheet(moshi: Moshi, configReader: ConfigReader): CVDRiskCalculationSheet? {
```
msasikanth, line 45: So, this happens on the main thread. Since this JSON is large, I am worried that it might cause ANR. How about we load it lazily and in a different thread and place where we use it?

msasikanth (changes requested):
Left few comments PTAL

app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculationSheet.kt
```diff
@@ -0,0 +1,36 @@
...
+    @Json(name = "50 - 54") val age50to54: List<RiskEntry>?,
+    @Json(name = "55 - 59") val age55to59: List<RiskEntry>?,
+    @Json(name = "60 - 64") val age60to64: List<RiskEntry>?,
+    @Json(name = "65 - 69") val age65to69: List<RiskEntry>?,
+    @Json(name = "70 - 74") val age70to74: List<RiskEntry>?
+)
+
+@JsonClass(generateAdapter = true)
+data class RiskEntry(
+    val sbp: String,
```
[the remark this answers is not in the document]
siddh1004: Done.

app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculator.kt
siddh1004: Done

app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculator.kt
siddh1004: We get this from server as "120-", so we can't change it to "<=120"

app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculator.kt
siddh1004: We get this from server as "20-", so we can't change it to "<=20"

app/src/test/java/org/simple/clinic/cvdrisk/CVDRiskCalculatorTest.kt
siddh1004: Thanks

app/src/test/java/org/simple/clinic/cvdrisk/CVDRiskCalculatorTest.kt
siddh1004: Done

app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculator.kt
msasikanth: Yeah, I understand that. I was talking about updating the spec from server and in the app.

app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculationSheet.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+class CVDRiskCalculationSheet(
+    val women: CVDRiskCalculationGenderSheet,
+    val men: CVDRiskCalculationGenderSheet,
+)
+
+@JsonClass(generateAdapter = true)
+data class CVDRiskCalculationGenderSheet(
```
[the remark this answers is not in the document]
siddh1004: I have renamed it to `CVDRiskCategory`

app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskCalculator.kt
siddh1004: Done.

msasikanth (approved):

PR #5163, opened 2025-01-10
Add smoking and bmi buttons to statin nudge

siddh1004:
https://app.shortcut.com/simpledotorg/story/14054/update-statin-nudge-design-in-patient-summary-screen-to-display-risk-scale

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffect.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffect.kt
@@ -122,3 +122,5 @@ data class ShowReassignPatientWarningSheet(
 data object ShowDiabetesDiagnosisWarning : PatientSummaryViewEffect()
 
 data class ShowHypertensionDiagnosisWarning(val continueToDiabetesDiagnosisWarning: Boolean) : PatientSummaryViewEffect()
+
+data object ShowSmokingStatusDialog : PatientSummaryViewEffect()
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffect.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryEffect.kt
@@ -126,3 +126,5 @@ data object ShowDiabetesDiagnosisWarning : PatientSummaryViewEffect()
 data class ShowHypertensionDiagnosisWarning(val continueToDiabetesDiagnosisWarning: Boolean) : PatientSummaryViewEffect()
 
 data object ShowSmokingStatusDialog : PatientSummaryViewEffect()
+
+data class OpenBMIEntrySheet(val patientUuid: UUID) : PatientSummaryViewEffect()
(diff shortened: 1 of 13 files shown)


app/src/main/java/org/simple/clinic/summary/PatientSummaryEffect.kt
```diff
@@ -64,6 +64,8 @@ data class CalculateCVDRisk(val patient: Patient) : PatientSummaryEffect()
 
 data class LoadStatinInfo(val patientUuid: UUID) : PatientSummaryEffect()
 
+data class UpdateSmokingStatus(val patientId: UUID, val isSmoker: org.simple.clinic.medicalhistory.Answer) : PatientSummaryEffect()
```
msasikanth, line 67: Can we use an import alias for the `Answer`, something like `MedicalHistoryAnswer`, wdyt?

app/src/main/java/org/simple/clinic/summary/compose/StatinNudgeView.kt
```diff
@@ -74,6 +79,7 @@ fun StatinNudge(
                 text = stringResource(R.string.statin_alert_at_risk_patient)
             )
           }
+          StainNudgeAddButtons(Modifier.padding(top = 16.dp), statinInfo, addSmokingClick, addBMIClick)
```
msasikanth: Can we put these params in separate lines and add param names, please?

app/src/main/java/org/simple/clinic/summary/compose/StatinNudgeView.kt
```diff
@@ -142,6 +148,54 @@ fun RiskProgressBar() {
...
 }
 
+@Composable
+fun StainNudgeAddButtons(
+    modifier: Modifier,
+    statinInfo: StatinInfo,
+    addSmokingClick: () -> Unit,
+    addBMIClick: () -> Unit,
+) {
+  if (statinInfo.cvdRisk != null) {
```
msasikanth, line 158: Technically we can pull this check into the parent composable, since there is no alternative layout in this composable function if it's false

msasikanth (changes requested):
Left couple of small comments, PTAL

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffect.kt
siddh1004: We also have `import org.simple.clinic.patient.Answer` in the same class that's why it's being used like here.

app/src/main/java/org/simple/clinic/summary/compose/StatinNudgeView.kt
siddh1004: Done

app/src/main/java/org/simple/clinic/summary/compose/StatinNudgeView.kt
siddh1004: Done

app/src/main/java/org/simple/clinic/summary/PatientSummaryEffect.kt
msasikanth: Yeah. That's why I was suggesting import aliases

https://kotlinlang.org/docs/packages.html#imports

We can use 'as' keyword

msasikanth (approved):

PR #5179, opened 2025-01-16
Update statin nudge risk scale V2

siddh1004:
https://app.shortcut.com/simpledotorg/story/14054/update-statin-nudge-design-in-patient-summary-screen-to-display-risk-scale

### DIFF
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
@@ -131,12 +131,12 @@ class PatientSummaryUpdate(
       }
 
       event.age > maxAgeForCVDRisk -> {
-        val updatedModel = model.updateStatinInfo(StatinInfo(canPrescribeStatin = hasCVD || hasDiabetes))
+        val updatedModel = model.updateStatinInfo(StatinInfo(canPrescribeStatin = hasCVD || hasDiabetes, hasCVD = hasCVD))
         next(updatedModel)
       }
 
       hasCVD || hasDiabetes -> {
-        val updatedModel = model.updateStatinInfo(StatinInfo(canPrescribeStatin = true))
+        val updatedModel = model.updateStatinInfo(StatinInfo(canPrescribeStatin = true, hasCVD = hasCVD))
         next(updatedModel)
       }
 
--- a/app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
@@ -125,18 +125,38 @@ class PatientSummaryUpdate(
         hasStatinsPrescribedAlready.not()
 
     return when {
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/simple/clinic/summary/compose/StatinNudgeView.kt
```diff
@@ -55,54 +63,105 @@ fun StatinNudge(
...
-              .padding(16.dp),
-          horizontalAlignment = Alignment.CenterHorizontally
+      BoxWithConstraints(
+          modifier = Modifier.padding(16.dp)
       ) {
-        Row {
-          Box(
-              modifier = Modifier
-                  .weight(2f)
+        val constraints = constraints
```
msasikanth, line 69: Is there a reason why we created a variable called `constraints`, when we can directly access `constraints` in the `BoxWithConstraints` scope?

app/src/main/java/org/simple/clinic/summary/compose/StatinNudgeView.kt
```diff
@@ -113,48 +172,105 @@ fun RiskProgressBar() {
...
+                    drawRect(
+                        color = color.copy(alpha = 1.0f),
+                        topLeft = Offset(x = visibleStart - segmentStartPx, y = 0f),
+                        size = Size(
+                            width = visibleEnd - visibleStart,
+                            height = size.height
+                        )
+                    )
+                  }
+                }
```
msasikanth, line 230: I am just curious, is it possible to replace this `Row` and `Box` usage and directly draw in canvas? Since we are already doing it in the `Box`?

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
```diff
@@ -131,12 +131,12 @@ class PatientSummaryUpdate(
       }
 
       event.age > maxAgeForCVDRisk -> {
-        val updatedModel = model.updateStatinInfo(StatinInfo(canPrescribeStatin = hasCVD || hasDiabetes))
+        val updatedModel = model.updateStatinInfo(StatinInfo(canPrescribeStatin = hasCVD || hasDiabetes, hasCVD = hasCVD))
```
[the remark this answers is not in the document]
siddh1004: Done

app/src/main/java/org/simple/clinic/summary/compose/StatinNudgeView.kt
siddh1004: Without creating a variable, it is not accessible in RiskText Which is inside a column. 
Error:- `'val constraints: Constraints' can't be called in this context by implicit receiver. Use the explicit one if necessary`

app/src/main/java/org/simple/clinic/summary/compose/StatinNudgeView.kt
siddh1004: I tried using canvas, and clipping and size of risk progress bar were not working properly. So i had to implement it using box and row.

msasikanth (approved):

PR #5229, opened 2025-01-31
Show smoking status dialog when cvd risk is `LOW-HIGH` or `MEDIUM-HIGH`

siddh1004:
https://app.shortcut.com/simpledotorg/story/14643/show-smoking-dialog-if-risk-is-low-high-or-medium-high?team_id=1&iteration_ids=14465&iteration_ids=14489

### DIFF
--- a/app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskRange.kt
+++ b/app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskRange.kt
@@ -4,7 +4,6 @@ import android.os.Parcelable
 import androidx.room.TypeConverter
 import com.squareup.moshi.FromJson
 import com.squareup.moshi.ToJson
-import kotlinx.parcelize.IgnoredOnParcel
 import kotlinx.parcelize.Parcelize
 
 @Parcelize
@@ -13,13 +12,11 @@ data class CVDRiskRange(
     val max: Int,
 ) : Parcelable {
 
-  @Transient
-  @IgnoredOnParcel
-  val level = CVDRiskLevel.compute(this)
+  val level: CVDRiskLevel
+    get() = CVDRiskLevel.compute(this)
 
-  @Transient
-  @IgnoredOnParcel
-  val canPrescribeStatin = max >= 10
+  val canPrescribeStatin: Boolean
+    get() = max >= 10
 
   class RoomTypeConverter {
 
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/simple/clinic/cvdrisk/CVDRiskRange.kt
```diff
@@ -12,6 +13,14 @@ data class CVDRiskRange(
     val max: Int,
 ) : Parcelable {
 
+  @Transient
+  @IgnoredOnParcel
+  val level = CVDRiskLevel.compute(this)
+
+  @Transient
+  @IgnoredOnParcel
+  val canPrescribeStatin = max >= 10
```
msasikanth, line 22: Instead of defining the variables directly, can we instead use the property getter like so?

```kotlin
  val canPrescribeStatin: Boolean
    get() = max >= 10
```

That way we don't have to deal with the serializer/parcelizer.

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
```diff
@@ -179,7 +181,20 @@ class PatientSummaryUpdate(
       event: StatinInfoLoaded,
       model: PatientSummaryModel
   ): Next<PatientSummaryModel, PatientSummaryEffect> {
-    return next(model.updateStatinInfo(event.statinInfo))
+    val canShowSmokingStatusDialog =
+        event.statinInfo.canPrescribeStatin &&
+        (event.statinInfo.cvdRisk?.level == CVDRiskLevel.LOW_HIGH ||
+        event.statinInfo.cvdRisk?.level == CVDRiskLevel.MEDIUM_HIGH) &&
+        event.statinInfo.isSmoker == MedicalHistoryAnswer.Unanswered &&
+        !model.hasShownSmokingStatusDialog
```
msasikanth, line 189: While this works, and we can leave this for now. The underlying problem still exists. The statin info is being reloaded every time patient info changes. So, the `LoadStatinPrescriptionCheckInfo` effect gets triggered every time.

msasikanth (changes requested):

msasikanth:
Please update the CHANGELOG as well

app/src/main/java/org/simple/clinic/summary/PatientSummaryUpdate.kt
siddh1004: Thanks for flagging this out. I will create a story and move to backlog.

msasikanth (approved):

PR #5390, opened 2025-04-24
Auto-select country and state when only one option is available

siddh1004:
https://app.shortcut.com/simpledotorg/story/15123/skip-country-and-state-selection-screen-in-the-wb-ncd-app

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,14 @@
 
 ## Next Release
 
-## Internal
+### Internal
 
 - Bump Gson to v2.13.0
 - Skip Sentry config during app startup in the debug mode
+
+### Features
+
+- Auto-select country and state when only one option is available
 
 ## 2025.04.07
 
--- a/app/src/main/java/org/simple/clinic/navigation/v2/Router.kt
+++ b/app/src/main/java/org/simple/clinic/navigation/v2/Router.kt
@@ -95,9 +95,13 @@ class Router(
       return
     }
 
-    val newHistory = history
-        .removeLast()
-        .add(navRequest)
+    val newHistory = if (history.requests.size > 1) {
+      history
+          .removeLast()
+          .add(navRequest)
+    } else {
+      History(listOf(navRequest))
+    }
 
     executeStateChange(newHistory, Direction.Replace, null, transactionOptions)
   }
--- a/app/src/main/java/org/simple/clinic/selectcountry/SelectCountryUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/selectcountry/SelectCountryUpdate.kt
@@ -2,6 +2,7 @@ package org.simple.clinic.selectcountry
 
 import com.spotify.mobius.Next
 import com.spotify.mobius.Update
+import org.simple.clinic.appconfig.Country
 import org.simple.clinic.mobius.dispatch
 import org.simple.clinic.mobius.next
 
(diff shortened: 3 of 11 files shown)


app/src/main/java/org/simple/clinic/selectcountry/SelectCountryUpdate.kt
```diff
@@ -16,10 +16,20 @@ class SelectCountryUpdate : Update<SelectCountryModel, SelectCountryEvent, Selec
...
+      CountrySaved -> countrySaved(model)
       RetryClicked -> next(model.fetching(), FetchManifest)
     }
   }
+
+  private fun countrySaved(model: SelectCountryModel): Next<SelectCountryModel, SelectCountryEffect> {
+    val effect = if (model.hasMoreThanOneCountry) {
+      GoToStateSelectionScreen
+    } else {
+      ReplaceCurrentScreenWithStateSelectionScreen
```
msasikanth, line 28: Let's rename this to `ReplaceWithStateSelectionScreen` to keep it a bit simple.

app/src/main/java/org/simple/clinic/selectcountry/UiActions.kt
```diff
@@ -2,4 +2,5 @@ package org.simple.clinic.selectcountry
 
 interface UiActions {
   fun goToStateSelectionScreen()
+  fun replaceCurrentScreenToStateSelectionScreen()
```
msasikanth, line 5: Same comment as above for renaming

app/src/main/java/org/simple/clinic/navigation/v2/Router.kt
```diff
@@ -95,9 +95,13 @@ class Router(
...
 
-    val newHistory = history
-        .removeLast()
-        .add(navRequest)
+    val newHistory = if (history.requests.size > 1) {
+      history
+          .removeLast()
+          .add(navRequest)
+    } else {
+      History(listOf(navRequest))
```
msasikanth, line 103: Instead of creating a new history, we probably can just return the existing history?

app/src/main/java/org/simple/clinic/selectcountry/SelectCountryUpdate.kt
```diff
@@ -31,10 +31,11 @@ class SelectCountryUpdate : Update<SelectCountryModel, SelectCountryEvent, Selec
   }
 
   private fun manifestFetched(model: SelectCountryModel, countries: List<Country>): Next<SelectCountryModel, SelectCountryEffect> {
+    val manifestFetchedModel = model.manifestFetched(countries)
     return if (countries.size > 1) {
-      next(model.manifestFetched(countries))
+      next(manifestFetchedModel)
     } else {
-      dispatch(SaveCountryEffect(countries.first()))
+      next(manifestFetchedModel, SaveCountryEffect(countries.first()))
```
msasikanth, line 38: Just curious, do we need to update the model at all? Can't we just perform the side effect and done with it? Even if the saving country takes a bit more time (highly unlikely) we just progress indicator instead of blank screen. WDYT? 🤔

CHANGELOG.md
```diff
@@ -2,11 +2,15 @@
...
 
-## Internal
+### Internal
 
 - Bump Gson to v2.13.0
 - Skip Sentry config during app startup in the debug mode
 
+### Features
+
+- Auto-select country and state when only one option is available
```
msasikanth, line 12: Did we add the auto state selection part in this PR? I don't see it.

msasikanth (changes requested):

app/src/main/java/org/simple/clinic/selectcountry/SelectCountryUpdate.kt
siddh1004: Yeah, we can do that. I updated the model so `SelectCountryInit` does not fetch manifest again and again as it depends on `val effects: Set<SelectCountryEffect> = if (!model.hasFetchedCountries()) setOf(FetchManifest) else emptySet()`. Also the screen will only be rendered when the count is greater than 1 i.e. `model.hasFetchedCountries() && model.hasMoreThanOneCountry -> {
        ui.displaySupportedCountries(model.countries!!, model.selectedCountry)
      }`

app/src/main/java/org/simple/clinic/navigation/v2/Router.kt
siddh1004: When the history contains only one request and we want to replace it with new screen., then we would have to create new history.

msasikanth (approved):

PR #5438, opened 2025-06-09
Add Compose components for Overdue screen

siddh1004:
https://app.shortcut.com/simpledotorg/story/15489/migrate-overdue-patient-item-list-to-use-compose?team_id=1&iteration_ids=15732


common-ui/src/main/java/org/simple/clinic/common/ui/components/SearchButton.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.simple.clinic.common.ui.theme.SimpleTheme
+
+@Composable
+fun SearchButton(
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+    enabled: Boolean = true,
+    content: @Composable RowScope.() -> Unit
+) {
+  androidx.compose.material.OutlinedButton(
```
msasikanth: We have a custom `OutlinedButton` in `Button.kt`, any reason we didn't use that here? Is it just the text style?

app/src/main/java/org/simple/clinic/home/overdue/compose/OverduePatientSearchButton.kt
```diff
@@ -0,0 +1,23 @@
...
+    onSearch: () -> Unit
+) {
+  SearchButton(
+      onClick = onSearch,
+      modifier = modifier
+          .fillMaxWidth()
+  ) {
+    Text(text = stringResource(id = R.string.overdue_search_patient_name_or_village))
+  }
+}
```
msasikanth: Since there aren't any customisations here, let's just use it in-place instead of creating another component.

msasikanth (changes requested):
LGTM. Left one comment, PTAL

common-ui/src/main/java/org/simple/clinic/common/ui/components/SearchButton.kt
siddh1004: @msasikanth There were two things different from `OutlinedButton` in `Button.kt` which are:

1. We needed 2 dp corner radius
2. We needed text style to be `body0` instead of `buttonBig`

app/src/main/java/org/simple/clinic/home/overdue/compose/OverduePatientSearchButton.kt
siddh1004: Done. Removed `OverduePatientSearchButton`

common-ui/src/main/java/org/simple/clinic/common/ui/components/SearchButton.kt
msasikanth: Actually let's replace the 2dp value, since it doesn't fit with the Material specs. Min is 4dp aka MaterialTheme.shapes.small.

But for text, we can just add that as a param for the `Button`s? Since we would have to use these buttons in other parts of the app where we will migrate screens to Compose. WDYT?

siddh1004: I think it's better to create a separate component for search button instead of updating the existing `OutlinedButton.kt`. Helps us to mimic the current design where radius is 2 dp and avoid unnecessary param addition to outlined button. Anyway, open to update the existing component as well. WDYT?

siddh1004: Removed `SearchButton`

msasikanth (approved):

PR #5440, opened 2025-06-10
Migrate medicines summary view to Jetpack Compose

msasikanth:
https://app.shortcut.com/simpledotorg/story/15751/migrate-medicines-summary-view-to-compose

### Pre Compose
drugs_summary_pre_compose

### Compose
drugs_summary_compose

While the text style of the drug item is changed, it matches the text style in Figma.


common-ui/src/main/java/org/simple/clinic/common/ui/components/Button.kt
```diff
@@ -90,9 +92,48 @@ fun FilledButton(
...
+      modifier = modifier.defaultMinSize(
+          minWidth = 64.dp,
+          minHeight = minHeight
+      ),
+      onClick = onClick,
+      enabled = enabled,
+  ) {
+    leadingIcon?.let {
+      it()
+      Spacer(modifier = Modifier.requiredWidth(8.dp))
```
siddh1004, line 121: Can we use dimenResource for this?

common-ui/src/main/java/org/simple/clinic/common/ui/components/Button.kt
```diff
@@ -90,9 +92,48 @@ fun FilledButton(
...
+    }
+    ProvideTextStyle(value = SimpleTheme.typography.buttonBig) {
+      content()
+    }
+    trailingIcon?.let {
+      Spacer(modifier = Modifier.requiredWidth(8.dp))
```
siddh1004, line 127: Can we use dimenResource for this?

siddh1004 (changes requested):

PR #5446, opened 2025-06-16
Migrate BP summary view to Compose

msasikanth:
https://app.shortcut.com/simpledotorg/story/15752/migrate-bp-summary-view-to-compose


app/src/main/java/org/simple/clinic/summary/bloodpressures/ui/BloodPressureSummary.kt
```diff
@@ -0,0 +1,176 @@
...
+          Text(text = label)
+        }
+      }
+    }
+
+    Divider(
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(horizontal = dimensionResource(id = R.dimen.spacing_8)),
+        thickness = 1.dp,
```
siddh1004: The default thickness of Divider is 1.dp, so maybe we can skip this parameter

app/src/main/java/org/simple/clinic/summary/bloodpressures/ui/BloodPressureSummary.kt
```diff
@@ -0,0 +1,176 @@
...
+          .then(modifier)
+  ) {
+    Row(
+        modifier = Modifier.fillMaxWidth(),
+        horizontalArrangement = Arrangement.SpaceBetween,
+    ) {
+      TextButton(
+          leadingIcon = {
+            Icon(
+                painter = painterResource(id = R.drawable.ic_add_circle_blue1_24dp),
```
siddh1004, line 50: Can we use the material icon instead of drawable if it's available in the library?

app/src/main/java/org/simple/clinic/summary/bloodpressures/ui/BloodPressureSummary.kt
```diff
@@ -0,0 +1,176 @@
...
+      ) {
+        val label = stringResource(id = R.string.patientsummary_bp_add_new).uppercase()
+        Text(text = label)
+      }
+
+      if (canShowSeeAllButton) {
+        TextButton(
+            trailingIcon = {
+              Icon(
+                  painter = painterResource(id = R.drawable.ic_chevron_right_blue1_24dp),
```
siddh1004, line 65: Same as above.

app/src/main/java/org/simple/clinic/summary/bloodpressures/ui/BloodPressureSummary.kt
```diff
@@ -0,0 +1,176 @@
...
+    Divider(
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(horizontal = dimensionResource(id = R.dimen.spacing_8)),
+        thickness = 1.dp,
+        color = SimpleTheme.colors.onSurface11,
+    )
+
+    if (summaryItems.isNotEmpty()) {
+      Column(
```
siddh1004, line 86: Can we use LazyColumn here?

app/src/main/java/org/simple/clinic/summary/bloodpressures/ui/BloodPressureSummaryItem.kt
```diff
@@ -0,0 +1,236 @@
...
+@Composable
+fun BloodPressureSummaryItem(
+    item: BloodPressureSummaryItem,
+    modifier: Modifier = Modifier,
+    onEdit: () -> Unit,
+) {
+  val interactionSource = remember { MutableInteractionSource() }
+  Row(
+      modifier = Modifier
+          .then(modifier)
```
siddh1004, line 38: Can't we directly use modifier.fillMaxWidth()...

app/src/main/java/org/simple/clinic/summary/bloodpressures/ui/BloodPressureSummary.kt
msasikanth: We cannot do that, it will throw an exception. LazyColumn doesn't calculate height based on the content inside it, instead it will have "infinite" content height so that it can scroll.


Since we are only showing fixed number of items(4 or 5), we want the content size to be wrapped based on the items height. So, `Column` fits the needs better here instead of `LazyColumn`

app/src/main/java/org/simple/clinic/summary/bloodpressures/ui/BloodPressureSummaryItem.kt
msasikanth: We don't have to here? Since we are using `Row` with a specific element weight set, it will automatically stretch to whatever the parent width is. So we don't have to set the width for individual item and also set the weight of the date time text view.

app/src/main/java/org/simple/clinic/summary/bloodpressures/ui/BloodPressureSummary.kt
msasikanth: We already have those drawables, I don't think it would change much in this PR? Since these icons are used in few others places. So, we cannot get rid of them as well in this PR. Ultimately, it doesn't provide much benefit in terms of performance or maintenance, so kept them as is.

app/src/main/java/org/simple/clinic/summary/bloodpressures/ui/BloodPressureSummary.kt
msasikanth: I know, but since these can be changeable by the library based on Material versions, I hardcoded the value. But I can remove it, I don't have strong opinion about that.

msasikanth: Done ✅

app/src/main/java/org/simple/clinic/summary/bloodpressures/ui/BloodPressureSummary.kt
msasikanth: Just to add, LazyColumn is better suited when we have lots of content and don't want all composables to be in memory even if they are outside the draw area. It's not suited here.

siddh1004 (approved):

PR #5452, opened 2025-06-18
Migrate medical history summary view to Jetpack Compose

msasikanth:
**Story:** https://app.shortcut.com/simpledotorg/story/15754/migrate-medical-history-to-compose

## Before
medical_history_pre_compose

## After
medical_history_compose


app/src/main/java/org/simple/clinic/medicalhistory/ui/MedicalHistoryQuestionItem.kt
```diff
@@ -0,0 +1,159 @@
...
+            onSelectionChange(Answer.Unanswered)
+          } else {
+            onSelectionChange(newAnswer)
+          }
+        }
+      )
+    }
+
+    if (showDivider) {
+      Divider(color = SimpleTheme.colors.onSurface11)
```
siddh1004, line 68: @msasikanth The default color for divider is `MaterialTheme.colors.onSurface.copy(alpha = DividerAlpha)` where alpha = 0.12f. Maybe we can use a default value only?

siddh1004 (changes requested):

app/src/main/java/org/simple/clinic/medicalhistory/ui/MedicalHistoryQuestionItem.kt
msasikanth: Our default divider color is surface11, let's just merge it in, we can make those changes through out app later instead of starting from now.

siddh1004 (approved):

PR #5605, opened 2025-09-17
Prevent multiple user lookup api calls 

siddh1004:
https://app.shortcut.com/simpledotorg/story/16445/prevent-users-to-click-next-button-when-a-call-is-already-in-queue

### DIFF
--- a/app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneUpdate.kt
+++ b/app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneUpdate.kt
@@ -40,6 +40,9 @@ class RegistrationPhoneUpdate : Update<RegistrationPhoneModel, RegistrationPhone
       event: EnteredNumberValidated
   ): Next<RegistrationPhoneModel, RegistrationPhoneEffect> {
     val updatedModel = model.phoneNumberValidated(event.result)
+    if (updatedModel.isRegistrationOngoing) {
+      return noChange()
+    }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/simple/clinic/registration/phone/RegistrationPhoneUpdate.kt
```diff
@@ -75,7 +78,7 @@ class RegistrationPhoneUpdate : Update<RegistrationPhoneModel, RegistrationPhone
       foundUserUuid: UUID,
       foundUserStatus: UserStatus
   ): Next<RegistrationPhoneModel, RegistrationPhoneEffect> {
-    val number = model.ongoingRegistrationEntry!!.phoneNumber!!
+    val number = model.ongoingRegistrationEntry.phoneNumber!!
```
igbanam, line 81: I see you've removed `!!` from `ongoingRegistrationEntry`; thereby accepting that `ongoingRegistrationEntry` could be null. This could potentially cause a null pointer exception.

How are we verifying that `ongoingRegistrationEntry` would never be null at this point?

igbanam: It happens in another place in the PR, but I did not want to duplicate comments

siddh1004:
It does look like one but it's not a regression function. O have just added a check which will prevent of the call is already in queue(registering patient )

igbanam (approved):

